Issue #7, opened 2020-09-24
As a data editor, I want to search across multiple metadata fields so that I can find geniza items by keyword or phrase.


kmcelwee:
When searching by description, bring up sorted matches with their associated PGPID, library, and shelfmark. 

- [x] display tags on search result
- [x] display document type on search result
- [x] include library in fields searched

kmcelwee:
Site is live! https://test-geniza.cdh.princeton.edu/

Testing notes:
- [ ] Can you successfully search on the most salient features of the metadata? (PGPID, description, library, shelfmark)
- [ ] To the best of your knowledge, is the metadata properly connected and formatted?

Also, we support more advanced search features that you're welcome to try. [Check out the documentation](https://prosody.princeton.edu/search/) from one of our other projects that shows some relevant functionality.

sluescher:
I searched for the terms "Arabic" and "court" (see screenshot attached), and the first result does not include "court," but "Arabic" in the shelf mark. I would expect the first result to contain both terms.
Here the PGP link as well: [https://geniza.princeton.edu/pgpsearch/?a=object&id=30854](https://geniza.princeton.edu/pgpsearch/?a=object&id=30854 )
![Arabic court search](https://user-images.githubusercontent.com/30683720/96177308-aa334900-0efb-11eb-9d86-07ca7452c667.png)

sluescher:
When searching for the library "Rylands," only results appear that either have "Rylands" in their description and/or shelf mark, but not those who have the value set as library (after | ).

kmcelwee:
**Re: "Arabic court"**
https://test-geniza.cdh.princeton.edu/?keywords=Arabic+court
Although the Solr search algorithm uses relevance of the entire input to rank the output, it looks like the first result was weighted higher than the second because the word "Arabic" occurs in both the fragment and the tags.

**Re: "Reylands"**
https://test-geniza.cdh.princeton.edu/?keywords=Rylands
I was unable to reproduce your results, would you mind sending me a link?

Both of these issues seem like they can be solved with faceting, which will be included in the next enhancement. @thatbudakguy any further thoughts on this?

rlskoeser:
I just checked the solr config, we're not currently searching on the library field. Indexed fields are: pgpid, type, shelfmark, tags, editors, translators

rlskoeser:
@kmcelwee I think it would displaying the tags on the search results would help with testing — I hadn't registered the fact that they weren't displayed until you mentioned it.

[...]

PR #17, opened 2020-10-15
Feature/index external links

kmcelwee:
* Here's the layout I landed on, I assume we'd want to iterate on this?
Screen Shot 2020-10-15 at 11 24 57 AM
* Do we want the external link icon? (the box with an arrow to the outside)
* Do we have a preference between null values and empty strings in Solr? Some tags are empty strings, and `link_s` is sometimes an empty string, but with my new `iiif_link_s` field, I set it to `None` so it doesn't appear in all solr records.
* The image server at Cambridge is currently down, so links are difficult to test right now.

### DIFF
--- a/scripts/index.py
+++ b/scripts/index.py
@@ -49,6 +49,11 @@ def index():
         'tags_txt': [tag.strip() for tag in row['Tags'].split('#')],
         'tags_ss': [tag.strip() for tag in row['Tags'].split('#')],
         'link_s': row['Link to image'],
+        'iiif_link_s': (
+            row['Link to image'].replace('/view/', '/iiif/') 
+            if 'cudl.lib.cam.ac.uk' in row['Link to image']
+            else None
+        ),
         'editors_txt': row['Editor(s)'],
         'translators_txt': row['Translator (optional)']
     } for row in rows], commitWithin=100)
--- a/scripts/index.py
+++ b/scripts/index.py
@@ -46,8 +46,8 @@ def index():
         'library_s': row['Library'],
         'shelfmark_s': row['Shelfmark - Current'],
         'shelfmark_txt': row['Shelfmark - Current'],
-        'tags_txt': [tag.strip() for tag in row['Tags'].split('#')],
-        'tags_ss': [tag.strip() for tag in row['Tags'].split('#')],
+        'tags_txt': [tag.strip() for tag in row['Tags'].split('#') if tag.strip() != ''],
+        'tags_ss': [tag.strip() for tag in row['Tags'].split('#') if tag.strip() != ''],
         'link_s': row['Link to image'],
         'iiif_link_s': (
             row['Link to image'].replace('/view/', '/iiif/') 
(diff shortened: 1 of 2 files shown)


scripts/templates/results.html
```diff
@@ -26,8 +26,14 @@
...
             <div class="fl w-80 pa2">
                 <span>{{ result.description_txt[0]|safe }}</span><br/>
-                <!-- source mss -->
-                <span class="id fl purple avenir">{{ result.shelfmark_s }} | {{ result.library_s }}</span><br>
+                <div class="w-100 flex justify-between">
+                    <!-- source mss -->
+                    <div class="id fl purple avenir">{{ result.shelfmark_s }} | {{ result.library_s }}</div><br>
+                    <!-- View external source -->
+                    {% if result.link_s %}
+                    <div><a class="id no-underline fl purple avenir" href="{{ result.link_s }}">View external record</a></div>
```
thatbudakguy, line 34: any reason for the `no-underline`? usually color alone isn't sufficient to indicate a link to all users.

thatbudakguy (approved):
idk if we need the external link icon now, but we should definitely make a note so @gissoo can either choose an appropriate icon or design one if none fit our purposes. I'd say `None` is preferable wherever possible, but defer to @rlskoeser.

rlskoeser:
How did I not know that sending a value of `None` in the dict results in Solr not storing a value? I could swear that a previous system I used stored that as the string "None" 😬 . Yes, let's use `None` instead of empty string for both links and iiif urls.

This display seems fine to me. While we're prototyping, let's keep the visuals as plain and simple as possible. No icons, minimal visuals. I don't want us to unintentionally make any visual decisions or assumptions that we'll have to try to forget!

rlskoeser:
@kmcelwee not specific to this PR, but is the library always repeated in the shelfmark? If so, let's drop it.

scripts/index.py
```diff
@@ -46,9 +46,14 @@ def index():
         'library_s': row['Library'],
         'shelfmark_s': row['Shelfmark - Current'],
         'shelfmark_txt': row['Shelfmark - Current'],
-        'tags_txt': [tag.strip() for tag in row['Tags'].split('#')],
-        'tags_ss': [tag.strip() for tag in row['Tags'].split('#')],
+        'tags_txt': [tag.strip() for tag in row['Tags'].split('#') if tag.strip() != ''],
```
rlskoeser, line 49: Thanks for catching this, had just noticed it myself. Just `if tag.strip()` is sufficient.

scripts/templates/results.html
rlskoeser: agree, let's drop the `no-underline`

rlskoeser (approved):

kmcelwee:
@rlskoeser The shelfmarks do not necessarily map to libraries in a consistent way, for example:
```
PGPID: 19961
T-S 12.399 | CUL

PGPID: 30854
JRL GENIZAH FRAGMENT ARABIC 202 | Rylands
```

I didn't have time to replace empty strings with None. Should we make a chore?

I think this should be merged? We haven't finished QA for #7, but as @rlskoeser said, it's probably useful for them to have tags while testing.

Issue #22, opened 2020-11-02
As a user, I want to see a list of topic clusters so I can get a sense of the major areas of content and relationships between them.


rlskoeser:
- [x] Create a new view & template in flask app
- [x] Use Solr facet pivot to get a list of tags that appear together, with some minimum (maybe 50, to start)
- [x] Output list of tag pairs & number of documents

rlskoeser:
- [ ] Create a new view & template in flask app
- [ ] Use Solr facet pivot to get a list of tags that appear together, with some minimum (maybe 50, to start)
- [ ] Output list of tag pairs & number of documents

rlskoeser:
I think normally this would be 2 points, but since we're not writing unit tests for experimental/prototype code I think it's 1 point.

rlskoeser:
@sluescher this is available on the test site now at https://test-geniza.cdh.princeton.edu/clusters/

@gissoo will be doing usability testing on the cluster search prototype, but to check the functionality please confirm that you see a list of document clusters, which are pairs of tags that occur on at least 50 document. For each cluster, you should see the two tags and the total number of documents.

sluescher:
I am unsure how this is different from #23? I do see the list of topic clusters. I am assuming there might be more than just sets of tags at some point? Or is the plan to only have two at a time?

rlskoeser:
@sluescher this one is the first step (just looking at the list of clusters) and #23 is the next step (selecting a single cluster to see the documents).

For the prototype, we decided to just use pairs of tags for our clusters. I think @gissoo is interested in having more than two pieces of information in a cluster (so it's more about connections within the data), and we may well generate clusters quite differently down the road, but this was a simple way to prototype the idea with the data we have now.

PR #29, opened 2020-11-03
List of tag-pair clusters with document counts

rlskoeser:
ref #22

### DIFF
--- a/scripts/server.py
+++ b/scripts/server.py
@@ -10,7 +10,6 @@ Run a debug server for development with:
     $ flask run
 
 '''
-
 from flask import Flask, g, render_template, request
 from parasolr.query import SolrQuerySet
 from parasolr.solr.client import SolrClient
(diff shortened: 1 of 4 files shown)


thatbudakguy (approved):
looks good! the solr output of the pivots is kind of weird, and it makes me wonder if the logic needed to parse it could get inefficient, but the actual query seemed to run really quickly for me.

scripts/server.py
```diff
@@ -65,6 +64,48 @@ def search():
...
+            # every pivot repeats the first tag; skip that one
+            if tag_b == tag_a:
+                continue
+            tag_pair = '/'.join([tag_a, tag_b])
+            # each pair will show up twice; only add it the first time
+            alt_tag_pair = '/'.join([tag_b, tag_a])
+            if alt_tag_pair in tag_pairs:
+                continue
+            tag_pairs.append(tag_pair)
+            tag_pair_counts[tag_pair] = pivot['count']
```
thatbudakguy, line 91: there might be a more efficient way to do this, but certainly not necessary to figure out now. it's weird how much information solr repeats in the response!

thatbudakguy (approved):

scripts/server.py
rlskoeser: @thatbudakguy I think the redundancy is because you usually pivot on different fields, not the same one. (Or at least, that's my sense of the typical use case for the functionality)

thatbudakguy: ah, that makes a lot of sense. I guess this is kind of a strange case

PR #40, opened 2020-11-17
Experiment/search clusters

rlskoeser:
Adds search functionality to the new cluster prototype.

Documents in sea rch results are _not_ grouped; they are displayed similar to the cluster browse view, with an indicator and link to any clusters that the document belongs to.


thatbudakguy (approved):
works exactly the way I expect! had trouble finding many documents that belonged to more than one cluster, incidentally. this might be because to do so, a document would need >4 tags, since each cluster is currently two tags.

Issue #36, opened 2020-11-16
As a content editor, I want to add data to the database in multiple languages so that I can fully represent existing project data.


rlskoeser:
# testing notes
**note** - this issue is a companion to #35, but they address different issues and function differently. this issue covers all the actual PGP data that can be stored in multiple languages, rather than text on the site.
## as a user
see "as a user" on #35; since this part works exactly the same way. then, pick a person from [the list of people](https://test-geniza.cdh.princeton.edu/en/people/). you'll see a basic testing page that tells you what profession that person had. currently, nobody has any profession. let's add one.
## as a content editor
1. go to the [admin backend for the site](https://test-geniza.cdh.princeton.edu/admin/) and click "people". click a person's name to edit them.
2. you'll now see the fields that are available on a person: their name, and their profession. go ahead and pick any profession from the list and save the model.
3. note also that there are three tabs here, labeled according to the language codes of the languages that you can currently browse the test site in. if you go to one of those tabs, the field will be empty, indicating there's no value in that language. we can control which languages are available at the level of each field, but the default is to allow every language that you can browse the site in.
4. go ahead and add a (fake or real) version of the person's name in another language and save the model.
5. if you go back to the public site and visit the page of the person you edited, you'll see a message like "X was a Y", where X is the person's name and Y is the profession you chose. if you switch language, both X and Y should now display in the translated versions you entered! if you didn't enter a translation for the language you chose, it will just display in english instead. note that the _middle_ part of this text ("was a") isn't actually "PGP data", and thus it would be translated via the methods covered in #35.

leave comments/opinions below on how this process went for you.

# dev notes
- [x] model editable in django admin with translation
- [x] simple list page
- [x] simple detail page

rlskoeser:
@thatbudakguy I think this was the library I was thinking of (but I don't know the maintenance status or versions of django it supports): https://django-modeltranslation.readthedocs.io/en/latest/

thatbudakguy:
some notes on how this is working:
- when you access an attribute of a model instance using its regular field name (`profession.title`), you get the value for the localized version of the field that matches Django's currently active language. This means that in templates and views, you usually just use the regular field name and let the language detection middleware do the rest (see #35).
- after you define the fields you want translated in an app's `translations.py`, you need to `makemigrations` so that a new migration can be generated to add the localized versions of the fields. if you change the available languages or default language in `settings.py` (uncommon), you also need to do this.
[...]

Issue #35, opened 2020-11-16
As a user, I want to read the geniza project in my native language so it's easier to understand.


rlskoeser:
# testing notes
**note** - this issue is a companion to #36, but they address different issues and function differently. this issue covers all the text on the site that isn't actual "PGP data" - in other words, nothing that would come from the database.

you might encounter documentation (or people) that use the terms `i18n` and `l10n`. these are lazy ways of writing the long words "internationalization" and "localization", where the numbers mean how many letters you skipped in the middle of the word. the former refers to writing code that _can be_ translated into multiple languages, while the latter refers to actually _doing_ the translation ("localizing") that code into some particular language. for the PGP, the developers will be doing the i18n, but the l10n can be done by the project team.

## as a user
1. visit [the test site](https://test-geniza.cdh.princeton.edu/) and you'll see a (very basic) homepage that tells you what language you're currently reading the PGP in.
2. you'll notice that the default is english; there's also now a `/en/` appended to the URL to indicate this.
3. check out the [list of professions](https://test-geniza.cdh.princeton.edu/en/people/professions/) in english and you'll see the transliterated profession names.
3. choose another language to read in using the dropdown at the top right, and click "Go." (these languages are configurable and we can choose as many or as few as we want).
4. you should now see the profession names change to reflect your choice, and the url suffix should also change to a language code (e.g. `/he/` for hebrew). note though that the actual URLs will still be in english (e.g. `/ar/people`). let us know in a comment if this makes sense - it's possible to instead do `/ar/اشخاص/` or `/ar/ashas` (idk if these are correct but you get the point).
5. go back to the homepage by clicking "home" in the top left, and you should see the text after "your language is" has also changed. once you pick a language, the website will "remember" it until you make another choice - including if you refresh, close the tab, close the browser, etc. you can remove the choice by clearing your cookies to default back to english, or just choose it from the dropdown.
6. note that the rest of the language on the homepage didn't change! to make that work, we have to do some extra work from the content editor's point of view.
## as a content editor
1. have a look at [the `locale` folder on github](https://github.com/Princeton-CDH/geniza/tree/experiment/i18n/geniza/locale). you'll see three folders: `ar`, `en`, and `he`, corresponding to our three language options. each has a folder inside called `LC_MESSAGES` (this is a standard name that's required to use).
[...]

PR #41, opened 2020-11-18
i18n experiments

thatbudakguy:
refs #35, #36 

# setup
- (optional) create a new virtual environment; I used python 3.8 (see `.python-version`)
- install python requirements
- run migrations
- download the .csv of professions and import with `python manage.py import_professions file.csv`
- download either or both .csv's of people ([editors](https://drive.google.com/file/d/1iaMdSJpaqo9AV1y-XLIxDobk6PGfWb2h/view?usp=sharing) or [authors](https://drive.google.com/file/d/1NXjBj0qWbLWGXuLGpkLAW11pjTp72WFH/view?usp=sharing)) and import with `python manage.py import_people file.csv`
- run `django-admin compilemessages` from inside the `geniza/` directory to generate the `.mo` binary files

# testing
- if you load the profession list and then change your language using the switcher at the top right, the profession names should match the language you chose
- if you go to the admin and create a new person or profession, you should see tabs for the localized versions of the model. translated fields will change, but other fields won't change.
- check out the generated `.po` files in the `locale/` folders for the main app and people app to get an idea of what the files that translators would edit look like. you can see comments and file references generated from the source files. if you add new items to be translated, you can regenerate these by running `django-admin makemessages --all` in the `geniza` app directory. it won't clobber existing translations.


geniza/locale/ar/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,62 @@
...
+#
+#, fuzzy
+msgid ""
+msgstr ""
+"Project-Id-Version: PACKAGE VERSION\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2020-11-18 16:44-0500\n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: LANGUAGE <LL@li.org>\n"
```
rlskoeser, line 14: Are these configurable somewhere in our settings? Doesn't matter for the experiment, just curious.

rlskoeser: Although it would be good to know which fields translators are supposed to update when they work on translations.

geniza/people/templates/people/person_detail.html
```diff
@@ -0,0 +1,16 @@
...
+{% block content %}
+<h2>{{ person.name }}</h2>
+<p>
+{% if person.profession %}
+{# Translators: description on a person's detail page #}
+{% blocktranslate with name=person.name profession=person.profession %}
+{{ name }} was a {{ profession }}.
+{% endblocktranslate %}
+{% else %}
+{% translate "No profession recorded." %}
```
rlskoeser, line 13: very cool

geniza/people/translation.py
```diff
@@ -0,0 +1,14 @@
...
+@register(Person)
+class PersonTranslationOptions(TranslationOptions):
+    fields = ("name",)
+    required_languages = ("en",)
+
+
+@register(Profession)
+class ProfessionTranslationOptions(TranslationOptions):
+    fields = ("title",)
+    required_languages = ("en",)
```
rlskoeser, line 14: Ah, interesting. So for every model that is translatable, we have to specify which fields and languages are required? That makes sense. I'm curious how we could simplify it; e.g. maybe a config setting, if the required languages are always the same. But that seems like something we could figure out down the road if and when it's a problem, once we see what the patterns are for how we're using it.

rlskoeser (commented):
This looks great.  Thanks for taking the time to create model and load sample data so it's clearer how this works and what it does. The tabbed admin interface seems pretty workable, and the model lists are switching based on the language I choose. 

I tried editing one of the `.po` files to add a translation for something you didn't include and ran the `compilemessages`, but I can't see the translation I added. And once I looked, I noticed that the home / people / professions nav doesn't seem to be switching based on the selected language, even though they are in the message files. Is that working for you?

geniza/locale/ar/LC_MESSAGES/django.po
thatbudakguy: I'm not sure where they come from...as far as I can tell they're defaults from gnu gettext, which django uses under the hood. I can see if there's some way to configure them, but for now at least if we edit them they won't be clobbered when we regenerate the file.

thatbudakguy:
I had the same experience as you - changed a `.po` file, recompiled, but didn't see the results. I was curious whether it'd work the same way for you. Not sure what's going on; need to investigate further.

geniza/locale/ar/LC_MESSAGES/django.po
thatbudakguy: As far as I can tell, translators only ever touch the `msgstr` field...maybe there are some docs somewhere we can point people to, since it's a common utility.

thatbudakguy: it looks like if we want to customize what `gettext()` is doing (e.g. to override the `.pot` template used to generate the `.po` files), we would need to [create our own management command](https://docs.djangoproject.com/en/3.0/topics/i18n/translation/#customizing-the-makemessages-command) that extends `makemessages`.

rlskoeser:
@kmcelwee tagging you for review just as a reminder for you to check this out later as a way of getting more familiar with django's internationalization.

geniza/people/models.py
```diff
@@ -0,0 +1,27 @@
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+
+class Person(models.Model):
+    """Person mentioned in or related to a geniza document."""
+    name = models.CharField(max_length=100)
+    profession = models.ForeignKey("Profession", null=True, on_delete=models.CASCADE)
```
kmcelwee, line 8: This means that if I delete the "Scribe" profession, that deletes all Persons who are scribes, correct? Is that the functionality we want?

kmcelwee (commented):
Great stuff! Would you mind throwing your setup notes into a readme or "devnotes"?

geniza/people/models.py
thatbudakguy: Probably not, I imagine we'd do `models.SET_NULL` in the real version!

.gitignore
```diff
@@ -0,0 +1,10 @@
+scripts/
+data/
+.vscode/
```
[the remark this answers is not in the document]
thatbudakguy: I do have one, just being lazy here because it's my branch ;)

thatbudakguy:
@kmcelwee definitely a good idea to preserve the setup notes, but I'm not sure how much they will change for the real version...I don't think we intend to merge this branch or keep any of the code in it, which is why it's a draft PR. I think we can just refer to this PR (even after it's closed) if we end up needing these specific notes.

@rlskoeser let me know if you agree.

rlskoeser:
@thatbudakguy I agree: no need to write up notes or worry about delete behavior since this is experimental code that we won't be using directly.

rlskoeser:
@thatbudakguy should we close this PR ? I think we're done reviewing at this point 🙂

rlskoeser:
Closing since review is complete and we don't actually want to merge this.

PR #44, opened 2020-11-24
Basic conversion of TEI transcriptions to IIIIF annotations

rlskoeser:
This is a basic, first-pass conversion of TEI transcriptions to IIIF annotations to help us all understand what transcriptions as annotations might look like and how they would work, and to get a better sense of how the current transcription data needs to be broken up.

Changes include:
- tweak the index script to fix the IIIF urls for Cambridge (I think they changed something recently on their end, because this was working previously)
- new script to generate iiif annotation lists for documents that have both IIIF and transcriptions (currently Cambridge only); for now, all annotations are added to the first image in the manifest and are attached to the whole page; multiple annotations are added when there are multiple blocks in the TEI document (as indicated by labels for new sections)
- new views in the flask server to serve out a mirador page that loads all the local manifests with annotations, so that the annotations created can be seen in context with the corresponding images

To try this out:
- update the index: `flask index`
- generate the transcription data file (or get a copy): `flask transcriptions`
- generate the annotation lists and local manifests: `flask transcriptions-iiif` (currently generates all 1176 of them; could set a max for testing)
- run the flask server and navigate to the transcriptions url (/iiif/); select a manifest to view the image + annotations

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 # local settings
 scripts/local_settings.py
-data/transcriptions.json
+data/
 
 # Webpack files
 webpack-stats.json
--- a/scripts/index.py
+++ b/scripts/index.py
@@ -54,7 +54,9 @@ def index():
         extlink = row['Link to image']
         iiif_link = None
         # cambridge iiif manifest links use the same id as view links
-        if 'cudl.lib.cam.ac.uk' in extlink:
+        # NOTE: exclude search link like this one:
+        # https://cudl.lib.cam.ac.uk/search?fileID=&keyword=T-s%2013J33.12&page=1&x=0&y=
+        if 'cudl.lib.cam.ac.uk/view/' in extlink:
             iiif_link = extlink.replace('/view/', '/iiif/')
             # view links end with /1 or /2 but iiif link does not include it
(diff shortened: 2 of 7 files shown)


rlskoeser:
Forgot to mention: currently the mirador page loads all the manifests that are generated, which means it is pretty slow. Obviously this isn't the solution we want to use eventually! 

Interested to hear if you think this good enough for testing or if we need to do something else to make it more usable in the short term. I can imagine creating a simple document detail page that loads mirador for the current document if we have iiif and transcription — but I think then we might need a filter to easily get to those documents from the search.

thatbudakguy (approved):
worked great once I set my `DATA_DIR` to an absolute path; we should probably document that.

this mirador viewer is way different than the one I'm used to; the interface is much nicer! worth thinking about how much we can customize it and where/how we want to embed it.

rlskoeser:
> this mirador viewer is way different than the one I'm used to; the interface is much nicer! worth thinking about how much we can customize it and where/how we want to embed it.

This is Mirador 3! It seems to be very configurable, and it seems like they are actively working on it. Unfortunately I did not find the documentation about how to configure it particularly comprehensive, but I did get answers pretty quickly when I asked on the IIIF Slack.

PR #59, opened 2021-02-04
Setup new django site

rlskoeser:
New django 3.1 site with basic configuration

I adapted i18n settings from @thatbudakguy's prototype, but I may not have gotten everything and I didn't add any instructions about it yet. (Not sure what we need at this point).

I'm trying out [django-split-settings](https://github.com/sobolevn/django-split-settings) for managing settings, interested to see what you guys think. I haven't split things out very much for now, and I'm really only using environment settings for unit testing.

I adapted GitHub Actions workflows for unit tests, dbdocs, and codeql. I think maybe pytest is erroring because I haven't written any tests yet.


rlskoeser:
I looked it up, pytest exit code 5 does indeed mean no tests were collected.

README.rst
```diff
@@ -0,0 +1,59 @@
...
+	cp geniza/settings/local_settings.py.sample geniza/settings/local_settings.py
+
+Remember to add a ``SECRET_KEY`` setting!
+
+- Create a new database and update your database settings accordingly
+
+- Run database migrations
+
+    python manage.py migrate
+
```
thatbudakguy, line 49: At some point we should add a reminder right after this to also run `cd geniza && django-admin compilemessages` which will compile the `.po` files so microcopy/translated content is available.

geniza/settings/components/base.py
```diff
@@ -0,0 +1,155 @@
...
+
+# SECURITY WARNING: don't run with debug turned on in production!
+DEBUG = True
+
+ALLOWED_HOSTS = []
+
+
+# Application definition
+
+INSTALLED_APPS = [
```
thatbudakguy, line 30: should we go ahead and throw `modeltranslation` on here too, and add to `requirements.py`?

geniza/settings/components/base.py
```diff
@@ -0,0 +1,155 @@
...
+    'django.contrib.messages',
+    'django.contrib.staticfiles',
+    'django_cas_ng',
+    'pucas',
+]
+
+MIDDLEWARE = [
+    'django.middleware.security.SecurityMiddleware',
+    'django.contrib.sessions.middleware.SessionMiddleware',
+    'django.middleware.common.CommonMiddleware',
```
thatbudakguy, line 44: needs `django.middleware.locale.LocaleMiddleware` for i18n; I think it needs to be after `SessionMiddleware`. I put it between `SessionMiddleware` and `CommonMiddleware` in the prototype.

geniza/settings/components/base.py
```diff
@@ -0,0 +1,155 @@
...
+    },
+]
+
+
+# Internationalization
+# https://docs.djangoproject.com/en/3.1/topics/i18n/
+
+LANGUAGE_CODE = 'en-us'
+
+TIME_ZONE = 'UTC'
```
thatbudakguy, line 120: is it bad to change this? could we just set it to `'EST'`? wondering if that would fix timestamps in the django logfile

geniza/settings/components/base.py
```diff
@@ -0,0 +1,155 @@
...
+LOCALE_PATHS = [BASE_DIR / "geniza" / "locale"]
+
+# Static files (CSS, JavaScript, Images)
+# https://docs.djangoproject.com/en/3.1/howto/static-files/
+
+STATIC_ROOT = BASE_DIR / "static"
+
+STATIC_URL = "/static/"
+
+
```
thatbudakguy, line 143: we had also set `STATICFILES_DIRS = [ BASE_DIR / "sitemedia" ]` here in the prototype, idk if that matters

thatbudakguy (approved):
lgtm; just a few questions about settings

README.rst
thatbudakguy: You also need to run `django-admin makemessages` from the `geniza` dir if you create any _new_ translatable content, but I don't think that one needs to be in the README.

README.rst
```diff
@@ -0,0 +1,59 @@
...
+Unit Tests
+----------
+
+Python unit tests are written with `py.test <http://doc.pytest.org/>`_
+and should be run with `pytest`.
+
+License
+-------
+This project is licensed under the `Apache 2.0 License <https://github.com/Princeton-CDH/mep-django/blob/main/LICENSE>`_.
```
kmcelwee, line 59: I've had issues with rst rendering on GitHub. Has it been just me? What are the unique benefits to rst?

kmcelwee (approved):

README.rst
rlskoeser: Thanks for noting these — if not in the readme, where should they go? seems important to document somewhere

geniza/settings/components/base.py
rlskoeser: I was thinking to leave it off until we actually start using it. Any value to adding it now?

geniza/settings/components/base.py
rlskoeser: oh, good point — I think we are supposed to change it for our local time zone

README.rst
thatbudakguy: imo the `compilemessages` one should go in the README, since it's required to get up and running. the `makemessages` one i view as analogous to `makemigrations`: not something we need to document explicitly since it occurs somewhat rarely, there's existing django documentation, and if you're in a position where you need it you probably already know. but open to changing that.

geniza/settings/components/base.py
thatbudakguy: i guess it doesn't make a difference

README.rst
thatbudakguy: One benefit I know of for us is that it allows sphinx to pick up the README and use it as the frontpage of the generated code documentation - here's [an example for PPA](https://princeton-cdh.github.io/ppa-django/). I think sphinx is still the de facto standard for autogenerated codedocs for python projects, which is why we use it (although I've been playing with [`pdoc3`](https://pdoc3.github.io/pdoc/) for DIRECT and like it so far).

rlskoeser: yes, exactly — I agree that GitHub's support for rst is fairly limited, but having the readme in rst means we can easily include it in our sphinx docs when we set them up. 

@thatbudakguy I saw you mentioned using pdoc3 for direct and wondered about it. I'd be open to trying that for geniza if you think it's a better choice and does everything we need it to!

README.rst
rlskoeser: @thatbudakguy I went ahead and added it to the readme for now, since it's not a regular part of our process yet like makemigrations is. We can always revise as we figure out what information is actually useful!

Issue #53, opened 2021-02-01
As a global admin, I want to be able to add a library collection not represented on the list in order to expand our content footprint and edit those already in the list.


sluescher:
## testing notes
log in on an account with **Content Admin** group permissions
- [x] list view should include library, collection name, library abbreviation, abbreviation, location
- [x] should be able to edit existing records or add new one
- [x] should be able to save a record with collection name but no library
- [x] should be able to save a record with library but no collection name
- [x] should not be allowed to save a record with both collection name and library empty
- [x] should be able to delete an existing record
- [x] should be able to add a new record
- [x] if you try to enter a collection with a library + collection combination that already exists, you should get an error

log in on an account with **Content Editor** group permissions
- [x] should be able to view collections but not add, edit, delete

---
For example, when new collections become available or there are name changes.

rlskoeser:
@sluescher in my draft db diagram I had the fields name, code, and url. I think based on your wording code should be abbreviation. Do you want the url field so that we can (eventually) link to libraries, or should I leave it out?

sluescher:
Yes, please with url!

sluescher:
Add the fields "Collection" (text, optional) and "Locations" (text).

sluescher:
Question: do we want Collection abbreviation?

rlskoeser:
@mrustow @sluescher we decided yesterday we only need a single abbreviation.  One more question: do you still want to call this "Library" or should it be called "Collection" ? Looking at your list, I think Collection might be a little clearer (especially for the one collection held jointly by two libraries), but I think either is fine. I just want to make sure to call it whatever makes the most sense to you.

sluescher:
But then we have a library with multiple collections. I do think Collections might be a more dynamic term than library, which feels stiff. This is for how we call the entity?

rlskoeser:
Yes, this is for the name of the entity. That's another good reason to go with Collection instead of Library — there are some libraries with more than one collection

sluescher:
✓ list view should include abbreviation, library, location, and optional collection name
✓ should be able to edit existing records or add new one; collection name should be optional
✓ should be able to delete an existing record
✓ (is included in #2 but works) should be able to add a new record
✓ if you try to enter a collection with an abbreviation that already exists, you should get an error
✓ if you try to enter a collection with a library + collection combination that already exists, you should get an error

Passes the test with flying colors 🌈.

sluescher:
[...]

Issue #52, opened 2021-02-01
As a global admin, I want a one-time import of the list of library collections that includes name, abbreviation and location, so that I can manage the library collection info in the database.


sluescher:
## testing notes

- Confirm that the list of libraries/collections has been correctly imported from the ontology spreadsheet
- Check that the edit history for a few records to confirm that there is a log entry showing the record was imported by script

rlskoeser:
@sluescher please confirm, should I be using the Libraries file in the ontologies directory in Google Drive? If so, is it ok if I publish to the web as CSV and use that for import? Alternatively I could download it as csv file if you don't expect any further changes.  

Also, do I need to do anything with the **Current List of Libraries** column?

sluescher:
Current List of Libraries is the info currently in the PGP spreadsheet in the column "Library." It is there as the checking list.

I think it should be fine to publish it (I'll double check with Marina). There might be changes happening.

Once we upload the metadata spreadsheet, will it be possible to link library and document automatically according to the library column?

rlskoeser:
@sluescher ah, I see — the first column is how they appear in the sheet now (mixture of names and abbreviations), and the other fields are how they should be listed in the database. 

Yes, we can use the Library sheet both to populate the fields in the database and as a lookup to match the libraries as listed in the sheet to the new versions in the database.

rlskoeser:
@sluescher sorry, more questions! I think both name and abbreviation should be required and unique, does that sound right? 

What should I do with the "ENA (should be JTS)" in your Libraries spreadsheet?

sluescher:
Well, ENA is in the library column, but it shouldn't be because it's part of the shelfmark, not the library. It should be JTS (Jewish Theological Seminary). So we need to correct that. I will set someone on it.

rlskoeser:
@sluescher we can convert ENA to JTS on import if it's consistent, no need to have someone make those changes manually

sluescher:
I will check how many instances there are (might also a good way of seeing whether there's other issues with the record). If it's 1 or 2, it's not a big deal. If it's a bunch of them, I will happily rely on the conversion

sluescher:
It is only one instance. taking care of it.

sluescher:
✓ Confirm that the list of libraries/collections has been correctly imported from the ontology spreadsheet
✓ Check that the edit history for a few records to confirm that there is a log entry showing the record was imported by script

At least one spelling mistake, but I assume that was imported 😆

[...]

PR #61, opened 2021-02-08
Library model, library admin config, import libraries

rlskoeser:
- new app `corpus` to hold document/fragment/library models and related information
- preliminary `data_exodus` script to import library information from provided csv
- new app `common` to configure user list in django admin and migrations to create content editor/admin groups and script user group

ref #53 #52


geniza/common/admin.py
```diff
@@ -0,0 +1,20 @@
...
+    '''Extends :class:`django.contribut.auth.admin.UserAdmin`
+        to provide additional detail for user administration.'''
+    list_display = UserAdmin.list_display + (
+        'is_superuser', 'is_active', 'last_login', 'group_names')
+
+    def group_names(self, obj):
+        '''Custom property to display group membership.'''
+        if obj.groups.exists():
+            return ', '.join(g.name for g in obj.groups.all())
+    group_names.short_description = 'groups'
```
thatbudakguy, line 16: this is a great idea!

geniza/common/migrations/0001_data_editor_group.py
```diff
@@ -0,0 +1,67 @@
...
+    ]
+}
+
+CONTENT_ADMIN = 'Content Admin'
+# additional permissions for content admin (also get content edit permissions)
+content_admin_perms = {
+    'corpus': [
+        'add_library', 'change_library', 'delete_library',
+    ]
+}
```
thatbudakguy, line 20: I thought the other permission level was called "global admin" – is that something else? Did we change it?

geniza/corpus/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
+
+
+class DocsConfig(AppConfig):
+    name = 'corpus'
```
thatbudakguy, line 5: Might be missing something, but why isn't this called `CorpusConfig`?

geniza/corpus/management/commands/data_exodus.py
```diff
@@ -0,0 +1,67 @@
...
+from django.core.management.base import BaseCommand, CommandError
+import requests
+
+from geniza.corpus.models import Library
+
+
+class Command(BaseCommand):
+    'Import existing data from PGP spreadsheets into the database'
+
+    logentry_message = 'Created via data exodus script'
```
thatbudakguy, line 17: this is very minor, but i actually kind of prefer "import" to "exodus" here because it's a little more clear about the data in question being external to django/the database. i can see reserving "exodus" for cases where we can't use a data migration, but still need to move data around in the database for whatever reason (a la cdhweb). but not a strong preference.

geniza/corpus/models.py
```diff
@@ -0,0 +1,15 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    abbrev = models.CharField('Abbreviation', max_length=255, unique=True)
+    url = models.URLField('URL', blank=True)
+
+    class Meta:
+        verbose_name_plural = 'Libraries'
+        ordering = ['name']
+
+    def __str__(self):
+        return self.name
```
thatbudakguy, line 15: super minor (and maybe I missed discussion on this) but I wonder if it'd be helpful to have it be `self.name (self.abbrev)` so they look like `British Library (BL)`. that might at least help us get more familiar with the abbreviations since we'll see them more.

thatbudakguy (approved):
good with merging as-is, just some notes/Qs. I really like the choice of `corpus` for this app, and `common` as well - i think we'll be hard-pressed to find anything else that works and is as succinct.

geniza/common/admin.py
rlskoeser: I copied it from one of our other projects — we haven't used it consistently, but I remember it being helpful when we did!

geniza/common/migrations/0001_data_editor_group.py
rlskoeser: @thatbudakguy to me global admin == superuser

Since they are making some controlled vocabulary not editable by content editors I thought it might be useful to make a content admin group so they don't have to make everyone superusers. Maybe worth discussing when we request review/testing?

geniza/corpus/apps.py
rlskoeser: Ha, good catch, thank you. (It's because I created it before I got the corpus name from Marina and I missed it when I renamed everything)

geniza/corpus/management/commands/data_exodus.py
rlskoeser: Fair point! I'm inclined to agree with you, import does seem more clear and obvious.

geniza/common/migrations/0001_data_editor_group.py
thatbudakguy: ah, I see - so we don't need to define perms for global admin since it's an existing django category, but we do need another level between that and content editor. that explanation makes sense, and i think this naming makes sense too – but agree good to discuss in testing.

geniza/corpus/models.py
thatbudakguy: oh, good point. if that's the case i agree it doesn't add much

kmcelwee (approved):
Branching off this makes way more sense for me!

PR #71, opened 2021-02-11
Define application version and make available via context processor

rlskoeser:
As @thatbudakguy pointed out somewhere, we need a python package version for the site.

Copied this code from other projects so it doesn't really need review, but thought a quick PR wouldn't hurt.

I decided to set the version as 0.3 since we've already created two prototypes, but that's kind of arbitrary. Not sure how much it matters. 😄

### DIFF
--- a/geniza/__init__.py
+++ b/geniza/__init__.py
@@ -1,0 +1,15 @@
+__version_info__ = (0, 3, 0, 'dev')
+
+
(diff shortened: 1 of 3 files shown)


rlskoeser:
Added a redirect view for the homepage to go to admin because otherwise the load balancer doesn't think there's a site to serve out.

thatbudakguy (approved):
lgtm!

PR #79, opened 2021-02-16
Refactor library model & library import to add fields and rename as collection

rlskoeser:


geniza/corpus/models.py
```diff
@@ -1,29 +1,41 @@
...
-    '''Library or archive that holds Geniza fragments'''
-    name = models.CharField(max_length=255, unique=True)
+class Collection(models.Model):
+    '''Collection at a library that holds Geniza fragments'''
+    library = models.CharField(max_length=255)
     abbrev = models.CharField('Abbreviation', max_length=255, unique=True)
-    url = models.URLField('URL', blank=True)
+    collection = models.CharField(
+        max_length=255, blank=True,
+        help_text='Collection name, if different than Library')
```
thatbudakguy, line 10: `Collection.collection` feels weird to me, but I'm guessing there's a reason it's not `Collection.name`?

geniza/corpus/models.py
```diff
@@ -1,29 +1,41 @@
...
+    location = models.CharField(
+        max_length=255, help_text='Current location of the collection')
```
thatbudakguy, line 12: idk what we're using this for -  if it's just display, `CharField` seems fine, but if we want to do anything with the actual info later we might wanna store real coords or geonames or something

geniza/corpus/models.py
```diff
@@ -1,29 +1,41 @@
...
+    display_name = models.CharField(
+        max_length=255, blank=True, unique=True, null=True,
         help_text="Option to override the autogenerated language-script name")
 
     class Meta:
         ordering = ['language']
-        constraints = [models.UniqueConstraint(fields=['language', 'script'], name='unique_language_script')]
+        constraints = [
+            models.UniqueConstraint(fields=['language', 'script'],
+                                    name='unique_language_script')
```
thatbudakguy, line 37: whoa, I've never seen this `constraints` API before! since you defined uniqueness here, should be trivial to also implement `get_natural_key()` – just a thought

thatbudakguy (approved):
the changes for `Collection` are a little strange to me but i'm sure they're based on convos with the team so it's just context I don't have; mentally going off of PEMM but that might not map to this model. would different collections at the same library have the the same abbrev?

geniza/corpus/models.py
rlskoeser: That's a good point; but name also seems a little weird, since this field is optional

geniza/corpus/models.py
rlskoeser: I thought about that too, and mentioned briefly in the working meeting when we discussed these changes; but I'm recommending we hold off on that until we have a use case, since our priority is getting the spreadsheet data migrated into the database.

geniza/corpus/models.py
rlskoeser: This was new to me too! Credit to @kmcelwee for this one, I would have just used `unique_together`, but apparently that's deprecated now and might go away eventually.

rlskoeser:
> the changes for `Collection` are a little strange to me but i'm sure they're based on convos with the team so it's just context I don't have; mentally going off of PEMM but that might not map to this model. would different collections at the same library have the the same abbrev?

Sounds like the changes aren't documented as well as they should be on the GitHub issue. Sorry about that.

We needed some adjustments to handle some outliers, including:
- multiple collections at the same institution
- one collection shared across two institutions

I asked about the abbreviation — it seems to be the collection abbreviation in some cases and the institution abbreviation in others (example: ENA vs JTS); I think that's ok as long as abbreviations are unique.  

Any thoughts or lessons learned from the PEMM collections that are relevant here?

rlskoeser:
@thatbudakguy added natural foreign keys for collection and languagescript; thanks for the suggestion. Anything else you think we should address before merging & delivering for testing?

thatbudakguy (approved):
let's merge it!

Issue #75, opened 2021-02-16
As a content editor, I want to create and edit fragment records so that I can document changes in the fragments.


sluescher:
## testing notes
**List display**
- [x] Ensure that fields that you want visible are available
- [x] Test that fragments are searchable by:
  - [x] shelfmark
  - [x] historic shelfark
  - [x] notes
- [x] Test that results for filtering by the following fields operate properly:
  - [x] collection
  - [x] Is a multifragment
  - [x] Has a url

**edit**
- [x] Test that changes to a fragment through the edit display are saved and reflected in the list view.
- [x] Test editing image url on the list view
- [x] Test that fragments with an associated IIIF url display a IIIF viewer with the page images
- [x] Test editing multifragment field

-----
## dev notes

revisions after testing:
- [x] fix multifragment list filter
- [x] fix iiif viewer mixed content error
- [x] make multifragment a boolean field and update admin filter/edit accordingly

**Fields**
Provide a list of the fields to be included on this model.
- Shelfmark
- Old shelfmark
- Library (abbreviation)
- Multi-fragment
- notes
- image url
- IIIF

**List Fields**
Which fields should be included on the django admin list view?
- Shelfmark
- library (abbreviation)
- image url 
- Multi-fragment (boolean)

**Edit Fields**
Which fields should be editable via django admin view? Please list them in the order you want them to appear. Indicate which fields are optional, and any fields that should be displayed but not editable.
- shelfmark
- historical shelfmark (optional) - populated if shelfmark field is changed
- library
- image url (optional)
- image thumbnail
- IIIF
- multi-fragment (optional)
- notes (optional)

**Search Fields**
Which fields should be searchable in the django admin list view?
- shelfmark
- historical shelfmark
- notes

**Default sort**
- Shelfmark

**Filters**
Which fields should be used for list filters?
- Library
- Multi-fragment (boolean)
- Image (boolean)

**List Fields (optional)**
Which fields (if any) should be editable on the django admin list view?
- Image url

richmanrachel:
- It's not obvious in the Edit Fragment view that the box for multifragment is supposed to be filled out as "Yes" (I tried it in a test, and it did create a check mark in the right place in the list view, but it wasn't initially clear that this is how that box is supposed to be used. Perhaps a checkbox here would be clearer.
- Filter by Multifragment (Yes/No) didn't work either, perhaps because the boxes referenced in my question above are not populated that way?
- I couldn't figure out a search term to bring up notes to test it.
[...]

Issue #68, opened 2021-02-09
As a content editor, I want to create, edit, filter and search documents so that I can add/edit information on documents in the database and find pertinent documents.


sluescher:
### Testing Notes
**List display:**
- [x] Ensure that all fields that you want visible are available
- [x] Test that Documents are searchable by:
  - [x] shelfmark
  - [x] tag
  - [x] description
  - [x] PGP team member who input data
- [x] Test that results for filtering by the following fields operate properly:
  - [x] document type
  - [x] language
  - [x] extent label
  - [x] multifragment

**Edit**
- [x] Test that changes to a document through the edit display are saved and reflected in the list view.
- [x] Test that read only fields are displayed and not editable
- [x] Test that text blocks for associated fragments are listed and can be updated
- [x] Test that you can't add unknown for a probable language
- [x] Test that you aren't allowed to set the same language+script as both language and probable language
- [x] Test that you aren't allowed to set unknown language+script as probable language
- [x] Test that multiple text blocks on the same shelfmark don't result in repeated shelfmark in document combined shelfmark
- [x] Test setting text block order and confirm that join shelfmark follows the specified order

-----

## dev notes

revisions after testing:
- [x] switch language and probable language to autocomplete
- [x] fix language/probable language validation
- [x] add help text for language and probable language (to be supplied by the team)
- [x] fix document shelfmark so it includes unique shelfmarks in order
- [x] add multifragment text field
- [x] make extent filter empty/not empty
- [x] add empty/not empty multifragment filter

**Fields**
Provide a list of the fields and entities to be included on this model.
- Fragments (entity)
- description
- side
- text-block
- Type (entity)
- Tags (entity)
- language
- Footnotes for edition and translation (entity)
- notes
- input by
- date entered

**List Fields**
Which fields should be included on the django admin list view?
- shelfmark
- description
- type
- tags
- language
- text-block (boolean)
- edition

**Edit Fields**
Which fields should be editable via django admin view? Please list them in the order you want them to appear. Indicate which fields are optional, and any fields that should be displayed but not editable.
- shelfmark (read only) at least one, possibly more
- historical shelfmark (read only)
- side (to indicate recto/verso)
- Image thumbnail (optional)
- type (optional)
- language
- description
- tags (optional)
- edition (optional)
- translation (optional)
- notes (optional)
- text-block (optional)
- Legacy input by (read only)
- Legacy date entered (read only) 

**Search Fields**
Which fields should be searchable in the django admin list view?
[...]

PR #80, opened 2021-02-17
models & admin for documents + fragments

rlskoeser:
- new models for document, fragment, document type, and document-fragment through model (currently calling this "text unit" but may want a better name)
- preliminary admin configuration for documents and fragments, with inline for text unit
- custom admin template for fragment editing that adds a universal viewer if there is a IIIF url
- methods on fragment to generate thumbnail images if there is a IIIF url; used in text unit inline

ref #68 #75


geniza/corpus/admin.py
```diff
@@ -12,3 +13,77 @@ class CollectionAdmin(admin.ModelAdmin):
...
+        ('created', 'last_modified')
+    )
+    filter_horizontal = ('languages', )
+    inlines = [
+        TextBlockInline,
+    ]
+
+    def get_queryset(self, request):
+        return super().get_queryset(request) \
+            .prefetch_related('tags', 'languages', 'textblock_set')
```
thatbudakguy, line 62: cool!

geniza/corpus/models.py
```diff
@@ -65,3 +71,161 @@ def __str__(self):
...
+
+    def get_by_natural_key(self, shelfmark):
+        return self.get(shelfmark=shelfmark)
+
+
+class Fragment(models.Model):
+    shelfmark = models.CharField(max_length=255, unique=True)
+    # multiple, semicolon-delimited values. Keeping as single-valued for now
+    old_shelfmarks = models.CharField(
+        'Historical Shelfmarks',
```
thatbudakguy, line 85: worth adding a docstring?

geniza/corpus/models.py
```diff
@@ -65,3 +71,161 @@ def __str__(self):
...
+        ordering = ['order']
+
+    def __str__(self):
+        # combine shelfmark, side, and optionally text block
+        parts = [self.fragment.shelfmark, self.get_side_display(),
+                 self.extent_label]
+        return ' '.join(p for p in parts if p)
+
+    def thumbnail(self):
+        return self.fragment.iiif_thumbnails()
```
thatbudakguy, line 231: confused about the function/naming of this - would it only ever return one thumbnail? seems like the method it delegates to could return multiple

geniza/templates/admin/corpus/fragment/change_form.html
```diff
@@ -0,0 +1,15 @@
+{% extends "admin/change_form.html" %}
+{% block extrahead %}{{ block.super }}
+<script type="text/javascript" id="embedUV" src="http://universalviewer.io/uv/lib/embed.js"></script>
+{% endblock %}
+
+{% block field_sets %}
+{% if original.iiif_url %}
+	<div class="wrapper">
+		<div id="change_form">{{ block.super }}</div>
+		<div class="uv" data-uri="{{ original.iiif_url }}" data-collectionindex="0" data-manifestindex="0" data-sequenceindex="0" data-canvasindex="0" data-rotation="0" style="width:100%; height:450px"></div>
```
thatbudakguy, line 10: probably not necessary at this point, but it might be useful to define the viewer height as a function of the viewport height, like `50vh` or something

thatbudakguy (approved):
lgtm; just one question about a method naming/function

geniza/corpus/models.py
rlskoeser: do you mean help text for the old shelfmarks field?

geniza/corpus/models.py
rlskoeser: good point, that was inconsistent of me — I have some thoughts on what this might look like later when we can identify which image from a manifest is referenced, but probably best to name it simply & accurately for now!

geniza/corpus/models.py
thatbudakguy: sorry, hard to indicate on github! I meant for the `Fragment` class itself, since the other models had them.

requirements/prod.txt
```diff
@@ -2,4 +2,7 @@ django>=3.1
 psycopg2
 pucas>=0.6
 django-split-settings
-requests
\ No newline at end of file
+requests
+django-taggit
+#next version of piffle
+git+https://github.com/princeton-cdh/piffle.git@feature/iiif_pres#egg=piffle
```
kmcelwee, line 8: Heads up: This doesn't exist anymore

rlskoeser: oh, thanks for pointing that out; I merged it the piffle branch, and I thought I updated the requirement for geniza (should be develop now), but either I didn't or I didn't push it yet...

geniza/corpus/admin.py
```diff
@@ -12,3 +13,79 @@ class CollectionAdmin(admin.ModelAdmin):
...
+@admin.register(DocumentType)
+class DocumentTypeAdmin(admin.ModelAdmin):
+    list_display = ('name',)
+
+
+@admin.register(Fragment)
+class FragmentAdmin(admin.ModelAdmin):
+    list_display = ('shelfmark', 'collection', 'url',
+                    'is_multifragment')
+    search_fields = ('shelfmark', 'old_shelfmarks', 'notes')
```
kmcelwee, line 76: Should we allow searching on collection?

kmcelwee: Ignore me it's a list filter

PR #81, opened 2021-02-23
Add django-postgres-extra; collection abbrev case-insensitive unique

rlskoeser:
Adds and enables django-postgres-extra and then uses their case insensitive unique index. 

Actual changes here are very small, but wanted another set of eyes and be aware of psqlextra. (Can you think of any concerns around using it?)


thatbudakguy (approved):
changes look good; no immediate concerns about using psqlextra - it seems well supported. this does make me think about what we could get out of database indices; i wonder if there's something there that could power an auto-highlight feature for shelfmarks or footnotes.

PR #87, opened 2021-03-03
Add save_as feature to document model

kmcelwee:
Customize this model's save_model function and then execute the existing admin.ModelAdmin save_model function. Borrows from existing the `save_as` functionality.

**Questions**
* I can't grab the original document object from the save_as request, any ideas? (`obj` is the new object, its pk is None)
* How would I go about testing this?

### DIFF
--- a/geniza/corpus/admin.py
+++ b/geniza/corpus/admin.py
@@ -65,6 +65,7 @@ class DocumentAdmin(admin.ModelAdmin):
     search_fields = ('fragments__shelfmark', 'tags__name', 'description',
                      'old_input_by')
     # TODO include search on edition once we add footnotes
+    save_as = True
 
     list_filter = (
         'doctype', 'languages', 'textblock__extent_label',
@@ -94,6 +95,15 @@ class DocumentAdmin(admin.ModelAdmin):
         return super().get_queryset(request) \
             .prefetch_related('tags', 'languages', 'textblock_set')
 
+    def save_model(self, request, obj, form, change):
+        '''Customize this model's save_model function and then execute the
+         existing admin.ModelAdmin save_model function'''
+        if '_saveasnew' in request.POST:
+            # HACK: I'm having trouble finding the original ID in the given parameters. Ideas?
+            original_doc = Document.objects.get(pk=int(request.POST['textblock_set-0-document']))
+            clone_message = f'Cloned from {str(original_doc)}'
+            obj.notes = obj.notes + '\n' + clone_message if obj.notes else clone_message
+        super().save_model(request, obj, form, change)
 
 @admin.register(DocumentType)
 class DocumentTypeAdmin(admin.ModelAdmin):
--- a/geniza/corpus/admin.py
+++ b/geniza/corpus/admin.py
@@ -2,7 +2,7 @@ from django.contrib import admin
 from django.db.models import Count
 
 from django.utils.html import format_html
-from django.urls import reverse
+from django.urls import reverse, resolve
 
 from geniza.corpus.models import Collection, Document, DocumentType, \
     Fragment, LanguageScript, TextBlock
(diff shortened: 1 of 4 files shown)


thatbudakguy:
to your first question, the [stackoverflow post you shared](https://stackoverflow.com/a/24162016) i think has a solution:

```python
    def save_model(self, request, obj, form, change):
        # Django always sends this when "Save as new is clicked"
        if '_saveasnew' in request.POST:
            # Get the ID from the admin URL
            original_pk = resolve(request.path).args[0]
            # Get the original object
            original_obj = obj._meta.concrete_model.objects.get(id=original_pk)

            # Iterate through all it's properties
            for prop, value in vars(original_obj).iteritems():
                # if the property is an Image (don't forget to import ImageFieldFile!)
                if isinstance(getattr(original_obj, prop), FieldFile):
                    setattr(obj,prop,getattr(original_obj, prop)) # Copy it!
        obj.save()
```
basically, you can scrape it from the request.

geniza/corpus/tests/test_corpus_models.py
```diff
@@ -139,7 +139,10 @@ def test_str(self):
         frag = Fragment.objects.create(shelfmark='Or.1081 2.25')
         doc = Document.objects.create()
         doc.fragments.add(frag)
-        assert str(doc) == doc.shelfmark
+        assert doc.shelfmark in str(doc) and str(doc.id) in str(doc)
+
+        unsaved_doc = Document()
+        assert str(unsaved_doc) == ' (??)'
```
kmcelwee, line 145: Is this extra space okay? Or should I add the extra logic to work around this?

geniza/corpus/models.py
```diff
@@ -172,11 +172,12 @@ class Meta:
         ordering = ['fragments__shelfmark']
 
     def __str__(self):
-        return self.shelfmark
+        return f"{self.shelfmark} ({self.id or '??'})"
 
     def clean(self):
-        if any([plang in self.languages.all() for plang in self.probable_languages.all()]):
-            raise ValidationError('Languages cannot be both probable and definite.')
+        if self.pk:
```
rlskoeser, line 178: this could just be a single condition — `if self.pk and any(...`

geniza/corpus/models.py
```diff
@@ -172,11 +172,12 @@ class Meta:
         ordering = ['fragments__shelfmark']
 
     def __str__(self):
-        return self.shelfmark
+        return f"{self.shelfmark} ({self.id or '??'})"
```
rlskoeser, line 175: Can we do (PGPID ##) to make this clearer?

geniza/corpus/tests/test_corpus_models.py
rlskoeser: Right, no shelfmark either when unsaved... what about `?? (PGPID ??)` for unsaved object?

geniza/corpus/admin.py
```diff
@@ -94,6 +95,17 @@ def get_queryset(self, request):
...
+    def save_model(self, request, obj, form, change):
+        '''Customize this model's save_model function and then execute the
+         existing admin.ModelAdmin save_model function'''
+        if '_saveasnew' in request.POST:
+            # Get the ID from the admin URL
+            original_pk = resolve(request.path).kwargs['object_id']
+            # Get the original object
+            original_doc = obj._meta.concrete_model.objects.get(id=original_pk)
+            clone_message = f'Cloned from {str(original_doc)}'
+            obj.notes = obj.notes + '\n' + clone_message if obj.notes else clone_message
```
rlskoeser, line 107: Please add a test case where document has existing notes

rlskoeser (commented):
Looks good, just a few small suggestions

geniza/corpus/admin.py
kmcelwee: Testing `notes` revealed that the object was not properly cloned in the first place. I tried debugging, but fixing this isn't straightforward. I can put an hour or so into this next week if you'd like me to ask Kate.

PR #92, opened 2021-03-08
Revise document/fragment admin

rlskoeser:
- shift `Document` language/probable language validation to admin `DocumentForm` because model `clean` method is insufficient
- configure language and probable languages as autocomplete instead of horizontal filter, to decrease space
- add validation to disallow unknown set as probable language, since admin autocomplete does not honor `limit_choices_to`
- use `EmptyFieldListFilter` instead of `BooleanFieldListFilter` for text field filters (multifragment & url)
- use https to load universalviewer script

### DIFF
--- a/geniza/corpus/admin.py
+++ b/geniza/corpus/admin.py
@@ -1,8 +1,9 @@
+from django import forms
 from django.contrib import admin
+from django.core.exceptions import ValidationError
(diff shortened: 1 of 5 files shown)


thatbudakguy (approved):
lgtm

PR #94, opened 2021-03-10
Revise collection model & import

rlskoeser:
- revise Collection model as discussed, to make library and collection name both optional (but at least one required)
- revise data import to handle collections better, including differentiating CUL items in collections based on shelfmark


geniza/corpus/management/commands/import_data.py
```diff
@@ -251,14 +257,33 @@ def import_documents(self):
...
+        if lib_code == 'CUL':
+            for cul_collection in ['T-S', 'CUL Or.', 'CUL Add.']:
+                if data.shelfmark.startswith(cul_collection):
+                    lib_code = 'CUL_%s' % cul_collection.replace('CUL ', '')
+                    break
+            # if code is still CUL, there is a problem
+            if lib_code == 'CUL':
+                self.stdout.write(self.style.WARNING(
+                    'CUL collection not determined for %s'
+                    % data.shelfmark))
```
thatbudakguy, line 284: idk if we're quite at this point yet, but using the actual `logging` lib has helped me a lot w/ cdhweb exodus, rather than writing to stdout

geniza/corpus/models.py
```diff
@@ -7,15 +7,18 @@
...
 class CollectionManager(models.Manager):
 
-    def get_by_natural_key(self, abbrev):
-        return self.get(abbrev=abbrev)
+    def get_by_natural_key(self, name, library):
+        return self.get(name=name, library=library)
 
 
 class Collection(models.Model):
     '''Collection at a library that holds Geniza fragments'''
```
thatbudakguy, line 15: docstring is maybe less accurate now since library isn't required

geniza/corpus/models.py
```diff
@@ -26,15 +29,21 @@ class Collection(models.Model):
...
         ordering = ['abbrev']
         constraints = [
-            models.UniqueConstraint(fields=['library', 'collection'],
-                                    name='unique_library_collection')
+            # require at least one of library OR name
+            models.CheckConstraint(
+                check=(models.Q(library__regex='.+') |
+                       models.Q(name__regex='.+')),
+                name='req_library_or_name'
+            ),
```
thatbudakguy, line 37: weird that this requires a regex, but glad we now have an example of how to do it!

geniza/corpus/tests/test_corpus_models.py
```diff
@@ -11,17 +12,54 @@
...
+        # library only
+        Collection.objects.create(library='British Library')
+        # collection only
+        Collection.objects.create(name='Chapira')
+        # library + collection
+        Collection.objects.create(library='Cambridge UL', name='Taylor-Schechter')
+
+        # one of library or name is required
+        with pytest.raises(IntegrityError):
+            Collection.objects.create(lib_abbrev='BL')
```
thatbudakguy, line 62: is this `IntegrityError` because it's a database-level constraint as opposed to an admin-level one which would be `ValidationError`? would it be simpler to implement validation at the admin level?

thatbudakguy (changes requested):
lgtm, just some questions

thatbudakguy (approved):

geniza/corpus/models.py
rlskoeser: I couldn't figure out any other way to look for non-empty content! Agree it is kind of weird.

geniza/corpus/management/commands/import_data.py
rlskoeser: not sure we are either, but thanks for the suggestion, will keep it in mind

PR #101, opened 2021-03-15
Add document suppression and include in admin

kmcelwee:
A lot of copying and pasting from PPA!  I didn't include any testing for this, should I? `is_public` doesn't really mean anything until we have views / other logic, right?


geniza/corpus/migrations/0005_add_suppressed_docs.py
```diff
@@ -0,0 +1,35 @@
...
+        migrations.AlterField(
+            model_name='document',
+            name='tags',
+            field=taggit_selectize.managers.TaggableManager(blank=True, help_text='A comma-separated list of tags.', through='taggit.TaggedItem', to='taggit.Tag', verbose_name='Tags'),
+        ),
+        migrations.AlterField(
+            model_name='fragment',
+            name='is_multifragment',
+            field=models.BooleanField(default=False, help_text='True if there are multiple fragments in one shelfmark', verbose_name='Multifragment'),
+        ),
```
thatbudakguy, line 34: this stuff seems unrelated to this PR. could we add it to `develop` instead, if the migrations need updating?

thatbudakguy (commented):
using a choice field might be overkill but it does leave us the option of easily adding other statuses, which is probably why we did it in PPA. seems fine to do here too!

i think the only test you need is basically a quick test of `is_public()` – just create one model in the test, check `is_public()` on the default value, change the value of `status` and confirm that `is_public()` also changes.

as a rule of thumb, we don't test things that Django handles 100% (model fields, admin config, etc.) but anytime we add extra logic (usually via methods to models, views, etc.) that usually needs testing.

thatbudakguy:
and yeah, `is_public` doesn't really "mean" anything right now, so no other work needed! eventually we'll need to respect it when we build a public site, but as far as this issue is concerned, it's implemented.

Issue #77, opened 2021-02-16
As a content editor, I want to be able to flag and annotate a fragment as needing examination by a global admin, so that the global admin can either resolve the issue or submit a request for changes to the CDH. 


sluescher:
Same idea as in #62 but for fragments.

## testing notes
- [x] edit a few fragments, add text to the 'needs review' field, and save
- [x] navigate to admin main index page: the fragments you edited should be listed as awaiting review
- [x] you should be able to click on the fragment to go straight to the edit page
- [x] you should be able to click on the fragment heading to go to a filtered fragment list view showing all fragments that need review
- [x] navigate to the corpus section of the admin site; confirm that the awaiting review section appears and functions the same as on the admin index page
- [x] create a test staff user account without permissions to view or edit fragments and login as that user to confirm that they do not see the list of fragments awaiting review

## dev notes
implementation is the same as #62

thatbudakguy:
@mrustow and others - given discussion on #78; could we think more about this one? under what circumstances would a fragment need review (rather than a document)? consider that if several documents on a single document needed to be marked for review, you could:
- go to the document list in the admin
- filter/search to limit to documents with the shelfmark of the fragment in question
- proceed to mark each of the relevant documents as needing review

if you already have some workflows in place for reviewing whole fragments at a time, this issue isn't hard to implement, but want to make sure that there's a unique use case separate from #62.

richmanrachel:
@thatbudakguy - I think because fragments are connected to shelfmarks and therefore to libraries, this type of flagging is still important so that a Content Editor can let the Admin know if a library has changed names, therefore making the shelfmark out of date. 

I agree with the discussions from #78 that we should icebox suppressing fragments, but I think flagging fragments as a whole is still useful.

thatbudakguy:
makes sense; thanks for explaining! @kmcelwee let's keep the fragment-related work in #102.

richmanrachel:
It's possible to add text to "Needs Review" and it shows up when done by Admin, but Content Editor flags are not showing up. Even for Content Editor, they only see Admin "needs review" and not their own.

Other things with checkmarks work.

But I'm confused by all the user permissions and don't know which groups/things to add to make the new staff detail work. Please help me, @rlskoeser or @thatbudakguy!

rlskoeser:
One clarification — anyone who can see the 'needs review' list will see _any_ document that needs review, no matter who added the note. 

I'm not sure what you mean by "content editor flags".

[...]

Issue #62, opened 2021-02-09
As a content editor, I want to be able to flag and annotate a document as needing examination by a global admin, so that the global admin can either resolve the issue or submit a request for changes to the CDH.


sluescher:
New field: needs_review text field
Ability to create a link to a filter search of things that need review on the dashboard

## testing notes
- [x] edit a few documents, add text to the 'needs review' field, and save
- [x] navigate to admin main index page: the documents you edited should be listed as awaiting review
- [x] you should be able to click on the document to go straight to the edit page
- [x] you should be able to click on the document heading to go to a filtered document list view showing all documents that need review
- [x] navigate to the corpus section of the admin site; confirm that the awaiting review section appears and functions the same as on the admin index page
- [x] create a test staff user account without permissions to view or edit documents and login as that user to confirm that they do not see the list of documents awaiting review

## dev notes 
- [x] add `needs_review` text field to `Document`; help text to indicate what it is for
-  [x] add empty/nonempty filter on `needs_review` to document list view
- [x] create new template snippet/include with a brief list of 10 most recent (based on last modified) documents with nonempty needs review; should report how many items total need review and link document list view filtered to see all items that need review
     * (Kevin):`Document.objects.exclude(needs_review='').order_by('-last_modified')[:10]`
- [x] extend corpus `app_index` admin template and include the document review snippet in the sidebar (see https://docs.djangoproject.com/en/3.1/ref/contrib/admin/#templates-which-may-be-overridden-per-app-or-model)
- [x] extend admin index and add document review in the sidebar after other sidebar content

May want to implement #77 at the same time, since the functionality is the same but for fragments

kmcelwee:
In the latest geniza meeting, the PGP team preferred the `needs_review` text to the "Last modified" text. This was updated: commit 3c8bbe06fd901d5285443e57191bc4652bc0904b

richmanrachel:
> you should be able to click on the document heading to go to a filtered document list view showing all documents that need review
- Works, but with long chain of documents, it ends up listing several times (one for each shelfmark in the document name).

Everything else works, except the test staff account, per #77

rlskoeser:
@richmanrachel yes, I noticed the duplicate document problem! It's because we're sorting documents by shelfmark, which is a many-to-many relationship with fragments. I made a new issue to address that #113

PR #102, opened 2021-03-15
Feature/62 document needs review

kmcelwee:
(Merging on branch `feature/63` since I wanted to keep migrations in order and I'm working on this alone.)


kmcelwee:
Screen Shot 2021-03-15 at 9 53 27 PM

geniza/admin.py
```diff
@@ -0,0 +1,21 @@
+from django.contrib import admin
+from geniza.corpus.models import Document, Fragment
+
+class GenizaAdminSite(admin.AdminSite):
+    '''A customized AdminSite to aid PGP workflow'''
+    def each_context(self, request):
+        '''Provide extra context dictionary to admin site'''
+        REVIEW_PREVIEW_MAX = 10
```
thatbudakguy, line 8: I'd define this at the level of `GenizaAdminSite` so it doesn't get re-declared on each call to `each_context()` (which we think happens every pageload in the admin).

geniza/admin.py
```diff
@@ -0,0 +1,21 @@
...
+
+        context = super().each_context(request)
+
+        docs_need_review = Document.objects.exclude(needs_review='').order_by('-last_modified')
+        context['docs_review_count'] = len(docs_need_review)
+        context['docs_need_review'] = docs_need_review[:REVIEW_PREVIEW_MAX]
+
+        fragments_need_review = Fragment.objects.exclude(needs_review='').order_by('-last_modified')
+        context['fragments_review_count'] = len(fragments_need_review)
+        context['fragments_need_review'] = fragments_need_review[:REVIEW_PREVIEW_MAX]
```
thatbudakguy, line 18: as written, I think #62 only covers documents. let's hold off on implementing review functionality for fragments (even though it might seem easy) given our conversation at iteration planning about not overcomplicating things – i think we need to learn more about the team's fragment workflow first.

geniza/corpus/migrations/0007_fragment_needs_review.py
```diff
@@ -0,0 +1,18 @@
...
+        ('corpus', '0006_document_needs_review'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='fragment',
+            name='needs_review',
+            field=models.TextField(blank=True, help_text='Enter text here if an administrator needs to review this fragment.'),
+        ),
+    ]
```
thatbudakguy, line 18: same comment here as above; let's wait on the fragment portion.

geniza/corpus/models.py
```diff
@@ -106,6 +106,8 @@ class Fragment(models.Model):
         'Multifragment', default=False,
         help_text='True if there are multiple fragments in one shelfmark')
     notes = models.TextField(blank=True)
+    needs_review = models.TextField(blank=True, 
+        help_text='Enter text here if an administrator needs to review this fragment.')
```
thatbudakguy, line 110: same comment here

thatbudakguy (changes requested):
fantastic work making this look good! it's exactly how i hoped it would appear. imo only barrier to merging is removing the fragment stuff, since that isn't technically a requested feature and we may not need it.

some suggestions if you want to go further with the appearance of the menu in the admin (not required):
- you could remove the "there are X documents awaiting review" text and simply add the counter number next to the "Documents" heading, so it would say e.g. `Documents (15)`
- you could make the "Documents" heading a link to open the documents admin list with the "needs review" filter already applied, using the same URL forming technique we used for the usage count on language/script
- you could use django's built-in [`pluralize`](https://docs.djangoproject.com/en/3.1/ref/templates/builtins/#pluralize) templatetag to avoid "1 documents"

thatbudakguy (commented):
based on discussion on #77 we can keep the fragment-related logic. once this has tests for the `each_context()` method, we can merge it.

kmcelwee:
@thatbudakguy testing added and here's the new view:
Screen Shot 2021-03-16 at 1 31 41 PM

thatbudakguy (approved):
looks good!

rlskoeser:
@kmcelwee this looks great — thanks for including the screenshots.  One small thing we forgot (I didn't think of it when I wrote up my notes) is to check logged in user permissions before displaying the list. I think users should at least have document/fragment view permissions, but maybe more than that. I'll ask the team.

PR #103, opened 2021-03-16
Feature/74 track historic shelfmarks

kmcelwee:
Things I'd especially like feedback about:
* I can see us going beyond 255 chars for old shelfmarks. I increased to 500, but a textfield seems excessive.
* I prevented a value from being equal to both shelfmark and in old_shelfmarks.
* I'm using `set()` to easily grab unique values which means that `';'.join(old_shelfmarks)` will be in random order. Should I be concerned?


geniza/corpus/models.py
```diff
@@ -144,6 +147,18 @@ def iiif_thumbnails(self):
...
         ))
 
+    def save(self, *args, **kwargs):
+        '''Remember how shelfmarks have changed by keeping a semi-colon list 
+        in the old_shelfmarks field'''
+        if self.pk and self.has_changed('shelfmark'):
+            if self.old_shelfmarks:
+                old_shelfmarks = set(self.old_shelfmarks.split(';'))
+                old_shelfmarks.add(self.initial_value('shelfmark'))
+                self.old_shelfmarks = ';'.join([sm for sm in old_shelfmarks if sm != self.shelfmark])
```
thatbudakguy, line 157: random thought: could you do `';'.join(old_shelfmarks - {self.shelfmark})`? seems a little clearer and might be slightly faster

thatbudakguy (approved):
everything looks good here, great work! about your questions...
- increasing the limit on the textfield seems prudent, but I'm not sure there will be cases over 255, let alone 500...let's ask rachel today.
- ensuring uniqueness seems like a good idea, and I like your usage of `set()` – it might even enable some more concise syntax.
- i don't think order will matter for this field.

geniza/corpus/models.py
kmcelwee: I don't know about faster, but it certainly is more elegant! I'll commit this and merge

PR #104, opened 2021-03-17
Chore/99 update dbdocs

kmcelwee:
Nothing really fancy here, just following documentation from https://dbdocs.io/guide, section 8. The build was successful: https://dbdocs.io/princetoncdh/geniza

### DIFF
--- a/.github/workflows/dbdocs.yml
+++ b/.github/workflows/dbdocs.yml
@@ -3,7 +3,7 @@ name: dbdocs
 on:
   push:
     branches:
-      - develop  # dev for now! switch to main later
+      - chore/99-update-dbdocs  # dev for now! switch to main later
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [ develop ]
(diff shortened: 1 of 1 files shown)


.github/workflows/dbdocs.yml
```diff
@@ -17,7 +17,7 @@ env:
 
 jobs:
   python-unit:
-    name: Python unit tests
+    name: Publish dbdocs
```
thatbudakguy, line 20: can we change the label here too? `python-unit` seems wrong

.github/workflows/dbdocs.yml
```diff
@@ -28,7 +28,9 @@ jobs:
         with:
           node-version: ${{ env.NODE_VERSION }}
       - name: Install dbdocs
-        run: npm install dbdocs
+        run: sudo npm install -g dbdocs
```
thatbudakguy, line 31: `sudo` is normally discouraged when using `npm` (for the same reason you never `sudo pip install ...`). does dbdocs require this?

thatbudakguy: just saw that they say it "may be required". can we try without? it's a CI runner so maybe not too much impact of `sudo install`ing it, but I'd rather not if we don't have to

kmcelwee: It was still successful after removing `sudo`. I also updated the label. Good to merge?

thatbudakguy (approved):
perfect, let's merge it!

Issue #113, opened 2021-03-22
fix document ordering so we don't get duplicate entries in django admin list view when sorting by shelfmark


rlskoeser:
we're currently sorting on a many to many field, which results in one entry for each shelfmark, so documents that are joins are repeated

rlskoeser:
I confirmed that this was working with the needs review queue that showed the problem so clearly before I re-ran the test import (which wiped the needs review test comments).  @richmanrachel do you want to check this to confirm, or ok to just close?

rlskoeser:
Confirmed in QA with help from @richmanrachel that this is working as expected.

PR #116, opened 2021-03-31
Revise document ordering to address duplicates in django admin #113

rlskoeser:


thatbudakguy (commented):
gnarly! could we just sort by shelfmark and then PGPID? or is this problem caused by trying to sort "through" the `TextBlock`? i don't think i totally understand what's going on.

rlskoeser:
> gnarly! could we just sort by shelfmark and then PGPID? or is this problem caused by trying to sort "through" the `TextBlock`? i don't think i totally understand what's going on.

The problem is that we're sorting by a value on a many-to-many relationship! I think it probably doesn't matter if we include the through model or not (I was thinking it would include the textblock order, but IDK why I was thinking that, it can't). In the document list in django admin, you get the same record showing up multiple times because of this multivalued sort. (Weirdly, it reports the correct total number of results. Also, distinct doesn't help any, AFAICT).  The duplicate record problem was most obvious when we were testing the "needs review" functionality, but I think it's probably happening any time we filter or search and bring back documents associated with multiple fragments.  (The document shows up once for each fragment!)

fwiw, we had a similar duplicate record issue in mep somewhere; I think we just decided to live with it there, but I don't think we had things showing up more than twice.

PR #118, opened 2021-04-01
Revise document language import

rlskoeser:
- revise to map languages to Language+Script based on display name or newly added spreadsheet_name (needed for a few cases)
- case-insensitive comparison to handle variation in the spreadsheet

### DIFF
--- a/geniza/corpus/management/commands/import_data.py
+++ b/geniza/corpus/management/commands/import_data.py
@@ -145,7 +145,7 @@ class Command(BaseCommand):
         languages = LanguageScript.objects.bulk_create([
             LanguageScript(language=row.language,
                            script=row.script,
-                           display_name=row.display_name)
+                           display_name=row.display_name or None)
             for row in language_data if row.language and row.script
         ])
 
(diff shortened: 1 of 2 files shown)


geniza/corpus/management/commands/import_data.py
```diff
@@ -142,19 +144,26 @@ def import_collections(self):
...
+            lang = LanguageScript.objects.create(
+                language=row.language,
+                script=row.script,
+                display_name=row.display_name or None)
+
+            # populate lookup for associating documents & languages;
+            # use lower case spreadsheet name if set, or display name
+            if row.display_name or row.spreadsheet_name:
+                self.language_lookup[(row.spreadsheet_name or
+                                      row.display_name).lower()] = lang
```
kmcelwee, line 162: oooo okay. Yeah this gives me a bit of a headache, but I agree this is the best way to handle it 🥴

kmcelwee (approved):
Looks good! Wish there was an easier way, but the complexity seems necessary 🤷

Issue #109, opened 2021-03-18
As a Content Editor I want to link a source to a document as a footnote, in order to show that the source is helpful for understanding the document.


thatbudakguy:
**Is your feature request related to a problem? Please describe.**
This depends on #96 and represents a relationship between a document and the scholarship model created in that issue.

## Testing Notes
- [x] Go to a document and create a footnote on that document (Make sure that you've already created a source). Click save ensure that footnote is attached to the document.
- [x] Create at least two footnotes with different document relationships and one footnote with more than one document relationship selected
- [x] Confirm that the footnote display on the document looks appropriate.
- [ ] Go to the Footnote admin list view. Confirm that the list page and the change page looks appropriate.
- [x] Test that you can filter correctly on document relation, including footnotes with more than one

## Dev Notes

- [x] Create a footnote model
   * Fields: Generic Foreign Key, ForeignKey(Source), Document Relation field, Notes, page range
- [x] Register footnote model in admin
- [x] Include footnote in document model as generic foreign key
- [x] Place inline footnote in document admin change view
- [x] Update permissions to include footnote model

mrustow:
I tried to link an existing source to a document. I searched for the source by author and it came up: yay! I chose it and added page numbers. But then, I hit save, and got an error message! Egads!
<img width="952" alt="Screen Shot 2021-04-09 at 4 47 58 PM" src="https://user-images.githubusercontent.com/73319225/114239022-82c1b400-9953-11eb-8f70-f3b7f21efd93.png">
<img width="953" alt="Screen Shot 2021-04-09 at 4 48 09 PM" src="https://user-images.githubusercontent.com/73319225/114239023-835a4a80-9953-11eb-8715-158a7e7c08c4.png">

rlskoeser:
I can't tell from the screenshot what the error is — what field is it complaining about? Do we have something set to required that shouldn't be?

mrustow:
I can't tell either. It's not pointing to a field, just complaining, kind of like our kids.

richmanrachel:
I was able to add a footnote as Content Admin, but Marina is still getting errors as Content Editor.

mrustow:
I tried it as content editor. I also tried it with and without footnotes. Then I tried adding a second source. No dice.

richmanrachel:
Feature idea: being able to see all the documents a source is related to from the source page.

rlskoeser:
[...]

Issue #96, opened 2021-03-10
As a content editor, I want to create and edit scholarship records so that I can keep track of relevant scholarship on documentary geniza fragments.


richmanrachel:
## Testing Notes
In the **footnotes** section in the admin, as a **Content Admin**:
- [x] Check that the provided list of **Source types** have been created
- [x] Add a new **Creator**
- [x] Add a new **Source Language**
- [x] Add a new **Source**. Confirm that the admin list display and edit fields are appropriate.
- [x] Enter values for the multiple languages in a Source. Confirm that it works as expected.
- [x] Confirm that you can add multiple authors to a Source and order them (drag using the gray bars on the left of the creator row after saving)
- [x] Add a new **Source type**. Confirm that the admin display is appropriate.
- [x] In the Source list, confirm you can sort by author and that it sorts by first author when there are multiple

As a **Content Editor**:
- [x] confirm you can view all of the source records (creator, language, type, source)
- [x] confirm you can add and edit creators, source types, source languages, sources
- [x] confirm you cannot delete any of the above

---

**Fields**
Provide a list of the fields to be included on this model. 

- Sourcetype (book, article, unpublished, dissertation, web/blog?)
- Author(Last name, First name)
- Title
- Date
- Edition Number (i.e. Gil #45)
- Volume
- Page range
- Edition/Translation/Discussion tickbox (mandatory to input at least one value, but allow all three)
- Language of the source

**List Fields**
Which fields should be included on the django admin list view?

- Author
- Title
- Date (year only)
- Edition Nr.
- Edition/Translation/Discussion (boolean)

**Edit Fields**
Which fields should be editable via django admin view? Please list them in the order you want them to appear. Indicate which fields are optional, and any fields that should be displayed but not editable.

- Sourcetype (book, article, unpublished, dissertation, web/blog?)
- Author
- Title
- Date (Month optional, Year optional)
- Edition Number (i.e. Gil #45) (optional)
- Volume (optional)
- Page range (optional)
- Edition/Translation/Discussion tickbox? (at least 1)
- Language of the source (at least one)

**Search Fields**
Which fields should be searchable in the django admin list view?

- Author
- Title
- Date (Year)

**Default sort**
Which field should be used for initial sorting?

- Author (Last name)

**Filters**
Which fields should be used for list filters?

- Edition
- Translation
- Discussion

**List Fields (optional)**
Which fields (if any) should be editable on the django admin list view?

[...]

PR #119, opened 2021-04-01
Feature/96 scholarship records

kmcelwee:
A lot is covered here! This addresses #96 and #109.

Two main remaining things to consider:
* How we are handling people? This is being discussed in the upcoming meeting
* How to test the custom Django filter. I've skipped it for now because I've had trouble replicating the admin functionality. It seemed like a big time sink. I can return to it while waiting for a review.


geniza/footnotes/models.py
```diff
@@ -0,0 +1,68 @@
...
+    # TODO: (RR) Should there be a default language?
+    language = models.ForeignKey('corpus.LanguageScript', on_delete=models.SET_NULL,
+        help_text='In what language was the source published?', null=True)
+
+    class Meta:
+        ordering = ['author__sort_name']
+
+    def __str__(self):
+        # TODO: Generate citation
+        return self.title
```
rlskoeser, line 39: I think a short citation similar to the display in the metadata spreadsheet would be fine for now; no need to worry about generating a formatted citation

geniza/footnotes/migrations/0005_add_permissions_source_sourcetype_footnotes.py
```diff
@@ -0,0 +1,77 @@
...
+CONTENT_ADMIN = 'Content Admin'
+# additional new permissions for content admin
+content_admin_perms = [
+    'delete_sourcetype',
+    'delete_source',
+    'delete_footnote',
+]
+
+
+def create_content_editor_groups(apps, schema_editor):
```
rlskoeser, line 40: Now that I see it here, it makes a lot more sense for the migration for footnote-specific permissions to be in the footnotes app than in the common one. 

Is it worth revising the documents-fragments permissions migration to follow that convention?

Also, I wonder if we want a migration utility method for adding permissions...

geniza/footnotes/models.py
```diff
@@ -0,0 +1,68 @@
...
+        return self.type
+
+
+class Source(models.Model):
+    # TODO: Multiple authors
+    author = models.ForeignKey(Person, on_delete=models.CASCADE)
+    title = models.CharField(max_length=255)
+    year = models.PositiveIntegerField(blank=True, null=True)
+    # TODO: `edition_number` isn't a number. Can we call it `edition` ?
+    edition_number = models.CharField(max_length=255, blank=True)
```
rlskoeser, line 23: just edition makes sense to me; we can always give it a different verbose name if they want it labeled with more/different detail

geniza/footnotes/models.py
```diff
@@ -0,0 +1,68 @@
...
+    # TODO: What kind of values would volume contain? Would they just be an int?
+    volume = models.CharField(max_length=255, blank=True)
```
rlskoeser, line 25: I don't think we can assume integer here, let's keep it agnostic for now

geniza/footnotes/views.py
```diff
@@ -0,0 +1,3 @@
+from django.shortcuts import render
+
+# Create your views here.
```
rlskoeser, line 3: let's delete —  I don't expect we'll have any views in the footnotes app

geniza/footnotes/models.py
```diff
@@ -0,0 +1,68 @@
...
+
+class SourceType(models.Model):
+    type = models.CharField(max_length=255)
+
+    def __str__(self):
+        return self.type
+
+
+class Source(models.Model):
+    # TODO: Multiple authors
```
rlskoeser, line 18: Can we use https://github.com/jazzband/django-sortedm2m for authors?

rlskoeser (commented):
> How we are handling people? This is being discussed in the upcoming meeting

Based on today's discussion, we should move our person model into the footnote model and rename it; I suggest `author` or `creator`. We want separate fields for first and last names.

I think we should have URIs for authors/creators eventually (e.g. to document VIAF and/or ORCID), but IDK if we need to add them now.

I also think we may want a 1-1 from author to django user, since we know there will be overlap (e.g. Alan, Marina); not sure if we should add it now or hold off.

Is there any more configuration needed for the model translation? Do we need to define the languages in settings or anything like that?

> How to test the custom Django filter. 

Here's the code where I tested an admin list filter in the S&co codebase, is it helpful at all?
https://github.com/Princeton-CDH/mep-django/blob/main/mep/people/tests/test_admin.py#L116 

If you don't have time to get to this one, I can pick it up.

kmcelwee:
@rlskoeser  I got that admin filter test working, the people app deleted and the new creator app created, collapsed the migrations, and allowed for multiple authors.

Remaining:
* Separate out the migration utility
* How do you sort sources by authors when it's a many to many field? It wouldn't accept my new `all_authors` property

I can write testing notes on Monday if you’d like, just let me know. And of course, you’re welcome to take over this branch and apply your own commits if there are any changes you wanted to see.

thatbudakguy (approved):
> Is there any more configuration needed for the model translation? Do we need to define the languages in settings or anything like that?

re: this question; it uses the defined `LANGUAGES` for i18n that you have in settings.py, which I think is the right values currently (english, hebrew, arabic). if you want you can change it to have different values from the ones the site is localized into, but i don't think we need to do that here.

rlskoeser:
Updated and made some changes:
- switched from sortedm2m to using a local through model with adminsortable2 to make the admin interface nicer (sortedm2m doesn't seem to be a proper m2m and doesn't want you to access the through model)
- figured out a way to sort by first author (requires queryset annotation, so not using as default)
- decided source language should be separate from Language+Script to avoid circular dependencies and because source languages should be much simpler
- shortened the variable name for document relationship

Not worrying about the permissions migration utility for now, this already took longer than I hoped and it's not urgent/required.

geniza/footnotes/admin.py
```diff
@@ -1,53 +1,81 @@
...
     list_display = (
-        'all_authors', 'title', 
+        'title', 'all_creators',
         'year',
         'edition',
     )
 
     search_fields = (
-        'title', 'all_authors', 'year'
+        'title', 'creators__first_name', 'creators__last_name' 'year'
```
thatbudakguy, line 30: is this missing a comma before `'year'`?

geniza/footnotes/migrations/0001_initial.py
```diff
@@ -1,21 +1,29 @@
-# Generated by Django 3.1.7 on 2021-04-01 22:12
+# Generated by Django 3.1.6 on 2021-04-07 13:20
```
thatbudakguy, line 1: why'd the minor version go down? that seems odd

geniza/footnotes/models.py
```diff
@@ -3,55 +3,92 @@
...
+            models.UniqueConstraint(fields=['first_name', 'last_name'],
+                                    name='creator_unique_name')
+        ]
+
+
+class Authorship(models.Model):
+    """Ordered relationship between :class:`Creator` and :class:`Source`."""
+    creator = models.ForeignKey(Creator, on_delete=models.CASCADE)
+    source = models.ForeignKey('Source', on_delete=models.CASCADE)
+    sort_order = models.PositiveSmallIntegerField(default=0)
```
thatbudakguy: Just to make sure I understand – this through-model creates an ordered relationship, but it also implies a nature to the relationship, right? It's only for authors, otherwise it'd be called `Creatorship`. And similarly, if we had editors or other kinds of creators, we'd need separate through-models for associating them, since there's no concept of "creator type" right now.

thatbudakguy (approved):
everything looks good!

geniza/footnotes/models.py
rlskoeser: I named it "creatorship" at first but that sounded terrible! (is that even a real word?) I didn't intend to imply that this was for authors only. Should I rename it back? 

Or should we ask what kind of creators the team cares about, and whether we should differentiate them? (🤔 They might only care about authors right now; and generating a citation properly requires we know which role a creator has...)

geniza/footnotes/migrations/0001_initial.py
rlskoeser: 😆 wow, eagle eyes. Kevin must have a newer version installed than I do!

geniza/footnotes/admin.py
rlskoeser: um yeah, how's that not throwing an error? thanks for catching

geniza/footnotes/models.py
rlskoeser: I looked at the user story again, and they explicitly requested author. This seems like a reasonable solution, doesn't it? — generic creators, but only define authorship for now and we can create more through models if there are other creator types we care about

thatbudakguy: Sounds good to me!

PR #128, opened 2021-04-08
Update document id sequence to max pgpid after document import

rlskoeser:


thatbudakguy (approved):
pretty cool! nice to know how to do this if we need to again.

Issue #112, opened 2021-03-18
As a Content Editor, I want to see who first input a document and who last edited it, and when, so that I can ensure records are kept up-to-date.


thatbudakguy:
# testing notes
check any documents you'd like in the admin and scroll down to the bottom of the fields to see the initial entry/latest revision dates.

- it should show the date the document was first added to PGP as the first date, including the original date from the "date entered" column in the spreadsheet
- it should show the latest revision and who made it, along with what the revision was (in most cases this will probably be the import event itself).
- you should be able to click the link to get the full history (does the same thing as clicking the "history" button at the top right)
- if you change something about the document and save it (e.g. edit the description or add a footnote) the latest revision should update accordingly.

# dev notes
**Is your feature request related to a problem? Please describe.**
See also and depends on #111.

**Describe the solution you'd like**
On the document change view, a content editor would be able to see the first and last of the log entries created as part of #111, displaying both date and the user who made the change.

richmanrachel:
@thatbudakguy - All four of your asked for inquiries work!

The only thing that concerns me is that in the document I looked at (AIU IX.B.27; PGP ID 
30737), we have Alan's input of the document into the PGP, but Matthew Dudley (MCD) did not annotate the editor column even though he did "sign" his transcription. It makes me wonder how much of the data is uneven in this way... Will there be a way to manually add this to the backend later, or does all of this need to be fixed before import?

thatbudakguy:
if i understand correctly, you're saying that changes to transcriptions count as "major revisions" for the purposes of tracking changes to a document, so Matthew's change/addition of a transcription should be shown in the "full history" for this document – is that right?

also, where did you see that Matthew transcribed this document? is that information available in the spreadsheet?

richmanrachel:
I'm sorry, I wrote "transcription" instead of "description." It would be helpful for both transcription and description changes to be marked in the history, and this might be another argument to include transcription editing in our backend.

rlskoeser:
@richmanrachel I agree transcription history is important, but I don't think it belongs here — we can aggregate them for display, but logging should happen on the transcription record wherever we end up storing it.

There's no good way to edit log entries once they're created, so any important errors or missing data is best added to or corrected in the spreadsheet now.

richmanrachel:
Closing this issue for now, but we will raise the desired chore of parsing initials out of description into history at the meeting.

PR #129, opened 2021-04-08
Preserve spreadsheet input dates via log entry change messages

thatbudakguy:
Realized there was a way to preserve the actual information entered in the spreadsheet so that it could be displayed for #112 – planning on adding some logic for that one to display the actual date entered (e.g. "January 1996") if it's present in the log entry, otherwise use the actual date of the log entry (for changes made after the spreadsheet). lmk if that doesn't make sense.


rlskoeser (approved):
Great idea! Glad you thought of a way to preserve the original date

PR #130, opened 2021-04-08
Add first input and last revised dates to Document admin

thatbudakguy:
- Adds new admin templates for displaying data after fields
- Adds convenience field to Document to get its log entries
- Adds first/last log entries to admin context with legacy dates

See #112

no tests yet; opening draft to check the approach – it seems to be working for me locally

### DIFF
--- a/geniza/corpus/admin.py
+++ b/geniza/corpus/admin.py
@@ -143,34 +143,6 @@ class DocumentAdmin(admin.ModelAdmin):
             .annotate(shelfmk_all=ArrayAgg('textblock__fragment__shelfmark')) \
             .order_by('shelfmk_all')
 
-    def change_view(self, request, **kwargs):
-        """Add this document's first and last log entries to context.
-        
-        If there was a legacy date included in the log entry, show that info
-        instead of the timestamp of the entry itself.
-        """
-
-        # get first/last log entries; we want the reverse of their default
-        # ordering which is chronological descending
-        obj = Document.objects.get(pk=kwargs["object_id"])
-        first_input = obj.log_entries.last()
-        last_revision = obj.log_entries.first()
-
-        # check for "dated " message in the log entry and add legacy date.
-        # template will use the actual timestamp as a fallback if no legacy date
-        for log in first_input, last_revision:
-            try:
-                log.legacy_date = log.get_change_message().split("dated ")[1]
-            except IndexError:
-                continue
-
-        # add to context & return
-        kwargs["extra_context"] = {
-            "first_input": first_input,  
-            "last_revision": last_revision
-        }
-        return super().change_view(request, **kwargs)
-
     def save_model(self, request, obj, form, change):
         '''Customize this model's save_model function and then execute the
          existing admin.ModelAdmin save_model function'''
--- a/geniza/corpus/admin.py
+++ b/geniza/corpus/admin.py
@@ -124,7 +124,6 @@ class DocumentAdmin(admin.ModelAdmin):
         # edition, translation
         'notes',
         # text block
-        ('created', 'last_modified')
     )
     autocomplete_fields = ['languages', 'probable_languages']
     # NOTE: autocomplete does not honor limit_choices_to in model
@@ -143,6 +142,34 @@ class DocumentAdmin(admin.ModelAdmin):
             .prefetch_related('tags', 'languages', 'textblock_set')  \
             .annotate(shelfmk_all=ArrayAgg('textblock__fragment__shelfmark')) \
             .order_by('shelfmk_all')
+
+    def change_view(self, request, **kwargs):
+        """Add this document's first and last log entries to context.
+        
+        If there was a legacy date included in the log entry, show that info
+        instead of the timestamp of the entry itself.
+        """
+
+        # get first/last log entries; we want the reverse of their default
+        # ordering which is chronological descending
+        obj = Document.objects.get(pk=kwargs["object_id"])
+        first_input = obj.log_entries.last()
+        last_revision = obj.log_entries.first()
+
+        # check for "dated " message in the log entry and add legacy date.
+        # template will use the actual timestamp as a fallback if no legacy date
+        for log in first_input, last_revision:
+            try:
+                log.legacy_date = log.get_change_message().split("dated ")[1]
+            except IndexError:
+                continue
+
+        # add to context & return
+        kwargs["extra_context"] = {
+            "first_input": first_input,  
+            "last_revision": last_revision
+        }
+        return super().change_view(request, **kwargs)
 
     def save_model(self, request, obj, form, change):
         '''Customize this model's save_model function and then execute the
--- a/geniza/corpus/templates/admin/corpus/document/change_form.html
+++ b/geniza/corpus/templates/admin/corpus/document/change_form.html
@@ -3,12 +3,15 @@
 <fieldset class="module aligned">
     <div class="form-row">
         <div class="fieldBox">
-            <label>First input:</label>
-            {% include "admin/corpus/document/rev_log_entry.html" with log_entry=first_input %}
+            <label>Initial entry</label>
+            {% include "admin/corpus/document/log_entry.html" with log_entry=original.log_entries.last %}
+            <a href="{% url "admin:corpus_document_history" object_id=original.pk %}">view full history</a>
         </div>
+    </div>
+    <div class="form-row">
         <div class="fieldBox">
-            <label>Last revision:</label>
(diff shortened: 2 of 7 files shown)


geniza/corpus/admin.py
```diff
@@ -143,6 +142,34 @@ def get_queryset(self, request):
             .annotate(shelfmk_all=ArrayAgg('textblock__fragment__shelfmark')) \
             .order_by('shelfmk_all')
 
+    def change_view(self, request, **kwargs):
```
rlskoeser, line 145: I'm a little surprised you wrote a method for this, I was just imagining using the instance in the view to display information `log_entries.last()` and `log_entries.first()`. (although I guess I didn't say that! sorry)

geniza/corpus/admin.py
```diff
@@ -143,6 +142,34 @@ def get_queryset(self, request):
...
+        first_input = obj.log_entries.last()
+        last_revision = obj.log_entries.first()
+
+        # check for "dated " message in the log entry and add legacy date.
+        # template will use the actual timestamp as a fallback if no legacy date
+        for log in first_input, last_revision:
+            try:
+                log.legacy_date = log.get_change_message().split("dated ")[1]
+            except IndexError:
+                continue
```
rlskoeser, line 164: Not sure it is worth this effort — can we just display the date in the date field _and_ revision message? It seems like knowing what was changed would be useful in every case.

geniza/corpus/templates/admin/corpus/document/rev_log_entry.html
```diff
@@ -0,0 +1,4 @@
+<div class="readonly">
+<time>{% if log_entry.legacy_date %}{{ log_entry.legacy_date }}{% else %}{{ log_entry.action_time|date:"DATETIME_FORMAT" }}{% endif %}</time>
```
rlskoeser, line 2: For first input can we just display date without time? Maybe a flag to the template to specify not to show time. (Asking because it's odd to see the `midnight` in the historic info)

geniza/corpus/templates/admin/corpus/document/change_form.html
```diff
@@ -0,0 +1,15 @@
+{% extends "admin/change_form.html" %}
+{% block after_field_sets %}
+<fieldset class="module aligned">
+    <div class="form-row">
+        <div class="fieldBox">
+            <label>First input:</label>
```
rlskoeser, line 6: Earliest revision, maybe?

geniza/corpus/templates/admin/corpus/document/change_form.html
```diff
@@ -0,0 +1,15 @@
...
+            {% include "admin/corpus/document/rev_log_entry.html" with log_entry=first_input %}
+        </div>
+        <div class="fieldBox">
+            <label>Last revision:</label>
```
rlskoeser, line 10: Most recent revision? (seems clearer than last to me, somehow)

geniza/corpus/templates/admin/corpus/document/change_form.html
```diff
@@ -0,0 +1,15 @@
+{% extends "admin/change_form.html" %}
+{% block after_field_sets %}
+<fieldset class="module aligned">
```
rlskoeser, line 3: I think these should be on two separate rows with date, user, change message

geniza/corpus/templates/admin/corpus/document/change_form.html
```diff
@@ -0,0 +1,15 @@
...
+    <div class="form-row">
+        <div class="fieldBox">
+            <label>First input:</label>
+            {% include "admin/corpus/document/rev_log_entry.html" with log_entry=first_input %}
```
rlskoeser, line 7: couldn't you just pass `instance.log_entries.first` here and not need the view method?

geniza/corpus/templates/admin/corpus/document/change_form.html
```diff
@@ -0,0 +1,15 @@
...
+        </div>
+        <div class="fieldBox">
+            <label>Last revision:</label>
+            {% include "admin/corpus/document/rev_log_entry.html" with log_entry=last_revision %}
+        </div>
+    </div>
```
rlskoeser, line 13: Can we construct a link to the record history? I know it's in the top right, but seems like it would be convenient to add here (something like "view full history")

rlskoeser (commented):
Approach seems sound; just a few suggestions to simplify and provide more information

geniza/corpus/templates/admin/corpus/document/change_form.html
thatbudakguy: Can it even be a revision if it's the first one?

geniza/corpus/templates/admin/corpus/document/change_form.html
thatbudakguy: Latest?

geniza/corpus/templates/admin/corpus/document/change_form.html
rlskoeser: 😆 🤔 too early to be asking existential questions like that!

first version? initial entry?

geniza/corpus/templates/admin/corpus/document/change_form.html
rlskoeser: I think latest works

geniza/corpus/admin.py
thatbudakguy: Sure, I like that idea! It only really makes sense for last revision though, right? The message would always be the same for first input, since there's only one kind of "first input" event.

geniza/corpus/admin.py
thatbudakguy: Part of the reason I resorted to this was because I couldn't access the object in context for some reason. I assumed it'd be `object`, which didn't work, tried a few other things, and gave up when I found docs describing this method but nothing else. Advice?

geniza/corpus/templates/admin/corpus/document/rev_log_entry.html
thatbudakguy: Makes sense. We don't have that info for the entire corpus now, and even for newly added documents it probably doesn't matter as much as the date

geniza/corpus/admin.py
thatbudakguy: Oh, although I guess we might want to indicate ones that were first input into the spreadsheet vs. created in the database.

geniza/corpus/admin.py
rlskoeser: I'm not sure where exactly it's documented explicitly, but you can see from [this example](https://docs.djangoproject.com/en/3.1/ref/contrib/admin/#overriding-vs-replacing-an-admin-template) or from my fragment template that the context variable is `original`.

geniza/corpus/templates/admin/corpus/document/change_form.html
rlskoeser: here's the list of admin url patterns, which includes history: https://docs.djangoproject.com/en/3.1/ref/contrib/admin/#reversing-admin-urls

thatbudakguy: I'm trying to figure out a way to make this less visually awkward. would it be with the text of one of the fields? It almost seems easier to just move these fields to the top so they're next to the history button.

rlskoeser: Can you share a screenshot to show how it's awkward?

If it doesn't make sense we don't need the history link, just thought it would be convenient.

thatbudakguy: is this what you were imagining?

Screen Shot 2021-04-14 at 5 35 28 PM

rlskoeser: that looks reasonable to me

geniza/corpus/templates/admin/corpus/document/rev_log_entry.html
thatbudakguy: Switched to `naturaldate` for both, since:
- for old dates (i.e. most "first input" dates) it looks the same
- for newer dates, it's easier to know how recently something was modified
- it seems unlikely that records would have many edits per day, so the timestamp isn't as important
- if you need the timestamp, you can view the full history

lmk if this reasoning makes sense to you; we could also ask the team today.

rlskoeser: naturaldate is a great solution

rlskoeser (approved):
Looks great

PR #133, opened 2021-04-13
Feature/110 document detail

kmcelwee:
Fields from Gissoo's design that were stubbed / not included:
* Editors
* Cluster name

Screenshot of stub:
Screen Shot 2021-04-13 at 11 57 17 AM


thatbudakguy (approved):
looks good, just some suggestions. i realize this page is preliminary, but i'm also tempted to say we shouldn't miss the opportunity to make it translatable so that we get into that workflow, since otherwise i don't know when we will... @rlskoeser thoughts?

geniza/corpus/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.urls import path
+
+from geniza.corpus.views import DocumentDetailView
+
+app_name = 'corpus'
+
+urlpatterns = [
+    path('documents/<int:pk>/', DocumentDetailView.as_view(), name='document_detail'),
```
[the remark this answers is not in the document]
kmcelwee: @rlskoeser brought up a similar concern. Would the `reverse` function then take `corpus:document:detail`?

geniza/templates/corpus/document_detail.html
```diff
@@ -0,0 +1,67 @@
...
+    <div><b>First input</b>: {{ document.log_entries.last.action_time.year }}</div>
+
+    {% if document.last_modified %}
+    <div><b>Last modified</b>: {{ document.last_modified }}</div>
+    {% endif %}
+
+    <div><b>Editor</b>: [editor]</div>
+    <div><b>Cluster Name</b>: [cluster name]</div>
+
+    <div><b>Tags</b>: {{ tags }}</div>
```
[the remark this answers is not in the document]
kmcelwee: Oh fantastic, thank you! I knew this was janky but didn't know a cleaner way

rlskoeser (commented):
A few small comments. Biggest feedback is that I would like to see the html look closer to Gissoo's wireframes. 

Please don't include labels or content not in the design (including last modified, list of fragments).  The bolded field name: value in particular is a peeve of mine, see  [Don't Bold the Field Name!](https://www.slideshare.net/simeonwarner/dont-bold-the-field-name-2011)). 

This is a chance to ask questions about the design if you think something is missing; from my brief look I have questions about PGPID and what the page title should be, you probably have more!

I would prefer not to see output for missing/placeholder content (although comments in the template would be useful).  I would also prefer to leave out the placeholder labels that Gissoo hasn't settled yet ('e.g. "description label"), but we could ask her if she has a preference.

geniza/corpus/urls.py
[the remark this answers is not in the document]
thatbudakguy: ah I'm wrong, sorry! i guess that's how django namespaces it. so you would want `document-detail` as `name` and then the `reverse` would be `corpus:document-detail`

rlskoeser:
> i'm also tempted to say we shouldn't miss the opportunity to make it translatable so that we get into that workflow, since otherwise i don't know when we will...

@thatbudakguy thanks for raising — do we have any fields available for translation yet? I think we might not be quite ready for that (Gissoo hasn't designed it yet either), but I like the idea of adding an issue to support it on this view so we keep it in mind.

geniza/corpus/urls.py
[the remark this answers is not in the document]
thatbudakguy: really like this.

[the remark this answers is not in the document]
rlskoeser: we typically don't include `detail` in the url names — I propose `corpus:document` in this case (detail / single document view implied by the lack of other qualifier)

geniza/urls.py
```diff
@@ -18,14 +18,13 @@
 from django.urls import include, path
 from django.views.generic.base import RedirectView
 
-
 urlpatterns = [
     # redirect homepage to admin site for now
-    path('', RedirectView.as_view(url='admin/')),
```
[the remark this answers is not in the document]
thatbudakguy: @kmcelwee to prevent it overriding the corpus URLs you could do a `path` of `"/"` (or use `re_path` to specify string terminator if that doesn't work)

thatbudakguy:
> do we have any fields available for translation yet?

sorry, I did not make that clear – I meant localizing UI elements/template values like the field names using django's i18n capabilities, not (just) translatable data. since this is the first public page, it's the first time it's come up!

geniza/corpus/tests/test_corpus_views.py
```diff
@@ -0,0 +1,25 @@
...
+from geniza.corpus.models import Document
+from pytest_django.asserts import assertContains
+
+class TestDocumentDetailView:
+    def test_get_queryset(self, db, client):
+        # Ensure page works normally when not suppressed
+        doc = Document.objects.create()
+        response = client.get(doc.get_absolute_url())
+        assert response.status_code == 200
+        assertContains(response, "Shelfmark")
```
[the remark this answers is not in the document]
rlskoeser: but if you keep it, would prefer to see a value specific to the actual document

rlskoeser:
> > do we have any fields available for translation yet?
> 
> sorry, I did not make that clear – I meant localizing UI elements/template values like the field names using django's i18n capabilities, not (just) translatable data. since this is the first public page, it's the first time it's come up!

oooh — great point. yes, let's do that! we need to start getting used to that, best if we start now!

thatbudakguy:
fyi @rlskoeser we couldn't get `makemessages` to generate the `.po` files in the corpus app unless corresponding files already existed at the top level (`geniza/geniza`)...no idea why! that's the reason for the `.keep` files.

geniza/settings/components/base.py
```diff
@@ -64,7 +64,6 @@
 TEMPLATES = [
     {
         'BACKEND': 'django.template.backends.django.DjangoTemplates',
-        'DIRS': [BASE_DIR / 'geniza' / 'templates'],
```
rlskoeser, line 67: This change lost the top-level admin index and login override templates (for pucas login). Maybe they are in the wrong location and should be moved, but we need them included somewhere.

geniza/corpus/locale/ar/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,40 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
```
rlskoeser, line 4: Do we need to configure something so that these will be populated with meaningful values? Or are we expected to edit and change them manually?

thatbudakguy: I think you have to edit them manually, unfortunately.

thatbudakguy: OK, not quite – you have to [implement a custom management command as detailed here](https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#customizing-makemessages) so that you can control the [parameters that are passed to GNU `xgettext`](https://www.gnu.org/software/gettext/manual/html_node/xgettext-Invocation.html). Which might be worthwhile so that we can also ignore files we don't want to translate, which is kind of a problem on the command-line.

kmcelwee: I searched through [django's settings](https://docs.djangoproject.com/en/3.1/ref/settings/#std:setting-USE_I18N) options and couldn't find anything. It also looks like Apple's open source asks their translators to manually enter this information! So maybe this isn't a feature that's been implemented yet? https://opensource.apple.com/source/subversion/subversion-52/subversion/TRANSLATING.auto.html

Should we just ignore this for now?

kmcelwee: Ignore me! I hadn't refreshed the page when @thatbudakguy had made his comment

thatbudakguy:
I ended up adding a lot more things to this...wasn't sure when they would happen otherwise and seemed good to get it out of the way.

- base template that implements html5 boilerplate, with extendable `<head>` and placeholders for header, footer, etc.
- fixtures for creating basic `Document` cases for testing
- basic template tests for the doc detail page template
- custom mgmt command to ignore virtual environment and pass extra metadata when running `makemessages`
- favicons (just for fun; we can replace them with real ones later)

geniza/corpus/locale/ar/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,41 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR The Trustees of Princeton University
+# This file is distributed under the same license as the geniza package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
```
thatbudakguy, line 4: We don't appear to be able to change the wonderful `SOME DESCRIPTIVE TITLE` or the authors list. I think the authors might be something translators are expected to edit as they contribute to the file, since it's a record of who helped localize something. Theoretically tools that help with this process like [poedit](https://poedit.net/) would handle this part.

geniza/common/management/commands/makemessages.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+
+    # For invocation options for xgettext, see:
+    # https://www.gnu.org/software/gettext/manual/html_node/xgettext-Invocation.html
+    xgettext_options = makemessages.Command.xgettext_options + [
+        "--copyright-holder=The Trustees of Princeton University",
+        "--msgid-bugs-address=cdhdevteam@princeton.edu",
+        f"--package-version={__version__}",
+        "--package-name=geniza",
+    ]
```
rlskoeser, line 17: nice!

geniza/corpus/models.py
```diff
@@ -264,6 +265,14 @@ def is_public(self):
     is_public.boolean = True
     is_public.admin_order_field = 'status'
 
+    def get_absolute_url(self):
+        return reverse('corpus:document', args=[str(self.id)])
+
+    def iiif_urls(self):
+        """List of IIIF urls for images of the Document's Fragments."""
+        return list(filter(None, self.fragments.values_list("iiif_url",
```
rlskoeser, line 273: If we access fragments via `text_block` it should honor order (if any is specified). Might not matter for now, not sure.

rlskoeser (approved):
Looks great!

I was testing locally to see what it looked like with the IIIF viewer, and on a few records I got this error:
> SyntaxError: Unexpected token / in JSON at position 0

I wasn't sure if this was an error with our code or something on the server side; did you run into it? On reload it seemed to work ok, which suggests it's probably something on the IIIF server.

geniza/corpus/models.py
thatbudakguy: oh good call, thx for the reminder

thatbudakguy:
Hmm, I didn't run into that – agree that it does sound like a IIIF-server side problem given it's a JSON parse error. My guess is the link temporarily 404'ed or returned some error code, so it tried to parse an HTML error page as JSON and gave up.

PR #134, opened 2021-04-13
Add excel and csv export to document admin actions

kmcelwee:
Awaiting headers from PGP team. Currently, we're just exporting the admin list view columns. If necessary, writing a custom action seems straightforward: https://github.com/LibraryOfCongress/django-tabular-export#admin-integration

This library seems old, but smoke testing didn't reveal any problems.


geniza/corpus/admin.py
```diff
@@ -161,6 +166,46 @@ def save_model(self, request, obj, form, change):
...
+
+    export_fields = [
+        'id', 'shelfmark', 'description', 'doctype',
+        'all_tags', 'all_languages', 'all_probable_languages',
+        'last_modified', 'notes', 'needs_review', 'status'
+    ]
+
+    def csv_filename(self):
+        '''Generate filename for CSV download'''
+        return f'geniza-documents-{now().strftime("%Y%m%dT%H%M%S")}.csv'
```
rlskoeser, line 179: maybe should use pgp instead of geniza here (more specific)

rlskoeser (approved):
Everything looks good to me

geniza/corpus/admin.py
kmcelwee: In a recent conversation, you preferred geniza. I'm happy to change this, but would you mind confirming?

rlskoeser: ha! I don't remember that. this seems fine!

kmcelwee:
Questions as we introduce more complicated columns:

* This refactoring isn’t the most perfectly efficient approach, but I get the sense that structuring it this way is more clear and will be easier for us in the long run. Do you agree?
* The data analyst in my mind is annoyed that we're oversimplifying multi-valued columns. If there are two fragments connected to a document, and one has a `iiif_url` and one doesn't, we won't know which fragment the url belongs to because using `';'.join()` will just output one url. I doubt this simplification will matter in the long run, but just wanted confirmation. 
* `shelfmark_historic` is a doubly-concatenated semicolon delimited list. Is that okay?
> footnotes (i.e editions & translations; include source, location, notes, and document relationship; pending data import)
* Should I just integrate this into just the string function for footnotes?

rlskoeser:
thanks @kmcelwee for the questions and notes

>     * This refactoring isn’t the most perfectly efficient approach, but I get the sense that structuring it this way is more clear and will be easier for us in the long run. Do you agree?

I think some version of this is probably what we need, but would like to think a little more about it. (I thought there might be a way to do some hybrid between this and the other, but IDK how readable that would be.)

FWIW, the method does need to yield rows and not return — we can't return a streaming http response otherwise.

>     * The data analyst in my mind is annoyed that we're oversimplifying multi-valued columns. If there are two fragments connected to a document, and one has a `iiif_url` and one doesn't, we won't know which fragment the url belongs to because using `';'.join()` will just output one url. I doubt this simplification will matter in the long run, but just wanted confirmation.

I agree that this is not ideal, but you're not omitting empty results so the indexes should still all match up when there are multiple.

>     * `shelfmark_historic` is a doubly-concatenated semicolon delimited list. Is that okay?

ugh! not great. should we use a secondary delimiter? | or something?  I don't know that we expect anyone to do much computation based on this version, but we do want them to be able to find things.  Alternatively, we could push back on this one and ask if they really need it in this download.

> > footnotes (i.e editions & translations; include source, location, notes, and document relationship; pending data import)
> 
>     * Should I just integrate this into just the string function for footnotes?

🤔 it seems like that might be reasonable? I see that the current implementation includes document information, but IDK if we actually need it in the str method or not. When I added the editor information to the document detail page, it seemed like it would be nice to be able to use the string method, but in that case I need to not show the doc relation information. I think the source str method should be pretty usable now for the first part.

kmcelwee:
> we can't return a streaming http response otherwise.
I'd love to learn the implications of this!

>  you're not omitting empty results so the indexes should still all match up when there are multiple
I am omitting empty results (?)

rlskoeser:
> > we can't return a streaming http response otherwise.
> 
> I'd love to learn the implications of this!

With a streaming response, the download starts as soon as we start generating rows of data. Without, download doesn't start until _all_ the data has been collected and converted.

geniza/corpus/admin.py
```diff
@@ -180,6 +185,84 @@ def save_model(self, request, obj, form, change):
...
+        for doc in queryset:
+            all_fragments = doc.fragments.all()
+            all_textblocks = doc.textblock_set.all()
+            all_footnotes = doc.footnotes.all()
+
+            row = {
+                'pgpid': doc.id,
+                # 'url': # site link
+                'iiif_urls': ';'.join([fragment.iiif_url for fragment in all_fragments]),
+                'fragment_url': ';'.join([fragment.url for fragment in all_fragments]),
```
thatbudakguy, line 214: should this be `fragment_urls`?

thatbudakguy (approved):
this approach seems sound to me – having everything be explicit in `admin.py` is simpler  than spreading the implementation out across multiple models/methods.

geniza/corpus/admin.py
```diff
@@ -180,6 +185,84 @@ def save_model(self, request, obj, form, change):
...
+                # input_by*
+                # date_entered (first log entry date)
+                'last_modified': doc.last_modified,
+                'status': 'Public' if doc.status == Document.PUBLIC else 'Suppressed',
+                'library': ';'.join([fragment.collection.lib_abbrev for fragment in all_fragments]),
+                'collection': doc.collection
+            }
+
+            rows.append(row.values())
+            columns = row.keys()
```
kmcelwee, line 239: TODO: This method assumes that the `row` dict will not change order when calling `values` and `keys`. Python seems to honor this, but it's not entirely safe. Use 
```python
row_values, columns = row.items()
```
to ensure that order is maintained

kmcelwee:
@rlskoeser Before I start writing tests, would you mind glancing through this?

Things to note:
* skipping footnote parsing, since `__str__` is addressed in one of @thatbudakguy 's PRs
* input_by and last_modified were replaced by initial_entry and latest_revision timestamps and usernames

geniza/corpus/admin.py
```diff
@@ -207,47 +215,45 @@ def tabulate_queryset(self, queryset):
             all_textblocks = doc.textblock_set.all()
             all_footnotes = doc.footnotes.all()
 
-            row = {
+            initial_entry = doc.log_entries.first()
+            latest_revision = doc.log_entries.last()
+
+            row = self.DocumentRow(**{
```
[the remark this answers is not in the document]
thatbudakguy: @kmcelwee wanted to preserve a mapping between field names and values, but without duplicating the list of valid field names. `export_to_csv` needs the full list of field names to write the column headers, but since they were only defined in a literal in `tabulate_queryset` you would otherwise need to duplicate them. the namedtuple was my suggestion to define the list of valid fields once so that it's accessible in both `tabulate_queryset` and `export_to_csv`, as well as throwing an error if you try to assign a field that doesn't exist.

geniza/corpus/admin.py
thatbudakguy: fyi @kmcelwee i think the `NamedTuple` preserves order so we're good here

geniza/corpus/admin.py
```diff
@@ -207,47 +215,45 @@ def tabulate_queryset(self, queryset):
...
                 'language_note': doc.language_note,
                 'notes': doc.notes,
                 'needs_review': doc.needs_review,
-                # 'url_admin': '',#reverse('admin:corpus_document_change', args=[doc.id]),
-                # input_by*
-                # date_entered (first log entry date)
-                'last_modified': doc.last_modified,
+                'url_admin': absolutize_url(reverse('admin:corpus_document_change', args=[doc.id])),
+                'initial_entry': f"{initial_entry.action_time}, {initial_entry.user.get_full_name() or initial_entry.user.get_username()}",
+                'latest_revision': f"{latest_revision.action_time}, {latest_revision.user.get_full_name() or latest_revision.user.get_username()}",
```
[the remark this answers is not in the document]
kmcelwee: And I'm indifferent. Who wants to be decisive about this? 😄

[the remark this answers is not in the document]
rlskoeser: I've thought about it more, and I would like to see the separate fields that Marina and I agreed on. I think that will be easier to work with. So that means initial entry date, last modified, and input by (unique list of all data editors).

rlskoeser (approved):
Tests look fine to me. Could have used some of the fixtures, maybe, but this seems fine.

PR #143, opened 2021-04-23
Import editor & translator information

rlskoeser:


thatbudakguy (approved):
from what i can tell, this approach makes sense. only thought is whether it might be helpful at this point to do the same type of thing you did with known authors/creators for known titles. what you could do is:
- start with a fixture of the most common titles we know for sure, which would hopefully cover the majority of the corpus
- make the import script only assign sources if the title is in the known whitelist; error otherwise 
- use the logging/errors to generate a list of titles it's trying to create, which should be hopefully small enough to go over manually: real titles get added to the whitelist
- once only real errors/bad titles remain, ask the team to edit those cells to adjust the formatting of the citations to make them more regular so that the script can parse them

geniza/corpus/templates/corpus/document_detail.html
```diff
@@ -21,6 +21,16 @@ <h1>{{ document.title }}</h1>
         {% comment %}Translators: Date document was first added to the PGP{% endcomment %}
         <dt class="sr-only">{% translate 'First input' %}</dt>
         <dd>{{ document.log_entries.last.action_time.year }}</dd>
+
+        {% if document.editions %}
+        {% comment %}Translators: Editor label{% endcomment %}
+        <dt>{% translate 'Editor' %}</dt>  {# optionally pluralize? #}
+        {% for ed in document.editions %}
+        <dd>Ed. {{ ed.source }}{% if ed.notes %}. {{ ed.notes }}{% endif %}</dd>
```
thatbudakguy, line 29: this "Ed." seems redundant with the label (and harder to translate, maybe)

geniza/corpus/templates/corpus/document_detail.html
```diff
@@ -21,6 +21,16 @@ <h1>{{ document.title }}</h1>
         {% comment %}Translators: Date document was first added to the PGP{% endcomment %}
         <dt class="sr-only">{% translate 'First input' %}</dt>
         <dd>{{ document.log_entries.last.action_time.year }}</dd>
+
+        {% if document.editions %}
+        {% comment %}Translators: Editor label{% endcomment %}
+        <dt>{% translate 'Editor' %}</dt>  {# optionally pluralize? #}
```
thatbudakguy, line 27: plurals when translating are...interesting, and probably something we should tackle (or at least get exposed to) earlier rather than later. i'm fine with just using the plural always and having it be a one-element list though, either way

geniza/corpus/templates/corpus/document_detail.html
rlskoeser: good point! I was following their current conventions, but I agree that doesn't make sense here

geniza/corpus/templates/corpus/document_detail.html
```diff
@@ -21,6 +21,16 @@ <h1>{{ document.title }}</h1>
         {% comment %}Translators: Date document was first added to the PGP{% endcomment %}
         <dt class="sr-only">{% translate 'First input' %}</dt>
         <dd>{{ document.log_entries.last.action_time.year }}</dd>
+
+        {% if document.editions %}
+        {% comment %}Translators: Editor label{% endcomment %}
```
thatbudakguy, line 26: my brain exploded when i realized this was a django i18n comment to go in the `.po` file, because some of the things are that are about to be listed below _are_ translators

geniza/corpus/templates/corpus/document_detail.html
rlskoeser: yeah, I looked it up but I also wasn't sure how to reliably figure out when something had multiple editors. Multiple editions? Multiple authors on any of the editions? What if there is a single-author edition with corrections from some one else?

rlskoeser:
Doesn't handle all cases, including at least:
- partial translations
- translation into [language]  (does language need to be on footnote and not just source?)
- identifying content that should go into notes is probably not working in all cases

geniza/corpus/management/commands/import_data.py
```diff
@@ -571,3 +608,306 @@ def update_document_id_sequence(self):
...
+    re_page_location = re.compile(
+        r'[,.] (?P<pages>((pp?|pgs)\. ?\d+([-–]\d+)?)|(\d+[-–]\d+))\.?',
+        flags=re.I)
+    re_doc_location = re.compile(
+        r'(, )?\(?(?P<doc>(Doc. #?|#)([A-Z]-)?\d+)\)?\.?',
+        flags=re.I)
+    # \u0590-\u05fe = range for hebrew characters
+    re_goitein_section = re.compile(
+        r' (?P<p>(\d+?[\u0590-\u05fe]|[\u0590-\u05fe]\d+)[\u0590-\u05fe]?)',
+        flags=re.I)
```
thatbudakguy, line 659: nice!!! I was wondering if something like this was possible; really cool solution

thatbudakguy (approved):
everything here looks good; seems like it's just a matter of continuing to expand your regexes for handling the notes and then finding some way to handle whatever else falls out. i'd say we're getting pretty close to the point where you can ask the team to reformat cells in order to fit these regexes, if you need to.

geniza/corpus/management/commands/import_data.py
rlskoeser: yes! found a good resource online that had the range

PR #147, opened 2021-04-26
Add migration check

kmcelwee:
See that the tests passed on the second commit (bd195db) after adding the appropriate migrations. I then removed those migrations since they are being applied in other branches.

I've placed this in the unit tests so that I could have easy access to `python manage.py` but that means that this test would be run on every push. Should it be run less frequently?


thatbudakguy:
fwiw you can [make a single job step conditional](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsif) on pull request/PR with `if: ${{ github.event_name == 'pull_request' }}`

rlskoeser (approved):

PR #148, opened 2021-04-26
Add footnote_count function to SourceAdmin

kmcelwee:
Two concerns, please tell me if they should be addressed:
1. I haven't written a test for this, should I?
2. I've placed this function in the admin, not on `Source` because it seems like the application of this method is only useful in the admin. Is that accurate?

### DIFF
--- a/geniza/footnotes/admin.py
+++ b/geniza/footnotes/admin.py
@@ -4,7 +4,10 @@ from django.contrib.admin import SimpleListFilter
 from django.contrib.contenttypes.admin import GenericTabularInline
 from django.contrib.postgres.aggregates import ArrayAgg
 from django.db import models
+from django.db.models import Count
 from django.db.models.functions import Concat
+from django.urls import reverse, resolve
+from django.utils.html import format_html
 from modeltranslation.admin import TabbedTranslationAdmin
 
 from geniza.footnotes.models import Authorship, Creator, Footnote, Source, \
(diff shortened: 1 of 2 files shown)


rlskoeser:
>     2. I've placed this function in the admin, not on `Source` because it seems like the application of this method is only useful in the admin. Is that accurate?

Agree it should be in admin, but please use queryset annotation like we did for the `LanguageScript` admin rather than individual counts — otherwise it will be too slow once we have all the content imported.

>     1. I haven't written a test for this, should I?

A small test to check that the query is working as expected would be good — could do something like we did for language+script, or could just test the queryset annotation works as expected. (What we want is to catch it if we break it in the future, e.g. if we add some prefetching to optimize the query.)

To make this field sortable, you'll need to define the `admin_order_field` like we did for Language+Script.

geniza/footnotes/admin.py
```diff
@@ -43,8 +47,19 @@ def get_queryset(self, request):
...
+                                          'authorship__creator__first_name'),
+            )
+
+    def footnotes(self, obj):
+        return format_html(
+            '<a href="{0}?source__id__exact={1!s}">{2}</a>',
+            reverse(self.footnote_admin_url), str(obj.id),
+            obj.footnote__count
+        )
+    footnotes.short_description = "# footnotes on which this source appears"
```
rlskoeser, line 61: probably something more like "footnotes for this source"?

rlskoeser (approved):

PR #149, opened 2021-04-27
Add certainty field, update str logic, and include in admin

kmcelwee:
I made some decisions that I'm not 100 percent confident with, I'd appreciate your feedback!


geniza/corpus/models.py
```diff
@@ -234,7 +234,8 @@ def shelfmark(self):
         # access via textblock so we follow specified order,
         # use dict keys to ensure unique
         return ' + '.join(dict.fromkeys(block.fragment.shelfmark
-                          for block in self.textblock_set.all()))
+                          for block in self.textblock_set.all()
```
rlskoeser, line 237: suggest filtering on certain=true on the textblock queryset instead of the extra test

geniza/corpus/models.py
```diff
@@ -311,7 +313,8 @@ class Meta:
 
     def __str__(self):
         # combine shelfmark, side, and optionally text block
-        parts = [self.fragment.shelfmark, self.get_side_display(),
+        certainty_str = '(?)' if not self.certain else ''
```
rlskoeser, line 316: nice, I like this

geniza/corpus/models.py
```diff
@@ -285,6 +286,7 @@ class TextBlock(models.Model):
     '''The portion of a document that appears on a particular fragment.'''
     document = models.ForeignKey(Document, on_delete=models.CASCADE)
     fragment = models.ForeignKey(Fragment, on_delete=models.CASCADE)
+    certain = models.BooleanField(default=True)
```
rlskoeser, line 289: Should we add brief help text? Or do we need to solicit language from the team?

geniza/corpus/admin.py
```diff
@@ -63,7 +63,7 @@ class TextBlockInline(admin.TabularInline):
     autocomplete_fields = ['fragment']
     readonly_fields = ('thumbnail', )
     fields = ('fragment', 'side', 'extent_label', 'multifragment',
-              'order', 'thumbnail')
+              'order', 'thumbnail', 'certain')
```
rlskoeser, line 66: I think display before the thumbnail (input fields before display/context fields)

rlskoeser (approved):
Everything here looks reasonable to me (although fyi I didn't run it locally to check).

What was it you weren't confident in?

geniza/corpus/models.py
kmcelwee: Please double check that I understood you correctly: a68757b. If it looks good to you, feel free to merge!

kmcelwee:
No I was just uncertain about the general structure. Happy to hear I didn't misunderstand anything.

PR #150, opened 2021-04-27
Create SourceFootnotesInline

kmcelwee:


rlskoeser (changes requested):
This works in that it displays current footnotes with document information, but it doesn't actually serve the purpose requested, which is the ability to add new footnotes in batches for a single source. 

From the work you've done, it looks like this may require a custom form that will let us add a document lookup, and then a save method to set content object based on the document in the lookup.

geniza/footnotes/admin.py
```diff
@@ -28,6 +29,53 @@ class AuthorshipInline(SortableInlineAdminMixin, admin.TabularInline):
...
+            content_obj._meta.app_label,
+            content_obj._meta.model_name,
+        )
+        edit_path = reverse(edit_url, args=[obj.object_id])
+        return format_html(
+            f'<a href="{edit_path}">{content_obj} '
+            + '<img src="/static/admin/img/icon-changelink.svg" alt="Change"></a>'
+        )
+
+    object_link.short_description = "object"
```
kmcelwee, line 62: When extra is not defined, it defaults to 3. I prefer fewer, even 0, but I don't think this matters too much right?

geniza/footnotes/admin.py
```diff
@@ -39,7 +87,7 @@ class SourceAdmin(TabbedTranslationAdmin, admin.ModelAdmin):
     fields = ("source_type", "title", "year", "edition", "volume", "languages", "notes")
```
kmcelwee, line 87: (not related to this PR, surprised this got past QA) The SourceAdmin `list_display` has no clickable item where a user can then see that source's changeview. At least it's possible to construct a Source with fields that offer no way to select it from the list view

kmcelwee (commented):
@rlskoeser

geniza/footnotes/admin.py
thatbudakguy: I set it to 0 for some other things in admin, fwiw

rlskoeser: The whole point of this story is for them to easily add more footnotes for a single source, so we _want_ extra footnotes — I thought three was a reasonable number.

kmcelwee: that makes sense, getting into the weeds, it's easy to forget what we're actually trying to accomplish!

rlskoeser: For forms where we expect them to regularly add more related data, I don't think we should set it to zero, but I'm ok with setting it to a smaller number especially if the change form is already long (like Document)

geniza/footnotes/admin.py
rlskoeser: oh — you mean it's possible to create a source without an author, which would give you nothing to click?

I think current data should always have an author (although we may have a handful where we have lost the information and can't track it down), but I do remember runing into this. Nrmally django should provide a default like "-" — I wondered if it wasn't doing this because author is a model translation field...

rlskoeser:
It occurred to me I should just share a screenshot.

@kmcelwee or @thatbudakguy does this seem good enough for now as a first implementation? It looks a little weird, but the search icon under the id will bring up a browser window where you can browse and select documents.

Screen Shot 2021-05-13 at 11 53 38 AM

PR #152, opened 2021-04-28
Stub FragmentTextBlockInline

kmcelwee:
Problems so far: 
* Selecting `document__description` raises an error, but `document` doesn't. Why is that?
* I've created a method `document_link` so that the read-only field links to the change page for that document, but I can't reference it in the inline. Can you not define new methods for inlines?


rlskoeser (commented):
looks good!

Issue #146, opened 2021-04-26
As a content editor, I want a preliminary version of transcription text linked to a scholarly source so that I know which documents have available transcriptions.


rlskoeser:
# testing notes
**this implementation is provisional!** the actual way in which we handle transcriptions will likely be quite different, and the transcription text that you will see was generated by an experimental script. no content is expected to be correct – we're testing that things show up in the right places.

- there should now be a boolean filter on the document admin list that allows you to filter by documents with a digitized transcription, along with a check mark indicator for those that have one.
- if you check the admin edit page for any of these documents, you should see a section labeled "transcription" below the main metadata area that displays the (experimental) transcription text with very light formatting (the text is assumed to be a right-to-left script and aligned accordingly). it is not intended to be editable; the content is not stored as editable text.
- in the list of footnotes for the document further down the page, you should also see a checkmark indicator showing which of the footnotes is the edition from which the transcription is taken.
- if you go to the footnotes admin list, you should also have the ability to see and filter on which footnotes provide a digitized transcription.
- if you edit a footnote, you should see the same transcription display as you would on the document edit page.

richmanrachel:
@rlskoeser - I love it! This looks great (and answers one of the questions I had on #139)!

Issue #154, opened 2021-04-29
As a user, I want documents that span fragments with consecutive shelfmarks to have their shelfmark displayed using a range, so that it's easier for me to read.


thatbudakguy:
## testing notes
- in the admin site:
 - [x] open a document for editing and confirm the new shelfmark override field is displayed
 - [x] confirm shelfmark override field is optional (save changes without adding one)
 - [x] confirm you can add a custom shelfmark 
 - [x] check that the shelfmark override value you added is displayed in the document list view
- on the public site:
 - [x] on the document details page, check that the shelfmark override value you set is displayed for the title and and the shelfmark metadata
 - on the document search
  - [x] confirm you can find your document by searching for your shelfmark override value (keyword search)
  - [x] confirm you can find your document by searching for individual shelfmarks for the fragments associated with the document (keyword search)
  - [x] confirm you can search on shelfmark override in the shelfmark field `shelfmark:...`
  - [x] confirm you can search on individual shelfmark in the shelfmark field `shelfmark:...`

**Is your feature request related to a problem? Please describe.**
As raised by alan in [the april 29th meeting](https://docs.google.com/document/d/1ZzoZJBqsseFCROEtFnP26G3baeZYT4vJGFDucYB3-nw/edit#heading=h.w8r3weqyj6tk), we have some documents that occur on fragments that have consecutive shelfmarks. Aan example is `Bodl. MS Heb. d 66/49–50`, which is a join between `Bodl. MS Heb. d 66/49` and `Bodl. MS Heb. d 66/50`.

Currently, the data import will be creating a single fragment with a shelfmark of `Bodl. MS Heb. d 66/49–50`, based on what's in the spreadsheet. However, we want to more accurately model this as a document that occurs across multiple distinct fragments, since there's technically no single object with the shelfmark `Bodl. MS Heb. d 66/49–50`. The team has agreed to handle this problem by "demerging" these fragments with dashes in the shelfmark after the data import (see [corresponding asana task](https://app.asana.com/0/187294619140624/1200267553437360)).

On the display side, however, it's better to show the document shelfmark _with_ the dash, because it's easier to read and to see that the fragments are in fact consecutive. Otherwise, these documents may have shelfmarks that are quite long and difficult to read.

**Describe the solution you'd like**
~The method that calculates `Document.shelfmark` should take account of whether the document's fragments have consecutive shelfmarks. For any consecutive ranges, we should display them using an en-dash, which is the convention for noting a numeric range.~

[...]

PR #156, opened 2021-05-03
Preliminary transcription import (#146)

thatbudakguy:
No tests yet.

- Add setting for path to transcriptions json file
- Look up transcriptions from json during import_data command
- Make transcription text visible in footnote admin
- Allow filtering documents by presence of transcription


geniza/corpus/admin.py
```diff
@@ -126,6 +128,7 @@ class DocumentAdmin(admin.ModelAdmin):
         'doctype', 'languages',
         'probable_languages',
         'status',
+        ('footnotes__content', admin.EmptyFieldListFilter),
```
thatbudakguy, line 131: this works but is labeled kinda weirdly, is there a better way?

geniza/corpus/admin.py
```diff
@@ -157,6 +160,11 @@ class Media:
             'all': ('css/admin-local.css', )
         }
 
+    def has_transcription(self, obj):
+        return any([obj.footnotes.values_list("content", flat=True)])
```
thatbudakguy, line 164: this implementation also seemed weird to me, lmk if there's a better way

geniza/corpus/management/commands/import_data.py
```diff
@@ -127,7 +128,8 @@ def source_setup(self):
 
         # delete source records created on a previous run
         Creator.objects.all().delete()
-        Source.objects.all().delete()  # should cascade to footnotes
+        Source.objects.all().delete()
+        Footnote.objects.all().delete()
```
thatbudakguy, line 132: Did this because deletions didn't actually seem to cascade (or at least PKs kept incrementing...is that normal?)

geniza/footnotes/admin.py
```diff
@@ -110,12 +110,21 @@ class FootnoteAdmin(admin.ModelAdmin):
...
 
+    def content_blocks(self, obj):
+        output = ""
+        for block in obj.content["blocks"]:
+            if block["label"]:
+                output += f"{block['label']}\n"
+            for line in block["lines"]:
+                output += f"{line}\n"
+        return output
+    content_blocks.short_description = "Content"
```
thatbudakguy, line 126: tried using some light html here but it didn't render; maybe `mark_safe()` is needed or something?

rlskoeser: I think we should probably do it in the html so we can right align and set rtl text direction. (Although it seems to already be displaying rtl? does the browser pick that up automatically?)  

I suggest setting rtl/right align for everything right now rather than trying to figure out what script it is.

rlskoeser (changes requested):
Generally looks good — great to have some provisional transcription content in the database.

Let's move the has_content boolean and filters to the footnote list and remove from the document list.

Also consider displaying the content in the footnote change template rather than making it a read only field. (Seems like it could make it easier).

geniza/corpus/admin.py
thatbudakguy: Pretty sure we need it on the doc view; we should ask the team. Being able to see if a document is transcribed or not seems like core functionality.

geniza/footnotes/admin.py
thatbudakguy: change template seems like it would give more opportunity to customize; I'll try that

geniza/settings/local_settings.py.sample
```diff
@@ -34,3 +34,6 @@ DATA_IMPORT_URLS = {
     'libraries': '',
     'languages': ''
 }
+
+# path to preliminary JSON transcription data
+TRANSCRIPTIONS_JSON_FILE = STATIC_ROOT / "transcriptions.json"
```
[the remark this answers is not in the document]
thatbudakguy: oops, good call. should we assume the existence of a `/data` directory, or something? I didn't want to add a new command-line param to `import_data`

geniza/corpus/admin.py
rlskoeser: I agree they need a way to see that, just not sure if we need it _here_. If it doesn't slow things down too much then let's keep it, since you've already implemented. (But we need one on the footnote view as well, and you'll have to figure out how to fix the label)

geniza/settings/local_settings.py.sample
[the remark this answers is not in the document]
rlskoeser: I think setting it to "data/transcriptions.json" is a reasonable default; we'll want to set it to something else for the deploys, so it's outside of any deployed version. Agreed about not wanting to make it a param!

thatbudakguy:
ok, lots more changes that need review – still no tests. some of this is based on conversations with @richmanrachel about the admin layout; some is speculative & may change pending meeting of the "problem documents working group" next week.

to view this, you need to start with a fresh database, run migrations, then run the data import (I used `-m 1000` instead of the whole thing). note that the path is now expected to be `data/transcriptions.json` in the project root.

- `TextBlock.extent_label` becomes `TextBlock.region`: this identifier can be things like `a` or `b` but also `top`, etc. and is a generic identifier for regions of a fragment or sub-fragment.
- `TextBlock.multifragment` becomes `TextBlock.subfragment`: this identifier is for which sub-fragment _of_ a multifragment a `Document` occurs on.
- new `CustomEmptyFieldListFilter` and helper function for changing the title and empty/non-empty messaging for `admin.EmptyFieldListFilter`, which was seeing a lot of use.
- document admin now shows, sorts, and filters on whether documents have an image and/or transcription, by request. this makes it easy to get a quick data download of e.g. "ladino documents with both an image and transcription" since you often want both.
- collapsed both language filters in document admin to only values that were actually in use, because they were long.
- removed ability to filter on documents that occur on multifragments (you can do this from fragment admin instead) and documents that occur on labeled regions of fragments (not useful as per @richmanrachel).
- used the same language/filters on both document and footnote admin for presence/absence of transcription.
- updated footnote change form to show a lightly formatted, RTL version of the transcription.
- avoid creating "Unknown" document type on import and instead set `empty_value_display` to "Unknown" in the admin

geniza/corpus/models.py
```diff
@@ -274,6 +267,20 @@ def iiif_urls(self):
             None,
             [b.fragment.iiif_url for b in self.textblock_set.all()])))
 
+    def has_transcription(self):
+        """Admin display field indicating if document has a transcription."""
+        return any(note.has_transcription() for note in self.footnotes.all())
```
rlskoeser, line 272: can this be done with queryset annotation so it will be more efficient ?

geniza/corpus/models.py
```diff
@@ -274,6 +267,20 @@ def iiif_urls(self):
...
+    has_transcription.short_description = "Transcription"
+    has_transcription.boolean = True
+    has_transcription.admin_order_field = "footnotes__content"
+
+    def has_image(self):
+        """Admin display field indicating if document has a IIIF image."""
+        return any(self.iiif_urls())
```
rlskoeser, line 279: same question — queryset anntotation for this too?

geniza/corpus/admin.py
```diff
@@ -82,8 +86,8 @@ class DocumentTextBlockInline(admin.TabularInline):
     model = TextBlock
     autocomplete_fields = ['fragment']
     readonly_fields = ('thumbnail', )
-    fields = ('fragment', 'side', 'extent_label', 'multifragment',
-              'order', 'certain', 'thumbnail')
+    fields = ('thumbnail', 'fragment', 'subfragment', 'side', 'region',
```
rlskoeser, line 89: I noticed that fragments with multiple images look a little funny here because they stack vertically and take up space — not sure if we should do anything at this point, just wanted to flag in case you hadn't run into it. (I'm not sure of an easy solution)

rlskoeser (commented):
I like these updates a lot — `subfragment` seems much clearer. Thanks for making the region change, it wasn't explicitly tracked anywhere and I never got to it.

Should footnote inline indicate if there is content (at least yes/no, maybe preview/snippet?)?

Hoping the new list display properties can be set on the queryset — seems a little sluggish for me on my local dev.

thatbudakguy:
> Should footnote inline indicate if there is content (at least yes/no, maybe preview/snippet?)?

I can't figure out how to make the inline display a computed property; I think it might not be possible with `GenericTabularInline`...maybe we should switch footnotes to non-generic.

rlskoeser:
> > Should footnote inline indicate if there is content (at least yes/no, maybe preview/snippet?)?
> 
> I can't figure out how to make the inline display a computed property; I think it might not be possible with `GenericTabularInline`...maybe we should switch footnotes to non-generic.

Hm, I would have thought you could customize the template at least — but could be that it's not possible with the generic inline. (I presume you tried read only field?). I'm starting to lean towards making it non-generic but not ready to make the call yet...

rlskoeser:
@thatbudakguy sorry, should have said explicitly — let's leave that one for now and not worry about displaying a content indicator there

geniza/corpus/admin.py
```diff
@@ -109,19 +113,26 @@ class DocumentTextBlockInline(admin.TabularInline):
...
+        "subfragment",
         "side",
-        "extent_label",
-        "multifragment",
+        "region",
         "order",
         "certain",
         "thumbnail",
     )
+    extra = 0
```
rlskoeser, line 123: why extra 0? seems like we should provide at least 1

geniza/corpus/admin.py
```diff
@@ -197,7 +218,16 @@ def get_queryset(self, request):
...
-            .prefetch_related("tags", "languages", "textblock_set")
+            .select_related("doctype")
+            .prefetch_related(
+                "tags",
+                "languages",
+                Prefetch(
+                    "textblock_set",
+                    queryset=TextBlock.objects.select_related("fragment"),
+                ),
+                "footnotes__content__isnull",
```
rlskoeser, line 229: Can you add a comment documenting this? I know you explained it to me but I already forgot

geniza/corpus/models.py
```diff
@@ -334,6 +333,15 @@ def shelfmark(self):
             )
         )
 
+    @property
+    def shelfmark_display(self):
+        """First shelfmark plus join indicator for shorter display."""
+        certain = self.textblock_set.filter(certain=True)
+        if not certain.exists():
+            return None
```
rlskoeser: Is this allowed / possible? Shouldn't we expect at least one definite shelfmark if we have a document?

geniza/corpus/models.py
```diff
@@ -334,6 +333,15 @@ def shelfmark(self):
             )
         )
 
+    @property
+    def shelfmark_display(self):
```
rlskoeser, line 337: this is definitely out of the scope of the issue you were nominally working on! do we have an issue for this? is it worth tracking?

geniza/corpus/models.py
```diff
@@ -386,11 +386,24 @@ def iiif_urls(self):
...
+    has_image.short_description = "Image"
+    has_image.boolean = True
+    has_image.admin_order_field = "textblock__fragment__iiif_url"
+
     @property
     def title(self):
         """Short title for identifying the document, e.g. via search."""
-        # NOTE preliminary, pending more discussion
-        return f"{self.doctype or 'Unknown'} ({self.id})"
+        return f"{self.doctype or 'Unknown'}: {self.shelfmark_display or '??'}"
```
rlskoeser, line 408: Isn't this still preliminary? It doesn't look quite like the format we finally settled on. (Also out of scope!)

Maybe you can make an issue for revising document title for the public site and note there's a preliminary implementation

rlskoeser (approved):

geniza/corpus/models.py
thatbudakguy: It might not be allowed, but it's definitely possible because we can't require that a document have any `TextBlocks`, right? Added because tests will fail on newly created documents since they have no associated fragments at all and it seemed better to handle it.

geniza/corpus/models.py
thatbudakguy: it's based on a convo we had at a meeting where we discussed notating joins with the plus sign; i thought it might be helpful to see it on the document detail pages to continue the conversation. you can also think of it as a stubbed implementation of #154; eventually this same function will need to collapse shelfmarks into ranges too.

geniza/corpus/models.py
thatbudakguy: Although maybe we should add a `save()` hook to just prevent that from happening, actually, since it definitely doesn't make sense to have a document without any textblocks

PR #158, opened 2021-05-05
Feature/editor import revisions

rlskoeser:
Improve editor & translator import to handle more cases. Specifically:
- smarter volume recognition
- more precise year recognition 
- when looking for existing sources, do case-insensitive exact search
- more variations of notes

I think some of the note content currently going into the sources should go into footnotes — but not _all_ of them. Leaving for now.


thatbudakguy (approved):
looks good, although not sure i could really tell if something was wrong at this point...

PR #166, opened 2021-05-10
164 implement black

kmcelwee:
- [x] Setup pre-commit hook as described in black documentation
- [x] Document install process
- [x] Apply quote / apostrophe standardization


rlskoeser (approved):
The code changes look good — the things that are different than my practice are manageable enough, and automating code cleanup for consistency and cleaner reviews seems well worth it.

Should we setup and manage the pre-commit hook this way instead? Seems better:
https://black.readthedocs.io/en/stable/version_control_integration.html

Also, optional — set git blame to ignore this change? (but GitHub may not ignore it)
https://black.readthedocs.io/en/stable/installation_and_usage.html#migrating-your-code-style-without-ruining-git-blame

rlskoeser:
> We'd have to add a `pyproject.toml` file to appropriately configure everything and pre-commit seems to have a bit of overhead, and adding one line to pre-commit seemed simpler. Thoughts?

Didn't seem like that much overhead — we add pre-commit to our dev dependencies, and adapt the example config files. Is there more setup required I'm missing?  

Seems like automating the commit hook setup would help with consistency and could be used on other projects.

@thatbudakguy any opinion?

kmcelwee:
@rlskoeser for a few reasons, it seems like implementing black won't be a one-liner anyway, so I'll set it up in the way they ask!

thatbudakguy:
I'm OK with the `pre-commit` dependency if it gets us "smart" behavior. @kmcelwee tried this and got it working with just the pre-commit hook file, but as he mentioned it seemed like it would require more setup to be very usable (e.g. we'd want it to only affect the changed files in the commit, and also automatically add them to the commit).

`pyproject.toml` is [a PEP that was finalized last year](https://www.python.org/dev/peps/pep-0621/) and lots of people/tools are using it already (e.g. poetry) so we may as well start now.

rlskoeser (approved):
Thees all look fine: .toml readme .git-blame-ignore-revs .pre-commit-config.yaml 
(we should probably add more project metadata at some point to the pyproject, but I'm fine with not doing it now if this is sufficient for immediate needs).

And I'm ok with acceding to black's opinionated cleaning where it differs from my own opinions!

Do you want me to handle the merge conflicts?

PR #169, opened 2021-05-12
Solr tweaks

rlskoeser:
- remove `SolrSchema` and make changes in the schema included in the configset instead
- use dynamic field for item_type (NOTE: requires latest version of parasolr develop)
- revise readme instructions to reflect the change for solr setup


solr_conf/conf/managed-schema
```diff
@@ -94,6 +94,12 @@
   <field name="_text_" type="text_general" multiValued="true" indexed="true" stored="false"/>
   <field name="_version_" type="long" indexed="false" stored="false"/>
   <field name="id" type="string" multiValued="false" indexed="true" required="true" stored="true"/>
+  <!-- local fields -->
+  <field name="last_modified" type="pdate" multiValued="false" indexed="true" required="true" stored="true" default="NOW"/>    
+
+  <copyField source="shelfmark_t" dest="shelfmark_ss" maxChars="30000" />
+  <copyField source="tag_t" dest="tag_ss" maxChars="30000" />
```
kmcelwee, line 101: Since we define the copy fields elsewhere, would it be possible to eventually add a feature to parasolr to automatically add these?

kmcelwee (approved):

solr_conf/conf/managed-schema
rlskoeser: I removed the other definition of them — we were automatically managing them with the schema api, but that isn't compatible with using a configset in zookeeper. When I redeployed, updating the configset clobbered the changes made via schema api.

I think we need to either use configset for everything or schema api for everything — using the schema api seems appealing, but right now we can't do everything we need to do through it (I'm not certain if everything we need is supported; likely it is, but we don't have wrapper code in parasolr yet for some of them), so seems best to just us the configset for everything for now.

PR #174, opened 2021-05-14
Improve inline editing UI for footnotes

thatbudakguy:
Some small changes to make the footnote inlines on both `Document` and `Source` a little easier to use.

- Remove `FootnoteInline` since it's been split into `SourceFootnoteInline` and `DocumentFootnoteInline`
- Re-add digitized transcription indicator to both inlines
- Tweak field sizing, order, and CSS so that inlines fit the screen width

Screenshot of source footnote inline:
Screen Shot 2021-05-14 at 3 21 01 PM

Screenshot of document footnote inline:
Screen Shot 2021-05-14 at 3 22 09 PM

### DIFF
--- a/geniza/footnotes/admin.py
+++ b/geniza/footnotes/admin.py
@@ -1,10 +1,13 @@
 from adminsortable2.admin import SortableInlineAdminMixin
+from django import forms
 from django.contrib import admin
 from django.contrib.admin import SimpleListFilter
 from django.contrib.contenttypes.admin import GenericTabularInline
 from django.db import models
 from django.db.models import Count
+from django.db.models.fields import CharField, TextField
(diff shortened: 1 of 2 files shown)


rlskoeser (commented):
Thanks for working on this and for including the screenshots in the PR — it looks much nicer!

My one quibble is that content type should be before object id — you _can't_ select object id until you've selected the type. Can you switch them easily?

thatbudakguy:
Yep – updated it so the magnifying glass is in between content type and object id now. I originally switched them because it was confusingly close to location, so I think this is a good compromise.

rlskoeser (approved):
Looks great, thanks for adjusting it

PR #175, opened 2021-05-17
Feature/145 demerge spreadsheet

kmcelwee:


geniza/corpus/management/commands/import_data.py
```diff
@@ -311,46 +373,21 @@ def import_documents(self):
...
-                doc, self.get_edit_history(row.input_by, row.date_entered, row.pgpid)
+
+        # update id sequence based on highest imported pgpid
+        self.update_document_id_sequence()
+
+        # demerged_metadata = reversed(sorted(demerged_metadata, key=lambda x: x.pgpid))
+        for row in demerged_metadata:
+            # overwrite document if it already exists
+            if row.pgpid:
+                Document.objects.filter(id=row.pgpid).delete()
```
kmcelwee, line 387: Documents in the demerge spreadsheet overwrite those in the metadata spreadsheet correct?

geniza/corpus/management/commands/import_data.py
```diff
@@ -294,8 +307,57 @@ def add_document_language(self, doc, row):
             doc.language_note = "\n".join(notes_list)
             doc.save()
 
+    def import_document(self, row, joins, docstats, recto_verso_lookup):
```
kmcelwee, line 310: some of these could become class attributes if you'd like this cleaned up?

geniza/corpus/management/commands/import_data.py
thatbudakguy: I think we avoid ever having conflicts, because if we detect a document that needs demerging in the metadata spreadsheet, we don't add it to the database at all

thatbudakguy: The assumption is that the demerge spreadsheet handles importing documents for all of those cases, and in fact we want to ensure that every demerge case detected from the main sheet is resolved via the demerge sheet

geniza/corpus/models.py
```diff
@@ -468,6 +468,12 @@ def index_data(self):
         # script+language when/if included in index data
     }
 
+    def save(self, *args, **kwargs):
+        # Auto-increment pk when none is provided
+        if self.id is None:
+            self.id = self.__class__.objects.all().order_by("-id")[0].id + 1
+        super(self.__class__, self).save(*args, **kwargs)
```
kmcelwee, line 475: Was there a more straightforward way to doing this?

rlskoeser: @kmcelwee django should do this automatically for us if no id is specified, we already have it set as an auto-increment. Did that not work?

kmcelwee: I removed this — I had changed two things at once, not realizing that casting an empty string as `None` was sufficient. Thanks!

geniza/corpus/management/commands/import_data.py
rlskoeser: ooh, I see my dev instructions about refactoring were vaguer than I thought! I did not want a method to import a single document, I wanted a method to import a set of documents. The inner loop of the current handle documents method.

 I thought the logic separation would be cleaner that way. Since you've worked on it, you may have a better sense about that?  Either way, this seems like a lot of stuff to be passing around — they should probably become class attributes if we need to pass them between methods.

kmcelwee: * I'm inclined to keep it as is, just because I feel like it'd be easier to test, but it shouldn't be hard to include the for loops. Let me know what you think.
* I'll move `joins` and `docstats` to the class, I don't see a way to refactor to get rid of those.
    * Making `joins` a class variable also allows us to place all the join logic in its own function. Let me know if you think that'd be a good idea.

geniza/corpus/management/commands/import_data.py
```diff
@@ -64,6 +64,19 @@
         "Editor(s)": "editor",
         "Translator (optional)": "translator",
     },
+    "demerged": {
```
rlskoeser, line 67: Aren't these all the same as the metadata fields? Can't we use the same list?

geniza/corpus/management/commands/import_data.py
```diff
@@ -294,8 +307,57 @@ def add_document_language(self, doc, row):
...
+            document=doc,
+            fragment=fragment,
+            # convert recto/verso value to code
+            side=self.recto_verso_lookup.get(row.recto_verso, ""),
+            region=row.text_block,
+            subfragment=row.multifragment,
+        )
+        self.add_document_language(doc, row)
+        if demerge:
+            self.docstats["demerged_documents"] += 1
```
rlskoeser, line 336: Do we really need a separate count? Since you're incrementing the total for documents, you can save the total before you run import the demerge and then compare the difference after.

geniza/corpus/management/commands/import_data.py
```diff
@@ -304,56 +366,27 @@ def import_documents(self):
...
-
-            # keep track of any joins to handle on a second pass
-            if row.joins.strip():
-                joins.append((doc, row.joins.strip()))
+        # update id sequence based on highest imported pgpid
+        self.update_document_id_sequence()
+
+        for row in demerged_metadata:
+            # overwrite document if it already exists
+            if row.pgpid and Document.objects.filter(id=row.pgpid).exists():
```
rlskoeser, line 383: Did this happen at all in your testing?

geniza/corpus/management/commands/import_data.py
```diff
@@ -517,8 +554,8 @@ def get_edit_history(self, input_by, date_entered, pgpid=None):
         """
 
         # split both fields by semicolon delimiter & remove whitespace
-        all_input = [i.strip() for i in input_by.split(";")]
-        all_dates = [d.strip() for d in date_entered.split(";")]
+        all_input = [i.strip() for i in input_by.split(";") if i]
+        all_dates = [d.strip() for d in date_entered.split(";") if d]
```
rlskoeser, line 558: Is this right? I thought we needed the two fields to match up (to the degree possible)

geniza/corpus/management/commands/import_data.py
```diff
@@ -294,8 +307,57 @@ def add_document_language(self, doc, row):
...
+        """Import a single document given a row from a PGP spreadsheet"""
+        if ";" in row.type:
+            logger.warning("skipping PGPID %s (demerge)" % row.pgpid)
+            self.docstats["skipped"] += 1
+            return
+
+        doctype = self.get_doctype(row.type)
+        fragment = self.get_fragment(row)
+        doc = Document.objects.create(
+            id=row.pgpid if row.pgpid else None,
```
rlskoeser, line 320: or just `row.pgpid or None` ?

geniza/corpus/management/commands/import_data.py
rlskoeser: Could specify via an optional param to `get_csv` to override the field config used; by default it would use the data name

rlskoeser (approved):
Everything looks reasonable, just a few small suggestions.

You're right, this refactor does look cleaner and I'm sure will be easier to test.

geniza/corpus/management/commands/import_data.py
kmcelwee: yes, should we give a heads up to the pgp team?
```
WARNING:import:Overwriting PGPID with demerge 31511
```

geniza/corpus/management/commands/import_data.py
kmcelwee: The demerge spreadsheet has `' '` in these columns, which threw an error. I realize that this quick fix may not be sufficient. This might be worth talking over.

geniza/corpus/management/commands/import_data.py
kmcelwee: oh that's smart 💡 thanks!

geniza/corpus/management/commands/import_data.py
rlskoeser: I just checked the spreadsheet — that one has two types separated by a semicolon, so I think we _should_ be skipping it. 

What happens if you don't check and you try to save a record for a pgpid that already exists? Does it error?

geniza/corpus/management/commands/import_data.py
rlskoeser: defer to @thatbudakguy on this one!

thatbudakguy: i think this is ok; we pad the lists to the same length later. also we might want to just duplicate/fill down the data in the demerge spreadsheet rather than leaving in the `' '`, which would eliminate this problem

geniza/corpus/management/commands/import_data.py
kmcelwee: The PGPID was manually defined twice in the demerge spreadsheet. That's what was causing the error. It was correctly skipped in the metadata spreadsheet.

Screen Shot 2021-05-17 at 4 46 06 PM

rlskoeser: ah, ok — great that you're checking, then! Thanks for finding and correcting it

rlskoeser (approved):
ok! (once the tests are passing, of course 🙂 )

kmcelwee:
Ran the full import process instead of just a subsection (took more than 15 minutes). This was the log output: 
```
0 were imported from the demerge spreadsheet, but -165 documents remain.
```
So something wrong :/

kmcelwee:
New, non-multiple-doctype demerges were recently added to the demerge spreadsheet. This is fine as long as the PGP team understands that anything in the demerge spreadsheet will overwrite the corresponding PGPID in the metadata spreadsheet

```
WARNING:import:Overwriting PGPID with demerge 8979
WARNING:import:Overwriting PGPID with demerge 19259
WARNING:import:Overwriting PGPID with demerge 19376
WARNING:import:Overwriting PGPID with demerge 6478
WARNING:import:Overwriting PGPID with demerge 19421
WARNING:import:Overwriting PGPID with demerge 6540
WARNING:import:Overwriting PGPID with demerge 6541
WARNING:import:Overwriting PGPID with demerge 19727
WARNING:import:Overwriting PGPID with demerge 19815
WARNING:import:Overwriting PGPID with demerge 5672
WARNING:import:Overwriting PGPID with demerge 12319
WARNING:import:Overwriting PGPID with demerge 6681
WARNING:import:Overwriting PGPID with demerge 2654
WARNING:import:Overwriting PGPID with demerge 11621
WARNING:import:Overwriting PGPID with demerge 6673
WARNING:import:Overwriting PGPID with demerge 11343
WARNING:import:Overwriting PGPID with demerge 730
WARNING:import:Overwriting PGPID with demerge 12446
WARNING:import:Overwriting PGPID with demerge 9799
WARNING:import:Overwriting PGPID with demerge 12025
WARNING:import:Overwriting PGPID with demerge 9798
WARNING:import:Overwriting PGPID with demerge 11211
WARNING:import:Overwriting PGPID with demerge 11812
WARNING:import:Overwriting PGPID with demerge 11848
WARNING:import:Overwriting PGPID with demerge 12264
WARNING:import:Overwriting PGPID with demerge 26705
WARNING:import:Overwriting PGPID with demerge 27044
WARNING:import:Overwriting PGPID with demerge 2882
WARNING:import:Overwriting PGPID with demerge 2886
WARNING:import:Overwriting PGPID with demerge 2898
WARNING:import:Overwriting PGPID with demerge 3379
WARNING:import:Overwriting PGPID with demerge 7361
WARNING:import:Overwriting PGPID with demerge 5292
WARNING:import:Overwriting PGPID with demerge 1965
WARNING:import:Overwriting PGPID with demerge 6046
WARNING:import:Overwriting PGPID with demerge 23481
WARNING:import:Overwriting PGPID with demerge 24597
WARNING:import:Overwriting PGPID with demerge 24599
WARNING:import:Overwriting PGPID with demerge 26272
WARNING:import:Overwriting PGPID with demerge 26375
WARNING:import:Overwriting PGPID with demerge 26377
WARNING:import:Overwriting PGPID with demerge 27896
INFO:import:Imported 30603 documents from the metadata spreadsheet and skipped 170. Imported 393 documents from the demerged spreadsheet. Overwrote 43 documents. Parsed 1008 joins.
```

Issue #121, opened 2021-04-05
As a Content Editor, I want to download a CSV version of all or a filtered list of documents in the backend, in order to data work or facilitate my own research.


richmanrachel:
## Testing Notes
There are two ways of downloading a document CSV in the document admin view, either by ticking the document checkboxes and selecting "Download as CSV" from the actions dropdown or by clicking the "Download all as CSV" in the top right corner of the page.

**Test the site:**
- [ ] Test the "Download all as CSV" button works.
- [x] Select a few documents and ensure that the actions dropdown works.

**Test the CSV**
- [x] Test the links in the CSV and ensure that they lead to the correct pages
- [ ] Confirm that all fields and header names are as expected.

---

**Fields**
Fields and field order as approved by PGP team

 - PGPID
 - link to public site (column name: url_public; or just url?)
 - fragment IIIF link* (column name: iiif_url)
 - link to view fragment on library site* (column name: fragment_url? view_fragment?)
 - shelfmark*
 - multifragment*
 - side*
 - extent_label* (rename to region?)
 - type
 - tags* (suggest displaying #tag1 #tag2 for consistency/familiarity; but could use semicolons)
 - description
 - footnotes (i.e editions & translations; include source, location, notes, and document relationship; pending data import)
 - historic shelfmarks* (column name: shelfmarks_historic)
 - languages*
 - probable languages* (column name: languages_probable)
 - language_note
 - notes (document only, not fragment notes)
 - needs_review (document only, not fragment)
 - link to admin change view (column name: url_admin)
 - input_by* (list all unique names of people who edited the record)
 - date_entered (first log entry date)
 - last_modified
 - status (i.e., public or suppressed — use display value, not code)
 - library abbreviation* (column name: library)
 - collection abbreviation* (column name: collection)

*multivalued or possibly multivalued fields; delimit with semicolons unless otherwise noted

**Default sort**
Which field should be used for initial sorting?

-----

## dev notes

We've used https://github.com/LibraryOfCongress/django-tabular-export in the past, and it does exactly what we need. It looks like it hasn't been updated in a while, but let's see if still works.  (I could ask on twitter if they're still using it or recommend something else; if appropriate, might be easy & nice to contribute an update for newer versions of django...)

fixes after first round testing:
- [x] for all multivalued fields (iiif urls, fragment url, subfragment): only generate if at least one fragment has a value (get rid of ";")
- [x] input by: output user fullname instead of user id
[...]

PR #176, opened 2021-05-18
Feature/121b csv admin export

kmcelwee:
Addresses post-testing issues outlined in #121.

### DIFF
--- a/geniza/corpus/admin.py
+++ b/geniza/corpus/admin.py
@@ -24,6 +24,7 @@ from geniza.corpus.solr_queryset import DocumentSolrQuerySet
 from geniza.common.admin import custom_empty_field_list_filter
 from geniza.footnotes.admin import DocumentFootnoteInline
 from geniza.common.utils import absolutize_url
+from django.contrib.auth.models import User
 
 
 class FragmentTextBlockInline(admin.TabularInline):
@@ -334,6 +335,7 @@ class DocumentAdmin(admin.ModelAdmin):
             all_fragments = doc.fragments.all()
             all_textblocks = doc.textblock_set.all()
             all_footnotes = doc.footnotes.all()
+            all_user_ids = set(doc.log_entries.values_list("user", flat=True))
 
             initial_entry = doc.log_entries.first()
             latest_revision = doc.log_entries.last()
@@ -371,8 +373,8 @@ class DocumentAdmin(admin.ModelAdmin):
                     "latest_revision": doc.log_entries.last(),
                     "input_by": ";".join(
                         [
-                            str(n)
-                            for n in set(doc.log_entries.values_list("user", flat=True))
+                            user.get_full_name() or user.username
+                            for user in User.objects.filter(pk__in=all_user_ids)
                         ]
                     ),
                     "status": "Public"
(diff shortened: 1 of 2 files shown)


geniza/corpus/admin.py
```diff
@@ -343,21 +345,33 @@ def tabulate_queryset(self, queryset):
                     "pgpid": doc.id,
                     "url": absolutize_url(doc.get_absolute_url()),
                     "iiif_urls": ";".join(
-                        [fragment.iiif_url for fragment in all_fragments]
+                        [
+                            fragment.iiif_url
+                            for fragment in all_fragments
+                            if fragment.iiif_url
```
rlskoeser, line 351: I don't think this is the logic we want (here and throughout), because if one fragment has a value and others don't we won't know which one it belongs to (unless we think they don't care about that?).

My recommendation would be to only generate the field if any fragments have the value (i.e. `any(fragment.iiif_url for fragment in all_fragments)`) but not to skip empty values.

geniza/corpus/admin.py
```diff
@@ -371,15 +385,19 @@ def tabulate_queryset(self, queryset):
                     "latest_revision": doc.log_entries.last(),
                     "input_by": ";".join(
                         [
-                            str(n)
-                            for n in set(doc.log_entries.values_list("user", flat=True))
+                            user.get_full_name() or user.username
+                            for user in User.objects.filter(pk__in=all_user_ids)
```
rlskoeser, line 389: No need to do a separate lookup on id, you can get it directly from the log entries (you just need to not get a flat  list of user ids). Try something like this:
`[entry.user.get_full_name() or entry.user.username for entry in doc.log_entries.all()]`

I think we only want the unique list, though, so wrap it in `set()`

geniza/corpus/admin.py
```diff
@@ -371,15 +385,19 @@ def tabulate_queryset(self, queryset):
...
                     "status": "Public"
                     if doc.status == Document.PUBLIC
                     else "Suppressed",
                     "library": ";".join(
-                        [fragment.collection.lib_abbrev for fragment in all_fragments]
+                        [
+                            fragment.collection.lib_abbrev
+                            for fragment in all_fragments
+                            if fragment.collection and fragment.collection.lib_abbrev
+                        ]
```
rlskoeser, line 400: Ooh, I didn't catch this on the review before — we should use the collection abbreviation if there is no lib abbreviation (and maybe also if there is).  And we should not have any empty values in this field.

I think there should always be at least one abbreviation, but I see it's not required in the model. The default string method for `Collection` uses abbreviations when set, so let's just use that.

geniza/corpus/admin.py
rlskoeser: One more thought — remove `script` after you've converted to a set, since pretty much every document will be modified by script at some point.

geniza/corpus/admin.py
kmcelwee: I wrapped them in a lambda function:
```python
semicolon_str = lambda x: "" if x == ";" else x
```
It seemed simpler to me, but I recognize it's atypical. Is that alright? (e7778bdc35a5bb02a24675081da64b117845fa35)

rlskoeser: seems ok. Please add a brief comment above the lambda so we remember what this is when we come back to it later 🙂

PR #177, opened 2021-05-18
Handle journals when importing editor/translator sources

rlskoeser:
- add optional `journal` field to `Source`
- use regular expressions to look for known journals when importing editors; sets type to article and imports journal and volume (if set), even if there is no title or title does not have quotes
- include journal and volume in source `str` output
- refinements to notes and document locations regexes to handle more cases

todos in comments are planned to be handled on other issues


geniza/corpus/tests/test_import_data.py
```diff
@@ -883,16 +884,27 @@ def test_get_source_creator_not_found():
...
-        'Mordechai Akiva Friedman, "Maimonides Appoints R. Anatoly Muqaddam of Alexandria [Hebrew]," Tarbiz, 2015, 135–61, at 156f. Awaiting digitization on PGP.',
+        'Mordechai Akiva Friedman, "Maimonides Appoints R. Anatoly Muqaddam of Alexandria [Hebrew]," Tarbiz, 2015',
         {
             "type": "Article",
             "authors": ["Friedman, Mordechai Akiva"],
             "title": "Maimonides Appoints R. Anatoly Muqaddam of Alexandria",
             "language": "Hebrew",
             "year": 2015,
+            "journal": "Tarbiz",
         },
```
thatbudakguy, line 895: should this also test for setting `f_location` to `135–61, at 156f` or something like that?

geniza/footnotes/models.py
```diff
@@ -109,29 +119,41 @@ def __str__(self):
...
         if self.year:
-            text = "%s (%d)" % (text, self.year)
-        return text
+            parts.append("(%d)" % self.year)
+
+        # title, journal, etc should be joined by spaces only
+        ref = " ".join(parts)
+
+        # delimit with comma whichever values are set
+        return ", ".join([val for val in (author, ref) if val])
```
thatbudakguy, line 156: this is a really complex `__str__()`! dunno if there's any way around it though. is the reason for trying to generate a format similar to the spreadsheet because it will help the team? i wonder if we could make it less complex by choosing a format that is more regular, if they don't absolutely need this format. unless it's just a problem of cobbling together many potentially missing pieces, which might not be resolvable.

thatbudakguy (approved):

geniza/corpus/tests/test_import_data.py
rlskoeser: I stripped it out because locations are handled in the other method (parse editor), before hand off to source parsing. (Location is stored in footnote, not source)

geniza/footnotes/models.py
rlskoeser: There's a lot of potential missing data, which does make it complicated. I'm open to revising the format in the future, but thought it would be helpful for now to stick with something similar to what the team is used to. (Also since I think we should provide something close to that version for populating the current PGP site).

PR #183, opened 2021-05-20
Feature/107 publish csv

kmcelwee:


geniza/corpus/views.py
```diff
@@ -15,3 +23,70 @@ def get_queryset(self, *args, **kwargs):
...
+
+        yield row
+
+
+def render_pgp_metadata_for_old_site(request):
+    """A view that streams a large CSV file."""
+
+    queryset = Document.objects.filter(status=Document.PUBLIC)
+
+    queryset = queryset.order_by("id")
```
rlskoeser, line 82: one line please 🙂 (doesn't even need to be a separate variable, could just pass to the tabulate method)

geniza/corpus/views.py
```diff
@@ -15,3 +23,70 @@ def get_queryset(self, *args, **kwargs):
...
+def tabulate_queryset(queryset):
+    reverse_recto_verso_lookup = {
+        code: label.lower() for code, label in TextBlock.RECTO_VERSO_CHOICES
+    }
+
+    for doc in queryset:
+        all_fragments = doc.fragments.all()
+        rectoverso_q = doc.textblock_set.filter(~Q(side="")).first()
+
+        row = DocumentRow(
```
rlskoeser, line 57: I don't remember why you guys decided to the named tuple for the admin export, but I don't want to devote that much space to this export! Let's just return (yield) a list of values for each row. You can add comments for the labels (where/if not obvious) so we don't lose track.

kmcelwee: Yeah, I can certainly yield a list instead of the dict, but we'll need to then list out the headers anyway in the `export_to_csv_response`, where we currently use `DocumentRow._fields`. I can do that, but I don't think it'll save that much space. Would you mind confirming?

rlskoeser: yes, please just provide a list of fields and generate a list of values

kmcelwee: ✅  c4b7918

geniza/corpus/tests/test_corpus_views.py
```diff
@@ -20,13 +25,35 @@ def test_get_queryset(self, db, client):
...
+    legal_doc = DocumentType.objects.create(name="Legal")
+    doc = Document.objects.create(id=36, doctype=legal_doc)
+    frag = Fragment.objects.create(shelfmark="T-S 8J22.21")
+    TextBlock.objects.create(document=doc, fragment=frag, side="r")
+    doc.fragments.add(frag)
+    doc.tags.add("marriage")
+
+    table_iter = tabulate_queryset(Document.objects.all())
+    row = next(table_iter)
+    print(row)
```
rlskoeser, line 39: leftover print statement

rlskoeser (approved):

Issue #170, opened 2021-05-13
As a content editor, I want duplicate joined documents to be automatically merged without losing their unique metadata, so that I don't have to merge them manually.


thatbudakguy:
# notes for testing
check a few different groups from the spreadsheet to confirm:
- [x] primary document should have all secondary document PGPIDs listed in historic PGPIDs
- [x] historic PGPID should be displayed in admin document edit form
- [x] all tags from any documents in a merged group should be present on the primary document
- [x] all descriptions (unless they are an exact duplicate) should be added to the description with a note about the source document
- [x] all footnotes from all documents in the merged group should be associated with the primary document; any footnotes that are exactly equal should only show up once
- [x] review note should indicate that the document has been merged
- [x] all primary and secondary languages from any documents in the group should be associated with the main document
- [x] all log entries from all documents in the merged group should be re-associated with the primary document (current behavior, need to decide if we want to revisit)
- [x] there should be a new log entry documenting the merge
 
also:
- [x] confirm that no action was taken for rows in the spreadsheet where action is not MERGE
---- 

### Is your feature request related to a problem? Please describe.
Because of the way the current PGP system records joins, there are many duplicate entries for joined documents in the metadata spreadsheet (see https://github.com/Princeton-CDH/geniza/issues/165). Over time, these entries have accumulated unique metadata describing the particular shelfmarks that make up the join.

The new system represents each document with a single PGPID, including joins. We need to merge the duplicate entries, but we want to preserve the unique fragment-level metadata that they have accumulated.

### Describe the solution you'd like
This could be implemented as a management command, and would be easier to accomplish once the data has already been migrated to the database.

In order to support this work, we need to be able to **track historic PGPIDs of a document**, because some PGPIDs will "go away" once their document is merged. We still want to be able to find the relevant information using that PGPID and redirect to the newly merged document.

The logic would be something like:

- [x] Find all documents that are joins, with duplicate shelfmarks, and where the duplicates also have the same document type
- [x] Add the PGPID of each of the duplicates to the historic PGPIDs list of the first/"original" document (use the lowest numeric pgpid)
- [x] Add all the unique tags from each of the duplicates to the tags of the primary document
- [x] Add the descriptions of each of the duplicates (if unique) to the description of the primary document, with a note identifying each description by shelfmark, e.g. `Description for ENA 2262.5: `
- [x] Consolidate edit history — reassociate any unique events for each duplicates with the primary document; delete any duplicates
[...]

PR #184, opened 2021-05-20
Feature/170 merge duplicate joins

kmcelwee:


geniza/corpus/models.py
```diff
@@ -264,6 +264,7 @@ class Document(ModelIndexable):
     appears on one or more fragments."""
 
     id = models.AutoField("PGPID", primary_key=True)
+    old_pgpids = models.CharField(max_length=255, blank=True)
```
rlskoeser, line 267: could we use a postgres arrayfield with integers here? https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/fields/#arrayfield

geniza/corpus/management/commands/merge_duplicate_joins.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    """Collapse PGPIDs that share the same joins"""
+
+    def get_documents_with_multiple_joins(self):
+        pass
```
rlskoeser, line 15: Here's the way I thought of for identifying candidates:  annotate document queryset with fragment count (something like `annotate(fragment_count=models.Count("fragments")`, filter to documents with more than one fragment, and then get a list of ids (`.value_list('pk', flat=True)`)

geniza/corpus/management/commands/merge_duplicate_joins.py
```diff
@@ -0,0 +1,45 @@
...
+        primary_doc = min(doc_list, lambda d: d.pgpid)
+        merge_docs = [doc for doc in doc_list if doc != primary_doc]
+
+        # add old ids
+        for doc in merge_docs:
+            primary_doc.add_old_pgpid(doc.pgpid)
+
+        # compress description
+        # confirm doctype the same
+        # compress unique tags
```
rlskoeser, line 28: I *think* you can just add all tags and taggit will do the right thing

geniza/corpus/management/commands/merge_joins.py
```diff
@@ -0,0 +1,250 @@
...
+                    "status",
+                    "role",
+                    "pgpid",
+                    "description",
+                ]
+            )
+            csvwriter.writerows(report_rows)
+        self.stdout.write("Report of merge candidates output as merge-report.csv")
+
+    def merge_documents(primary_doc, merge_docs, rationale):
```
rlskoeser, line 221: I've been thinking about it, and I think `group_merge_candidates` should return a list of tuples of primary document, one or more merge documents, and text/string for the merge rationale. Then that list could be used either to generate the csv report or as input for actually doing the merge, depending on the mode. (The rationale would be used in the log entry documenting the change).

thatbudakguy: Interesting...I guess that would allow you to manually remove false positive merges from the spreadsheet in between running the two commands, so that you only merged ones that you had reviewed. Makes sense!

thatbudakguy:
- [ ] Add group IDs to generated report
- [ ] Implement `Document.merge_with()`, which takes a group of candidates and a rationale (string), and follows the logic outlined in #170
- [ ] Add a method to the `merge_joins` management command that parses a report CSV (generated from its `report` subcommand) and calls `Document.merge_with()` on the primary, passing the rest of the candidates in that group (by group ID) with the `MERGE` action specified in that group, and the rationale (from the primary row). Make sure that the command doesn't assume group IDs are sequential or that groups are always together in the spreadsheet.

rlskoeser:
@thatbudakguy I've finished implementing and testing the `merge_with` function, if you want to review

geniza/corpus/models.py
```diff
@@ -533,6 +536,108 @@ def index_data(self):
...
+        for doc in merge_docs:
+            # add merge id to old pgpid list
+            self.old_pgpids.append(doc.id)
+            # add any tags from merge document tags to primary doc
+            self.tags.add(*doc.tags.names())
+            # add description if set and not duplicated
+            if doc.description and doc.description not in self.description:
+                description_chunks.append(
+                    "Description from %s:\n%s" % (doc.shelfmark, doc.description)
+                )
```
thatbudakguy, line 570: i realized on looking at the issue that adding the shelfmark indicator might not actually help...if we're merging two docs and they both have essentially the same shelfmark (because it's a join), then the description will say:
```
[primary description]
Description from ENA XXX + CUL YYY
[secondary description]
```
...which doesn't make much sense because _both_ docs are ENA XXX + CUL YYY. their shelfmark will be same. so we might want to change it to:
```
[primary description]
Description from PGPID [merged PGPID]
[secondary description]
```
...or something else. lmk if this doesn't make sense

geniza/corpus/models.py
```diff
@@ -533,6 +536,108 @@ def index_data(self):
...
+            # add any notes
+            if doc.notes:
+                notes.append("Notes from %s:\n%s" % (doc.shelfmark, doc.notes))
```
thatbudakguy, line 573: same deal here

geniza/corpus/models.py
```diff
@@ -533,6 +536,108 @@ def index_data(self):
...
+            # (i.e., for newly discovered joins)
+            # does not deal with discrepancies between text block fields or order
+            for textblock in doc.textblock_set.all():
+                if textblock.fragment not in self.fragments.all():
+                    self.textblock_set.add(textblock)
+
+            # reassociate all footnotes
+            for footnote in doc.footnotes.all():
+                # any good way to check if any are exactly the same?
+                self.footnotes.add(footnote)
```
thatbudakguy, line 596: we could implement `__eq__` for `Footnote`. i think the combination of content object, source, relation type, and location would constitute uniqueness

thatbudakguy (approved):
`Document.merge_with()` looks good!

geniza/corpus/models.py
rlskoeser: yeah, good point — I had it as pgpid at first and then saw in the issue we wanted shelfmark; but in all of these cases the shelfmark won't be helpful since they're all joins. I'll change it back to what I had before.

geniza/corpus/models.py
rlskoeser: Yeah, I thought about that too, but didn't want to take it on right now. I was thinking that at least for now we can ask them to check footnotes when reviewing merged documents. Although implementing it might not be all that hard... (and even if we do they would probably still need to check the footnotes). What do you think?

thatbudakguy: i think we should assume that there may actually be many exact duplicate footnotes, since for a decent number of the merge groups they went to the trouble of ensuring that the description was also duplicated exactly. that being the case, it seems worthwhile to implement `__eq__` so we can do a check here and maybe save some work

rlskoeser: @thatbudakguy what do you think, should notes and content (i.e. provisional transcription content) be included in footnote equality check?

rlskoeser: Actually, to use this for the merge check, the simplest thing would be if footnote equality does to _not_ compare content object — but I'm not sure if that's logical for any other use case... What do you think?

rlskoeser:
@thatbudakguy could you review the recent changes? I'm basically leaving the reporting functionality as we left it (but added some light testing), since we already generated a report. I revised the merge functionality as we discussed for footnotes; we may still want to revise the log entry handling, but Rachel and I want to see what it looks like in test before deciding.

geniza/corpus/models.py
thatbudakguy: were we still planning to change this back to PGPID rather than shelfmark?

geniza/corpus/models.py
```diff
@@ -585,6 +588,110 @@ def index_data(self):
...
+            # (i.e., for newly discovered joins)
+            # does not deal with discrepancies between text block fields or order
+            for textblock in doc.textblock_set.all():
+                if textblock.fragment not in self.fragments.all():
+                    self.textblock_set.add(textblock)
+
+            # combine footnotes
+            current_footnotes = self.footnotes.all()
+            for footnote in doc.footnotes.all():
+                # if footnote is not present (based on footnote eqauality check), add it
```
thatbudakguy, line 648: eqauality

geniza/corpus/models.py
thatbudakguy: pgpid here too

geniza/footnotes/models.py
```diff
@@ -224,6 +224,15 @@ def __str__(self):
...
 
+    def __eq__(self, o):
+        """Footnotes should be considered if content source,
+        location, document relation type, notes, and content all match.
+        Does not include associated content object in comparison check."""
+        return all(
+            getattr(self, val) == getattr(o, val)
+            for val in ["source", "location", "doc_relation", "notes", "content"]
+        )
+
```
thatbudakguy, line 235: i think "other" is a little more readable, unless "o" is a django convention or something

thatbudakguy: also i think docstring should say "considered equal"

thatbudakguy (approved):
footnote comparison logic & tests look good!

geniza/footnotes/models.py
rlskoeser: The `o` was from a vscode template that popped in when I started the method! I should have revised, IDK why I didn't. And thanks for catching the missing word equal there

geniza/corpus/models.py
rlskoeser: good catch! thank you, I'd forgotten about this one

rlskoeser:
@thatbudakguy the merge script is currently named `merge_joins` but I think it should actually be `merge_documents` since there are non-join duplicate docs that need to be merged. The reporting logic is looking for joins, but the merge logic doesn't care about them. Any opinion?

thatbudakguy:
makes sense to me; the code doesn't care whether the merged docs are joins or not.

PR #185, opened 2021-05-21
Chore/editor import revisions1000

rlskoeser:
further refinements to editor import, notably:
- create unknown source & footnote if there is transcription content even if the editor text is in the ignore list
- add book section source type and recognize book sections similar to journals
- generalize language recognition so it will handle more cases
- put unparsed editor information into an "other_info" field for extra citation details (e.g. dissertation institution); this should be included in the display method when we get that branch merged
- clean up source notes so they aren't as messy and don't include redundant information

(Yes, more complications to the already very complicated editor parsing code. This is the last round, I promise! And we can delete it in the next release.)


geniza/corpus/management/commands/import_data.py
```diff
@@ -298,7 +299,8 @@ def add_document_language(self, doc, row):
 
     def import_document(self, row):
         """Import a single document given a row from a PGP spreadsheet"""
-        if ";" in row.type:
+        # skip any row with multiple types or flagged for demerge
+        if ";" in row.type or "DISAGGREGATE" in row.notes:
```
kmcelwee, line 303: "DISAGGREGATE" occurs in the "Technical notes (optional)" column for 8918 rows, whereas in "Notes2 (optional) it occurs in 1925 rows. Should we consider both? Has that already been discussed?

kmcelwee (approved):

geniza/corpus/tests/test_import_data.py
```diff
@@ -297,6 +297,7 @@ def test_import_document():
 
     DocumentCSVRow = namedtuple(
         "DocumentCSVRow",
+        # NOTE: could we use import_data.csv_fields['metadata'].values() ?
```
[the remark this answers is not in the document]
rlskoeser: yeah — I tried it and broke things! so I put it back 😂 

I guess we can clean it up when we remove all the import code in the next release

geniza/corpus/management/commands/import_data.py
rlskoeser: @kmcelwee the last time I looked, there were no more DISAGGREGATE entries in the technical notes (and I just confirmed via colab). I think they've been handled, and thus were removed. Otherwise, we should definitely check there also! 

Can you easily check if the records we're skipping match up with what's in the demerge spreadsheet? Or would the import reporting flag it if they weren't ?

rlskoeser: I'm going to go ahead and merge — we can tweak the disaggregate skipping logic if needed

Issue #107, opened 2021-03-18
As a user, I want to see updates and changes made in the new database in the current pgp site while the new website is still in development so that I can reference current information. 


thatbudakguy:
**Is your feature request related to a problem? Please describe.**
During the transition period after the data has been imported to the new database, changes won't be manifested on the current PGP site unless the metadata google sheet is also updated.

**Describe the solution you'd like**
The existing PGP site should be updated to automatically pull a new CSV or JSON feed with data from the database that matches the fields used from the existing spreadsheet. 

We should create a new view that returns CSV as a streaming response so that data can be consumed before the full result is generated.  Use https://github.com/LibraryOfCongress/django-tabular-export and integrate into a view instead of django admin

**Additional context**
Details provided by Ben:

> It can be CSV or JSON, whichever you prefer. I can also test the ingest from your data at any time with a local copy of the database, not a problem.
> The following are the bits of info I getting from the spreadsheet and the corresponding column in the spreadsheet (starting with 0 for the first)

pgpid    column 0
~notes    column 1~
library    column 2
shelfmark    column 3
shelfmark_alt    column 5
recto_verso     column 7
type    column 9
tags    column 10
joins    column 11
description    column 12
editor    column 13

## dev notes
- [x] create new corpus view and configure a url
- [x] use `document` queryset to get a list of public documents
- [x] use the queryset to create a generator with the needed fields
- [x] use `tabular_export.core.export_to_csv_response` to render
 
open questions / suggestions:
- current pgp models documents and joins slightly differently; propose we return the first shelfmark only in the shelfmark column and the full shelfmark in the join
- suggest we create a footnote method to output edition information in a similar form to what was in the spreadsheet; does not have to match exactly.

rlskoeser:
@mrustow The list of fields Ben says he uses from the current spreadsheet includes the **Technical Notes** column, which as far as I know we were not planning to import into the database. Please advise.

mrustow:
Technical notes is a mess, but contains valuable info on action items. Would be easier to get these sorted out in the spreadsheet, I would think — not sure.

rlskoeser:
Confirmed with Ben we can drop the technical notes field from the new version of the data export.

richmanrachel:
@rlskoeser - just to clarify, this is for Ben to test, not me?

rlskoeser:
@richmanrachel yes! working on getting him access, sorry if adding the label was confusing

rlskoeser:
[...]

PR #191, opened 2021-05-24
Revise old pgp csv metadata view; add prefetching, tweak output #107

rlskoeser:
- added prefetching to make the download generate faster
- renamed some of the methods to make the names more accurate/specific, and make it clearer they are associated with the old pgp metadata export
- adjusted the output of some of the fields to better match what I expected the content to be (my fault for not documenting more clearly and not catching in review)

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -329,7 +329,8 @@ class Document(ModelIndexable):
         return " + ".join(
             dict.fromkeys(
                 block.fragment.shelfmark
-                for block in self.textblock_set.filter(certain=True)
+                for block in self.textblock_set.all()
+                if block.certain  # filter locally instead of in the db
             )
         )
 
--- a/geniza/corpus/views.py
+++ b/geniza/corpus/views.py
@@ -1,20 +1,14 @@
-import csv
-from collections import namedtuple
-
-from django.shortcuts import render
-from django.http import Http404, StreamingHttpResponse
-from tabular_export.admin import export_to_csv_response
+from django.db.models.query import Prefetch
 from django.views.generic.detail import DetailView
+from tabular_export.admin import export_to_csv_response
 
-from django.db.models import Q
 from geniza.corpus.models import Document, TextBlock
 from geniza.corpus.admin import DocumentAdmin
 from geniza.footnotes.models import Footnote
 
-from unittest.mock import Mock
-
 
 class DocumentDetailView(DetailView):
+    """public display of a single :class:`~geniza.corpus.models.Document`"""
 
     model = Document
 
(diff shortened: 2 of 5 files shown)


geniza/corpus/models.py
```diff
@@ -329,7 +329,8 @@ def shelfmark(self):
         return " + ".join(
             dict.fromkeys(
                 block.fragment.shelfmark
-                for block in self.textblock_set.filter(certain=True)
+                for block in self.textblock_set.all()
+                if block.certain  # filter locally instead of in the db
```
thatbudakguy, line 333: weird that this was faster...did it also change/speed up for e.g. the admin list view? I thought at one point I optimized the `get_queryset()` for a list view to prefetch textblocks so that this would be faster, so it'd be a shame to have lost that

geniza/corpus/views.py
```diff
@@ -29,51 +23,78 @@ def get_queryset(self, *args, **kwargs):
...
 
 
-def tabulate_queryset(queryset):
+def old_pgp_tabulate_data(queryset):
+    """Takes a :class:`~geniza.corpus.models.Document` queryset and
+    yields rows of data for serialization as csv in :method:`pgp_metadata_for_old_site`"""
     # NOTE: This logic assumes that documents will always have a fragment
     for doc in queryset:
         primary_fragment = doc.textblock_set.first().fragment
+        num_fragments = len([tb for tb in doc.textblock_set.all() if tb.certain])
```
thatbudakguy, line 53: the naming of `num_fragments` suggests this counts fragments, but doesn't it count `TextBlock`s? currently we can have multiple `TextBlock`s between a single `Document` and `Fragment` pair.

thatbudakguy (approved):

geniza/corpus/models.py
thatbudakguy: but i guess prefetching only certain ones is impossible, so it's one query either way, and i can see how making a bunch more per document might be inefficient

rlskoeser: you added optimization specifically for this? 

I have a separate branch where I'm working on the admin csv download speedup, and will share all my refactoring and time testing notes soon if you're interested :-) 

I did add some prefetching and select related to the queryset used for this export; full thing is downloading locally for me in a little under 2 minutes, which seems like it might be good enough? (although I wish it was faster)

geniza/corpus/views.py
rlskoeser: I don't actually care how many they are, I just wanted a quick check to see if there is more than one — it's used to determine whether we put the combined shelfmark in the join column; I thought it would be faster to count vs  generating the full shelfmark and checking for the presence of "+"

thatbudakguy: right, but in the case described above, wouldn't that be erroneously counted as a join? we might not actually have any at the moment, not sure

rlskoeser: oh! I see what you mean. yes, it might ... I don't think we have any like that currently. Suggestion for a quick revision that would be more future proof?

thatbudakguy: i think it's just the same logic that gets used to generate the shelfmark – the only surefire way to know if something is a join is if it has more than one unique, certain fragment. order doesn't matter here though, so you could do:
```py
set([tb.fragment for tb in doc.textblock_set.all() if tb.certain])
```

PR #192, opened 2021-05-25
Add old pgpid field with postgres array

kmcelwee:
How it would currently be visible in the admin interface: 

Screen Shot 2021-05-24 at 10 24 49 PM

Styling this might isn't a priority, but the `{}` isn't too pretty. I am also assuming that we won't be searching on this field. 

Should I overwrite the `save()` function to append the former pgpid using the `TrackChangesModel` we've already used to track old shelfmarks for Fragment?


geniza/corpus/admin.py
```diff
@@ -167,14 +167,15 @@ class DocumentAdmin(admin.ModelAdmin):
...
-    readonly_fields = ("created", "last_modified", "shelfmark", "id")
+    readonly_fields = ("created", "last_modified", "shelfmark", "id", "old_pgpids")
     search_fields = (
         "fragments__shelfmark",
         "tags__name",
         "description",
         "notes",
         "needs_review",
         "id",
+        "old_pgpids",
```
rlskoeser, line 178: ooh, I should have added a comment here — these search fields are ignored since we use solr for document search. We should add old pgpids to the index data and admin search, though. If that's not obvious to you I can throw it in!

rlskoeser (commented):
Thanks for working on this @kmcelwee !

old pgpid does need to be searchable, LMK if you want me to add that

> Styling this might isn't a priority, but the `{}` isn't too pretty. I am also assuming that we won't be searching on this field.

I agree, not great — could you add a quick display version that converts to delimited string?

> Should I overwrite the `save()` function to append the former pgpid using the `TrackChangesModel` we've already used to track old shelfmarks for Fragment?

No need! There's no way for a user to change the id, so it wouldn't happen there (right now it can only be set on import or by the automerge script we're going to write).

This looks good but I think we need search first — although could be persuaded to merge it without so we can proceed with the work that needs this field 🙂

geniza/corpus/admin.py
richmanrachel: Not sure if I'm jumping into the wrong conversation, but saw on Slack the earlier comment... @kmcelwee - I think we do need to be able to search for old PGP IDs! Especially with all of the demerge work happening, it's crucial we have access to searching this way.

kmcelwee:
Screen Shot 2021-05-25 at 10 40 58 AM

I'm kinda confused... I didn't do anything but the display is fixed now (?!)

PR #193, opened 2021-05-25
Feature/179b parse technical notes

kmcelwee:
Questions are noted inline.

### DIFF
--- a/geniza/corpus/management/commands/import_data.py
+++ b/geniza/corpus/management/commands/import_data.py
@@ -305,11 +305,7 @@ class Command(BaseCommand):
         return None
         # ?? Should this be []
 
-    def parse_notes(self, doc, row):
-        if "PGPID" in row.notes:
-            old_pgpids = row.notes.split(": ")[-1]
-            doc.old_pgpids = [int(old_pgpid) for old_pgpid in old_pgpids.split(", ")]
-
+    def get_notes(self, notes, tech_notes):
         IGNORE = [
             "India",
             "India; unsure of date entered by Geniza Lab team (historical, before 2015)",
@@ -319,8 +315,8 @@ class Command(BaseCommand):
         ]
 
         notes_to_set = []
-        if row.notes not in IGNORE and "PGPID" not in row.notes:
-            notes_to_set.append(row.notes)
+        if notes not in IGNORE and "PGPID" not in notes:
+            notes_to_set.append(notes)
 
         ## TECHNICAL NOTES
         # Unique values:
@@ -339,33 +335,23 @@ class Command(BaseCommand):
             scanned in drive (TRANSCRIPTION & TRANSLATION)
         """
 
-        if row.tech_notes == "not in Gil":
+        if tech_notes == "not in Gil":
             GIL_NOTE = "Not published by Gil, pace FGP."
             notes_to_set.append(GIL_NOTE)
 
-        if "scanned in drive" == row.tech_notes.strip("?"):
+        if "scanned in drive" == tech_notes.strip("?"):
             SCANNED_GOITEIN_NOTE = "Look for Goitein scan in this folder: https://drive.google.com/drive/folders/1ZAWSK3ILoRyll0FafU61V5zdccx0CgaP?usp=sharing"
             notes_to_set.append(SCANNED_GOITEIN_NOTE)
-        elif "scanned in drive" in row.tech_notes:
+        elif "scanned in drive" in tech_notes:
             detail = " and ".join(
-                [
-                    t
-                    for t in ["transcription", "translation"]
-                    if t in row.tech_notes.lower()
-                ]
+                [t for t in ["transcription", "translation"] if t in tech_notes.lower()]
             )
             DETAIL_GOTEIN_SCAN_NOTE = (
                 f"There is a {detail} in Goitein's notes that should be digitized."
             )
             notes_to_set.append(DETAIL_GOTEIN_SCAN_NOTE)
 
-        # NOTE: This assumes that this is the only (or at least first) time
-        #  the notes field is set.
-        doc.notes = "\n".join(notes_to_set)
-        doc.save()
-        # ??: Should there be a doc.save() here?
-        #  It's addition was necessary to get test_import_document passing.
-        #  Should there be one at the end of the script?
+        return "\n".join(notes_to_set)
 
     def import_document(self, row):
         """Import a single document given a row from a PGP spreadsheet"""
@@ -388,6 +374,7 @@ class Command(BaseCommand):
             doctype=doctype,
             description=row.description,
             old_pgpids=self.get_old_pgpids(row.notes),
+            notes=self.get_notes(row.notes, row.tech_notes),
         )
         doc.tags.add(*[tag.strip() for tag in row.tags.split("#") if tag.strip()])
         # associate fragment via text block
@@ -400,7 +387,6 @@ class Command(BaseCommand):
             subfragment=row.multifragment,
         )
         self.add_document_language(doc, row)
-        self.parse_notes(doc, row)
         self.docstats["documents"] += 1
         # create log entries as we go
         self.log_edit_history(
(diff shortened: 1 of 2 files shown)


geniza/corpus/management/commands/import_data.py
```diff
@@ -297,6 +298,68 @@ def add_document_language(self, doc, row):
...
+            scanned in drive (TRANSCRIPTION + TRANSLATION)
+            scanned in drive (TRANSCRIPTION & TRANSLATION)
+        """
+
+        if row.tech_notes == "not in Gil":
+            GIL_NOTE = "Not published by Gil, pace FGP."
+            notes_to_set.append(GIL_NOTE)
+
+        if "scanned in drive" == row.tech_notes.strip("?"):
+            SCANNED_GOITEIN_NOTE = "Look for Goitein scan in this folder: https://drive.google.com/drive/folders/1ZAWSK3ILoRyll0FafU61V5zdccx0CgaP?usp=sharing"
```
kmcelwee, line 340: Should this be an `a` tag? And should this link be included in the other note as well?

rlskoeser: no, notes doesn't support formatting; it's fine to leave it as is

geniza/corpus/management/commands/import_data.py
```diff
@@ -297,6 +298,68 @@ def add_document_language(self, doc, row):
             doc.language_note = "\n".join(notes_list)
             doc.save()
 
+    def parse_notes(self, doc, row):
+        if "PGPID" in row.notes:
+            old_pgpids = row.notes.split(": ")[-1]
+            doc.old_pgpids = [int(old_pgpid) for old_pgpid in old_pgpids.split(", ")]
+
```
rlskoeser, line 305: Can we split this into two separate functions for old pgpids and notes so that it can return the value (if any) to use when _creating_ the document, so that we don't have to update it again after the fact? It will help if we can avoid an additional save call, and it seems like it will be easier to test the logic too.

What I imagined is something like this:
```
doc = Document.objects.create(
                  id=row.pgpid or None,
                  old_pgpids=self.get_oldpgids(row.notes),
                  doctype=doctype,
                  description=row.description,
                  notes=self.get_notes(row)
              )
```

rlskoeser: Does this work for all cases? PGPID doesn't occur anywhere else in the notes as a non-old PGPID?

geniza/corpus/management/commands/import_data.py
```diff
@@ -297,6 +298,68 @@ def add_document_language(self, doc, row):
...
+            "India",
+            "India; unsure of date entered by Geniza Lab team (historical, before 2015)",
+            "India: addendum to IB IV",
+            "Rustow, Lost Archive",
+            "DISAGGREGATE",
+        ]
+
+        notes_to_set = []
+        if row.notes not in IGNORE:
+            notes_to_set.append(row.notes)
```
rlskoeser, line 316: Will this also ignore the old pgpids? They shouldn't be added to the notes text

geniza/corpus/management/commands/import_data.py
```diff
@@ -297,6 +298,68 @@ def add_document_language(self, doc, row):
...
+            Transcription needs to be uploaded*
+            Transcription needs to be uploaded
+            scanned in drive (TRANSCRIPTION)
+            scanned in drive (TRANSLATION)
+            not in Gil
+            Scanned in drive
+            scanned in drive 
+            scanned in drive (TRANSCRIPTION + TRANSLATION)
+            scanned in drive (TRANSCRIPTION & TRANSLATION)
+        """
```
rlskoeser, line 333: Simpler to make this a dict lookup of old text to new text, at least for the simple cases?

geniza/corpus/tests/test_import_data.py
```diff
@@ -821,6 +823,69 @@ def test_update_document_id_sequence():
...
+    doc = Document.objects.create()
+    row = AttrMap({"notes": "India; Old PGPID: 9242", "tech_notes": ""})
+    import_data.Command().parse_notes(doc, row)
+    assert 9242 in doc.old_pgpids
+
+    # Make sure ignored notes aren't included
+    doc = Document.objects.create()
+    row = AttrMap({"notes": "DISAGGREGATE", "tech_notes": ""})
+    import_data.Command().parse_notes(doc, row)
+    assert doc.notes == ""
```
rlskoeser, line 853: great to see all these use cases tested

please add test to confirm old pgpid string doesn't get included in note text

geniza/corpus/management/commands/import_data.py
```diff
@@ -297,6 +298,68 @@ def add_document_language(self, doc, row):
...
+            detail = " and ".join(
+                [
+                    t
+                    for t in ["transcription", "translation"]
+                    if t in row.tech_notes.lower()
+                ]
+            )
+            DETAIL_GOTEIN_SCAN_NOTE = (
+                f"There is a {detail} in Goitein's notes that should be digitized."
+            )
```
rlskoeser, line 352: this seems fine, but happy to help with a regex if you want to go that route 🙂

geniza/corpus/management/commands/import_data.py
kmcelwee: No. All instances when "PGPID" is in the note2 column:
```
Old PGPID: 6160
Old PGPID: 6263
Old PGPID: 8843
India; Old PGPID: 9242
India; Old PGPID: 9261
India; Old PGPID: 9263
Old PGPID: 9563
Old PGPID: 11525
India; old PGPID: 6928
Old PGPID: 5711
India; old PGPID: 6580
India; old PGPID: 6589
India; Old PGPID: 9243
India; old PGPID: 16829
India; old PGPID: 9626
India; old PGPID: 5773
India; old PGPID: 30077
India; old PGPID: 6951
"Old PGPIDs: 9382, 16089"
Old PGPID: 7200
Old PGPID: 5861
Old PGPID: 5876
Old PGPID: 7988
India; Old PGPID: 8011
Old PGPID: 8037
Old PGPID: 7966
India; Old PGPID: 9207
Old PGPID: 7971
OId PGPID: 1727
Old PGPID: 7705
Old PGPID: 7627
Old PGPID: 7630
Old PGPID: 7631
Old PGPID: 7622
Old PGPID: 7731
Old PGPID: 7733
Old PGPID: 7742
India; Old PGPID: 7737
Old PGPID: 7738
India; Old PGPID: 8600
Old PGPID: 8631
India; old PGPID: 9636
India; old PGPID: 9416
old PGPID: 6948
```

geniza/corpus/management/commands/import_data.py
kmcelwee: mmmmmmm yeah I don't know. It works as it is, but I agree that a lookup would be just as good 🤷‍♀️

geniza/corpus/management/commands/import_data.py
kmcelwee: Sure, if it's worth it for tomorrow 🤷‍♀️

geniza/corpus/management/commands/import_data.py
rlskoeser: you're right! if it's working and we don't need to mess with it, just get it done for tomorrow

geniza/corpus/management/commands/import_data.py
```diff
@@ -297,6 +298,61 @@ def add_document_language(self, doc, row):
             doc.language_note = "\n".join(notes_list)
             doc.save()
 
+    def get_old_pgpids(self, notes):
+        if "PGPID" in notes:
+            old_pgpids = notes.split(": ")[-1]
+            return [int(old_pgpid) for old_pgpid in old_pgpids.split(", ")]
+        return None
```
rlskoeser, line 305: from my experiment with the array field, it looks like `None` is fine?

rlskoeser (approved):
looks great

PR #195, opened 2021-05-25
Include footnote url in document and source inlines

rlskoeser:
@richmanrachel I'm adding url as an editable field to the inlines on document and source pages

Here's how it looks on the document edit page:
Screen Shot 2021-05-25 at 12 01 49 PM

And here's how it looks on the source edit page:
Screen Shot 2021-05-25 at 12 01 41 PM

Do these look ok?


richmanrachel:
@rlskoeser - Looks perfect! Thanks!

PR #197, opened 2021-05-25
Include old PGPIDs in solr index data and add to default search fields

rlskoeser:
I haven't tested locally by updating my configset and reindexing (and don't have any records with old pgpids yet to test with), but I think this should do what we need.

### DIFF
--- a/geniza/corpus/solr_queryset.py
+++ b/geniza/corpus/solr_queryset.py
@@ -18,7 +18,7 @@ class DocumentSolrQuerySet(AliasedSolrQuerySet):
         "notes": "notes_t",
         "needs_review": "needs_review_t",
         "pgpid": "pgpid_i",
-        "old_pgpids": "old_pgpid_i",
+        "old_pgpids": "old_pgpid_is",
     }
 
     # (adapted from mep)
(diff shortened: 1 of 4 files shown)


geniza/corpus/solr_queryset.py
```diff
@@ -18,6 +18,7 @@ class DocumentSolrQuerySet(AliasedSolrQuerySet):
         "notes": "notes_t",
         "needs_review": "needs_review_t",
         "pgpid": "pgpid_i",
+        "old_pgpids": "old_pgpid_i",
```
kmcelwee, line 21: Is this supposed to be `old_pgpid_is`?

kmcelwee (approved):

geniza/corpus/solr_queryset.py
rlskoeser: yes!

thatbudakguy, line 21: should this be `old_pgpid_is` (not `old_pgpid_i`) to match the field def in `Document.index_data()`?

thatbudakguy (approved):

Issue #182, opened 2021-05-20
As a user, I want to search documents by keyword or phrase so that I can find materials related to my interests.


rlskoeser:
## testing notes

Test at https://test-geniza.cdh.princeton.edu/documents/ 

This is a preliminary version of the keyword search, and is missing a lot of functionality (see below for some of the many things that are out of scope). Also, be aware that not all edits in the database will immediately propagate to the search index (notably, changes to scholarship records currently do not); we will set up a cron job to reindex nightly until we're able to get on-demand indexing implemented for footnote/source edits.

You should be able to search across content in all of the following fields:
- [x] description
- [x] pgpid
- [x]  type
- [x] library / collection
- [x] shelfmark (including join notation  — the + should be ignored)
- [x] tags
- [x] scholarship record (source/footnote, including notes; try searching a scholar like Udovitch not mentioned in description)
- [x] old pgpid. Here are a few you can test with: 6160 (5575);  9382 or 16089 (2752), 9626 (9220)

Search result information should include:
- [x] document type + shelfmark (this is preliminary, does not look exactly like the final display in the design)
- [x] input date if known
- [x] pgp id
- [x] description excerpt
- [x] indicator and count if document has transcription, translation, discussion (AFAIK there are no footnotes in the db currently marked discussion; if you want to test this, you can edit one to change it and ask me to reindex)
- [x] link to document detail page

--- 

## dev notes

Preliminary implementation for public keyword search

Intentionally keeping this limited to try to keep it manageable. Out of scope functionality includes: pagination, sorting, image display, transcription search, keywords in context

richmanrachel:
@rlskoeser 
- Type isn't working. ("Legal" produces a bunch of documents because it's used in descriptions and tags, but "paraliterary" produces zero results.)
- Shelfmark doesn't work. I got way too many results for a solitary shelfmark and no results for a shelfmark join that I copy-pasted from that previous results list:

![image](https://user-images.githubusercontent.com/68028846/120532778-0a54fd80-c3ae-11eb-9af4-31316763899b.png)


![image](https://user-images.githubusercontent.com/68028846/120532713-f5786a00-c3ad-11eb-806e-7a194fe14b8b.png)

richmanrachel:
Everything else works!

rlskoeser:
@richmanrachel 

> * Type isn't working. ("Legal" produces a bunch of documents because it's used in descriptions and tags, but "paraliterary" produces zero results.)

[...]

Issue #114, opened 2021-03-25
As a content editor, I want to be able to filter/search within the “needs review” field of the list of items "awaiting review," so that I can more efficiently review issues of the same type in batches (docs that need fixes to description, shelfmark, etc).


mrustow:
**Fields**
Provide a list of the fields to be included on this model. 
- field 
- field

**List Fields**
Which fields should be included on the django admin list view?
- field
- field 

**Edit Fields**
Which fields should be editable via django admin view? Please list them in the order you want them to appear. Indicate which fields are optional, and any fields that should be displayed but not editable.
- field
- field 
- field (optional)
- field (read only)

**Search Fields**
Which fields should be searchable in the django admin list view?
- field
- field 

**Default sort**
Which field should be used for initial sorting?

**Filters**
Which fields should be used for list filters?

**Date Hierarchy (optional)**
Which field (if any) should be used for date hierarchy? (must be a date object)

**List Fields (optional)**
Which fields (if any) should be editable on the django admin list view?

**Field translation (optional)**
Which fields (if any) do you want to provide in multiple languages?

**Related models**
List any directly related models (database tables) that would be helpful to edit when on the same page when editing this one.

**Additional context**
Add any other context about the model or database here. Include a link to a database diagram if one is available.

mrustow:
@rlskoeser which of the fields do we actually need to specify for this?

rlskoeser:
@mrustow if you give us a list of keywords that you want available for filtering within the `needs_review` field (and develop a workflow to make sure your team uses those keywords), we could write a custom filter that would do this.  However, this approach would require us to update the list of terms in the code whenever you want to add a new item to the list, which is obviously not ideal for anyone.

One solution that other project teams have used is to develop a list of codes or keyword terms that you can search for without getting irrelevant matches for terms used in other fields (e.g., could be something like NEED_DESCR or NEED_SHLFMRK).

I've made the needs review field search able in admin. Would you try adding some test review notes and using the existing needs review filter & keyword search to see if that gets you the functionality you need?

rlskoeser:
duplicate of #383 and also probably addressed/improved by field-based advanced saerch

PR #198, opened 2021-05-25
preliminary keyword search for public site

rlskoeser:
This adds a preliminary keyword search for the public site; see #182 for the constraints I added to try to make this a minimal first implementation. I don't have a sense of how much work is left for it to be useful or for us to be comfortable with merging it.

I haven't added any tests yet, and got hung up trying to add signal-based indexing for footnotes/sources/creators, but it occurred to me we could set up a cron job to do nightly reindexing for now.


rlskoeser:
@thatbudakguy @kmcelwee I've updated the `index_data` a bit more and added tests for the new fields, but have not added any tests for the view or form. Would appreciate feedback

geniza/corpus/forms.py
```diff
@@ -0,0 +1,75 @@
...
+class DocumentSearchForm(forms.Form):
+    query = forms.CharField(
+        label="Keyword or Phrase",
+        required=False,
+        widget=forms.TextInput(
+            attrs={
+                "placeholder": "search by keyword",
+                "aria-label": "Keyword or Phrase",
+            }
+        ),
```
thatbudakguy, line 48: I tested adding `"type": "search",` here to get [search input behavior](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search), which is nice – it remembers previous entries, and on some platforms also adds a little search indicator or "clear field" button on its own.

geniza/corpus/forms.py
```diff
@@ -0,0 +1,75 @@
...
+        Override to set choices dynamically based on form kwargs.
+        """
+        super().__init__(data=data, *args, **kwargs)
+
+        # if a keyword search term is not present, relevance sort is disabled
+        if data and not data.get("query", None):
+            self.fields["sort"].widget.choices[0] = (
+                self.SORT_CHOICES[0][0],
+                {"label": self.SORT_CHOICES[0][1], "disabled": True},
+            )
```
thatbudakguy, line 75: when i tested, relevance was set by default when the page loads, despite there being no query present (although i don't think it's actually sorting by relevance). maybe just doing
```py
if not data or not data.get("query", None):
```
would fix this?

geniza/corpus/models.py
```diff
@@ -434,28 +451,78 @@ def editions(self):
...
+            "languages",
+            "footnotes",
+            "log_entries",
+            Prefetch(
+                "textblock_set",
+                queryset=TextBlock.objects.select_related(
+                    "fragment", "fragment__collection"
+                ),
+            ),
+        )
```
thatbudakguy, line 468: Love the idea to prefetch/optimize these index-time queries, and would be great to see if we can extract common parts as queryset filters!

geniza/corpus/models.py
```diff
@@ -434,28 +451,78 @@ def editions(self):
...
-                # TODO: editors/translators/sources
             }
         )
 
+        # count scholarship records by type
+        footnotes = self.footnotes.all()
+        counts = defaultdict(int)
+        for fn in footnotes:
+            for val in fn.doc_relation:
+                counts[val] += 1
```
thatbudakguy, line 498: wonder if this thing could also become a method that uses annotation/aggregation to accomplish this faster, although maybe we won't really have enough footnotes of each type to make it worthwhile

geniza/corpus/models.py
```diff
@@ -434,28 +451,78 @@ def editions(self):
...
+            index_data["input_year_i"] = last_log_entry.action_time.year
+        # image indexing was causing timeouts with cudl manifests,
+        # hold off on that for now
+        # images = img_labels = []
+        # for fragment in self.fragments.all():
+        #     imgs = fragment.iiif_images()
+        #     if imgs:
+        #         images.extend(imgs[0])
+        #         img_labels.extend(imgs[1])
+        # index_data['images_s'] = images
```
thatbudakguy, line 523: what kind of data actually gets indexed in this field?

geniza/corpus/templates/corpus/document_list.html
```diff
@@ -0,0 +1,59 @@
...
+
+    <form>
+    {{ form }}
+    {% for field, errors in form.errors.items %}
+    {# no clean way to get field labels here, so we just collapse the error lists #}
+        {% for error in errors %}
+            <li>{{ error }}</li>
+        {% endfor %}
+    {% endfor %}
+    </form>
```
thatbudakguy, line 21: could we get a "go" button or similar? might end up hiding it later depending on the implementation, but it's nice to have for now (not everyone will think to press <kbd>enter</kbd>) and we'll need it later regardless

geniza/corpus/templates/corpus/document_list.html
```diff
@@ -0,0 +1,59 @@
...
+        <li class="result">
+
+            <dl class="tags">
+                <dt class="sr-only">{% translate 'Tags' %}</dt>
+                {% for tag in document.tags %}
+                <dd class="tag">#{{ tag }}</dd>
+                {% endfor %}
+            </dl>
+
+            <h2><b>{{ document.type }},</b> {{ document.shelfmark|join:" + " }}</h2>
```
thatbudakguy, line 41: I think we want the shorter form here that displays a plus or some other indicator for joins, but not every shelfmark

geniza/corpus/templates/corpus/document_list.html
```diff
@@ -0,0 +1,59 @@
...
+            <p>{% translate 'Input date' %} {{ document.input_year|default:'unknown' }}
+            <p>{% translate 'PGP ID' %} {{ document.pgpid }}</p>
+            <p>{{ document.description.0|truncatechars:135 }}</p>
```
thatbudakguy, line 44: what's the `.0` do?

geniza/corpus/templates/corpus/document_list.html
```diff
@@ -0,0 +1,59 @@
...
+    <ol>
+    {% for document in documents %}
+        <li class="result">
+
+            <dl class="tags">
+                <dt class="sr-only">{% translate 'Tags' %}</dt>
+                {% for tag in document.tags %}
+                <dd class="tag">#{{ tag }}</dd>
+                {% endfor %}
+            </dl>
```
thatbudakguy, line 39: it's still a bit weird to me that these go here. semantically, the heading should be the first thing inside its parent element (the `<li>`). we could use CSS to move the tags above the record, but that feels a little weird too.

geniza/corpus/tests/test_corpus_models.py
```diff
@@ -403,15 +403,22 @@ def test_index_data(self, document):
...
             assert tag.name in index_data["tags_ss"]
         assert index_data["status_s"] == "Public"
         assert not index_data["old_pgpids_is"]
 
+        # test with notes and review notes
+        document.notes = "FGP stub"
+        document.needs_review = "check description"
+        index_data = document.index_data()
+        assert index_data["notes_t"] == document.notes
+        assert index_data["needs_review_t"] == document.needs_review
```
thatbudakguy, line 420: why do we index the "needs review" notes?

geniza/corpus/urls.py
```diff
@@ -1,10 +1,16 @@
...
+    DocumentDetailView,
+    DocumentSearchView,
+    pgp_metadata_for_old_site,
+)
+
 
 app_name = "corpus"
 
 urlpatterns = [
+    path("documents/", DocumentSearchView.as_view(), name="document-search"),
```
thatbudakguy, line 13: +1 for explicitly calling this `SearchView` instead of `ListView`!

geniza/corpus/views.py
```diff
@@ -1,11 +1,82 @@
...
+    model = Document
+    form_class = DocumentSearchForm
+    context_object_name = "documents"
+    template_name = "corpus/document_list.html"
+
+    # map form sort to solr sort field
+    solr_sort = {
+        "relevance": "-score",
+        #        'name': 'sort_name_isort'
+    }
```
thatbudakguy, line 22: does input date not need to go in here also?

thatbudakguy: sorry, can't tell how much sorting is actually implemented

thatbudakguy (approved):
tried locally and it does seem to be working the way I expect (though I'm sure once it gets further the team will find problem cases I don't know how to test for).

geniza/corpus/tests/test_corpus_models.py
thatbudakguy: oh, is it for #114? makes sense, if so

rlskoeser: yes, exactly — we need admin fields and suppressed documents indexed in solr for the admin document search, and will have to make sure we exclude them from the public search

geniza/corpus/templates/corpus/document_list.html
rlskoeser: I agree — I struggled a little with the counter placement too, until I figured out I could add it relative to the h1. Ok for a first implementation though, right?

geniza/corpus/templates/corpus/document_list.html
rlskoeser: displays the first element in a list — we're using `*_t` which is multivalued text

geniza/corpus/models.py
rlskoeser: hmm, interesting — we can probably generate counts in the database, but that would complicate the case when we index a single record

geniza/corpus/templates/corpus/document_list.html
thatbudakguy: yes, for sure – and nice work on the counter placement!

geniza/corpus/templates/corpus/document_list.html
thatbudakguy: maybe a dumb question, but why's the description multivalued?

geniza/corpus/forms.py
rlskoeser: I didn't really implement sort yet — think it should be a separate issue. Wasn't sure the best way to communicate that it shouldn't be expected to work  — disable the input, maybe?

The default page load isn't supposed to bring back any records (we display the tag network instead); it seemed simpler to me to leave them there for now. What do you think, is it problematic at all?

geniza/corpus/templates/corpus/document_list.html
rlskoeser: I don't think there's a text field that's not multivalued! But I can double check that, it would be nice if it wasn't

geniza/corpus/forms.py
thatbudakguy: ah, I'd forgotten about the tag network view. i think leaving the results there is fine, but for the sort dropdown I'd either not render it at all, or add a `hidden` attribute/completely hide it via CSS. we've used the disabled select to mean "this is active, but you can't change it" for sort before (and presumably will again here) so i think until it can function it's better to not show it.

thatbudakguy: alternatively we could just make sure there's no available options other than relevance, and keep it disabled – then only add sort options to the form once they're already implemented

geniza/corpus/templates/corpus/document_list.html
rlskoeser: I checked, all of the dynamic fields for text are multivalued. I guess we could think about adding one ourselves, but going to leave this for now.

geniza/corpus/models.py
rlskoeser: I was thinking to index the IIIF image urls so we could easily generate the thumbnails in the search results without an additional call, but stopped and decided it was out of scope when I got throttled by Cambridge when I was trying to index. I don't think indexing image urls on documents is the right solution, just not sure what is.

Issue #189, opened 2021-05-24
As a Content Editor, I want to download a CSV version of all or a filtered list of footnotes in the backend, in order to data work or facilitate my own research.


kmcelwee:
In the admin footnote list, you should have a button to download all footnotes as csv and a drop-down action to download footnotes sources. Here are the fields included in the csv:

- document
- source 
- location
- doc_relation (edition, translation, etc)
- notes
- url
- content (simple version of preliminary content from TEI, as displayed on admin site)
- admin_url

Please confirm you can download the whole thing as well as a selected set, and that the content looks correct.

richmanrachel:
@rlskoeser - Works perfectly! Closing :)

PR #199, opened 2021-05-25
Quick admin csv export for footnotes adapted from source/docs csv #189

rlskoeser:


geniza/footnotes/admin.py
```diff
@@ -356,6 +356,67 @@ def doc_relation_list(self, obj):
...
+                footnote.doc_relation,
+                footnote.notes,
+                footnote.url,
+                "\n".join(footnote.content["lines"]) if footnote.content else "",
+                f"{url_scheme}{site_domain}/admin/footnotes/footnote/{footnote.id}/change/",
+            ]
+
+    def export_to_csv(self, request, queryset=None):
+        """Stream source records as CSV"""
+        queryset = self.get_queryset(request) if queryset is None else queryset
```
thatbudakguy, line 397: does this do the same thing as `queryset = queryset or self.get_queryset(request)`?

geniza/footnotes/admin.py
```diff
@@ -356,6 +356,67 @@ def doc_relation_list(self, obj):
...
+        return export_to_csv_response(
+            self.csv_filename(),
+            self.csv_fields,
+            self.tabulate_queryset(queryset),
+        )
+
+    export_to_csv.short_description = "Export selected sources to CSV"
```
thatbudakguy, line 404: aren't we technically exporting footnotes?

geniza/footnotes/admin.py
```diff
@@ -356,6 +356,67 @@ def doc_relation_list(self, obj):
...
+
+    def get_urls(self):
+        """Return admin urls; adds a custom URL for exporting all sources
+        as CSV"""
```
thatbudakguy, line 408: same q about footnote vs. source

geniza/footnotes/admin.py
```diff
@@ -356,6 +356,67 @@ def doc_relation_list(self, obj):
...
+        site_domain = Site.objects.get_current().domain.rstrip("/")
+        # qa / prod always https
+        url_scheme = "https://"
+
+        for footnote in queryset:
+            yield [
+                footnote.content_object,
+                footnote.source,
+                footnote.location,
+                footnote.doc_relation,
```
thatbudakguy, line 388: `doc_relation` actually prints something like `Edition,Translation`, right? i remember it works a little weirdly with `__str__` where it's sometimes a friendly value and sometimes just `["E", "T"]`

thatbudakguy (approved):

geniza/footnotes/admin.py
rlskoeser: yeah, good catch — should refactor all three of these

geniza/footnotes/admin.py
rlskoeser: yeah, thanks. working too quickly!

geniza/footnotes/admin.py
rlskoeser: django-tabular-export converts to string for us, I'm getting the full label in the export

PR #200, opened 2021-05-26
last minute import tweaks

rlskoeser:
- propagate fgp stub technical note to db notes
- order document editions so the one with content is listed first


thatbudakguy (approved):
looks good!

PR #206, opened 2021-05-27
Feature/187 import iiif csv

kmcelwee:


geniza/corpus/management/commands/import_iiif_urls.py
```diff
@@ -0,0 +1,77 @@
...
+            csvreader = csv.reader(f)
+            header = next(csvreader)
+
+            # Create a namedtuple based on headers in the csv
+            # and local mapping of csv names to access names
+            CsvRow = namedtuple("IiifCsvRow", header)
+
+            # iterate over csv rows and yield a generator of the namedtuple
+            for row in csvreader:
+                yield CsvRow(*row)
```
thatbudakguy, line 54: would [`DictReader`](https://docs.python.org/3/library/csv.html#csv.DictReader) be simpler than creating a `namedtuple` type for rows? it should automatically parse the headers.

geniza/corpus/management/commands/import_iiif_urls.py
```diff
@@ -0,0 +1,77 @@
...
+        iiif_link = re.sub(r"/\d$", "", iiif_link)
+        return iiif_link
+
+    def import_iiif_url(self, row):
+        try:
+            fragment = Fragment.objects.get(shelfmark=row.shelfmark)
+        except Fragment.DoesNotExist:
+            # logger.warning(
+            #     f"Fragment with shelfmark {row.shelfmark} does not exist in the database."
+            # )
```
thatbudakguy, line 68: this seems like a useful warning!

geniza/corpus/management/commands/import_iiif_urls.py
```diff
@@ -0,0 +1,77 @@
...
+            return
+
+        if not fragment.iiif_url or self.overwrite:
+            fragment.iiif_url = self.view_to_iiif_url(row.url)
+
+            if self.dryrun:
+                logger.info(f"Set {fragment} url to {row.url}")
+            else:
+                fragment.save()
```
thatbudakguy, line 77: nice, this is cool

geniza/corpus/management/commands/import_iiif_urls.py
```diff
@@ -0,0 +1,77 @@
...
+class Command(BaseCommand):
+    """Given a CSV of fragments and IIIF URLs, add those IIIF urls to their respective fragments in the database"""
+
+    def __init__(self, *args, **options):
+        self.csv_path = options.get("csv")
+        self.overwrite = options.get("overwrite")
+        self.dryrun = options.get("dryrun")
+
+    def add_arguments(self, parser):
+        parser.add_argument("-c", "--csv", type=str, required=True)
```
thatbudakguy, line 26: since we always expect exactly one csv path, i'd consider this a positional argument rather than a flag

geniza/corpus/tests/test_import_iiif_urls.py
```diff
@@ -0,0 +1,95 @@
...
+    command.csv_path = "foo.csv"
+    csv_data = "\n".join(
+        [
+            "shelfmark,url",
+            "T-S NS 305.65,https://cudl.lib.cam.ac.uk/view/MS-TS-NS-00305-00065",
+            "T-S NS 305.69,https://cudl.lib.cam.ac.uk/view/MS-TS-NS-00305-00069",
+            "T-S NS 305.75,https://cudl.lib.cam.ac.uk/view/MS-TS-NS-00305-00075",
+        ]
+    )
+    mockfile = mock_open(read_data=csv_data)
```
thatbudakguy, line 21: did not know `mock_open` was a thing!! super useful

geniza/corpus/management/commands/import_iiif_urls.py
rlskoeser: fwiw, the namedtuple approach in the data import was a workaround so we could have nice names for the fields in  code without changing the headers in spreadsheet; I usually prefer dictreader

geniza/corpus/management/commands/import_iiif_urls.py
```diff
@@ -0,0 +1,77 @@
...
+
+            # iterate over csv rows and yield a generator of the namedtuple
+            for row in csvreader:
+                yield CsvRow(*row)
+
+    def view_to_iiif_url(self, url):
+        iiif_link = url.replace("/view/", "/iiif/")
+        # view links end with /1 or /2 but iiif link does not include it
+        iiif_link = re.sub(r"/\d$", "", iiif_link)
+        return iiif_link
```
rlskoeser, line 60: this logic is specific to cudl urls, but I don't think the script should be

kmcelwee: would you mind finishing this thought 😅 Should I add an `if 'cudl' in url` or something?

kmcelwee: I grabbed this from `import_data`. Should I also put this in `common.utils` or something?

rlskoeser: ha, sorry for being unclear! it looks like you didn't grab all the logic from the import script — please include the url check we have there

rlskoeser: I don't know yet if it belongs in a common util! I don't have a sense of whether we'll use it again anywhere else.

kmcelwee: ✅ d82f05e41179d6c65c8dab476d6372f6ea825795

geniza/corpus/management/commands/import_iiif_urls.py
kmcelwee: There are 22375 items in the shelfmark CSV, only ~7k of which are in our database. This floods the logs. I'm gonna leave it just commented out. Let me know if you still think it's worth keeping.

rlskoeser: oh, of course, because they have tons of non-documentary stuff

Probably remove entirely,  instead of leaving commented out.

Can you add very light-weight summary reporting at the end of the script? something like # added/updated/skipped

kmcelwee:
Pandas is telling me there should be ~7k matches, but here are the logs:
```
INFO:import:URLs added: 9
INFO:import:URLs updated: 32
INFO:import:Fragments not found: 22334
INFO:import:Fragments skipped: 0
```

kmcelwee:
Final logs look as expected from testing locally.
```
INFO:import:URLs added: 5981
INFO:import:URLs updated: 1540
INFO:import:Fragments not found: 14854
INFO:import:Fragments skipped: 0
```

Should we update existing urls?

geniza/corpus/management/commands/import_iiif_urls.py
```diff
@@ -37,6 +38,11 @@ def handle(self, *args, **options):
             for row in csvreader:
                 self.import_iiif_url(row)
 
+        logger.info(f"URLs added: {self.stats['added']}")
+        logger.info(f"URLs updated: {self.stats['updated']}")
+        logger.info(f"Fragments not found: {self.stats['not-found']}")
+        logger.info(f"Fragments skipped: {self.stats['skipped']}")
+
```
rlskoeser, line 45: FWIW, I prefer that we use django manage command output methods unless we have a compelling reason not to https://docs.djangoproject.com/en/3.2/howto/custom-management-commands/

rlskoeser (commented):
please add help text to the command that documents what headers are expected in the input csv

geniza/corpus/management/commands/import_iiif_urls.py
```diff
@@ -0,0 +1,85 @@
...
+LOG_LEVELS = {0: logging.ERROR, 1: logging.WARNING, 2: logging.INFO, 3: logging.DEBUG}
+
+
+class Command(BaseCommand):
+    """Given a CSV of fragments and IIIF URLs, add those IIIF urls to their respective fragments in the database"""
+
+    def __init__(self, *args, **options):
+        self.csv_path = options.get("csv")
+        self.overwrite = options.get("overwrite")
+        self.dryrun = options.get("dryrun")
```
rlskoeser, line 23: command options aren't available here, only in `handle`

geniza/corpus/management/commands/import_iiif_urls.py
```diff
@@ -0,0 +1,85 @@
...
+        parser.add_argument("-d", "--dryrun", action="store_true")
+
+    def handle(self, *args, **options):
+        self.csv_path = options.get("csv")
+        self.overwrite = options.get("overwrite")
+        self.dryrun = options.get("dryrun")
+
+        with open(self.csv_path) as f:
+            csvreader = csv.DictReader(f)
+            for row in csvreader:
```
rlskoeser, line 38: should probably bail out with an error if the csv headers aren't as expected (I say this since I just tried it with the old csv with wrong headers)

geniza/corpus/management/commands/import_iiif_urls.py
kmcelwee: Following the documentation isn't working for me (?) Getting the following error 
```
AttributeError: 'Command' object has no attribute 'stdout'
```
I simply tried replacing `logger.info` with `self.stdout.write`. If my mistake is obvious, let me know. Otherwise I'm just going to put this to the side.

rlskoeser: It's because you overrode the init method without calling super (sorry I didn't catch this before). I think it's not recommended to override init in manage commands, but you could check the django docs

rlskoeser:
@thatbudakguy this is ready for final review, would you take a look? The cambridge csv is on the qa server in the data directory if you want to test running it

geniza/corpus/management/commands/add_fragment_urls.py
```diff
@@ -0,0 +1,144 @@
...
+    """Takes a CSV of shelfmarks and view URLs and/or IIIF URLs, update
+    corresponding Fragment records in the database with those URLs.
+    Expects CSV headers 'shelfmark' and one or both of 'url' and 'iiif_url'"""
+
+    help = __doc__
+
+    def __init__(self, *args, **options):
+        super().__init__(*args, **options)
+
+        self.stats = defaultdict(int)
```
thatbudakguy, line 25: obvi not a blocker for merging, but for stat collection things like this I've used [`Counter`](https://docs.python.org/3/library/collections.html#collections.Counter), which behaves the same (counts of nonexistent elements are zero instead of `KeyError`) and I think might have a few optimizations under the hood

geniza/corpus/management/commands/add_fragment_urls.py
```diff
@@ -0,0 +1,144 @@
...
+                for row in csvreader:
+                    if "shelfmark" not in row or not (
+                        "url" in row or "iiif_url" in row
+                    ):
+                        raise CommandError(
+                            "CSV must include 'shelfmark' and one or both of 'url' and 'iiif_url'"
+                        )
+                    self.add_fragment_urls(row)
+        except FileNotFoundError as err:
+            raise CommandError(err)
```
thatbudakguy, line 55: fwiw when i tested this pattern on `merge_joins`, the `FileNotFoundError` doesn't actually report the path, so the error text is kinda useless. I just raised `CommandError(f"csv file not found: {self.csv_path}")` or something like that to make sure that it was displayed

geniza/corpus/management/commands/add_fragment_urls.py
```diff
@@ -0,0 +1,144 @@
...
+                    log_message.append("updated IIIF URL")
+
+        if save_needed:
+            if self.dryrun:
+                self.stdout.write(f"Set {fragment} url to {url} and iiif to {iiif_url}")
+            else:
+                fragment.save()
+                self.log_change(fragment, " and ".join(log_message))
+        else:
+            self.stats["skipped"] += 1
```
thatbudakguy, line 133: this is nice; i like avoiding the `.save()` call on things we don't change. probably speeds it up too

geniza/corpus/tests/test_corpus_admin.py
```diff
@@ -178,7 +178,7 @@ def test_save_model(self, db):
     def test_get_search_results(self, document, join):
         # index fixture data in solr
         Document.index_items([document, join])
-        time.sleep(1)
+        time.sleep(2)
```
thatbudakguy, line 181: low priority, but is there a way to poll solr in order to know when the operation finishes? or is it more like an eventually-consistent database, where the request will always complete but the database operation is never guaranteed to have happened at any particular time (just "eventually")

thatbudakguy (approved):
looks good!

geniza/corpus/tests/test_corpus_admin.py
rlskoeser: I don't know! The tests started failing in CI and it seemed like this was the culprit — I think it was after the indexing added more data.  However... I remember now we added a solr indexing config for tests in one of the other projects, maybe we can add that. (I don't think we've added it here)

geniza/corpus/management/commands/add_fragment_urls.py
rlskoeser: ah, thanks — I learned about this recently but I forget

Issue #207, opened 2021-05-27
Sorting fragments by collection raises a 500 error


kmcelwee:
## testing notes
in admin on the test site, go to the fragments list view and try to sort by collection — it should work and not raise an error. (Feel free to compare with production to see the difference)

--- 
Clicking on collection sort in the fragment admin causes an error. https://test-geniza.cdh.princeton.edu/admin/corpus/fragment/?o=2.1

Local debugging error:
```
Cannot resolve keyword 'lib_abbrev' into field.
```

richmanrachel:
@rlskoeser - it works fine for me on both the test and non-test site. Closing!

PR #208, opened 2021-05-27
Add workaround to sort fragments by collection in admin list #207

rlskoeser:


geniza/corpus/models.py
```diff
@@ -2,7 +2,7 @@
 
 from django.db import models
 from django.urls import reverse
-from django.db.models.functions import Concat
+from django.db.models.functions import Concat, Coalesce
```
thatbudakguy, line 5: does `Coalesce` get used?

thatbudakguy (approved):

geniza/corpus/models.py
rlskoeser: good catch, thanks — I tried it but it didn't help, guess I forgot to remove!

kmcelwee (approved):

PR #212, opened 2021-05-27
New Crowdin updates

thatbudakguy:

PR #213, opened 2021-05-27
feature/crowdin l10n

thatbudakguy:
This PR implements localization via an external platform (crowdin). You can access the project at https://crowdin.com/project/geniza (might need an invite from me; let me know if that's the case).

The basic workflow (feel free to try it) is as follows:
1. a translator or team member logs into the crowdin platform
2. they choose a language to translate
3. they select a file from the list (there will be only one file for each language, so it's impossible to select the wrong thing)
4. they proceed to translate each of the strings in the file, with automatic machine-assisted translations available
5. they finish translating and close the editor

Every ten minutes (configurable), crowdin will then push the changes made to the `.po` files to a branch in github. You can have one of these `l10n` branches per normal branch in your repo. For this example, I've set up one called `l10n_test` which is targeted at this branch (`feature/crowdin-l10n1`). Most likely we will ultimately want just one that is targeted at `develop`.

Crowdin automatically opens a pull request showing the changes that have been made, which we can merge at any time (a new PR will be opened once new changes appear). Have a look at https://github.com/Princeton-CDH/geniza/pull/212 to see what that looks like (I added some comments). We don't delete the branch after merging; instead we just leave it open so that the translators can always work in their own branch.

The `crowdin.yml` file controls what files crowdin looks at, and where it should save the changes that the translators make. It also has a setting to append text to a commit, which I set up to use the `[ci skip]` idiom so that CI doesn't run every time new translations are added (though it will still run on PRs between the `l10n` branch and `develop`).

This PR also gets rid of the app-specific `locale` folders and centralizes everything in a top-level `locale` folder. There's no need to run `makemessages` from within apps; you can just run it at the top level and everything should work.

Handling "microcopy"/text that we want the team to be able to change on their own is done by re-localizing the app into english: if you add an english translation, crowdin will store it in the same `en/django.po` file that is used as a source for the other translations. I'm not sure this is the best way to handle this – it's not how I saw other apps configured – but I also don't think other teams are using localization in order to accomplish this, and it does seem to work for now.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,7 +27,11 @@ Python 3.8 / Django 3.1 / Postgresql / Solr 8.6
 
 .. image:: https://img.shields.io/badge/code%20style-black-000000.svg
     :target: https://github.com/psf/black
-    :alt: code style: blacm
+    :alt: code style: Black
+
+.. image:: https://badges.crowdin.net/geniza/localized.svg
+    :target: https://crowdin.com/project/geniza
+    :alt: Localization status
 
 Development instructions
 ------------------------
--- a/README.rst
+++ b/README.rst
@@ -64,6 +64,7 @@ Remember to add a ``SECRET_KEY`` setting!
 - Create Solr collection with the configured configset (use create_core with Solr standalone and create_collection with SolrCloud)::
 
     solr create_core -c geniza -n geniza
+
 - Index content in Solr::
 
     python manage.py index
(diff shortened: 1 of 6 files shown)


README.rst
```diff
@@ -74,9 +79,13 @@ Internationalization & Translation
 
 This application has internationalization and translation enabled.
 
-- If you create any new translatable content, you should run `makemessages <https://docs.djangoproject.com/en/3.1/ref/django-admin/#makemessages>`_ to create or update message files.
+- If you create any new translatable content, you should run `makemessages <https://docs.djangoproject.com/en/3.1/ref/django-admin/#makemessages>`_ to create or update message files. We use a customized version of this command, available in ``/geniza/common/management/commands/makemessages.py``.
+
+	django-admin makemessages
+
+- Before running the app, you should run `compilemessages <https://docs.djangoproject.com/en/3.1/ref/django-admin/#compilemessages>`_ to generate compiled translations.
```
kmcelwee, line 86: Could you add this to `DEPLOYNOTES.md`?

kmcelwee (approved):
It's awesome how simple this is!

README.rst
thatbudakguy: I think we usually use `DEPLOYNOTES` for one-off tasks that need to run after a particular version is deployed, rather than tasks that should run every deploy. I'm working on a `django` role for our `cdh-ansible` repo that will run this command on every deploy (among other things): https://github.com/Princeton-CDH/cdh-ansible/pull/73

rlskoeser (approved):
Changes here all look reasonable! Can you add me to the project on crowdin so I can see how it looks there? (I created an account but I don't have access)

Issue #188, opened 2021-05-24
Stub all templates that use localizable site text


thatbudakguy:
See @gissoo's [doc with site text](https://docs.google.com/document/d/11aCR1nGLpgBHar_0nkXuh9Y0bL1oh-X-ZQvOQM6809k/edit?usp=sharing).

we need to create stub template files with placeholders for all these bits of UI text, which can first be implemented for English, and then later also localized.

<s>see also the Crowdin platform, where we could create a project and then [request open source access](https://crowdin.com/page/open-source-project-setup-request) to manage site text (and make it available in designs via a figma plugin).</s>

see [project on Crowdin](https://crowdin.com/project/geniza/) <s>GitHub integration is not yet configured</s>

## templates
- [x] doc detail page #110
- [x] scholarship records page a5f5c9a
- [x] external links page (no unique content)
- [x] doc search page #182

## snippets
- [x] doc detail page navigation a5f5c9a
- [x] main navigation 2c085c7

rlskoeser:
@thatbudakguy is this complete now?

thatbudakguy:
yep!

PR #214, opened 2021-05-27
feature/scholarship records

thatbudakguy:
this PR adds some new templates that have localizable content (for #188) and also implements a first-pass version of the scholarship records page (#119). the new `document_tabs.html` snippet handles rendering of the sub-page navigation for documents (details, scholarship records, external links).

no tests yet.

### DIFF
--- a/geniza/corpus/templates/corpus/document_scholarship.html
+++ b/geniza/corpus/templates/corpus/document_scholarship.html
@@ -1,0 +1,26 @@
+{% extends 'base.html' %}
+{% load static i18n %}
+
+{% block title %}{{ document.title }}{% endblock title %}
+
+{% block main %}
+<!-- document scholarship records -->
+<div class="container">
+    <h1>{{ document.title }}</h1>
+    {% include "corpus/snippets/document_tabs.html" %}
+    <ol>
+        {% for fn in document.footnotes.all %}
+        {% spaceless %}
+        <li><p class="footnote">
+        {% if fn.source.title %}<span class="title">{{ fn.source.title }}</span>{% endif %}
+        {% if fn.source.authors.exists %}<span class="author">{{ fn.source.all_authors }}</span>{% endif %}
+        {% if fn.source.year %}<span class="year">{{ fn.source.year }}</span>{% endif %}
+        <span class="relation">{{ fn.doc_relation }}</span>
+        {% if fn.location %}<span class="location">{{ fn.location }}</span>{% endif %}
+        {% if fn.url %}<a href="{{ fn.url }}">{% translate "Link to source" %}</a>{% endif %}
+        </p></li>
+        {% endspaceless %}
+        {% endfor %}
+    </ol>
+</div>
+{% endblock main %}
--- a/geniza/corpus/urls.py
+++ b/geniza/corpus/urls.py
@@ -3,14 +3,19 @@ from django.urls import path
 from geniza.corpus.views import (
     DocumentDetailView,
     DocumentSearchView,
+    DocumentScholarshipView,
     pgp_metadata_for_old_site,
 )
-
 
 app_name = "corpus"
 
 urlpatterns = [
     path("documents/", DocumentSearchView.as_view(), name="document-search"),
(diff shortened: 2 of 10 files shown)


geniza/corpus/views.py
```diff
@@ -19,6 +19,19 @@ def get_queryset(self, *args, **kwargs):
...
+class DocumentScholarshipView(DocumentDetailView):
+    """List of :class:`~geniza.footnotes.models.Footnote`s for a Document"""
+
+    template_name = "corpus/document_scholarship.html"
+
+    def get_queryset(self, *args, **kwargs):
+        """Don't show document if there are no footnotes"""
+        # prefetch footnotes since we'll render all of them in the template
+        queryset = super().get_queryset(*args, **kwargs)\
+                          .prefetch_related("footnotes")
```
rlskoeser, line 31: this needs a `.distinct()` added — I got an error when I tried to load this page for a document with two footnotes.

Also should filter on document status set to public

geniza/corpus/templates/corpus/document_scholarship.html
```diff
@@ -0,0 +1,26 @@
...
+        <li><p class="footnote">
+        {% if fn.source.title %}<span class="title">{{ fn.source.title }}</span>{% endif %}
+        {% if fn.source.authors.exists %}<span class="author">{{ fn.source.all_authors }}</span>{% endif %}
+        {% if fn.source.year %}<span class="year">{{ fn.source.year }}</span>{% endif %}
+        <span class="relation">{{ fn.doc_relation }}</span>
+        {% if fn.location %}<span class="location">{{ fn.location }}</span>{% endif %}
+        {% if fn.url %}<a href="{{ fn.url }}">{% translate "Link to source" %}</a>{% endif %}
+        </p></li>
+        {% endspaceless %}
+        {% endfor %}
```
rlskoeser, line 23: I see this implementation matches Gissoo's designs — we'll probably want to revise that now that we know more about what the source data looks like, but seems fine for current purposes.

rlskoeser (commented):
Seems reasonable!

Can you add some minimal tests before merging? 404 on suppressed, 404 on no footnotes, basic display for a document with 2 footnotes?

geniza/corpus/views.py
thatbudakguy: Inheriting from `DocumentDetailView` should already filter on status, no? That's what I was hoping it would do

rlskoeser: oh, I missed that ! yes, perfect

thatbudakguy: Also, wouldn't `distinct()` just operate on the level of the document? `DetailView`s should only ever return one object in the queryset otherwise you'd get `MultipleObjectsReturned`, I think

rlskoeser: you can get multiple rows if there are multiple footnotes — the distinct is to collapse them back to a single document. multiple objects returned was exactly what I got for the document with two footnotes

geniza/corpus/templates/corpus/document_scholarship.html
thatbudakguy: agree; it looks awkward for some current data but seemed OK for a first pass

thatbudakguy:
Added more tests, including implementation and testing of a function that gets a document's external links (fragment URLs) so that the template could render how many there are.

geniza/corpus/urls.py
```diff
@@ -3,14 +3,19 @@
...
 )
 
-
 app_name = "corpus"
 
 urlpatterns = [
     path("documents/", DocumentSearchView.as_view(), name="document-search"),
     path("documents/<int:pk>/", DocumentDetailView.as_view(), name="document"),
+    path(
+        "documents/<int:pk>/scholarship",
```
rlskoeser, line 16: should have a trailing slash here

rlskoeser (approved):
Looks great to me. Just need to add a trailing slash on the new url. (That way it will work with or without, since we have django configured to redirect to add a slash if it's missing in the request)

PR #226, opened 2021-06-02
merge from report

thatbudakguy:
Opened a PR to check the approach. The thing I had been thinking of doing (and ended up doing) wasn't a namedtuple but instead a dataclass - it lets you have a mutable record datatype with some light type-checking and you get `__init__`, `__str__`, `__repr__`, and `__eq__` for free. I think it'll end up being useful in testing; the default `__repr__` has already been useful for me.

### DIFF
--- a/geniza/corpus/management/commands/merge_joins.py
+++ b/geniza/corpus/management/commands/merge_joins.py
@@ -5,11 +5,17 @@ for data as tracked in the spreadsheet.
 
 To generate a report of potential merges and actions to be taken::
 
-    python manage.py. merge_joins report
+    python manage.py merge_joins report
+
+To execute the actions detailed in a report (by default, merge-report.csv)::
+
+    python manage.py merge_joins merge report.csv
 
 """
 import csv
 from collections import defaultdict
+from dataclasses import dataclass, field
+from typing import List
 import re
 
 from django.core.management.base import BaseCommand
@@ -18,12 +24,28 @@ from django.db.models import Count
 from geniza.corpus.models import Document
 
 
+@dataclass
+class MergeGroup:
+    """Group of Documents that are ready to be merged: candidates in `to_merge`
+    will be combined with `primary` using Document.merge_with, creating log
+    entries to document the merge reasoning using `rationale`."""
+
+    primary: int = None  # pgpid
+    to_merge: List[int] = field(default_factory=list)  # list of pgpid
+    rationale: str = ""
+
+    def merge(self):
+        """Execute the merge using `Document.merge_with()`."""
+        Document.objects.get(self.primary).merge_with(self.to_merge, self.rationale)
+
+
 class Command(BaseCommand):
     """Merge documents that are variations of the same joins, based on
     shelfmark, document type, and description"""
 
     def add_arguments(self, parser):
         parser.add_argument("mode", choices=["report", "merge"])
+        parser.add_argument("path", type=str, nargs="?", default="merge-report.csv")
 
     def get_merge_candidates(self):
         """identify merge candidates from the database. Looks for documents
@@ -79,6 +101,7 @@ class Command(BaseCommand):
         """process candidates identified in :meth:`get_merge_candidates`
         to determine which ones should be merged"""
         same_desc = 0
+        group_id = 1
 
         # TODO: group documents to merge into a structure that can be
         # used for reporting OR to do the actual merge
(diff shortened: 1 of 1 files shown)


geniza/corpus/management/commands/merge_joins.py
```diff
@@ -18,12 +20,28 @@
...
 
+@dataclass
+class MergeGroup:
+    """Group of Documents that are ready to be merged: candidates in `to_merge`
+    will be combined with `primary` using Document.merge_with, creating log
+    entries to document the merge reasoning using `rationale`."""
+
+    primary: int = None  # pgpid
+    to_merge: List[int] = field(default_factory=list)  # list of pgpid
+    rationale: str = ""
```
thatbudakguy, line 31: the reason for the seemingly useless defaults here is so that later on we can do a `defaultdict(MergeGroup)` and assign to the inner `to_merge` list without needing to initialize each one manually. that way, the rows from the .csv can be read in any order (for example, a merge candidate could be listed before the group's primary) and we will still get the right data structure

geniza/corpus/management/commands/merge_joins.py
```diff
@@ -18,12 +24,28 @@
...
+
+    def merge(self):
+        """Execute the merge using `Document.merge_with()`."""
+        Document.objects.get(self.primary).merge_with(self.to_merge, self.rationale)
```
thatbudakguy, line 39: i don't think we decided whether the API for this function would accept an iterable of `Document` or an iterable of pgpids for the first param. i think it might be more consistent to do the former, so that the caller is responsible for retrieving all the objects from the db (since it already has to retrieve the primary). if that sounds good to you I can change it here, and add a bit of error handling in the manage command in case we get a `Document.DoesNotExist` from a bad pgpid.

rlskoeser: From what we discussed, I expected it to be given a list of documents (not ids). I agree that it would be good to have the error handling external to the function

geniza/corpus/management/commands/merge_joins.py
rlskoeser: I looked at this briefly at the end of the day yesterday and it wasn't making much sense — maybe because I'm not familiar with dataclass (seems interesting, though).

This seems more complicated than we nee. Why not just load the spreadsheet rows into a list of dicts and then use django regroup? Then you don't have to worry about getting them in order, you will always process them as complete group.

geniza/corpus/management/commands/merge_joins.py
```diff
@@ -187,27 +210,27 @@ def group_merge_candidates(self, joins):
...
-        if self.mode == "report":
-            self.generate_report(report_rows)
-
-    def generate_report(self, report_rows):
-        # output report of what would be done when in report mode
-
-        with open("merge-report.csv", "w") as csvfile:
+    def generate_report(self, report_rows, path):
+        """Generate a .csv report of identified candidate groups for merging."""
+        with open(path, "w") as csvfile:
```
rlskoeser, line 228: should add error handling here — raise a `CommandError` if the csv file is not found

rlskoeser:
Did you mean to add the merge report to git?

thatbudakguy:
nope 😂 just removed it

geniza/corpus/management/commands/merge_joins.py
thatbudakguy: should've mentioned this at standup, sorry – what I wanted to do was process the rows (in whatever order they occur in the spreadsheet) in one pass and then be done; no sorting or grouping needed afterwards. `defaultdict` seems like a natural choice because we already have the keys specified, and i potentially need to assign/append to the `to_merge` property without it being initialized as a list first. the value type for the `defaultdict` needs to be mutable, so I picked a solution that would give me:
- `__init__` for free, so that `to_merge` would be an empty list
- `__repr__` for free, because i wanted to see a representation in the console that was compact like a tuple
- access to members as attributes, like `MergeGroup.primary`
- type annotation for the members
- a place to document what the record is supposed to do
- associated behavior (the `MergeGroup.merge()` function)

i don't think this adds too much complexity in return for being explicit about the data type, and i think it'll help in tests.

rlskoeser:
I've pulled in & adapted the changes from this branch that I needed, and was able to cherry-pick the error handling added most recently.

Issue #215, opened 2021-05-28
As a content editor, I would like to input dates in a separate field, so that both content editors and site users can sort and filter documents by date.


mrustow:
## testing notes

I've implemented a super-simple version of this so you can start using it, and then we can revisit and refine after you've used it for a bit. 

- [x] confirm that document now has fields for original document date, calendar, and standardized date
- [x] confirm you can save without adding them (should be optional)
- [x] confirm you can add data to those fields (try searching for a description with a date, e.g. by "seleucid" and then add the date information in the description into the date fields)
- [x] confirm that admin document export includes the new date fields for the record you just edited

---- 

Here are the fields I had in mind (with explanations for those fields in parens). 

1. Date on document (for when there is an explicit date)

2. Calendar (according to which the document gives the date; these could be radio buttons)

- hijrī
- kharajī
- Seleucid
- anno mundi

3. Standardized date (CE, so Julian before 1582, Gregorian after 1582)

4. Inferred date (when there's no date on the doc, but someone has deduced one; these will often be ranges; it would be good to footnote inferred dates, and these, too, can be radio buttons)

- document editor (e.g., the transcription is by Moshe Gil, and he's also the one who inferred the date)
- team member who input the document
- computationally inferred (for later, when we can infer dates based on people's floruits, or even paleography)

rlskoeser:
@mrustow I think we could give you a simple version of 1-3 this week if you need no or very little validation for 3. Implementing a simple version now will let you start inputting the data in a more structured form and will help us when we come back to it later to refine our solution and do more with the dates.

Am I understanding correctly that 1-3 all pertain to the document date?

I'm not sure yet how to implement 4 and think it should be a separate issue, but is good to have documented!

richmanrachel:
Looks great! Small details:
- Kharājī needs to have the macron over the second a. 
- I think we need a bit more help text for the calendar selection (so undergrads can more easily translate what's in the description):
1. Hijrī (AH)
2. Kharājī (rare - mostly for fiscal docs)
3. Seleucid (sometimes listed as Minyan Shetarot) 
4. Anno Mundi (Hebrew calendar) 
- Current CE help text is good, but before "Julian" please add "covert to Julian..."

Could we easily add a "Has date" "Does not have date listed" Boolean filter in the Document List View for now?

richmanrachel:
Re-opened for small changes above.

rlskoeser:
[...]

PR #227, opened 2021-06-03
Preliminary, simple implementation with fields for document dates #215

rlskoeser:
Simple first-pass implementation for document date fields; all optional, no validation (for now)

I added them to the document admin edit page and included in document admin csv export.

I would like to add them to the document detail page (and maybe search results also), but I think there are template changes in a different branch that I'd like to get first. Was thinking we probably need a document date display method to put them together, since they're all optional right now.

Does this seem reasonable for now? Anything important to add/change?


thatbudakguy (approved):
everything here looks reasonable, just some questions about implementation:

- which fields will be populated via a data migration (or similar) from the descriptions? would it only ever be the original date and/or calendar? could it be all of them?
- could we avoid populating the standardized date unless we're putting well-formatted dates into it, or are there cases where descriptions record a standardized date that doesn't fit the format we want (and we want to populate it)?

i think the use case for the calendar and original date is obvious since those fields are going to be used to eventually infer the standardized date later on, but if we aren't actually enforcing any kind of validation on the standardized date, i'm not sure what it will be immediately useful for (unless the use case is just pulling more dates out of descriptions).

rlskoeser:
> * which fields will be populated via a data migration (or similar) from the descriptions? would it only ever be the original date and/or calendar? could it be all of them?

Right now, none — I wanted to do the simplest possible version, so they can start using it before we pause development. They can start pulling the original date, calendar, and standardized dates from descriptions where they're documented and putting them in these fields.

> * could we avoid populating the standardized date unless we're putting well-formatted dates into it, or are there cases where descriptions record a standardized date that doesn't fit the format we want (and we want to populate it)?

I want to add validation on this field eventually, but I wanted to let them start using it first without any validation because I think any validation we apply now will be too specific.  If we come back to it after they've used it for a few months, we can use the preliminary data to help figure out what constraints and range of information we need to handle.

Also, I haven't decided yet how I want to handle uncertain and partial dates, date ranges, etc (whether adapting S&co partial date code, using EDTF, or something else). I'm currently thinking that eventually this set of three fields would be refactored into a reusable multi-field, and we'll add logic to calculate the standardized date from the original one (where possible).

Issue #234, opened 2021-06-04
admin document csv export has wrong date for first input


rlskoeser:
**Describe the bug**
First input column in the admin document csv export is the date from the _most recent_ log entry instead of the _first_ entry

**To reproduce**
1. Login to django admin and navigate to documents list
2. Select any subset of documents and choose export to csv 
3. Look at input date
4. See that the input date matches exactly the last modified date

**Expected behavior**
First input date should be earliest log entry date, not last modified date

richmanrachel:
@rlskoeser - It's still not working :( 

![image](https://user-images.githubusercontent.com/68028846/121424681-98495f00-c93f-11eb-81ea-bdb506e8829b.png)

![image](https://user-images.githubusercontent.com/68028846/121424705-9f706d00-c93f-11eb-8640-d43b1c6e8b8d.png)

richmanrachel:
Success now!
![image](https://user-images.githubusercontent.com/68028846/121425441-671d5e80-c940-11eb-867b-7f024e90d50e.png)

PR #237, opened 2021-06-07
Correct first input date in admin document csv export #234

rlskoeser:


thatbudakguy (approved):

geniza/corpus/tests/test_corpus_admin.py
```diff
@@ -254,6 +254,10 @@ def test_tabulate_queryset(self):
                 f"https://example.com/admin/corpus/document/{doc.id}/change/"
                 in doc_data
             )
+            # initial input should be before last modified
+            # (document fixture has a log entry, so should have a first input)
+            if doc_data[-6]:
+                assert doc_data[-6] < doc_data[-5]
```
thatbudakguy, line 260: could throw a message on here for when the assert fails so that we know the pgpid, just in case

Issue #239, opened 2021-06-14
Fragment reassociation doesn't update the search index


thatbudakguy:
**Describe the bug**
Changing the associated fragment for a document doesn't result in a corresponding change to the search index – the document is still findable under its old shelfmark instead of the new one. See [slack conversation](https://cdhprinceton.slack.com/archives/C016091ASAZ/p1623656912301500).

**Expected behavior**
Reassociating a document with a different fragment should immediately trigger an update to the search index, so that the document can be found using its new shelfmark.

richmanrachel:
Worked immediately! Thanks!

PR #242, opened 2021-06-15
Trigger document reindex on textblock save #239

rlskoeser:
I'm not sure why this hasn't come up in other projects, but triggering reindex on textblock change should resolve the problem that Alan noticed.

Eventually it would be nice to generalize for a solution that can live in parasolr, but I'm not going to worry about that right now.

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -233,7 +233,7 @@ class DocumentSignalHandlers:
         "fragment": "fragments",
         "tag": "tags",
         "document type": "doctype",
-        # log entry ?
+        "Related Fragment": "textblock",  # textblock verbose name
     }
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


thatbudakguy (approved):
looks good; agree that this solution is fine until/unless it can be addressed more systematically in parasolr. will the problem appear anywhere else? maybe `Authorship`, if that determines any data that gets indexed?

rlskoeser:
> will the problem appear anywhere else? maybe `Authorship`, if that determines any data that gets indexed?

Yes, it might. I haven't set up signal-based indexing for footnotes content because I don't think the current approach we're using works as well for those relationships (current parasolr approach is to retrieve them from model attributes, but for those it would be a lot easier to just pass in classes).

I'm hopeful that at some point we can look at the signal-based indexing we've implemented across projects to see what's common and adapt that for parasolr, but I'm not sure we're far enough along for that quite yet (nor do we have the time).

PR #244, opened 2021-06-16
Basic main navigation template

thatbudakguy:
adds a basic main navigation similar to gissoo's design. no styles for mobile yet. no tests, since i think this snippet may change a fair bit.


rlskoeser (approved):
Changes look reasonable to me. Agree on no tests at this point since it's so preliminary.

Issue #256, opened 2021-07-13
Rename language fields


richmanrachel:
# testing notes
Note that the existing field has been renamed, so any languages that were previously entered as probable languages will now be secondary languages.

- [x] confirm both languages are present in the document edit form and labeled correctly
- [x] confirm you can make changes to both language fields
- [x] confirm you are allowed to set unknown language as secondary
- [x] check the language+script admin list view has been updated with revised names and works correctly (counts, links)
- [x] confirm that admin csv export includes the updated labels and correct values for primary and secondary languages

---- 
After discussing the best way to handle source languages, we've decided that it's more useful for researchers to distinguish between Primary and Secondary languages on documents (rather than language/probable language). We have now done the cleanup necessary to change the names of these fields.

Please change:
- "Languages" to "Primary Langauge"
- "Probable Language" to "Secondary Languages" 

@mrustow and I will follow up by creating guidelines for what constitutes a primary vs. a secondary language.

rlskoeser:
@richmanrachel I'm finally working on this and as I'm making the changes I remembered we have a count and link from the language admin list for "probable documents". Should I rename / relabel this for the change to secondary language or should we drop it?  If you want to keep it, please supply the text you want me to use.

![Screen Shot 2021-08-12 at 11 11 54 AM](https://user-images.githubusercontent.com/691231/129221910-99f41edb-c023-4a75-af60-3d28f3bfdcaf.png)

richmanrachel:
@rlskoeser - just to be clear, are you asking what text we should use for the headers on the final two columns in the image above?

rlskoeser:
@richmanrachel yes! sorry. specifically:

1. do you still want both of these last two columns?
2. if so, how do you want them labeled?

richmanrachel:
@rlskoeser 
1. Yes, these columns are still useful. 
2. Let's rename to "# of documents where this is the primary language" and "# of documents where this is a secondary language"
3. I agree that it now makes sense to allow an unknown secondary language.
@mrustow - do you agree with these decisions?

richmanrachel:
@rlskoeser - everything looks good except the Unknown language for secondary. I got a Server Error 500.

rlskoeser:
@richmanrachel there was an old constraint still in place that disallowed unknown language! Thanks for catching. I've updated with a change that should fix this.

richmanrachel:
@rlskoeser - it works now! Closing :)

PR #258, opened 2021-08-12
Rename probable languages to secondary #256

rlskoeser:


geniza/corpus/migrations/0016_rename_probable_languages_to_secondary.py
```diff
@@ -0,0 +1,25 @@
...
+            old_name="probable_languages",
+            new_name="secondary_languages",
+        ),
+        migrations.AlterField(
+            model_name="document",
+            name="languages",
+            field=models.ManyToManyField(
+                blank=True, to="corpus.LanguageScript", verbose_name="Primary Languages"
+            ),
+        ),
```
thatbudakguy, line 24: do you know why one of these ended up as a `RenameField` and the other as an `AlterField`? is it because we only changed verbose name for `languages` but changed the actual field name for `secondary_languages`?

thatbudakguy (approved):

geniza/corpus/migrations/0016_rename_probable_languages_to_secondary.py
rlskoeser: @thatbudakguy correct

Issue #259, opened 2021-08-13
As a content editor, I want a one time consolidation of India Book sources so that the source list correctly represents the book volumes.


richmanrachel:
## testing notes
- [x] check that india book sources have been consolidated correctly
- [x] check that footnotes have been preserved & associated correctly

--- 
It seems the import accidentally made each India Book reference its own separate source, so we now have 352 IB sources: 62 sources for IB1; 79 sources for IB2; 83 sources for IB3; and 123 sources for IB4. There are also 102 without a volume number (though Marina just gave me an index that will help sort these).

Luckily, all of the separate sources are in the same format (eg. India Book 1, India Book 2). 

[For whatever reason, India Book 5 and 7 did import properly, and 6 only has two outliers which we are fixing manually.]

**Could you please help us merge all of the India Book 1, 2, 3, and 4s into their proper groupings?** 

Once that is done it will be easier to do the manual work of the volume-less India Book documents into the proper locations.

richmanrachel:
@rlskoeser - I'm very confused... everything looks almost perfect (one India Book 3 footnote is now showing up as India Book III, but that can be a manual fix). All the numbers for the Volumes add up except that I can no longer find the 102 India Book sources that are missing a volume number (including on the current, non-test website)... I'll check in with Jake, but I swear I verified that number myself!

richmanrachel:
There seem to be no problems with the code, so after re-checking with data from 8/10 when I got the 102 number, we can close.

PR #268, opened 2021-08-18
Data migration to merge India Book sources #259

rlskoeser:


geniza/footnotes/tests/test_footnotes_migrations.py
```diff
@@ -0,0 +1,78 @@
...
+from django.db.migrations.executor import MigrationExecutor
+import pytest
+
+
+# migration test case adapted from
+# https://www.caktusgroup.com/blog/2016/02/02/writing-unit-tests-django-migrations/
+# and copied from mep-django
+
+
+@pytest.mark.last
```
thatbudakguy, line 14: does the decorator do anything here? does this thing actually get run as a (no-op) test?

thatbudakguy (approved):
everything looks great here! migration logic is really easy to follow.

geniza/footnotes/tests/test_footnotes_migrations.py
thatbudakguy: relatedly: if we apply the decorator here do we also need it on tests that subclass this?

rlskoeser: I remember testing before to see if I could just put it on the base classes and not the subclasses, and that didn't work; it needs to be on each specific test class. It probably doesn't do anything on this one (except possibly serve as a reminder that subclasses should be run last). Should I remove and add a comment?

thatbudakguy: yeah, that sounds right. i think removing and adding a comment would be perfect

PR #273, opened 2021-08-19
Apply djlint to django templates

rlskoeser:
I started looking at setting up a pre-commit hook for djlint, but thought first I should apply it to our existing templates so we can see if we like what it does. (I'm not sure that I do...)

### DIFF
--- a/geniza/corpus/templates/admin/corpus/app_index.html
+++ b/geniza/corpus/templates/admin/corpus/app_index.html
@@ -1,22 +1,17 @@
 {% extends "admin/index.html" %}
 {% load i18n %}
-
 {% block bodyclass %}{{ block.super }} app-{{ app_label }}{% endblock %}
-
 {% if not is_popup %}
-{% block breadcrumbs %}
-<div class="breadcrumbs">
-<a href="{% url 'admin:index' %}">{% translate 'Home' %}</a>
-&rsaquo;
-{% for app in app_list %}
-{{ app.name }}
-{% endfor %}
-</div>
-{% endblock %}
+    {% block breadcrumbs %}
+        <div class="breadcrumbs">
+            <a href="{% url 'admin:index' %}">{% translate 'Home' %}</a>
+            &rsaquo;
+            {% for app in app_list %}{{ app.name }}{% endfor %}
+        </div>
+    {% endblock %}
 {% endif %}
-
 {% block sidebar %}
-<div id="content-related">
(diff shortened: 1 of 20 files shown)


thatbudakguy (approved):
I see what you mean! I'm OK with these changes, tbh. It definitely results in more spread-out/verbose templates, but I'd guess the reasoning behind that is so that diffs are easier to read/calculate, and I'm willing to give up some efficiency/conciseness in exchange for consistent style. Too often I think about "should I indent this or not" and it's almost worth it just so that I don't have to care anymore.

rlskoeser:
@thatbudakguy what about whitespace / indentation for inline tags like `<a>`, `<b>`, etc? Won't these cause display problems?

thatbudakguy:
i wondered about that, too. I found only one instance where there was a difference. It's the whitespace before the `:` in the input date fields (first one is djlint):

Screen Shot 2021-08-23 at 3 34 18 PM
Screen Shot 2021-08-23 at 3 33 57 PM

Inspecting the actual rendered HTML was...not encouraging. All lines with comments or django template syntax are just removed from the output, but the result isn't reformatted in any way, so there are a ton of whitespace "holes" — and many more of them with djlint. I think what's actually happening is that once the django templatetags and comments are subtracted, the newlines that are left in their place are making the browser insert whitespace where we don't want.

I looked at some other pages to see what HTML was actually served (view source) and it seems that many people minify their HTML into a single line! This makes sense, since there's no reason to serve the client extraneous whitespace — we already minify CSS and JS, why not HTML? [This StackOverflow answer](https://stackoverflow.com/questions/48669439/django-best-practice-to-minify-html-to-improve-page-speed-loading) suggests that serving gzipped HTML resolves the problem on its own, since gzipping removes that whitespace. Idk if we serve gzipped assets now (we might handle it at the nginx level?) but maybe it's something to think about. If we did that, we could keep the templates formatted however we want in github, but they'd effectively be minified at the point we serve them to the client.

...that might've been more investigation than this needed, and maybe we should just skip djlint for now. idk. i was just dismayed to see how weird (and maybe slightly inefficient) the actual HTML we're serving is.

rlskoeser:
Thanks for investigating — that's exactly what I was concerned about. I was also leaning towards skipping djlint, glad to have it confirmed.

I'm more interesting in looking into using `prettier` or finding a plugin for it for our templates, because I think the the html formatting is probably going to be a lot better in a tool like that with wide adoption.

PR #281, opened 2021-08-30
261 frontend builds

### LINKED ISSUE #261: set up js & css builds
- [x] set up minification
- [x] set up sourcemaps
<s>set up compression</s> handle via nginx; see #262

thatbudakguy:
closes #261 

- Set up webpack sources in sitemedia/
- Set up frontend compilation pipeline
- Add basic styles/scripts managed via webpack
- Default to prod bundles and allow live reload via local_settings
- Add frontend dev info to README
- Add comments to webpack config

this configuration borrows a lot from MEP & PPA but it has one important difference: style & script builds only ever happen locally, not in tests or on the server. this lets us avoid needing to deploy and manage node.js or any dependencies. the idea is:

- if you do nothing other than run a django dev server locally, you will see the currently checked-in versions of the bundled frontend files. this is what happens in CI also. unless you're changing styles or scripts, you can just proceed to develop normally — no need to run a server.
- if you need to make a small change to a style or script, you can edit the source files in `sitemedia/` and then run `npm run build` once to regenerate the bundles. later, you can commit the changes, and since the bundles are checked into git there's nothing else to do: they will get deployed with the rest of the code whenever you release.
- if you are making many changes to styles/scripts, you can uncomment a line in your local settings to tell django to look at a separate manifest file (which is _not_ checked in) that tells it to read bundles from memory, and then run a dev server with `npm run start`.

note also that everything is now kept in `sitemedia/`. source files are distinguished by their file extensions (`.scss`, `.esm.js`) and the directories they live in. `collectstatic` knows not to collect these files, but it will collect everything else, including the generated bundles.

check out the new README section for more info about how all this works.


geniza/apps.py
```diff
@@ -1,5 +1,10 @@
 from django.contrib.admin.apps import AdminConfig
+from django.contrib.staticfiles.apps import StaticFilesConfig
 
 
 class GenizaAdminConfig(AdminConfig):
     default_site = "geniza.admin.GenizaAdminSite"
+
+class GenizaStaticFilesConfig(StaticFilesConfig):
+    # don't collect frontend source files when running collectstatic
+    ignore_patterns = StaticFilesConfig.ignore_patterns + ["*.esm.js", "*.scss"]
```
rlskoeser, line 10: oh, cool — wasn't aware that static files could be configured like this

README.rst
```diff
@@ -77,6 +77,29 @@ Remember to add a ``SECRET_KEY`` setting!
     python manage.py index
 
 
+Static Files
+------------
+
+Static files are stored in `sitemedia/` and bundled by Webpack. The `webpack-bundle-tracker` plugin generates a JSON manifest file listing the name and location of bundled files. This file — `webpack-stats.json` in production/QA/test and `webpack-stats-dev.json` for local development — is read by Django using `django-webpack-loader` so that the relevant files can be included in the template's `<head>`.
+
+Bundled files and the production JSON manifest are checked into git and will be collected by Django's `collectstatic` command. They should be recompiled and any changes should be committed as part of a release. To recompile bundles for production::
+
```
rlskoeser, line 86: Let's add the `npm install` command somewhere to the development instructions. Maybe after installing python packages?

README.rst
```diff
@@ -77,6 +77,29 @@ Remember to add a ``SECRET_KEY`` setting!
...
+    npm run build
+
+When actively developing stylesheets and scripts, you can instead run a development Webpack server with live reload capability. To do this, first uncomment or add the line to your `local_settings.py` that tells Django to read from the development Webpack manifest instead::
+
+    WEBPACK_LOADER["DEFAULT"]["STATS_FILE"] = BASE_DIR.parent / "sitemedia" / "webpack-stats-dev.json"
```
rlskoeser, line 91: this should just be `BASE_DIR`, not `BASE_DIR.parent`

geniza/settings/local_settings.py.sample
```diff
@@ -23,6 +23,14 @@ DATABASES['default']['PASSWORD'] = ""
 # SOLR_CONNECTIONS['default']['COLLECTION'] = ''  # default geniza
 # SOLR_CONNECTIONS['default']['CONFIGSET'] = ''   # default geniza
 
+# Development webpack config: don't cache bundles
+WEBPACK_LOADER["DEFAULT"]["CACHE"] = False
+
+# Uncomment below to switch to the development webpack manifest, which expects
+# a running webpack dev server to serve files from memory and enable autoreload.
+# WEBPACK_LOADER["DEFAULT"]["STATS_FILE"] = (
+#     BASE_DIR.parent / "sitemedia" / "webpack-stats-dev.json"
```
rlskoeser, line 32: here too, just `BASE_DIR`

rlskoeser: and `WEBPACK_LOADER` needs to be imported from base settings

README.rst
```diff
@@ -77,6 +77,29 @@ Remember to add a ``SECRET_KEY`` setting!
...
+
+Then, run a development server to generate the manifest::
+
+    npm start
```
rlskoeser, line 95: When I ran this I was prompted to install `webpack-dev-server`; should that be included in npm packages ?

rlskoeser (commented):
Got an error at first but I think it's a path problem in the config, see my notes. 

I was able to run the django app with checked in bundles, was able to generate locally, and was able to run the dev server and make a change and see it (once I followed the directions and made the edits in the right project 🤦‍♀️).

When I ran `npm start` I was prompted to install `webpack-dev-server`, should that be added to dev dependencies?

I like the idea of not building the files on deploy servers a lot, but I'm not totally sold on checking them into version control. How will we make sure they stay up to date? Is there an easy check we could run in ci to ensure that bundles have been regenerated if source files have changed? (modification time comparison?) Do the filenames change every time you rebuild? 

I'm wondering if we could use github actions & artifact uploads, or something like that. Automate building bundles, download them automatically for deploys ? Setting that up now is  out of scope, but if you think it could work maybe we could talk through approach.

README.rst
thatbudakguy: I agree, but something super weird is happening with importing that variable — it has different values in different settings files (for me, anyway). Would love to pair on it so I can show you at some point; suspect it has to do with how `django-split-settings` works.

README.rst
thatbudakguy: That's odd; it worked for me without installing, but I couldn't tell if that package was automatically installed along with `webpack` or not. I'll add it to `package.json` regardless

README.rst
kmcelwee: Since `local_settings.py` is in `./geniza`, isn't `BASE_DIR` here `./geniza`? and since `sitemedia` is at the top of the repo, you'd have to go up one?

thatbudakguy:
ok, ready for final review.

- we now only use one webpack manifest (if you switch to the dev server it just overwrites the file; no need to keep a separate manifest) and it's not checked in
- built static files are no longer checked in
- the README has been updated to add the `npm install` step and revise the commands for local dev
- there's now a dummy `webpack-stats-ci.json` file that can be used for CI purposes so that tests pass without needing to compile static files (this is the same approach recently used on PPA)

i haven't found anything worthwhile re: remote static file deployment so for now let's stick to what we know and run the build on our VM.

kmcelwee (approved):

rlskoeser (approved):
This all looks good. Was able to pull locally and run it (once I cleaned up the stuff from the previous version in my local settings that was causing some path problems).  Sorry to lose the idea of prebuilding assets instead of building them on the app server, but now that you've planted the seed maybe we'll be able to circle back to that.

What does the dummy ci webpack do? Does it just ignore / no assets for CI? Any implications for this if/when we want to set up lighthouse or something like that for accessibility checking?

thatbudakguy:
> What does the dummy ci webpack do? Does it just ignore / no assets for CI?

that's basically it. when the django dev server runs, it will look for a webpack manifest (`webpack-stats.json`) because of `django-webpack-loader`. if it doesn't find one, it will throw errors because we try to include built files in the `<head>` in templates, but it doesn't know where to find those files. the dummy one creates just enough of a manifest to avoid erroring, but it doesn't actually point to any real files, so the tags in the `<head>` will be empty — which is fine, because all we need to do is run a few template unit tests.

when we have lighthouse or percy set up, we will actually need to include node and run the build to produce real static files, because we care about how the site actually appears in a browser. there are two different ways we could approach that (separate github actions workflows vs. separate jobs in a single unified CI workflow). the former would mean that our unit tests workflow would still need the dummy file; the latter wouldn't. happy to talk more about the implications of both!

PR #282, opened 2021-09-02
chore/setup lighthouse

thatbudakguy:
- Refactor pytest fixtures; create make_test_site command
- Set up lighthouse CI workflow

### DIFF
--- a/geniza/corpus/templates/corpus/document_list.html
+++ b/geniza/corpus/templates/corpus/document_list.html
@@ -1,14 +1,12 @@
 {% extends 'base.html' %}
 {% load static i18n %}
 
-{% comment %}Translators: Title for document search page{% endcomment %}
-{% translate 'Search Documents' as doctitle%}
-
-{% block title %}{{ doctitle }}{% endblock title %}
+{% block meta_title %}{{ page_title }}{% endblock meta_title %}
+{% block meta_description %}{{ page_description }}{% endblock meta_description %}
 
 {% block main %}
 <div class="container search">
-    <h1>{{ doctitle }}</h1>
+    <h1>{{ page_title }}</h1>
 
     <form>
     {{ form }}
@@ -31,12 +29,14 @@
     {% for document in documents %}
         <li class="result">
 
+            {% if document.tags.exists %}
             <dl class="tags">
                 <dt class="sr-only">{% translate 'Tags' %}</dt>
                 {% for tag in document.tags %}
                 <dd class="tag">#{{ tag }}</dd>
                 {% endfor %}
             </dl>
+            {% endif %}
 
             <h2><b>{{ document.type }},</b> {{ document.shelfmark|join:" + " }}</h2>
             <p>{% translate 'Input date' %} {{ document.input_year|default:'unknown' }}</p>
--- a/geniza/corpus/templates/corpus/document_list.html
+++ b/geniza/corpus/templates/corpus/document_list.html
@@ -29,7 +29,7 @@
     {% for document in documents %}
         <li class="result">
 
-            {% if document.tags.exists %}
+            {% if document.tags %}
             <dl class="tags">
                 <dt class="sr-only">{% translate 'Tags' %}</dt>
                 {% for tag in document.tags %}
--- a/geniza/corpus/templates/corpus/snippets/document_tabs.html
+++ b/geniza/corpus/templates/corpus/snippets/document_tabs.html
@@ -11,7 +11,7 @@
         {% with n_links=document.fragment_urls|length %}
         {# Translators: n_links is number of external links #}
         {% blocktranslate asvar elink_text %}External Links ({{ n_links }}){% endblocktranslate %}
-        <li>{% if n_links > 0 %}<a>{{ elink_text }}</a>{% else %}<span>{{ elink_text }}</span>{% endif %}</li>
+        <li>{% if n_links > 0 %}<a href="#">{{ elink_text }}</a>{% else %}<span>{{ elink_text }}</span>{% endif %}</li>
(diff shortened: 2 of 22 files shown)


thatbudakguy:
this is now running correctly — lighthouse has identified some problems which I'll track on a separate issue so this one can be merged.

geniza/corpus/templates/corpus/document_list.html
```diff
@@ -31,12 +29,14 @@ <h1>{{ doctitle }}</h1>
     {% for document in documents %}
         <li class="result">
 
+            {% if document.tags.exists %}
```
rlskoeser, line 32: does this work? tags is a list in a solr result, not a queryset (IIRC)

geniza/corpus/templates/corpus/snippets/document_tabs.html
```diff
@@ -11,7 +11,7 @@
         {% with n_links=document.fragment_urls|length %}
         {# Translators: n_links is number of external links #}
         {% blocktranslate asvar elink_text %}External Links ({{ n_links }}){% endblocktranslate %}
-        <li>{% if n_links > 0 %}<a>{{ elink_text }}</a>{% else %}<span>{{ elink_text }}</span>{% endif %}</li>
+        <li>{% if n_links > 0 %}<a href="#">{{ elink_text }}</a>{% else %}<span>{{ elink_text }}</span>{% endif %}</li>
```
rlskoeser, line 14: do we need to do anything to disable this for MVP or do you think this implementation is fine?

geniza/corpus/views.py
```diff
@@ -84,17 +92,50 @@ class DocumentDetailView(DetailView):
...
         """Don't show document if it isn't public"""
         queryset = super().get_queryset(*args, **kwargs)
         return queryset.filter(status=Document.PUBLIC)
 
+    def get_context_data(self, **kwargs):
+        context_data = super().get_context_data(**kwargs)
+        context_data.update(
+            {
+                "page_title": self.page_title(),
+                "page_description": self.page_description(),
```
rlskoeser, line 111: Great to have these conventions established early — should every page have these?

geniza/corpus/views.py
```diff
@@ -84,17 +92,50 @@ class DocumentDetailView(DetailView):
...
+        # Translators: title of document scholarship page
+        return _("Scholarship on %(doc)s") % {"doc": self.get_object().title}
+
+    def page_description(self):
+        doc = self.get_object()
+        count = doc.footnotes.count()
+        # Translators: description of document scholarship page, for search engines
+        return ngettext(
+            "%(count)d scholarship record for %(doc)s",
+            "%(count)d scholarship records for %(doc)s",
```
rlskoeser, line 132: is this repeated for singular/plural? worth a comment or just a convention we'll get used to once we use it more?

rlskoeser (approved):
This all looks good to me. Great to have it setup.

I'd like to know more about what we need to do to maintain/extend it as we build out the site. What do you think about adding a block to the developer instructions in the readme about lighthouse testing and adding it to other pages as needed? 

Any idea if there will be a problem once we add more pages, is there a limit to how many we can add or will it slow down the build?

geniza/corpus/views.py
thatbudakguy: I think so, yeah!

geniza/corpus/views.py
thatbudakguy: It is, yeah. Maybe a convention we'll get used to; the docs explain more about it but I'm still figuring out how to do it correctly in templates

geniza/corpus/templates/corpus/snippets/document_tabs.html
thatbudakguy: probably we should just comment out the whole thing, I guess? that's a good point

geniza/corpus/templates/corpus/document_list.html
thatbudakguy: it appeared to...but I'll switch to just `document.tags`

thatbudakguy:
So, this probably shouldn't be merged yet — I added the correct CSP config, but there appears to be a bug in `django-csp` that is preventing it from working, and I think that needs resolving before we proceed. I think it's this: https://github.com/mozilla/django-csp/issues/167

basically, the new config has every script that's included add a `nonce=` value that `django-csp` generates for us (it's available in context as `request.csp_nonce`). anything that gets pulled in via something with a nonce (more scripts, styles, etc). is automatically considered safe. anything else isn't. this way, we don't have to whitelist where we want things to come from.

the issue is that the nonce isn't actually available on the request. it will output in a template, but trying to use it in the `<head>` it appears to be empty. I _think_ this might be related to it being a `LazyObject`. once we get it to work, all the remaining lighthouse issues should go away, since they're either related to CSP config or to errors raised in the console because of things being (erroneously) blocked by CSP due to the missing nonce.

rlskoeser:
Some notes on what I've figured out so far:
-  problem is not the nonce; it's present, it's just not available when you use developer tools to inspect the dom — it's hidden from dom inspection tools for security reasons, but it's included in the output when you view source
- duplicated the site locally as it's set up in lighthouse-ci and ran lighthouse locally in my browser; the error on the home page points to "issues" output, which is an error about [accessing the deprecated user-agent variables](https://blog.chromium.org/2021/05/update-on-user-agent-string-reduction.html) — as far as I can tell, this is either in webpack or django-webpack code somewhere; I think we should configure our build to ignore it, but haven't found the right way to specify that (and am leary of ignoring all console output)
- it looks like the major problems on the document detail page are related to mirador 😞

Issue #252, opened 2021-06-18
omit volume when outputting footnote/source string for unpublished sources (i.e. Goitein "typed texts")


mrustow:
## testing notes

Check the public display of editor/source information on a document that has an unpublished source with a volume defined in the database and visible in the admin site – the volume should not be displayed on the public site.

Here's a link to the test site for the example referenced below when this issue was reported: [Letter CUL Or.1080 J86](https://test-geniza.cdh.princeton.edu/en/documents/4016/)

---- 

Check out CUL Or. 1080 J86: back end says "Goitein, typed texts CUL," but front end says "Goitein, typed texts (2019)." 

<img width="405" alt="image" src="https://user-images.githubusercontent.com/73319225/122498624-71cd9880-cfbd-11eb-8bfe-694acd8d68b8.png">
<img width="535" alt="image" src="https://user-images.githubusercontent.com/73319225/122498642-7a25d380-cfbd-11eb-88f8-eddfd4d39a96.png">

While I'm here: can we change the template to "Goitein, typed texts (CUL)" (add parens)? 

## dev notes

- [x] suppress volume for unpublished sources

Other source/footnote formatting is now tracked on #313

mrustow:
the error seems to not to be limited to this one document or run of shelfmarks; I just found it in T-S NS J24.

rlskoeser:
@mrustow I think at least two things are going on here:

1. The current PGP site was still not regularly updating whenever you checked this, so you were seeing out of date information (there were definitely parsing errors; this one at least has been cleaned up in the database).
2. The formatting for sources and footnotes needs to be revised. I think I did this on my own (trying to roughly match the format in the metadata spreadsheet) without input or testing from you or Rachel.

Could you provide the source formatting you'd like to see that will work for all our different sources?  Or suggest something and we can figure out what the variations need to be?  It's a little complicated since we have multiple different types and a lot of the fields are optional.  For example, in the case of Goitein's typed texts we put the shelfmark prefixes into the volume field — I can put it in parentheses, but I'm not sure we want volume in parentheses for all sources. Maybe it would work if we did that for all unpublished sources, since I think none of the others have volumes.

(I think somewhere in asana you shared the format you wanted for articles in relation to the scholarship record cleanup work, but I didn't make a note of it.)

mrustow:
Book: 
[...]

PR #320, opened 2021-10-18
draft changes for footnotes.Source str method issue #252

apjanco:


geniza/footnotes/models.py
```diff
@@ -141,16 +140,49 @@ def __str__(self):
...
+            # Moshe Gil, Palestine during the First Muslim Period, 634–1099, in Hebrew (Tel Aviv, 1983), vol. 2, doc. 134"""
+            return f'{author}, {self.title} {"in "+languages+")" if languages else ""}, {self.volume if self.volume else ""} {"("+str(self.year)+")" if self.year else ""}{": "+str(self.page_range) if self.page_range else ""}'
+
+        if self.source_type.type == "Book Section":
+            # S. D. Goitein, "New Documents on the Gaonate in Palestine," in Salo Baron Jubilee Volume on the Occasion of His Seventy-fifth Birthday, ed. Arthur Hyman (New York, 1975), 2:55–74
+            # TODO need field for place of publication
+            return f'{author}, "{self.title}," {"in <i>"+self.journal+"</i>" if self.journal else ""} {"("+str(self.year)+")," if self.year else ""}{" "+self.volume+": " if self.volume else ""}{str(self.page_range) if self.page_range else ""}'
+
+            # TODO: formatted version with italics for book/journal title.
+            # AJ (10/18/21) Added <i> but will require |safe in templates, will appear as html in admin, is this the right solution?
```
rlskoeser, line 176: Let's keep the formatted version out of scope for now — we'll need to add it eventually, and I think it will probably need to be two methods, but it isn't needed to fix the current bug.

geniza/footnotes/models.py
```diff
@@ -141,16 +140,49 @@ def __str__(self):
...
-        # TODO: formatted version with italics for book/journal title
+        if self.languages:
+            if not self.languages.all():
+                languages = None
+
+            if len(self.languages.all()) > 1:
+                # Separate languages with commas
+                languages = ",".join([l.name for l in self.languages.all()])
+            else:
+                languages = " ".join([l.name for l in self.languages.all()])
```
rlskoeser, line 158: the if/else not needed, you can just join the list — even if there's 1 or zero entries you get the right result... unless I'm missing something; why join with a space in the second case? I'm not sure that's doing anything.

```suggestion
            # Separate languages with commas
            languages = ", ".join([l.name for l in self.languages.all()])
```

geniza/footnotes/models.py
```diff
@@ -141,16 +140,49 @@ def __str__(self):
...
+
+        # Source formatting for given source type, see #252
```
rlskoeser, line 160: We don't typically reference issues in comments; I think because in the long term it's not a particularly helpful resource — any details that are needed to understand the code should be in the comments, which you've done.

geniza/footnotes/models.py
```diff
@@ -141,16 +140,49 @@ def __str__(self):
...
+            return f'{author}, "{self.title}"{"("+languages+")" if languages else ""}, {self.journal} {self.volume if self.volume else ""} {"("+str(self.year)+")" if self.year else ""}{": "+str(self.page_range) if self.page_range else ""}'
+
+        if self.source_type.type == "Book":
+            # Moshe Gil, Palestine during the First Muslim Period, 634–1099, in Hebrew (Tel Aviv, 1983), vol. 2, doc. 134"""
+            return f'{author}, {self.title} {"in "+languages+")" if languages else ""}, {self.volume if self.volume else ""} {"("+str(self.year)+")" if self.year else ""}{": "+str(self.page_range) if self.page_range else ""}'
+
+        if self.source_type.type == "Book Section":
+            # S. D. Goitein, "New Documents on the Gaonate in Palestine," in Salo Baron Jubilee Volume on the Occasion of His Seventy-fifth Birthday, ed. Arthur Hyman (New York, 1975), 2:55–74
+            # TODO need field for place of publication
+            return f'{author}, "{self.title}," {"in <i>"+self.journal+"</i>" if self.journal else ""} {"("+str(self.year)+")," if self.year else ""}{" "+self.volume+": " if self.volume else ""}{str(self.page_range) if self.page_range else ""}'
```
rlskoeser, line 173: I find the f strings with all the conditionals really hard to read.  Can we combine the common logic outside of the if statements? Everything starts with author, title right? And we should handle missing title or author gracefully, which these don't. So maybe join those two together as the base string that other methods expand on. Are there any other common bits to these formats? 

I will ask Marina if we can standardize the way languages are included.

geniza/footnotes/models.py
```diff
@@ -141,16 +140,49 @@ def __str__(self):
...
             # if this is an article, wrap title in quotes
             if self.source_type.type == "Article":
                 parts.append('"%s"' % self.title)
             else:
                 parts.append(self.title)
 
-        # TODO: formatted version with italics for book/journal title
+        if self.languages:
+            if not self.languages.all():
+                languages = None
```
rlskoeser, line 152: They way I usually see this logic is to set languages to None outside the conditional, and then set the appropriate value if there is one — it reduces an extra check.

I think I forgot to mention — we only want to include the language in the reference if it is not English.

geniza/footnotes/models.py
```diff
@@ -141,16 +140,49 @@ def __str__(self):
...
+            return f'{author}, "{self.title}," {"in <i>"+self.journal+"</i>" if self.journal else ""} {"("+str(self.year)+")," if self.year else ""}{" "+self.volume+": " if self.volume else ""}{str(self.page_range) if self.page_range else ""}'
+
+            # TODO: formatted version with italics for book/journal title.
+            # AJ (10/18/21) Added <i> but will require |safe in templates, will appear as html in admin, is this the right solution?
+
+        # is Blog
+        # Source objects exist with type Blog, but no formatting is given in #252
+
+        if self.source_type.type == "Dissertation":
+            # Ṣabīḥ ʿAodeh, "Eleventh Century Arabic Letters of Jewish Merchants from the Cairo Geniza" (PhD diss. Tel Aviv University, 1992), ??? what is doc. 5 above?
```
rlskoeser, line 182: "doc 5" in the example is the location — it's similar to a page number; that's set on the footnote, not the source, so you can ignore it here. Footnote string method combines source string and location, so that should work properly once the source string method is corrected.

geniza/footnotes/tests/test_footnote_models.py
```diff
@@ -33,7 +39,7 @@ def test_str(self, source, twoauthor_source, multiauthor_untitledsource):
         )
 
         # two authors
-        assert str(twoauthor_source) == "%s and %s, %s" % (
+        assert str(twoauthor_source) == "%s and %s, %s," % (
```
rlskoeser, line 42: why this change?

geniza/footnotes/tests/test_footnote_models.py
```diff
@@ -45,32 +51,104 @@ def test_str(self, source, twoauthor_source, multiauthor_untitledsource):
...
-            article.year,
-        )
-        # no volume
-        article.volume = ""
-        assert str(article) == "%s, %s (%s)" % (
-            article.authors.first().firstname_lastname(),
-            article.journal,
-            article.year,
-        )
+        #                                            #
```
rlskoeser, line 54: This looks like you're duplicating a lot of the logic in the code you're trying to test, which isn't a good approach.

We have several pyest source fixtures in `geniza.footnotes.conftest` — you should test against them, similar to the way the previous `test_str_article` worked. You include the fixture name as an argument to your test method, and pytest will create the fixture so you can test against it. I think having methods for each type would make sense, and you can add new fixtures for types we don't already have represented.

geniza/footnotes/models.py
```diff
@@ -141,16 +140,49 @@ def __str__(self):
...
+        if self.source_type.type == "Book Section":
+            # S. D. Goitein, "New Documents on the Gaonate in Palestine," in Salo Baron Jubilee Volume on the Occasion of His Seventy-fifth Birthday, ed. Arthur Hyman (New York, 1975), 2:55–74
+            # TODO need field for place of publication
+            return f'{author}, "{self.title}," {"in <i>"+self.journal+"</i>" if self.journal else ""} {"("+str(self.year)+")," if self.year else ""}{" "+self.volume+": " if self.volume else ""}{str(self.page_range) if self.page_range else ""}'
+
+            # TODO: formatted version with italics for book/journal title.
+            # AJ (10/18/21) Added <i> but will require |safe in templates, will appear as html in admin, is this the right solution?
+
+        # is Blog
+        # Source objects exist with type Blog, but no formatting is given in #252
```
rlskoeser, line 179: I think if we combine the common parts a bit more (author, title, language) then new source types should look decent without custom formatting provided. (I hope)

PR #327, opened 2021-10-19
Configure pre-commit hooks for prettier & djhtml

rlskoeser:
- configure pre-commit hooks to apply [prettier](https://prettier.io/) to css/js/json/yaml etc and [djhtml](https://github.com/rtts/djhtml) to django template files
- apply new formatting to all files
- add the git revisions to ignore to the existing git blame ignore file


blms (approved):
This all looks good to me, very helpful hooks. I especially appreciate the `djhtml` hook properly indenting Django templates, since those can get messy using Prettier alone.

Issue #292, opened 2021-10-07
As a user, when I search on shelfmark I want to see documents associated directly with that fragment before documents that include the shelfmark in a description or notes, so I can easily find documents by shelfmark.


rlskoeser:
## testing notes
In the [public document search](https://test-geniza.cdh.princeton.edu/documents/), keyword and phrase matches on shelfmarks will now be "boosted" so they are more relevant than matches elsewhere. In particular, you should try searching on a shelfmark that occurs in a description — the document associated directly with that shelfmark should be listed before the document that references it in a description. But try other shelfmark searches as well and see how it behaves.

## dev notes
- [x] add boosting to shelfmark fields in solrconfig.xml `keyword_qf` and `keyword_pf`
- [x] write a unit test (~ integration test) to confirm the behavior to avoid/identify regressions in future

PR #330, opened 2021-10-21
Configure & test shelfmark boosting in solr #292

rlskoeser:


geniza/corpus/tests/test_corpus_views.py
```diff
@@ -202,6 +204,29 @@ def test_get_context_data(self, rf):
...
+        # in solr configuration
+
+        # create a second document with a different shelfmark
+        # that references the shelfmark of the first in the description
+        related_doc = Document.objects.create(
+            description="See also %s" % document.shelfmark
+        )
+        TextBlock.objects.create(document=related_doc, fragment=multifragment)
+        # wait to make sure solr index is updated with document fixtures
+        sleep(1)
```
blms, line 218: Will this sleep command guarantee the necessary amount of wait time in all environments? I wonder if there could be a way to run a loop checking for the fixture's presence in the index instead of sleeping—though I'm not familiar enough with solr to know how doable that is.

solr_conf/conf/solrconfig.xml
```diff
@@ -738,8 +738,8 @@
         type_s
         type_t
         collection_ss
-        shelfmark_t
-        shelfmark_ss
+        shelfmark_t^80
```
blms, line 741: Boosts look good. As a developer, do I need to copy this file again to my solr home directory and/or run any solr commands for the changes to go into effect?

geniza/corpus/tests/test_corpus_views.py
rlskoeser: Good question — we have the test solr configured with a pretty aggressive commit time, you can see it here https://github.com/Princeton-CDH/geniza/blob/main/geniza/settings/components/base.py#L107-L110 

So I think it _should_ be enough, but there have been cases in the past where things fail — e.g., a test that works locally but fails in CI. This is the approach we've been using in other projects, but even having the sleep isn't ideal, it would probably be better to use a solr commit that would ensure the index is current.

I'll see if I can add a solr commit here instead.

solr_conf/conf/solrconfig.xml
rlskoeser: yes, you'll have to update the configset or core config with the changed files and reload the solr core.

I was thinking to warn everyone about the change in slack once this gets merged in; but we might want to think about a better approach for that, since we will likely be making more changes to the solr config

Issue #303, opened 2021-10-12
As a user I would like to see a permalink for each document so that I can easily document, remember and share links.


gissoo:
## testing notes
Visit a document detail page:
- [x]  Verify that there is a permalink in the metadata section, and that the permalink correctly points to the absolute URL of the document detail page
- [x] Use browser tools to view page source: in the HTML source verify that there is a `<link>` within the `<head>` block with `rel="canonical"` and `href=` the same absolute URL.

## dev notes
Add absolute url for current document to detail template with `rel=bookmark` and include in header as a `link` with `rel=canonical`

Use `absolutize_url` method in `geniza.common.utils` to get the absolute url.

Should be listed after editor & input date and before tags:
![Screen Shot 2021-10-18 at 5 41 48 PM](https://user-images.githubusercontent.com/691231/137810807-68e84408-6170-4453-a344-00ac8ce9a429.png)

Add tests to `TestDocumentDetailView` to confirm behavior

PR #331, opened 2021-10-21
#303 - Add permalink to document detail view

blms:
### What this PR does

- Per #303:
  - Adds permalink (generated with the `absolutize_url` function) to document detail view, in both the head and body of the template
  - Adds a test for the above to `TestDocumentDetailView`

### Additional notes

- I renamed the `extrahead` template block to `extrascript`, following the convention of `extrastyle`, since it is located in the scripts portion of the HTML head. I also created a new block called `extrameta` which I used to include the `<link>`.

### Steps to test

1. Visit a document detail page
2. Verify that there is a permalink in the metadata section, and that the permalink correctly points to the absolute URL of the document detail page
3. In the HTML source, verify that there is a `link` within the `<head>` block with `rel="canonical"` and `href=` the same absolute URL.

### DIFF
--- a/geniza/corpus/views.py
+++ b/geniza/corpus/views.py
@@ -6,7 +6,6 @@ from django.views.generic import DetailView, ListView
 from django.views.generic.edit import FormMixin
 from tabular_export.admin import export_to_csv_response
 
-from geniza.common.utils import absolutize_url
 from geniza.corpus.forms import DocumentSearchForm
 from geniza.corpus.models import Document, TextBlock
 from geniza.corpus.solr_queryset import DocumentSolrQuerySet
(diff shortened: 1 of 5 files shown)


geniza/corpus/views.py
```diff
@@ -99,6 +100,9 @@ def page_title(self):
     def page_description(self):
         return Truncator(self.get_object().description).words(20)
 
+    def permalink(self):
+        return absolutize_url(self.get_object().get_absolute_url())
```
rlskoeser, line 104: I love that you made this a method — but I'd like to have it on the `Document` object. It will eventually be useful in other contexts also, like the data exports. And that way we won't have to pass anything else in the view context, just reference `document.permalink` in the template

Issue #307, opened 2021-10-12
Implement the alignment, spacing and margins for the search page for desktop and mobile based on the designs


gissoo:

rlskoeser:
visual review signed off via percy; successfully deployed to qa

Issue #305, opened 2021-10-12
Document Detail View


gissoo:
## dev notes:
- rm doc type and shelfmark from metadata section in detail view
- sort tags on document detail view

Issue #309, opened 2021-10-12
Implement type sizes for the search page on mobile and desktop based on the designs


gissoo:

blms:
Reviewed in QA deployment 👍

Issue #306, opened 2021-10-12
Implement content alignment and margins for the document detail view on mobile, and spacing on mobile and desktop based on the designs


gissoo:
## Testing notes

- [x] Check document detail views, verify that alignment and margins match the designs

blms:
Reviewed in QA deployment. All set except for #357, which has its own issue, so closing this one.

PR #336, opened 2021-10-28
#307 - Search results: Alignment, spacing, margins

blms:
### What this PR does
- Per #307:
  - Implements alignment, spacing, and margin for search results page
    - Implements a minimal amount of typography for those to look right
    - Adds `breakpoints.respond_to` SASS mixin
  - Reorders `document_list.html ` template to match order of elements in the design
    - Uses individual form elements instead of `{% form %}` for finer control over layout
  - Separates `document_result.html` into two containers, which act as columns in desktop and rows in mobile

### Additional notes
- Because [element widths are defined in relation to character count](https://github.com/Princeton-CDH/geniza/blob/27c6f62c130c1a12a2d5a969ce13b79e3d9f4783/sitemedia/scss/base/_container.scss) and [spacing is defined in relation to 1rem](https://github.com/Princeton-CDH/geniza/blob/27c6f62c130c1a12a2d5a969ce13b79e3d9f4783/sitemedia/scss/base/_spacing.scss), it seemed to go against the code philosophy to define these things in terms of exact pixel count.
  - What I ended up doing is defining an exact-pixel container width for what I call a "two-column" layout, the children of which should not exceed 60 characters in width, thus still incorporating the philosophy that text containers should remain readable.
  - For all spacing, I used the closest existing value in `_spacing.scss` to approximate the pixel values used in the design.
- I used the view context to pass the id `"search"` to the base template `body` element based on the comment:
    ```css
    // rules go here once body context ID is applied
    ```
    I'm not sure if that is actually desirable or makes sense, and I may have completely misunderstood the comment, but for now I've tried to make a guess at what was intended.
- Document detail view was slightly touched by this PR but I'll leave fixes for that until #305.

### DIFF
--- a/geniza/corpus/templates/corpus/document_list.html
+++ b/geniza/corpus/templates/corpus/document_list.html
@@ -5,6 +5,7 @@
 {% block meta_description %}{{ page_description }}{% endblock meta_description %}
 
 {% block main %}
+    <h1 class="sr-only">{{ page_title }}</h1>
     <form>
         <fieldset>
             {{ form.q }}
--- a/geniza/corpus/templates/corpus/document_list.html
+++ b/geniza/corpus/templates/corpus/document_list.html
@@ -5,11 +5,24 @@
 {% block meta_description %}{{ page_description }}{% endblock meta_description %}
 
 {% block main %}
-    <h1>{{ page_title }}</h1>
-    <div class="container">
-
-        <form>
-            {{ form }}
+    <div class="document-list">
+        <form class="search-form">
+            <div class="search-bar">
+                {{ form.q }}
+                <input type="submit" value="search" class="search-button">
+            </div>
+            <div class="results-count">
+                {% comment %}Translators: number of search results{% endcomment %}
+                {% blocktranslate count counter=total trimmed %}
+                    1 result
+                {% plural %}
+                    {{ counter }} total results
+                {% endblocktranslate %}
+            </div>
+            <div class="sort-by-container">
+                {{ form.sort.label }}
+                {{ form.sort }}
+            </div>
             {% for field, errors in form.errors.items %}
                 {# no clean way to get field labels here, so we just collapse the error lists #}
                 {% for error in errors %}
@@ -17,13 +30,6 @@
                 {% endfor %}
             {% endfor %}
         </form>
-
-        {% comment %}Translators: number of search results{% endcomment %}
-        {% blocktranslate count counter=total trimmed %}
-            1 result
-        {% plural %}
-            {{ counter }} total results
-        {% endblocktranslate %}
 
         <ol class="search-results">
             {% for document in documents %}
--- a/geniza/corpus/templates/corpus/document_list.html
+++ b/geniza/corpus/templates/corpus/document_list.html
@@ -5,36 +5,33 @@
 {% block meta_description %}{{ page_description }}{% endblock meta_description %}
 
 {% block main %}
-    <div class="document-list">
-        <form class="search-form">
-            <div class="search-bar">
-                {{ form.q }}
-                <input type="submit" value="search" class="search-button">
-            </div>
-            <div class="results-count">
-                {% comment %}Translators: number of search results{% endcomment %}
-                {% blocktranslate count counter=total trimmed %}
-                    1 result
-                {% plural %}
-                    {{ counter }} total results
-                {% endblocktranslate %}
-            </div>
-            <div class="sort-by-container">
-                {{ form.sort.label }}
-                {{ form.sort }}
-            </div>
-            {% for field, errors in form.errors.items %}
-                {# no clean way to get field labels here, so we just collapse the error lists #}
-                {% for error in errors %}
-                    <li>{{ error }}</li>
-                {% endfor %}
+    <form>
+        <fieldset>
+            {{ form.q }}
+            {{ form.sort }}
+            <label for="sort">{{ form.sort.label }}</label>
+            <input type="submit" value="search" />
+        </fieldset>
+        {% for field, errors in form.errors.items %}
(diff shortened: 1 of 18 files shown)


blms:
@gissoo @rlskoeser At the moment, in order to revise search results by applying a sort, one still has to return to the search bar and click the "search" button after choosing the sort type. So perhaps it makes sense to leave the sort selection coupled with the search bar until the floating sort field (and live results sorting or a second search button?) is implemented.

blms:
@rlskoeser weird lighthouse error just now—"Failed to determine the port of a local Solr instance, cannot create geniza!"

I don't think I changed any Solr config in my code… should I make an empty commit to retrigger that workflow?

rlskoeser:
> @rlskoeser [weird lighthouse error](https://github.com/Princeton-CDH/geniza/runs/4070189967?check_suite_focus=true) just now—"Failed to determine the port of a local Solr instance, cannot create geniza!"
> 
> I don't think I changed any Solr config in my code… should I make an empty commit to retrigger that workflow?

No, don't do an empty commit — I can trigger it to run again from github actions. I'll do that now, and can investigate if it still fails

blms:
Thanks @rlskoeser, it worked! Also looks like adding `unpkg.com` to the allowed hostnames in Percy fixed the arrow issue.

geniza/corpus/templatetags/corpus_extras.py
```diff
@@ -0,0 +1,12 @@
...
+register = template.Library()
+
+
+@register.filter
+def alphabetize(value):
+    """Lowercases, then alphabetizes, a list of strings"""
+    if isinstance(value, list) and all(isinstance(s, str) for s in value):
+        return sorted([s.lower() for s in value])
+    else:
+        raise TypeError("Argument must be a list of strings")
```
rlskoeser, line 12: I'm a little surprised you needed a custom template tag for this. I'm not sure that lower-casing the tags is the desired behavior, either... oh, but of course we're displaying all caps so case doesn't matter.

Ok, makes sense and I'm good with this approach.  

For consistency, we should apply to the document detail page as well — but if it doesn't work as-is we can hold off (I think we have a queryset of taggit tags instead of a list of strings, so may need some adjustment).

rlskoeser (approved):
Code changes look good to me, and I'm in agreement with holding off on moving the sort filter for now.  

Noticed a small style thing on percy, not sure it's relevant for this pr or where to track: in every browser but Firefox, the link underline style on the "view document details" link continues under the phosphor icon, but disjointed.

geniza/corpus/templates/corpus/document_list.html
```diff
@@ -5,30 +5,33 @@
 {% block meta_description %}{{ page_description }}{% endblock meta_description %}
 
 {% block main %}
-    <h1>{{ page_title }}</h1>
```
rlskoeser, line 8: I almost missed that you removed this! I know there is no title in Gissoo's designs (which now I'm kind of questioning a little bit); I think it's still valuable for screen readers, and maybe search engines. I propose we preserve the page-level h1 and make it screen-reader only.

geniza/corpus/templatetags/corpus_extras.py
blms: Yeah, weirdly Django's only builtin filter/template tag for sorting is `sortdict`, which requires a `dict` input rather than a `list`. And exactly on the lowercasing—I was so confused for a minute about why they weren't displaying alphabetically, then I realized some had an uppercase first letter and some didn't!

Indeed it's not a list of strings in the document detail page, so probably good to hold off for #306 or thereabouts—went ahead and added it into the PR desc there so I don't forget.

geniza/corpus/templates/corpus/document_list.html
```diff
@@ -5,36 +5,33 @@
...
+        <fieldset>
+            {{ form.q }}
+            {{ form.sort }}
+            <label for="sort">{{ form.sort.label }}</label>
+            <input type="submit" value="search" />
+        </fieldset>
+        {% for field, errors in form.errors.items %}
+            {# no clean way to get field labels here, so we just collapse the error lists #}
+            {% for error in errors %}
+                <li>{{ error }}</li>
```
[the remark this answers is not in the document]
blms: I believe these are builtin Django form errors, so they should be automatically translated by Django based on the locale. Not entirely certain—I don't see any logic about it in our code. [Django docs](https://docs.djangoproject.com/en/3.2/ref/forms/fields/#error-messages) say overrides should go in an `error_messages` dict for each field in the form, if we want to override default behavior in the future.

blms:
@gissoo should be ready for another visual review now!

gissoo:
@blms thank you, Ben! I think the only two things that do not match the designs now are:
1. the arrow icon next to the "view document details" appears to have a separate underline on all browsers except FF.
Screen Shot 2021-11-01 at 3 20 39 PM

2. @blms thanks for explaining your logic for the sort option placement ...I think we will need to revisit the designs on mobile ...if we are implementing a second submit button for the sort options it would help the current design decisions but if not then the designs will need to be revised a bit on mobile as this was not discussed previously when reviewing the designs. usability-wise I strongly suggest a second submit button for the sort options as it would better support the users' preference of navigating between search and sort (according to usability testing done for a previous project as well.) – let's go through with it for now but flag to fix it in the future.

blms:
> @blms thank you, Ben! I think the only two things that do not match the designs now are:
> 
> 1. the arrow icon next to the "view document details" appears to have a separate underline on all browsers except FF.

Thanks @gissoo. I wonder if this one might make more sense for #309 since it's about typography, while this one was more focused on alignment and spacing. It's not sizing related, though the arrow should be larger to match the design as well, so maybe it makes sense to roll that all into #309.

gissoo:
@blms I agree! Then I approve! Thanks!!

PR #340, opened 2021-10-28
Localize search form and regenerate messages

rlskoeser:
Was looking at the localization and noticed that we missed adding it to the UI text in the search form, and I think also had not regenerated the message files in a while.


blms (approved):
This all looks correct to me, and I'll be on the lookout for this for future work with UI text.

Strange result with Percy, but I don't see any visual changes caused by this PR.

Issue #235, opened 2021-06-07
As a user, I want to see document titles that include shelfmark and type so I can distinguish documents at a glance.


rlskoeser:
## testing notes

visit [document type list](https://test-geniza.cdh.princeton.edu/admin/corpus/documenttype/) on the admin site:
- [x] confirm that display label is listed, and has been populated for types where display is different than type (should match Marina's list below)
- [ ] confirm you can edit or create a document type with display label — confirm you can add/or update it or leave it blank

Visit the public [document search page](https://test-geniza.cdh.princeton.edu/documents/)
- [x] Verify that document titles display in the format "**Document type**; shelfmark"
- [x] Verify that the document types display according to the labels specified below 
- [x] Verify that documents with unknown type read "Unknown type", rather than just "Unknown"
- Navigate to a document detail page
  - [x] Verify that document titles display in the same format as on the search page
  - [x] Verify document types display the correct labels
  - [ ] Verify documents with unknown type display correctly

## dev notes
Revise document display title to match the designs, and use the following labels provided by @mrustow based on document type.

| short name for back end | full name for front end |
|---|---|
| legal | legal document |
| letter | letter |
| list or table | list or table |
| literary | literary text |
| paraliterary | paraliterary text |
| state | state document |
| unknown | unknown type |

This applies to document detail page and search results

## dev notes
- add optional `display_label` field to document type model
- write a data migration to populate display labels that differ from short name
- include display label in admin list and edit form for document types
- revise `Document` `title` property to display label and follow format from the designs
- include document title in `index_data`
- display document title in search results

## Steps to test

1. Visit the document search page
2. Verify that document titles display in the format "**Document type**; shelfmark"
3. Verify that the document types display according to the labels specified in #235 
4. Verify that documents with unknown type read "Unknown type", rather than just "Unknown"
4. Navigate to a document detail page
5. Verify that 2-4 remain true on this page

## Deploy notes 

- Solr reindexing is required on deploy.

richmanrachel:
@rlskoeser - For document types, the save button to change the labels doesn't work (it took a long time to load then landed on a 405 not allowed page).

richmanrachel:
The unknown type also isn't showing up on the document detail view in the details (but the header is correct):
![image](https://user-images.githubusercontent.com/68028846/139945570-18a83467-aba0-450e-be1d-f5b57d57ed7b.png)

richmanrachel:
Just for comparison, here's a letter:
[...]

PR #341, opened 2021-11-02
#235 - Add display_label field to document type, update index and views

blms:
### What this PR does
- Per #235:
  - Adds optional `display_label` field to `DocumentType` model
  - Updates `DocumentType` `__str__` to use `display_label` if available
  - Adds data migration to add the provided labels
  - Updates all instances of "Unknown" to "Unknown type"
  - Updates `Document` solr index to use the new `__str__` method
  - Fixes tests with duplicate DocumentTypes
  - Adds SASS mixin for properly formatted doctype in titles, used in both search and detail view

### Steps to test

1. Visit the document search page
2. Verify that document titles display in the format "**Document type**; shelfmark"
3. Verify that the document types display according to the labels specified in #235 
4. Verify that documents with unknown type read "Unknown type", rather than just "Unknown"
4. Navigate to a document detail page
5. Verify that 2-4 remain true on this page

### Deploy notes 

- Solr reindexing is required on deploy.


geniza/corpus/templates/corpus/document_detail.html
```diff
@@ -18,7 +18,14 @@
     <!-- document details -->
 
     {# title #}
-    <h1>{{ document.title }}</h1>
+    <h1>
+        {% if document.doctype %}
+            <span class="doctype">{{ document.doctype }}</span>
+        {% else %}
+            <span class="doctype">Unknown type</span>
```
rlskoeser, line 25: "unknown type" needs to be marked for translation here and in the model code

geniza/corpus/templates/corpus/document_detail.html
```diff
@@ -18,7 +18,14 @@
...
+        {% endif %}
```
rlskoeser, line 26: This can be simpler too, actually, although I'm not sure how you integrate it with the translation (maybe need to get the translation text first).
```suggestion
         <span class="doctype">{{ document.doctype|default:"Unknown type" }}</span>
```

geniza/corpus/tests/test_corpus_models.py
```diff
@@ -208,6 +208,12 @@ def test_str(self):
         doctype = DocumentType(name="Legal")
         assert str(doctype) == doctype.name
 
+    def test_display_label(self):
+        legal = DocumentType(name="Legal", display_label="Legal document")
+        assert legal.display_label == "Legal document"
+        no_label = DocumentType(name="test")
+        assert not no_label.display_label
```
rlskoeser, line 215: What are you testing here?

What you need to test is the new case for `str` when `display_label` is set

geniza/corpus/views.py
```diff
@@ -137,6 +137,7 @@ def get_context_data(self, **kwargs):
             {
                 "page_title": self.page_title(),
                 "page_description": self.page_description(),
+                "id": "detail",
```
rlskoeser, line 140: Let's specify document or document-detail, since we could have other kinds of detail pages.

And now that I see it again I'm not sure about using `id` for the attribute here, that seems potentially misleading. Could you come up with something a bit more specific to how this is used?

sitemedia/scss/main.scss
```diff
@@ -14,3 +14,4 @@
 
 // Page-specific styles
 @use "pages/search";
+@use "pages/detail";
```
rlskoeser, line 17: here too, I think document would be clearer than detail

rlskoeser (changes requested):
In general this looks good, just a few things that need a bit of revision.

geniza/corpus/migrations/0020_populate_documenttype_display_label.py
```diff
@@ -0,0 +1,39 @@
...
+    DocumentType = apps.get_model("corpus", "DocumentType")
+    DocumentType.objects.update_or_create(
+        name="Legal", defaults={"display_label": "Legal document"}
+    )
+    DocumentType.objects.update_or_create(
+        name="Letter", defaults={"display_label": "Letter"}
+    )
+    DocumentType.objects.update_or_create(
+        name="List or table", defaults={"display_label": "List or table"}
+    )
```
rlskoeser, line 16: one more: there's no need to set the display label when it is the same as the name, so please drop these two

geniza/corpus/models.py
```diff
@@ -524,7 +527,7 @@ def index_data(self):
         index_data.update(
             {
                 "pgpid_i": self.id,
-                "type_s": self.doctype.name if self.doctype else "Unknown",
+                "type_s": str(self.doctype) if self.doctype else "Unknown type",
```
rlskoeser, line 530: We _might_ need to differentiate between the actual document type and the display label for the type — we're going to have a facet/filter on document type and I'm not sure which one makes most sense in that context. Seems reasonable to assume for now that they should be consistent, so let's leave this for now.

geniza/corpus/models.py
```diff
@@ -221,8 +221,11 @@ class DocumentType(models.Model):
     """The category of document in question."""
 
     name = models.CharField(max_length=255, unique=True)
+    display_label = models.CharField(max_length=255, blank=True)
```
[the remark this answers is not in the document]
blms: Thanks! How do we handle translations for help text, by the way?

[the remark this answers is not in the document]
rlskoeser: We need to start tagging them! I was just reading more about django localization because I had a question about how to test it, and I saw that they recommend you explicitly set model verbose names and make them translatable — which we have _not_ done yet. I think hold off on making this one translatable and I'll add a chore to set it up across all the models (verbose names, help text, etc.)

geniza/corpus/views.py
blms: I ended up going with `page_type` here, let me know if that doesn't make sense!

geniza/corpus/migrations/0019_documenttype_display_label.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ("corpus", "0018_document_doctype_help_link"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="documenttype",
+            name="display_label",
+            field=models.CharField(blank=True, max_length=255),
```
rlskoeser, line 16: It irks me that django creates migrations for meta changes like help text, and I try to avoid creating unnecessary migrations, especially within a feature branch.

Please integrate this help text change to the initial migration, then delete 0021 and confirm it's no longer needed.

```suggestion
            field=models.CharField(
                blank=True,
                help_text="Optional label for display on the public site",
                max_length=255,
            ),
```

geniza/corpus/models.py
```diff
@@ -486,7 +492,7 @@ def has_image(self):
     @property
     def title(self):
         """Short title for identifying the document, e.g. via search."""
-        return f"{self.doctype or 'Unknown'}: {self.shelfmark_display or '??'}"
+        return f"{self.doctype or _('Unknown type')}; {self.shelfmark_display or '??'}"
```
rlskoeser, line 495: 🤔 I wonder if we need to translate "??"

rlskoeser (approved):
Looks great, thanks for making the changes; `page_type` makes sense to me, much more meaningful than `id`!

Requesting one small change with the migrations, then we can merge.

blms:
All set, thanks @rlskoeser! Let me know also if I should be merging things once all have approved; am I right that the QA workflow is merge, move to "dev complete" in ZenHub, and include testing instructions in the PR (or does that go in the issue)?

PR #345, opened 2021-11-03
#309 - Search page type sizes

blms:
### What this PR does
- Per #309:
  - Implements type sizes for the search page according to the designs
  - Removes the underline from "View document details" link per designs
- Slightly reorganizes CSS to prevent the same selector from having two separate definitions

### Additional notes
- As I mentioned previously, the type sizes are defined relative to the page root—in terms of `rem`—rather than absolute `px` values. I went ahead and tried to match the values as closely as I could. I mapped it like this:
  - 16px --> `$text-size-md: 1rem;`
  - 18px --> `$text-size-2md: 1.15rem;`
  - 20px --> `$text-size-lg: 1.25rem;`
  I don't believe there are any font sizes besides those in the designs.
- I feel like the "View document details" type looks massive on desktop, but it's matching the design at 20px, the same size as the document title. It could be that the lack of an image is making it look larger than it should. On mobile it's smaller than the doc title, which makes it look nicer on mobile, in my opinion.
- I noticed that the description text is in "redacted script" but is size 20px on desktop/18px on mobile, in other words, the same size as the doc title. I wasn't sure if this was intended—I left it untouched at `text-size-md` (actually the "body" style mixin) but I can change it to match the doc title if need be.

### DIFF
--- a/sitemedia/scss/base/_typography.scss
+++ b/sitemedia/scss/base/_typography.scss
@@ -2,8 +2,9 @@
 // Typographic styles (size, weight, etc.) used across the project.
 // -----------------------------------------------------------------------------
 
-@use "fonts";
+@use "breakpoints";
 @use "colors";
+@use "fonts";
 
 // Type scale: generate sizes according to harmonic ratio. This scale is based
 // on the "Major Third" scale, with a scale factor of 1.25. See:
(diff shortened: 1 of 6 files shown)


rlskoeser (approved):
code changes all look fine to me; defer to @gissoo on visual review and your questions about sizes

gissoo:
@blms Thank you!! Yes, everything you mentioned above is intentional. We are very close to coming to a decision about typefaces, I suggest you don't spend more time on this now till we can replace the redacted script with the new typeface. Would that be ok? I think the sizes will stay the same which is amazing but just in case (efficiency-wise).

blms:
Fantastic, thanks @gissoo! I'll leave the type as-is until I hear more about this, then.

Issue #311, opened 2021-10-12
Implement styles for a search submit button on mobile and desktop based on the designs


gissoo:

blms:
Reviewed in QA deployment 👍

Issue #310, opened 2021-10-12
Implement styles for the search box on mobile and desktop


gissoo:

blms:
Reviewed in QA deployment 👍

Issue #338, opened 2021-10-28
Revise the scholarship records and the external links current format in the search results, and the document detail view navigation designs so that it's more readable for the users


gissoo:
I have made updates on the [search results](https://www.figma.com/file/HpGBOZi9lO8B3nCAx3d4Fj/Princeton-Geniza-(Project-team)?node-id=721%3A134)

and the [document detail view wireframes](https://www.figma.com/file/HpGBOZi9lO8B3nCAx3d4Fj/Princeton-Geniza-(Project-team)?node-id=572%3A0), where the page navigation for "scholarship records" and the "external links" exist.

## dev notes:

- Also translate aria-label for doc detail view navigation

gissoo:
the tabs are now back to having the count at the end with () – and in the search results the counts come first with no ()

Issue #312, opened 2021-10-12
Implement styles for site navigation on mobile 


gissoo:
## testing notes
- [ ] On desktop, visit any user-facing page (e.g. https://test-geniza.cdh.princeton.edu/documents/) and hover over the "About" section of the top navigation menu, and verify that a sub-menu dropdown appears with more items
- [ ] On mobile, visit any user-facing page and click the hamburger menu icon next to "GENIZA" at the top of the screen.
  - [ ] Verify that clicking on this icon brings up a menu with several options
  - [ ] Verify that clicking "About" in this menu brings up a sub-menu with more items
  - [ ] Verify that clicking the back icon (<) returns to the previous menu
  - [ ] Verify that clicking the close icon (X) closes all menus

## Dev notes
- [x] Also translate aria-label for nav

gissoo:
- Thank you, @rlskoeser ! I verified everything and it works well on firefox and safari. However, on Chrome mobile nav when I select the "About" no submenu items show up 

- (may not be useful: if you're on chrome desktop, hover over the about drop down, and then shrink the page to get the mobile nav, select the mobile nav it already opens in "about" expanded and does not show the default menu items unless you select the back/left chevron)

blms:
@gissoo Thanks for testing! Strange, I can't replicate that Chrome issue. Which OS and version of Chrome are you using? And are you testing on a mobile device or the responsive design mode of the desktop browser?

gissoo:
@blms I can't replicate the issue on mobile, it's just on my desktop and I think my browser is older. If you can't replicate the issue on desktop then ignore it, there shouldn't be a problem.

PR #346, opened 2021-11-03
#310, #311 - Implement search box and button styles

blms:
Might have made more sense to roll this in with #345, but I had already created a new feature branch for it!

### What this PR does
- Per #310:
  - The bulk of this was already accomplished in #336
  - Fixes some issues with form elements on different screen sizes via better use of `max-width`

- Per #311:
  - Changes the search submit button into an icon button using the Phosphor magnifying glass
  - Hides the search button on mobile, but leaves it accessible to screen readers

### Additional notes
- Question for @rlskoeser: Should `aria-label` names be translated?

### DIFF
--- a/geniza/corpus/templates/corpus/document_list.html
+++ b/geniza/corpus/templates/corpus/document_list.html
@@ -11,7 +11,9 @@
             {{ form.q }}
             {{ form.sort }}
             <label for="sort">{{ form.sort.label }}</label>
-            <button type="submit" aria-label="search">
+            {# Translators: Search submit button #}
+            {% translate 'Submit search' as search_label %}
+            <button type="submit" aria-label="{{ search_label }}">
                 <i class="ph-magnifying-glass"></i>
             </button>
         </fieldset>
--- a/geniza/corpus/templates/corpus/document_list.html
+++ b/geniza/corpus/templates/corpus/document_list.html
@@ -11,7 +11,9 @@
             {{ form.q }}
             {{ form.sort }}
             <label for="sort">{{ form.sort.label }}</label>
-            <button type="submit"><i class="ph-magnifying-glass"></i></button>
+            <button type="submit" aria-label="search">
+                <i class="ph-magnifying-glass"></i>
+            </button>
         </fieldset>
         {% for field, errors in form.errors.items %}
             {# no clean way to get field labels here, so we just collapse the error lists #}
--- a/geniza/corpus/templates/corpus/document_list.html
+++ b/geniza/corpus/templates/corpus/document_list.html
@@ -11,7 +11,7 @@
             {{ form.q }}
             {{ form.sort }}
             <label for="sort">{{ form.sort.label }}</label>
-            <input type="submit" value="search" />
+            <button type="submit"><i class="ph-magnifying-glass"></i></button>
         </fieldset>
         {% for field, errors in form.errors.items %}
             {# no clean way to get field labels here, so we just collapse the error lists #}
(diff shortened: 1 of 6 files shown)


rlskoeser:
> Question for @rlskoeser: Should `aria-label` names be translated?

Good question, @blms!  I did some quick searching and confirmed what I suspect: yes, it should be translated. Thanks for keeping an eye on this — interesting to see the overlap between accessibility and localization!

blms:
Done, thanks @rlskoeser! Also noticed some untranslated in [`document_tabs.html`](https://github.com/Princeton-CDH/geniza/blob/develop/geniza/corpus/templates/corpus/snippets/document_tabs.html#L3) and [`nav.html`](https://github.com/Princeton-CDH/geniza/blob/develop/geniza/templates/nav.html#L2), but those could be dealt with separately, perhaps in #338 for `document_tabs.html` and #312 for `nav.html`.

geniza/corpus/templates/corpus/document_list.html
```diff
@@ -11,7 +11,11 @@ <h1 class="sr-only">{{ page_title }}</h1>
             {{ form.q }}
             {{ form.sort }}
             <label for="sort">{{ form.sort.label }}</label>
-            <input type="submit" value="search" />
+            {# Translators: Search submit button #}
+            {% translate 'Submit search' as search_label %}
+            <button type="submit" aria-label="{{ search_label }}">
+                <i class="ph-magnifying-glass"></i>
+            </button>
```
rlskoeser, line 18: I was testing a little bit with both screen reader and keyboard navigation; the tab order is confusing since it doesn't match the visual layout — although of course it makes sense to have a submit button after the end of the form so you can submit any changes. I'm not requesting any changes now, but I think we'll need to add that second submit button at the bottom of the form sooner then later, and we should move this one so it's position in the DOM is consistent with where it's displayed.

sitemedia/scss/components/_searchform.scss
```diff
@@ -47,20 +51,34 @@
...
             margin-left: 0;
             background-color: colors.$gray-30;
             border: 1px solid colors.$gray-50;
-            height: spacing.$spacing-xl;
+            font-size: typography.$text-size-xl;
+            display: flex;
+            align-items: center;
+            justify-content: center;
+            @include a11y.sr-only;
+            @include a11y.sr-only-focusable;
```
rlskoeser, line 71: Do we need both? Should `sr-only-focusable` imply `sr-only` ?

I see from the note on the existing class/mixin that they're meant to be applied topegher, just wondering if you can think of any case when we would want `sr-only-focusable` without `sr-only`

rlskoeser (approved):
html/scss changes look fine to me

geniza/corpus/templates/corpus/document_list.html
blms: Yes, I wanted at least the DOM to have the submit button at the end of the form, since it might be confusing for screen readers to encounter the submit button before some form fields. But the tradeoff is that it does make keyboard navigation somewhat confusing. I wonder how most people handle this—it doesn't seem like WCAG at least requires you to put the submit button at the end of the form, so maybe it could just exist in the middle of it even in the DOM.

sitemedia/scss/components/_searchform.scss
blms: I can't, so maybe good to go ahead and just include the `sr-only` mixin inside of the `focusable` one.

PR #347, opened 2021-11-03
Add intcomma to search results total

rlskoeser:
small change to humanize the total results on the document list, so it's more readable


blms (approved):
Smart! Didn't know about django `humanize` but may come in handy in the future.

Issue #308, opened 2021-10-12
Implement type sizes for the document detail view on mobile and desktop based on the designs


gissoo:
## Testing notes

- [x] Check document detail views, verify that type sizes match the designs

blms:
Reviewed in QA deployment 👍

Issue #257, opened 2021-08-10
Multi-word tags get broken up into single-word tags


alanelbaum:
**Describe the bug**
When editing a document record, if you add a tag that is more than one word (e.g., "Arabic script") and save your changes, the tag automatically breaks down into multiple single-word tags (e.g., "Arabic" and "script"). It does not do this when there's a colon (e.g., "illness: eye"). I haven't been able to identify any workaround. A corollary issue is that I've created many dozens of bogus tags this way, so will need to delete those and reaggregate the single-word tags into multi-word tags.

**To reproduce**
1. Go to any document record
2. Add any tag that consists of 2 or more words separated by spaces
3. Scroll down and click 'save'

**Expected behavior**
If you add a tag such as "Arabic script" or "Ibn Habib's patient," the multi-word tag should not be broken up into smaller tags. 

**Screenshots**
This document (id20426) needs a tag to be properly associated with all the other fragments tagged "Ibn Habib's patient." So for now it has 3 erroneous tags ("Ibn" and "Habib" and "patient" which will need to be deleted and replaced with the single tag "Ibn Habib's patient."

<img width="1098" alt="Screen Shot 2021-08-10 at 1 46 54 PM" src="https://user-images.githubusercontent.com/1414260/128925200-1a91f927-b8fa-478a-83b1-35998fd79b8c.png">

**Device information**
 - Device: MacBook Air
 - OS and version: macOS 10.15.7
 - Browser and version: Firefox 90 and Chrome 92
 - Site version: v 0.4.0

**Additional context**

alanelbaum:
(Ignore previous comment, sorry! It seemed like it was working for some tags, but then I got this: "power, of, attorney"
<img width="843" alt="Screen Shot 2021-08-15 at 5 28 02 PM" src="https://user-images.githubusercontent.com/1414260/129495858-5aef8c2b-f144-4105-849a-d3abeea9001b.png">

kmcelwee:
This problem with `django-taggit` has been encountered before: https://stackoverflow.com/questions/30513783/django-taggit-how-to-allow-multi-word-tags

richmanrachel:
@kmcelwee - adding a new tag was no problem! I did get this server error 500 when I tried to add the same new tag to a document with existing tags, though... 
![image](https://user-images.githubusercontent.com/68028846/148300927-5a6814ce-e41f-48a3-8a11-6a3795ca1843.png)

richmanrachel:
[...]

PR #348, opened 2021-11-04
#306, #308 - Add document detail view styles

blms:
### What this PR does
- Per #306:
  - Implements spacing, alignment, and margins for the document detail view
- Per #308:
  - Implements type sizes for document detail view
- Implements new link colors and styling
- Implements alphabetical ordering of tags in document detail view
- Removes document type and shelfmark from metadata section of document detail view


rlskoeser:
@blms ooh, sorry we did not tell you! the colors in the scss we have now are placeholders, Gissoo is still deciding on the colors — she had a set she was working on and refining, but then colors got back-burnered because of other work!

blms:
> @blms ooh, sorry we did not tell you! the colors in the scss we have now are placeholders, Gissoo is still deciding on the colors — she had a set she was working on and refining, but then colors got back-burnered because of other work!

Ah, OK! Good to know, I'll revert link styling back to what it was before for the time being.

geniza/corpus/models.py
```diff
@@ -442,6 +443,9 @@ def all_tags(self):
 
     all_tags.short_description = "tags"
 
+    def alphabetized_tags(self):
+        return self.tags.order_by(Lower("name"))
+
```
rlskoeser, line 448: I feel like this would make more sense to put on the tag manager, but I'm not sure it's worth the extra effort to extend and customize at this point.

There is some possibly related work with taggit that I should look into — turning on `TAGGIT_CASE_INSENSITIVE` or possibly even forcing tags to lower case (which would simplify this!). The particular issue I'm thinking of is #257 ; we should probably discuss whether that should be a technical or data entry change.

geniza/corpus/templates/corpus/snippets/document_tabs.html
```diff
@@ -4,13 +4,21 @@
...
         {% endwith %}
         {% with n_links=document.fragment_urls|length %}
-            {# Translators: n_links is number of external links #}
-            {% blocktranslate asvar elink_text %}External Links ({{ n_links }}){% endblocktranslate %}
+            {% if n_links == 1 %}
+                {% blocktranslate asvar elink_text %}1 External Link{% endblocktranslate %}
+            {% else %}
+                {# Translators: n_links is number of external links #}
+                {% blocktranslate asvar elink_text %}{{ n_links }} External Links{% endblocktranslate %}
+            {% endif %}
```
rlskoeser, line 21: Gissoo and I discussed this change and decided it made more sense the way it was and the tabs don't need to match the search result display (since here they are page labels, where on the search result they are describing the document). Sorry for that, please revert the change.

For future reference, here's an example of a block translate that needs to handle plural and singular; you have to pass a counter https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#blocktranslate-template-tag

geniza/corpus/templates/corpus/document_detail.html
```diff
@@ -29,42 +29,49 @@ <h1>
...
-        {# metadata #}
-        <dl class="metadata">
-            <dt class="sr-only">{% translate 'Shelfmark' %}</dt>
-            <dd>{{ document.shelfmark }}</dd>
-            {% if document.doctype %}
-                <dt class="sr-only">{% translate 'Document Type' %}</dt>
-                <dd>{{ document.doctype }}</dd>
+        <section>
+            <h2 class="sr-only">
+                {# Translators: label for metadata section #}
```
rlskoeser, line 34: Not how clear this will be for translators, let's add a little more detail:

```suggestion
                {# Translators: label for document metadata section (editor, date, input date) #}
```

blms:
All set on those changes, thanks @rlskoeser, should be ready for @gissoo now.

gissoo:
@blms Thanks!! This looks great!! The only thing that looks a bit odd is the spacing between the tabs on mobile except on Safari
Screen Shot 2021-11-05 at 5 38 15 PM

blms:
Thanks @gissoo, I think those are fixed now! Would you be able to do another review just to make sure everything's all set?

PR #349, opened 2021-11-04
keyword in context highlighting for search term matches in description

rlskoeser:
- configure solr highlighting on search term in description text
- render highlighting in results if present, otherwise display description excerpt as usual

I ported a `dict_item` template from one of our other codebases for an easy way to get the highlighted content for a given document.

I made a slight update to a pytest fixture in `parasolr` (our locally developed python solr client) because it made testing the assertions on the solr queryset much easier and more readable — so tests will require the develop version of that package until I release a new version.


blms (approved):
All looks good to me! That `dict_item` filter looks useful. Nice improvements to the existing template tests as well. I'll be sure to update my local parasolr package!

Issue #339, opened 2021-10-28
localize search form


rlskoeser:
UI text in the search form needs to be made available for translation

rlskoeser:
reviewed in qa; we don't have translations yet for the search form fields, but it looks like everything is setup properly once we do, and now we an easily test by editing the url

PR #351, opened 2021-11-04
Configure public site urls for i18n

rlskoeser:
When I went to test the form localization (#339) I discovered I'd missed a step with the i18n setup in the url configuration; this remedies that.

I revised the permalink url logic so it does _not_ include a language code: this way all versions of a document will show the same url for citation purposes, but it will resolve to the user's preferred language or the site default.

Note that this means there's now a difference between `get_absolute_url` and `permalink`; this is because I think the absolute urls we're generating elsewhere should be relative to the currently active language.

changes in this PR:
- adds `i18n_patterns` to main url config for public site urls (currently only `corpus`)
- revise document permalink behavior so it is language independent 
- adjust tests for the change to permalink

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -12,6 +12,7 @@ from django.db.models.functions import Concat
 from django.db.models.query import Prefetch
 from django.urls import reverse
 from django.utils.safestring import mark_safe
+from django.utils.translation import activate
 from django.utils.translation import gettext as _
 from parasolr.django.indexing import ModelIndexable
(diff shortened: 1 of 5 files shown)


blms (approved):
This all looks good, and I like the trick of activating `en` and then removing it from the URL to get the generic one :)

rlskoeser:
> This all looks good, and I like the trick of activating `en` and then removing it from the URL to get the generic one :)

😁 Thanks! Surprisingly django does not seem to have a built in way of generating the url without the language prefix.

Issue #294, opened 2021-10-07
As a user, I want to sort search results by the number of scholarship records so I can easily find documents with scholarly work available or that have not been written about.


rlskoeser:
## testing notes
- [ ] visit the document search page, and without entering a query, confirm that results are sorted by "Scholarship Records (Most–Least)"
- [x] verify that you can select "Scholarship Records (Most–Least)", "Scholarship Records (Most–Least)", and "Relevance" in the "Sort by" dropdown
- [x] after typing a query and selecting a sort option, click the search submit button, and confirm that the results are as expected
- [x] confirm that sorting by "Scholarship Records (Most–Least)" displays the expected results
- [x] confirm that sorting by "Scholarship Records (Least-Most)" displays the expected results
- [x] confirm that sorting by "Relevance" displays the expected results

## dev notes
- [x] add sort options to the form : "Scholarship Records (Most–Least)" and "Scholarship Records (Least–Most)"
- [x] add logic to the view to sort the solr queryset appropriately using the total scholarship field that is already indexed
- [x] confirm the rendered form on the search results page displays the correct sort status
- [x] add unit tests for the sort logic in the view

richmanrachel:
@rlskoeser - It looks good!  It defaults to relevance, but we can switch the sort-by, and honestly relevance is better as a default.

PR #353, opened 2021-11-05
#294 - Implement sort by scholarship, asc and desc

blms:
### What this PR does

- Per #294:
  - Adds sort options "Scholarship Records (Most–Least)" and "Scholarship Records (Least–Most)"
  - Adds tests for the above
  - Sets initial sort option (w/ empty query) as "Scholarship Records (Most–Least)"
  - Removes disabled "relevance" sort and related tests

### Additional notes

- I made the decision to set "Scholarship Records (Most–Least)" as the initial ordering of items when a query is not input manually. Not sure what makes the most sense as a default for an empty query, but this is probably better than "Relevance."
- I also removed the logic to disable the "Relevance" sort option when there is no search query in the submitted form (and related tests). To me, this was confusing from a user standpoint, because it meant that the user would need to click the search button and submit the form once _before_ it was possible to enable the "Relevance" option, even though they might have typed something. In other words, I imagine a user would want to be able to choose that option between typing a query and clicking the submit button, even when they're on a URL with no query param.
  - I also set "Relevance" to be the default selection in the dropdown when one is about to perform a search.
  - I could see it working such that the relevance option is disabled when the search box is empty, and re-enables via JS when one starts typing into the search box, but then the user would still have to manually pick it from the dropdown—whereas this way, it's already the selected option.
  - All of this might be irrelevant if the goal is to have a "live sort" where results update automatically on changing the sort dropdown. For the time being though, this seems to make sense with a form submission.


geniza/corpus/tests/test_corpus_views.py
```diff
@@ -255,6 +315,76 @@ def test_get_context_data(self, rf):
             == docsearch_view.queryset.get_highlighting.return_value
         )
 
+    def test_scholarship_sort(self, empty_solr, source):
+        """integration test for sorting by scholarship asc and desc"""
+
+        doc_no_record = Document.objects.create()
```
rlskoeser, line 321: you could use the existing `document` and `join` fixtures here; I don't feel strongly, not sure how much it helps with readabiilty

geniza/corpus/tests/test_corpus_views.py
```diff
@@ -255,6 +315,76 @@ def test_get_context_data(self, rf):
...
+        doc_one_record = Document.objects.create(description="My record count is one")
+        Footnote.objects.create(
+            content_object=doc_one_record,
+            source=source,
+            doc_relation=Footnote.EDITION,
+        )
+        assert doc_one_record.index_data()["scholarship_count_i"] == 1
```
rlskoeser, line 328: this should be tested elsewhere

geniza/corpus/views.py
```diff
@@ -37,17 +40,12 @@ def get_form_kwargs(self):
...
+        # sort by relavance if query text exists and no sort chosen
+        elif form_data.get("q", None):
             form_data["sort"] = "relevance"
-        # sorting TODO
-        # else:
-        # form_data['sort'] = self.initial['sort']
-
-        # use initial values as defaults
-        # for key, val in self.initial.items():
-        # form_data.setdefault(key, val)
```
rlskoeser, line 49: the default initial sort isn't getting set; I had to look at the django docs again to be reminded of this:
> These values are only displayed for unbound forms, and they’re not used as fallback values if a particular value isn’t provided.

So if we're setting form data and binding the form, we need to pass the initial sort value into our form data.

geniza/corpus/views.py
```diff
@@ -37,17 +40,12 @@ def get_form_kwargs(self):
         # use GET instead of default POST/PUT for form data
         form_data = self.request.GET.copy()
 
-        # always use relevance sort for keyword search;
-        # otherwise use default (sort by name)
-        if form_data.get("q", None):
+        # sort by chosen sort
+        if "sort" in form_data:
+            form_data["sort"] = form_data.get("sort", None)
+        # sort by relavance if query text exists and no sort chosen
```
rlskoeser, line 46: typo
```suggestion
        # sort by relevance if query text exists and no sort chosen
```

geniza/corpus/views.py
```diff
@@ -69,9 +67,11 @@ def get_queryset(self):
...
                     .also("score")
                 )  # include relevance score in results
-
-            # sorting TODO; for now, order by relevance
-            documents = documents.order_by("-score")
+            if search_opts["sort"]:
+                documents = documents.order_by(self.solr_sort[search_opts["sort"]])
+            else:
+                initial_sort = self.initial["sort"]
+                documents = documents.order_by(self.solr_sort[initial_sort])
```
rlskoeser, line 74: Initial sort needs to be set in `get_form_kwargs` so that it is always included in the search opts from the form. I guess right now it is doing the right thing but displaying the wrong value in the sort select in the rendered form.

rlskoeser (changes requested):
Something isn't working quite right with the initial form sort value. The tests look a little verbose/redundant to me but not sure anything needs to change. The rest looks ok.

I've discussed with Gissoo, and we are in agreement about making most scholarship records the default sort, at least for now.

You're right about the problems with relevance sorting logic — it's something we've debated about on other projects too; see for example the member search on Shakespeare and Company Project: https://shakespeareandco.princeton.edu/members/  (although that's simpler because there are only two sort options; I think we decided users don't get to choose on that one to avoid complications.)

My worry is that users will not know or take the time to switch the sort to relevance; but it seems ok to live with that for now until we are ready to add javascript to enhance the form behavior.

geniza/corpus/tests/test_corpus_views.py
blms: You're right—already tested in `test_corpus_models`!

rlskoeser (approved):
updates look good

Issue #354, opened 2021-11-08
change document title to use colon instead of semicolon


rlskoeser:
@rlskoeser and @gissoo - One additional comment from Marina after our meeting this morning: she would prefer that the punctuation between the Type and Shelfmark is a colon (not a semi-colon). I agree with this change.

_Originally posted by @richmanrachel in https://github.com/Princeton-CDH/geniza/issues/235#issuecomment-959537255_

## dev notes
- [x] change title on document search results
- [x] change title on document details page (display version and metadata head title)
- [x] update any unit tests impacted by the change

rlskoeser:
confirmed updated display on the test site

PR #355, opened 2021-11-08
#354 - Replace semicolon with colon in doc titles

blms:
### What this PR does
- Per #354:
  - Updates `Document` model `title` property, which is used in document detail view head metadata 
  - Updates `_typography.scss`, which is used in search results page, and document detail view title header
  - Updates tests to use colon


rlskoeser (approved):
Thanks for taking care of this so quickly!

Issue #298, opened 2021-10-12
As a user I would like to see which page I'm on when viewing search results and navigate between pages so I can see more results.


gissoo:
## testing notes
test on the document search https://test-geniza.cdh.princeton.edu/en/documents/
- [x] results should be paginated
- [x] test entering a search query with greater than 50 results, should have pages that you can click through
- [x] test entering a search query with fewer than 50 results, pagination should disappear

note that some work on design/style is forthcoming in #361

## dev notes
- [x] use django paginator on solr queryset search in document search view — should be similar to this code in ppa: https://github.com/Princeton-CDH/ppa-django/blob/main/ppa/archive/views.py#L347-L350
- [x] include needed pagination variables in template context
- [x] create template snippet to render pagination per design and include at top and bottom of search results; adapt from ppa pagination https://github.com/Princeton-CDH/ppa-django/blob/main/ppa/archive/templates/archive/snippets/pagination.html
- [x] unit tests to confirm pagination works: displays correctly, links work, can retrieve a secondary page

richmanrachel:
Perfect!

PR #356, opened 2021-11-09
#298 - Paginate search results

blms:
### What this PR does
- Per #298:
  - Adds pagination control to document search results

### Additional notes
- @rlskoeser: I'm not 100% sure about the way I split up logic between `get_queryset` and `get_context_data`. In general I tried to follow the example from the PPA project, so my `get_queryset` returns the entire queryset, and `paginate_by`/`get_context_data` does the actual pagination. I guess by using `paginate_by` I didn't even have to pass the Paginator result into the context manually.


blms:
I wonder if we should include Percy tests with several pages of search results?

gissoo:
> I wonder if we should include Percy tests with several pages of search results?
> 

I wonder that too – would help show the states visually (current page vs other ones) + the margins + the spaces between the numbers – what I currently see on Percy does not match the designs margin-wise, I can tell because on mobile the arrows are closer to the edges.

blms:
Thanks @gissoo, I'll take another crack at the margins on mobile! Just to clarify, they should be closer to the edges on mobile?

Another question that this brings up is what the pagination control should look like when there are 5 or fewer pages of results. Should desktop be modified such that the arrows are spaced how they are on mobile currently? Or should they be tighter with the number(s) like they are now?

geniza/corpus/views.py
```diff
@@ -75,20 +77,27 @@ def get_queryset(self):
...
-    def get_context_data(self):
-        context_data = super().get_context_data()
-        # should eventually be handled by paginator, but
-        # patch in total number of results for display for now
+    def get_context_data(self, **kwargs):
+        context_data = super().get_context_data(**kwargs)
+
+        paginator = Paginator(self.queryset, per_page=self.paginate_by)
+        page_num = self.request.GET.get("page", 1)
+        current_page = paginator.page(page_num)
```
rlskoeser, line 87: per django documentation, paginator object and current page object should already be set in context data; let's use those

https://docs.djangoproject.com/en/3.2/ref/class-based-views/mixins-multiple-object/#django.views.generic.list.MultipleObjectMixin.get_context_data

geniza/corpus/views.py
```diff
@@ -75,20 +77,27 @@ def get_queryset(self):
...
             {
-                "total": self.queryset.count(),
-                "page_title": self.page_title,
+                "highlighting": highlights,
                 "page_description": self.page_description,
+                "page_number": page_num,
+                "page_title": self.page_title,
                 "page_type": "search",
-                "highlighting": self.queryset.get_highlighting(),
+                "result_offset": (page_num - 1) * self.paginate_by,
```
rlskoeser, line 99: You need to get this value from the paginator; should use current page object start_index method `https://docs.djangoproject.com/en/3.2/ref/paginator/#django.core.paginator.Page.start_index

Doesn't need to be passed in separately, since the page object will already be in context.

(This matters if we configure a non-zero value for `orphans`, which we should probably think about doing.)

geniza/corpus/views.py
```diff
@@ -75,20 +77,27 @@ def get_queryset(self):
...
+
+        highlights = paged_result.get_highlighting() if paged_result.count() else {}
+        page_num = int(self.request.GET.get("page", 1))
         context_data.update(
             {
-                "total": self.queryset.count(),
-                "page_title": self.page_title,
+                "highlighting": highlights,
                 "page_description": self.page_description,
+                "page_number": page_num,
```
rlskoeser, line 96: I think it's better to get this from the page object, which is already included in context data; should be `Page.number`

geniza/corpus/templates/corpus/snippets/pagination.html
```diff
@@ -0,0 +1,67 @@
+{% load i18n corpus_extras %}
+{% spaceless %}
```
rlskoeser, line 2: Should we not show any pagination when `is_paginated` is false? (i.e., only one page of results)

Maybe this is a question for @gissoo

rlskoeser (changes requested):
You split the logic correctly — `get_queryset` should return the entire queryset, which then gets paginated.

Needs some revision to take better advantage of built in pagination functionality.

rlskoeser:
> > I wonder if we should include Percy tests with several pages of search results?
> 
> @blms I wonder that too – would help show the states visually (current page vs other ones) + the margins + the spaces between the numbers – what I currently see on Percy does not match the designs margin-wise, I can tell because on mobile the arrows are closer to the edges.

I agree, it would be more useful if we had more results. We need more results to test pagination anyway, don't we?  (We could set `paginate_by` to a lower value for testing purposes so we don't have to add a huge amount of data).

@gissoo can you help brainstorm the different cases we'd want to include? We can turn this into a chore to add them.

blms:
Thanks @rlskoeser, I didn't know about all that convenience with the Paginator object!

Quick question, do you think this makes sense in testing the Document Result template? `self.page_obj = Paginator([1], 1).page(1)` Just to have some paginator object to pass into the context.

rlskoeser (approved):

rlskoeser:
> Thanks @rlskoeser, I didn't know about all that convenience with the Paginator object!
> 
> Quick question, do you think this makes sense in testing the Document Result template? `self.page_obj = Paginator([1], 1).page(1)` Just to have some paginator object to pass into the context.

yes, that seems reasonable enough

blms:
Will leave this open for now until we add some Percy testing for multiple pages, and @gissoo weighs in on the questions above.

gissoo:
> > > I wonder if we should include Percy tests with several pages of search results?
> > 
> > 
> > @blms I wonder that too – would help show the states visually (current page vs other ones) + the margins + the spaces between the numbers – what I currently see on Percy does not match the designs margin-wise, I can tell because on mobile the arrows are closer to the edges.
> 
> I agree, it would be more useful if we had more results. We need more results to test pagination anyway, don't we? (We could set `paginate_by` to a lower value for testing purposes so we don't have to add a huge amount of data).
> 
> @gissoo can you help brainstorm the different cases we'd want to include? We can turn this into a chore to add them.
>

@rlskoeser @blms sure! thinking of showing what it looks like when:
1. you're on the first page 
2. you're on the last page 
3. you're on the middle page (equally far from the first and the last page)
4. you're on the first quarter
5. you're on the last quarter 
6. you're on a page with a 2-digit number and next to it can still see pages with 1-digit number 
7. you're on a page with a 3-digit number and next to it can still see pages with 2-digit numbers 

:D 
Are these good?

rlskoeser:
@gissoo do we need to test all of those on percy every time?

Sorry, my question was vague! what I meant was, what are the variations in documents that would be useful to include in our visual review? The things that occurred to me so far
- with transcription / without
- multiple shelfmarks
- single shelfmark

gissoo:
> Thanks @gissoo, I'll take another crack at the margins on mobile! Just to clarify, they should be closer to the edges on mobile? Or are they too close currently?
> 

@blms they are closer to the edges now than they should be – in the designs they look more compact together and far from the edges. 

> Another question that this brings up is what the pagination control should look like when there are 5 or fewer pages of results. Should desktop be modified such that the arrows are spaced how they are on mobile currently? Or should they be tighter with the number(s) like they are now?
>

@blms thanks! I don't think I designed for a case of 5 or fewer pages of results – we can add it as a chore for me to work on – the desktop looks like it matches the design, the mobile doesn't.

gissoo:
> > @gissoo do we need to test all of those on percy every time?
> > Sorry, my question was vague! what I meant was, what are the variations in documents that would be useful to include in our visual review? The things that occurred to me so far
> > ```
> > * with transcription / without
> > 
> > * multiple shelfmarks
> > 
> > * single shelfmark
> > ```
> 
> @rlskoeser 
These seem accurate – I can't think of anything right now as we don't have anything else in mvp – will let you know if I think of something

geniza/corpus/templates/corpus/snippets/pagination.html
blms: @gissoo What do you think here? I'm leaning towards not showing pagination controls when there's only one page.

blms:
Struggled for a while with CSS here, as I ran into problems with overflow when just directly applying the same styling as desktop to mobile. `justify-content: space-evenly` fixes this for the most part, but with fewer than 5 pages the page numbers still get rather too far spaced out. In that case the only way it seems to work is to apply the same styling as desktop, so I added a special case for it.

blms:
@gissoo Looks like Lighthouse doesn't approve of the search nav for mobile. Any ideas for how we might improve this?

Screen Shot 2021-11-09 at 2 54 42 PM

gissoo:
> @gissoo Looks like Lighthouse doesn't approve of the search nav for mobile. Any ideas for how we might improve this?
> 
> Screen Shot 2021-11-09 at 2 54 42 PM
>

@blms when you say "search nav" you're only talking about the pagination right? or also other things in the search and filter features? 

thanks for this! I knew the min was 40px for usability on mobile, interesting that they mention 48 – I can revise the design, @blms ! It would be a simple chore

blms:
@gissoo Sorry for the ambiguity! Yes, I meant the pagination controls only. That would be great if you could revise; I'll leave this PR open for the time being.

geniza/corpus/templates/corpus/snippets/pagination.html
rlskoeser: I've thought about this more — I don't think we should show pagination controls when there's only one page. Only thing to be aware of is we don't want the page content to shift too much if pagination appears/disappears depending on how the search is filtered, so will probably want to do something to keep the spacing the same even when pagination is not included.

geniza/common/management/commands/visual_tests.py
```diff
@@ -35,6 +35,9 @@ def take_snapshots(self, browser):
         browser.get("http://localhost:8000/documents/?q=physician")
         percy_snapshot(browser, "Document Search")
 
+        browser.get("http://localhost:8000/documents/?per_page=2")
+        percy_snapshot(browser, "Document Search (paginated)")
+
```
rlskoeser, line 40: Is there a way we can combine these two snapshots, or do we need both?

blms: Is my solution OK? Not the most meaningful search 😆 but at least it has a query and shows pagination with 5+ pages.

rlskoeser: 😆 Thanks for working on consolidating this. It's ok to have a meaningless search for testing purposes! But can we also include a search term that gets highlighted in the description to show keywords-in-context styles? (I guess we didn't have that in the previous example...?)  Sorry this is kind of a pain. Maybe we can just add another search term.

blms: That was my intention, and oddly enough it _is_ highlighted on my local! I'll try adding a second keyword.

blms: Strange, I added a second keyword and it's still not highlighting the result in Percy. But it is working as expected for me locally!

Issue #357, opened 2021-11-10
Fix document detail view spacing for entries with editors


blms:
Current spacing does not match design:

<img width="751" alt="Screen Shot 2021-11-10 at 9 33 43 AM" src="https://user-images.githubusercontent.com/7234006/141132320-9c45f025-8cd4-49a4-921f-b835cdb4ed47.png">

blms:
Reviewed in QA deployment 👍

PR #359, opened 2021-11-10
#357 - Fix spacing for document detail view

blms:
### visual review notes
[Current Percy build](https://percy.io/3201ecb4/geniza/builds/13935264)
- Only the document detail view is in scope for this PR—specifically, the change to the spacing of Editors

### What this PR does
- Per #357 
  - Fixes issue with document detail view spacing when the document has editors
  - Also fixes other minor spacing and alignment issues in that view
  - ~~Adds a test fixture for a document with an editor~~

### DIFF
--- a/geniza/common/management/commands/create_test_site.py
+++ b/geniza/common/management/commands/create_test_site.py
@@ -4,6 +4,7 @@ from django.core.management.base import BaseCommand
 
 from geniza.corpus.tests.conftest import (
     make_document,
+    make_document_with_editor,
     make_fragment,
     make_join,
     make_multifragment,
@@ -33,6 +34,7 @@ class Command(BaseCommand):
         multifrag.iiif_url = ""
         multifrag.save()
         make_document(multifrag)
+        make_document_with_editor()
         make_join(frag, multifrag)
 
         # index everything in solr
--- a/geniza/common/management/commands/create_test_site.py
+++ b/geniza/common/management/commands/create_test_site.py
@@ -2,13 +2,6 @@ from django.contrib.sites.models import Site
 from django.core.management import call_command
 from django.core.management.base import BaseCommand
 
-from geniza.corpus.tests.conftest import (
-    make_document,
-    make_fragment,
-    make_join,
-    make_multifragment,
-)
-
 
 class Command(BaseCommand):
     def handle(self, *args, **options):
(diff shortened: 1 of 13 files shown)


blms:
@rlskoeser Looks like the document in the fixture does not contain any editors. Should I add editors to the existing fixture, or create a new fixture for a document with editors?

rlskoeser:
> @rlskoeser Looks like the document in the fixture used by Percy does not contain any editors. Should I add editors to the existing fixture, or create a new fixture for a document with editors?
> 
> Edit: I went ahead and created a new fixture, let me know if that doesn't make sense and I can revert!

Sorry for the delay in responding — I think a larger fixture generated from real data would be better and answer more of our problems, so let's revert this one.

blms:
@gissoo I think this should be good to go. Wanted to get your feedback on spacing with more than one editor—I believe the document detail view in Percy right now has two editors listed, and they're spaced pretty close together.

gissoo:
> @gissoo I think this should be good to go. Wanted to get your feedback on spacing with more than one editor—I believe the document detail view in Percy right now for this PR has two editors listed, and they're spaced pretty close together.
>

@blms Thanks for sharing!! Can we have 4px spacing between each editor? I think that would solve it

blms:
Thanks @gissoo! Feel free to take a look at this Percy build now and let me know if all looks good, and if so I'll merge 👍

gissoo:
@blms Thanks! I left the following note on the percy page as well but I'll put here too: 

Ben, I'm a bit confused, the original version here looks like what we want, not the "new" – but also in the "original" the spacing between the second editor and the Input Date is gone, we should preserve that

blms:
@gissoo, ah, I can explain that!

Basically what happened in the "original" was that the full vertical spacing intended to separate different metadata sections was being applied between 2 different editors. So you'll notice the spacing between those 2 editors is the same as the spacing between "2017" and "Permalink". In the "new", I've fixed it so that there is now roughly 4px spacing between 2 editors, and the proper full spacing between the last editor listed and "Input Date".

But we can definitely increase the spacing to be greater than 4px for the spacing between editors if you think that makes sense!

gissoo:
@blms wow thanks!! That's 4px??!! ...it looked way tighter than in the designs but it's an illusion on my side ...the format of the content doesn't make them easy to read , can we increase it to 8px only between the two editors?

blms:
@gissoo Increased the spacing to about 7.11 px (`spacing-xs`), here's the [latest Percy build](https://percy.io/3201ecb4/geniza/builds/13999094). The next size up (`spacing-sm`) is about 10.5px if you want to bump it up further!

gissoo:
@blms this is so much better thank you!!!

Issue #296, opened 2021-10-12
As a user I would like to filter my search by document type so that I can view specific types of documents


gissoo:
## testing notes
test on the document search https://test-geniza.cdh.princeton.edu/en/documents/
- [x] should be able to filter by one or more document types
- [x] when no document type is selected, should show results of all types

## dev notes
- [x] add document type facet field to search form (adapt from S&co or PPA)
- [x] add faceting on document type field in solr query
- [x] display facet field with totals
- [x] when type is set in request, filter queryset 
- [x] display selected document type in the form

richmanrachel:
Omg, @mrustow you are going to jump for joy when you see this! I LOVE the addition of numbers for how many results fall into which categories (and I clearly need to change back Funsies -> Lists, haha). This will truly improve everyone's research experience. Thank you!!!

![image](https://user-images.githubusercontent.com/68028846/142683605-14e90594-7ad0-49c3-aea8-44f6476752ee.png)

PR #362, opened 2021-11-11
Preliminary document type search filter

rlskoeser:
## visual review notes
[Current Percy build](https://percy.io/3201ecb4/geniza/builds/13964560)
- Only the document type filter is in scope for this PR (the super long vertical screenshot)
- One option, "Legal document", is selected (but not applied—no filter applied in results)

## dev notes

initial functionality towards implementing #296 
- adapts `FacetChoiceField` from other projects (have not yet pulled in tests)
- new radio select subfield to add counts
- preliminary display of document type filter on the form
- enable solr faceting on document type field
- populate facet choice field with returned facet values and counts
- filter queryset by document type if specified in the request

I've tried a couple of different ways to render the choices — initially thought a `select` based on the designs, but I don't think there's a good way to display the counts. Currently I have the counts both in a span of the label and on a data-count attribute on the option; I thought perhaps the data attribute could be used by css for display, but it seems that `option` elements don't have an `::after`. I switched to a details/summary with a radio select widget, but I don't know if that's the best solution.

No unit tests, only very preliminary/placeholder styles.


blms:
For UI, I wonder if something like [this](https://hybrismart.com/2019/02/13/facet-search-the-most-comprehensible-guide-best-practices-design-patterns/#h.t9a1mzolpkdu) might make sense?

geniza/corpus/views.py
```diff
@@ -76,7 +76,14 @@ def get_queryset(self):
...
 
-            documents = documents.order_by(self.solr_sort[search_opts["sort"]])
+            documents = documents.order_by(
+                self.solr_sort[search_opts["sort"]]
+            ).facet_field("type", exclude="type", sort="value")
+            # exclude type filter when generating counts
+
+            # filter by type if specified
+            if search_opts["doctype"]:
+                documents = documents.filter(type=search_opts["doctype"], tag="type")
```
blms, line 86: This only seems to work when I filter by doctype "Letter". Were you experiencing that too or have I misconfigured something? The only reason I can think of for this is that it doesn't contain multiple words…

rlskoeser: oh, good catch — it needs to be quoted since some of these are multiple words

blms: Something like this?
```suggestion
                documents = documents.filter(type=f'"{search_opts["doctype"]}"', tag="type")
```

rlskoeser:
Regarding the UI: your screenshot is fairly similar to what we did on the Shakespeare and Company Project — see the members page, toggle open the demographics filters: https://shakespeareandco.princeton.edu/members/

Gissoo and I discussed and we think the document type should be an exclusive filter, not a combination filter; that's why I went with radio buttons instead of checkboxes. 

Have you looked at the designs on Figma on the Filters + Search Results page?

blms:
> Gissoo and I discussed and we think the document type should be an exclusive filter, not a combination filter; that's why I went with radio buttons instead of checkboxes.
> 
> Have you looked at the designs on Figma on the Filters + Search Results page?

I see! Just took a look at those designs. I think radio buttons do make sense, then. It also does look in the design like it was maybe meant as a collapsible/expandable set of options rather than a select dropdown, at least based on the Sort By dropdown design next to it.

> Regarding the UI: your screenshot is fairly similar to what we did on the Shakespeare and Company Project — see the members page, toggle open the demographics filters: https://shakespeareandco.princeton.edu/members/

This looks great, will probably come in handy as a reference even if not for this exact element.

rlskoeser:
> I see! Just took a look at those designs. I think radio buttons do make sense, then. It also does look in the design like it was maybe meant as a collapsible/expandable set of options rather than a select dropdown, at least based on the Sort By dropdown design next to it.

Oh wow, you're right! That's what Gissoo said too when I talked to her about what kind of element it should be — I see now that I was not looking at the designs closely enough, the contrast between the sort and the other filters is clear.

blms:
@rlskoeser I think this is pretty much ready to go, with a couple of questions:

1. Does the way I've added "All" as an option make sense? I basically wanted a way to unapply the filter without restarting the entire search.
2. Since Percy can't display interactions, here are some supplemental screenshots showing different states of the filter. Should I request Gissoo's review on this? Maybe design for this should be a separate issue.

   a. Default, no filter selected ("All")

   Screen Shot 2021-11-15 at 4 32 10 PM

   b. Filter selected, focused, and active (I believe radio button needs to be focusable for accessibility)

   Screen Shot 2021-11-15 at 4 32 30 PM

   c. After filter applied

    Screen Shot 2021-11-15 at 4 32 00 PM

blms:
Just noticed things are breaking on Safari desktop, and all non-Chrome mobile! Will take a look at what's causing those issues.

blms:
Quick update, I can't seem to replicate the Safari desktop issue at all. But I am able to replicate the mobile issues, so I'll take a crack at them tomorrow.

Also noticed that on Safari, I wasn't able to navigate inside of the expanded `<details>` element with the keyboard. Some blog posts I found on the subject:
https://www.hassellinclusion.com/blog/accessible-accordions-part-2-using-details-summary/
https://www.hassellinclusion.com/blog/accessible-accordion-pattern/

Maybe it would make sense to go with their preferred pattern.

rlskoeser:
> Also noticed that on Safari, I wasn't able to navigate inside of the expanded `<details>` element with the keyboard. Some blog posts I found on the subject (though they don't mention this issue specifically, so maybe there's something else wrong with my implementation)

Interesting. I experimented with this a little bit too, and found this guide for configuring tabbing properly in OSX: https://www.scottohara.me/blog/2014/10/03/link-tabbing-firefox-osx.html 

Once I changed that setting I was able to navigate within the choices in the details.

I don't love the accessible pattern you found — their solution for no javascript is just to have the widget always open. If we want to make this fully accessible, we should probably use the deque pattern library https://dequeuniversity.com/library/aria/expand-collapse

rlskoeser:
Thanks for the screenshots — and clever use of summary/details in your github comment!

The input needs should be focusable for everyone, not just screen readers; removing that mixin locally and hiding the radio input in a way that doesn't prevent it from being focused seemed like it worked (I used width & opacity, but IDK what's best) that was more what I expected visually. I did not test with Voice Over, but I can if that would be helpful.

I get the logic for having an "all" choice since there's no easy way to deselect a type once you've chosen one. I wonder if using check boxes and allowing multiple types would be nicer? It seems weird to have an "all" choice without a count. I would like to hear what @gissoo thinks.

When there is a filter is applied, it would be nice to have some indication when the disclosure widget is collapsed — either the specific document type (on screens where there is room) or some other visual. I don't think that's an MVP feature, though.

blms:
> Interesting. I experimented with this a little bit too, and found this guide for configuring tabbing properly in OSX: https://www.scottohara.me/blog/2014/10/03/link-tabbing-firefox-osx.html
> 
> Once I changed that setting I was able to navigate within the choices in the details.

Ah, it worked, thanks!

> I don't love the accessible pattern you found — their solution for no javascript is just to have the widget always open. If we want to make this fully accessible, we should probably use the deque pattern library https://dequeuniversity.com/library/aria/expand-collapse

I agree, this is much better. Should I include the CSS/JS in `sitemedia`?

> The input needs should be focusable for everyone, not just screen readers; removing that mixin locally and hiding the radio input in a way that doesn't prevent it from being focused seemed like it worked (I used width & opacity, but IDK what's best) that was more what I expected visually. I did not test with Voice Over, but I can if that would be helpful.

I think I may have written an inaccurate comment in the CSS, since I'm able to focus those radio inputs without a screen reader. Though, I did "unhide" them when focused, as I thought that was more accessible—maybe it is fine to hide them? Not sure what's the best practice there.

> I get the logic for having an "all" choice since there's no easy way to deselect a type once you've chosen one. I wonder if using check boxes and allowing multiple types would be nicer? It seems weird to have an "all" choice without a count. I would like to hear what @gissoo thinks.

Checkboxes may be nicer, though it may still be nice to have an "all" or "none" control (or both) for that.

As far as the count, I think I should be able to just add the count in the form logic. Either by including that explicitly in the `facet_dict` or by adding up all the counts from that dict.

> When there is a filter is applied, it would be nice to have some indication when the disclosure widget is collapsed — either the specific document type (on screens where there is room) or some other visual. I don't think that's an MVP feature, though.

I agree, that would be nice, but should probably have a @gissoo design and maybe is outside MVP scope.

blms:
Also, it's bizarre, I can't replicate or resolve that issue with Safari that's showing up on Percy. Maybe it's an issue fixed in more recent Safari versions, but I can't seem to install an old version of Safari to test with.

rlskoeser:
@blms yes, let's go ahead and try out the deque pattern library and apply it to the details/summary. 

I don't think the input needs to be visible for accessibility as long as we have a focus indicator — I was thinking of the selection styles as standing in for that, but that's not quite right, is it. Can we apply a focus indicator to the label?

I was thinking if we use check boxes we wouldn't need an "all" option because you can deselect more easily. But I'm not sure allowing to select multiple types is worth that trade off, because I'm not sure it makes sense. Let's wait for Gissoo to weigh in on this before we speculate further. 😁 

Regarding the Safari issue in Percy : they only recently added support for Safari, so maybe it is still a little experimental. I don't see anywhere in the project settings anything about what version they're testing with. If you can't duplicate it then I'm ok with ignoring it.

@gissoo do we have focus styles yet, or is it on your list somewhere?

blms:
@rlskoeser Implemented the Deque pattern. As a positive side effect, it's resolved the Safari issue. As a negative side effect, it fails a Lighthouse test for unused JS. Should we disable that rule, dive into the JS and figure out what can go, or something else?

rlskoeser:
> @rlskoeser Implemented the Deque pattern. As a positive side effect, it's resolved the Safari issue. As a negative side effect, it fails a Lighthouse test for unused JS. Should we disable that rule, dive into the JS and figure out what can go, or something else?

I think I want to backtrack and drop the deque pattern library — I looked more closely at the link you shared, and it seems that at least some of the issues identified there have been resolved (certainly the NVDA one). And the deque solution (at least as you've implemented it) isn't no-script friendly. Can you revert the deque changes?

What's the safari issue specifically? I can take a look on my machine and see if I can duplicate

blms:
The safari issue is this strange line breaking behavior with the expand/collapse indicator:

Screen Shot 2021-11-16 at 1 47 23 PM
Screen Shot 2021-11-16 at 1 47 27 PM

rlskoeser:
> The safari issue is this strange line breaking behavior with the expand/collapse indicator:

Ok, I can duplicate that locally — I've got Safari 14.1.2 and it renders the same way. I can investigate.

rlskoeser:
@blms look at my changes, see what you think.  I also took the liberty of coloring the select widget for the sort to make it match the color scheme.

blms:
Looks great, thank you for the quick fix and additional work. I love the CSS transition! 😄

blms:
Is all that's left for this to make a decision on radio buttons vs checkboxes, and whether or not to keep "All" for radio buttons (and whether it should include count)?

gissoo:
@blms @rlskoeser why do we need the "all" option? Ben, I don't think I understand what you wrote above: "I basically wanted a way to unapply the filter without restarting the entire search." – aren't they all selected by default? What would the results be like if none of them are selected?

blms:
@gissoo If the options were checkboxes, that would be correct—all would be checked by default. 

However, since it is now implemented as radio buttons instead, only one can be chosen at a time. Therefore, if one selects an item from the list, they have no way to return to the state where all document types are included in the search. The "All" button would allow for that.

rlskoeser:
@blms ok, Gissoo and I just discussed this. Let's do the following:
- remove the "all" option
- switch from radio buttons to check boxes
- change view logic to do a filter in list (parasolr has a queryset `__in` filter we should be able to use)
- display phosphor icon check-circle-fill to the left of choices when they are selected; set to use secondary theme color (prefer using css `::before` if we can; see how I did the triangle for the summary/details; we could make a phosphor mixin to set the font if it's useful)

gissoo:
@rlskoeser @blms  thanks for writing!! 

Ben, following what RSK has written above, as I discussed with her I see your point and it's functional but I don't think having an "All" option is really what we need here because by default they are all selected so it could be confusing if the user sees that option – and checkboxes seem to be more useful here.

blms:
Thanks both, I agree with all these decisions!

rlskoeser (commented):
@blms Code changes look good, but when I test it locally the styles for interaction don't seem quite right.

- it looks like bold is being used as the focus style, document types should be bold when they are selected (both bold text + phosphor icon)
- we don't have a focus style yet — Gissoo and I discussed this briefly yesterday and I forgot to relay; would you add a generic one for now, and make sure it works on our checkbox labels? something like a solid or dashed border using primary or secondary color.
- minor, but the phosphor icon looks too small — I had trouble recognizing it at first

blms:
@rlskoeser Made those changes, thanks! Let me know if the phosphor icon is still too small. Right now it's the same font size as the text.

PR #364, opened 2021-11-15
Manage command to sync TEI transcriptions to html content in footnotes (WIP)

rlskoeser:
Preliminary work on a new manage command to sync TEI transcription content to content associated with footnotes that we can use for the MVP site. There are still a number of questions that need to be resolved, but I thought it would be helpful to get eyes on what I've drafted so far.

I haven't written any unit tests yet, and I'm not sure how much I want to write — this is intended to be an interim solution and I'm not sure yet if we'll use any of this code for the "real" solution; possibly none.

### DIFF
--- a/geniza/corpus/management/commands/sync_transcriptions.py
+++ b/geniza/corpus/management/commands/sync_transcriptions.py
@@ -1,7 +1,10 @@
+import glob
 import os.path
+from collections import defaultdict
 
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
+from django.db import models
 from git import Repo
 
 from geniza.corpus.models import Document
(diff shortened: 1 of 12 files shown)


geniza/corpus/management/commands/sync_transcriptions.py
```diff
@@ -0,0 +1,151 @@
...
+                # print('more than one edition for %s' % xmlfile)
+                # print(list(ed.source for ed in editions))
+                # try filtering by current text content
+                editions_with_content = editions.filter(content__isnull=False)
+                # print('editions with content')
+                # print(list(ed.source for ed in editions_with_content))
+                stats["multiple_editions"] += 1
+                if editions_with_content.count() > 1:
+                    stats["multiple_editions_with_content"] += 1
+                    multiedition_docs.add(doc.id)
```
blms, line 85: I noticed that in this case, it seems `footnote` is not set. Is that because the intended behavior for multi-edition docs isn't known yet? I guess if the edition has to be matched to a document, and there's more then one edition, it's unclear which edition should get the HTML formatted TEI content from the repo.

geniza/corpus/management/commands/sync_transcriptions.py
```diff
@@ -0,0 +1,151 @@
...
+            else:
+                stats["one_edition"] += 1
+                # if only one edition, update the transciption content there
+                footnote = editions.first()
+
+            # if we identified the correct footnote, update it
+            if footnote:
+                html = tei.text_to_html()
+                if html:
+                    footnote.content = {"html": html}
```
blms, line 102: In my local, it looks like the footnotes' content is always a dict with the `lines` key, which is an array of lines. Is the idea that this should be overwritten with the new HTML, or should that other representation be preserved?

geniza/corpus/management/commands/sync_transcriptions.py
```diff
@@ -0,0 +1,151 @@
...
+
+        # if directory does not yet exist, clone repository
+        if not os.path.isdir(local_path):
+            self.stdout.write(
+                "Cloning TEI transcriptions repository to %s" % local_path
+            )
+            Repo.clone_from(url=gitrepo_url, to_path=local_path)
+        else:
+            # pull any changes since the last run
+            Repo(local_path).remotes.origin.pull()
```
blms, line 144: This is neat, I didn't know about this library!

geniza/corpus/tei_transcriptions.py
```diff
@@ -0,0 +1,85 @@
...
+            text_lines = " <ul>%s</ul>" % "".join(
+                f"\n <li value='{line_number}'>{line}</li>"
+                for line_number, line in block["lines"]
+                if line.strip()
+            )
+            output.append(text_lines)
+            output.append("</section>")
+            html.append("\n".join(output))
+
+        return "\n".join(html)
```
blms, line 85: All the HTML conversion looks good. On another project I'm working on, we prettify the HTML before storing it anywhere. Not sure if that's needed/desired here but might be worth consideration.

blms (approved):
All looks good to me so far!

geniza/corpus/management/commands/sync_transcriptions.py
rlskoeser: correct.  I have some open questions for Rachel and Marina about how we want to determine the correct footnote (among other questions)

geniza/corpus/management/commands/sync_transcriptions.py
rlskoeser: The lines/blocks json data was the transcription we intitially populated when we imported the spreadsheet data. My plan is to replace that one with the html version — the html is still an interim solution, but it adds line numbers which are a requirement and were not included in the initial lines/blocks implementation.

FWIW, I started working on indexing and highlighting transcription text and I think I'm going to populate a plain text version of the transcription content at the same time as the html version — seems easier to do it here and combine the logic than doing it at index time.  I'll add that in a subsequent PR, though.

geniza/corpus/tei_transcriptions.py
rlskoeser: That's smart. I'm not going to worry about that right now since I don't know how much of this we'll use when we implement a proper transcription conversion, but will keep it in mind!

rlskoeser:
I have tests for tei xml logic but not for the sync script itself; given that this is an interim solution, and in the interest of getting an initial version out for testing, I'm going to go ahead and merge. If tests are useful on the script when we revisit to handle the other cases (multiple footnotes, no footnotes, etc) we can add tests as needed at that point.

PR #365, opened 2021-11-15
Revise & correct scholarship counts for search

rlskoeser:
updates the scholarship display on the public search
- fix a bug where 1-counts for scholarship records were not displayed
- revise layout to match the latest revisions in the design (#338) and update tests accordingly
- use `blocktranslate` to handle pluralization properly


blms (approved):
Great, this template is much improved now.

Issue #231, opened 2021-06-03
As an admin, I want an easy way to get from the public document view to the edit view on the admin site, so I can make edits and correct errors..


rlskoeser:
## testing notes
- before testing, login to the admin page: https://test-geniza.cdh.princeton.edu/admin
- then visit a document detail page (e.g. https://test-geniza.cdh.princeton.edu/en/documents/3504)
- [x] A link to "edit" with a pencil icon should appear in the upper right
- [x] This link should take you to the admin edit view for this document
- then test again as an anonymous user (log out or test in a different browser or incognito window)
- [x] confirm anonymous user does not see an edit link

## dev notes
- [x] update document detail template to add a link to the admin change view for that document
- [x] use template permissions check to only display the link if user has change permissions on documents
- [x] add template test to confirm the conditional display

We don't have any designs for this, but it's not a public feature anyway. I suggest putting it somewhere in the top right corner of the document detail page; label with text and an appropriate phosphor icon; make sure it looks reasonable on both desktop and mobile.

richmanrachel:
Exciting update! Works well. Can we perhaps add something to @gissoo's design to show if you're logged in, or not?

PR #366, opened 2021-11-16
#231 - Add "edit" link to document detail view for admins

blms:
### What this PR does

- Per #231:
  - Adds an "Edit" link (with Phosphor pencil icon) to the document detail view template for those with appropriate permissions
  - Adds template test for the above

### DIFF
--- a/geniza/corpus/tests/test_corpus_templates.py
+++ b/geniza/corpus/tests/test_corpus_templates.py
@@ -1,3 +1,4 @@
+from django.contrib.auth.models import User
 from django.template.loader import get_template, render_to_string
 from django.urls import reverse
 from pytest_django.asserts import assertContains, assertNotContains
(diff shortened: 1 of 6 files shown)


geniza/corpus/tests/test_corpus_templates.py
```diff
@@ -58,6 +59,18 @@ def test_multi_viewer(self, client, join):
         second_url = join.textblock_set.last().fragment.iiif_url
         assertContains(response, f'data-iiif-urls="{first_url} {second_url}"')
 
+    def test_edit_link(self, client, document):
```
rlskoeser, line 62: pytest-django has an `admin_client` fixture https://pytest-django.readthedocs.io/en/latest/helpers.html#admin-client-django-test-client-logged-in-as-admin 

let's use that here — although it doesn't test the exact permission, that seems fine

geniza/corpus/tests/test_corpus_templates.py
```diff
@@ -58,6 +59,18 @@ def test_multi_viewer(self, client, join):
...
+        """Edit link should appear if user is admin, otherwise it should not"""
+        edit_url = reverse("admin:corpus_document_change", args=[document.id])
+        response = client.get(document.get_absolute_url())
+        assertNotContains(response, f'<a class="edit-link" href="{edit_url}">')
```
rlskoeser, line 66: let's just check for the presence of the url, we don't really care what the markup is for this test

rlskoeser (approved):
Looks good to me! Just two small suggestions for the unit test.

PR #372, opened 2021-11-18
#312 - Mobile site navigation

blms:
### Visual review notes
[Percy build](https://percy.io/3201ecb4/geniza/builds/14271917/)
In scope for this PR are only the following:
- Navigation menu
- "About" submenu

The "About" submenu is only visible in the snapshot titled "About submenu", and the mobile navigation menu is only visible on snapshot titled "Mobile menu". The desktop navigation menu is visible on all screenshots.

### What this PR does

- Per #312:
  - Adds mobile site navigation
- Updates desktop site navigation to match designs

<details>
<summary>Old description</summary>

Just wanted to get a draft up of my approach so far, and check in with @rlskoeser that it makes sense.

Basically I wanted a compromise between accessibility, no JS, and reuse of the existing nav. I ended up going with a pattern from [here](https://css-tricks.com/three-css-alternatives-to-javascript-navigation/) using a checkbox input and CSS to expand and collapse the menu. Though that doesn't focus on accessibility, I think the keyboard focusing and aria-labels I've implemented should help with that—it seems to work well with VoiceOver and keyboard navigation.

</details>


rlskoeser:
I remember that Nick didn't like the checkbox toggle for displaying a menu like this — IIRC he thought it was stretching the functionality beyond what it was meant for. It doesn't bother me as much.

Short answer: I think this approach is ok.

Both keyboard nav and voiceover do seem to work; I worry a little bit that it's an unusual pattern, which would make it harder to use. It would be nice if the enter key worked as well as space for toggling the menu. (It occurs to me, I think we don't yet have "skip to main" keyboard shortcut; do people still do a "skip to nav", and would that be useful ? Maybe not, since there's not much to skip to get to the nav)

I definitely appreciate having the same nav used for different screens. Could you add placeholder entries for pages that will live under the About so we can see how it will work?

FWIW, for the main nav on shakespeare and company project we used an anchor link and then used css `:target` to adjust the styles (works without js but pollutes the browser history), then some typescript to make it work a little nicer when js is available. I don't know if this is relevant for our menu here, and we don't need to use it; just sharing in case it's useful.

html: https://github.com/Princeton-CDH/mep-django/blob/main/templates/snippets/nav.html
css: https://github.com/Princeton-CDH/mep-django/blob/main/srcmedia/scss/common/_nav.scss#L86-L90
typescript: https://github.com/Princeton-CDH/mep-django/blob/main/srcmedia/ts/components/MainMenu.ts

blms:
Thanks @rlskoeser! Actually, I like that approach from S&co more. I think the checkbox is not only an unusual pattern, but it also requires some strange stuff with positioning elements. I'll give that pattern a try.

Also looked up "skip to nav"—what I'm seeing a lot of is the opposite, a way to skip _past_ the nav so you don't have to encounter it on every page when using a screen reader. I like that a lot and will give it a try.

blms:
Hi @rlskoeser, I've updated this to use roughly the pattern from S&Co, and added the About submenu.

The JS is optional and just improves accessibility when JS is available.

A couple questions:
1. Does it make sense the way I've implemented it? It's not quite as easily keyboard navigable as the previous pattern, not sure how to mitigate.
2. Should I add Percy snapshots for the mobile menus, and the about menu open on desktop?

rlskoeser:
> The JS is optional and just improves accessibility when JS is available.

I love how minimal the JS is, and that is isn't required!

>     1. Does it make sense the way I've implemented it? It's not quite as easily keyboard navigable as the previous pattern, not sure how to mitigate.

This seems reasonable to me! Let's use it for now; we can always revise/expand the JS if we want to enhance it more later on.

>     2. Should I add Percy snapshots for the mobile menus, and the about menu open on desktop?

Please go ahead and add a Percy snapshot to capture the menus; I'm not sure if we need both of them or not. How does it  apply across mobile and desktop, is there a way to get one screenshot that's meaningful for both? Maybe the about menu expanded on desktop? @gissoo do you want percy visual review on both the main menu and the about nav on every page?

rlskoeser (approved):
Everything here looks reasonable to me.

The transition from main nav to about nav looks a bit weird right now, would that be easy to fix?

geniza/templates/nav.html
```diff
@@ -1,17 +1,100 @@
...
+                    </li>
+                    <li class="menu-item" role="none">
+                        {# Translators: label for link to 'Technical' page in main navigation #}
+                        <a href="#" role="menuitem">{% translate "Technical" %}</a>
+                    </li>
+                    <li class="menu-item" role="none">
+                        {# Translators: label for link to 'FAQ' page in main navigation #}
+                        <a href="#" role="menuitem">{% translate "FAQ" %}</a>
+                    </li>
+                </ul>
```
rlskoeser, line 96: I don't know if these should use the translate tag or if we want to use wagtail's i18n — probably we'll need the wagtail functionality (was just reading up, it looks pretty powerful https://docs.wagtail.io/en/stable/advanced_topics/i18n.html#wagtail-s-approach-to-multi-lingual-content ) — but we may not have that for MVP so in the meantime this probably makes sense.

gissoo:
> > The JS is optional and just improves accessibility when JS is available.
> 
> I love how minimal the JS is, and that is isn't required!
> 
> > ```
> > 1. Does it make sense the way I've implemented it? It's not quite as easily keyboard navigable as the previous pattern, not sure how to mitigate.
> > ```
> 
> This seems reasonable to me! Let's use it for now; we can always revise/expand the JS if we want to enhance it more later on.
> 
> > ```
> > 2. Should I add Percy snapshots for the mobile menus, and the about menu open on desktop?
> > ```
> 
> Please go ahead and add a Percy snapshot to capture the menus; I'm not sure if we need both of them or not. How does it apply across mobile and desktop, is there a way to get one screenshot that's meaningful for both? Maybe the about menu expanded on desktop? @gissoo do you want percy visual review on both the main menu and the about nav on every page?
>

@rlskoeser @blms sure! that would be helpful.

blms:
@rlskoeser I've managed to get Percy to show the About menu on desktop, but I'm having trouble showing either menu on mobile since Percy [pauses all CSS animations](https://docs.percy.io/docs/animations). Any ideas for a workaround?

rlskoeser:
> @rlskoeser I've managed to get Percy to show the About menu on desktop, but I'm having trouble showing either menu on mobile since Percy [pauses all CSS animations](https://docs.percy.io/docs/animations). Any ideas for a workaround?

Wow, interesting! That makes sense. Let's try [percy specific sss](https://docs.percy.io/docs/percy-specific-css) to disable the css transition.

PR #376, opened 2021-11-19
Update highlighting to work with latest parasolr

rlskoeser:
update template highlighting for latest version of parasolr `AliasedQuerySet`, which now handles aliased fields in highlighting results


blms (approved):
Nice! Those field aliases are super useful, great to support them in highlighting too.

PR #381, opened 2021-11-29
search & keywords in context on transcriptions

rlskoeser:
included in this PR:
- revises `sync_transcription` script to generate and store a plain-text version of transcription content (no line numbers or block labels)
- add ISO code field to LanguageScript model
- data migration to populate ISO code for Language+Script instances present in transcriptions
- index plain text transcription and language codes in solr
- configure keyword search to include transcription
- configure and display keyword highlighting in transcription text (requires latest develop version of parasolr)


geniza/corpus/migrations/0020_populate_documenttype_display_label.py
```diff
@@ -36,5 +36,5 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunPython(populate_document_types),
+        migrations.RunPython(populate_document_types, migrations.RunPython.noop),
```
blms, line 39: Oh! I didn't know about this. This is super helpful for another PR I'm working on.

geniza/corpus/tests/test_corpus_views.py
```diff
@@ -319,28 +319,35 @@ def test_get_queryset(self, mock_solr_queryset):
...
-            docsearch_view.object_list = docsearch_view.queryset
 
             context_data = docsearch_view.get_context_data()
-            assert context_data["total"] == 22
             assert (
                 context_data["highlighting"]
                 == docsearch_view.queryset.get_highlighting.return_value
             )
             assert context_data["page_obj"].start_index() == 0
+            # NOTE: test paginator isn't initialized properly from queryset count
```
blms, line 349: Is this a bug with the `get_context_data`/paginator logic, or is it just happening because the paginator's not picking up the mock solr queryset? I see in my code I had used a separate `total` context var (didn't know about `paginator.count`), which just uses the `queryset.count`, so it makes sense that would have worked.

geniza/common/management/commands/visual_tests.py
```diff
@@ -42,7 +42,7 @@ def take_snapshots(self, browser):
         percy_snapshot(browser, "Document Search filter")
 
         # document search
-        browser.get("http://localhost:8000/documents/?q=the+writer&per_page=2")
+        browser.get("http://localhost:8000/documents/?q=the+writer+באנפנא&per_page=2")
```
blms, line 45: I wonder if we should have a query here that also shows the Hebrew being highlighted? Not strictly necessary though.

blms (approved):
Looks great! Exciting to have so much more powerful of a search.

geniza/corpus/tests/test_corpus_views.py
rlskoeser: I think it's because we're not running the normal progression of view methods that would happen if the view was dispatched normally, so the paginator isn't initialized from our mock. I hoped that setting the return from `get_queryset` would help, but I think there's something else not being called — I looked at the generic view mixin documentation for a bit and it wasn't obvious, so I decided not to worry about it for this test.

I was just simplifying the context data to remove the redundant total, since we can get it from the paginator.

geniza/common/management/commands/visual_tests.py
rlskoeser: Yeah, I was trying to get that but with the per-page results it didn't show up. Should I tweak the query and try to get it in the first two? I think we'll want to see it for testing fonts.

geniza/corpus/tests/test_corpus_views.py
blms: Makes sense to me!

geniza/common/management/commands/visual_tests.py
blms: Yes, I wonder if just using "באנפנא" as the first keyword would work? Might also be helpful to include an English word that's going to exist in the description of a document containing "באנפנא", then you could omit "writer" and we can see both English and Hebrew highlighted.

blms (approved):
New query is perfect!

rlskoeser:
> New query is perfect!

Thanks for suggesting a search term that would match the description for the same document!

Issue #390, opened 2021-12-03
As a front-end user, I want to see brief citations in the Document Detail view, more concise than those in Scholarship Records.


blms:
## testing notes
- [x] Check various document detail pages, e.g. https://test-geniza.cdh.princeton.edu/en/documents/5504/, and verify that the brief citations under "Editor"/"Editors" look correct, and are not missing any citation fields or including unwanted ones
- [x] Verify that the "Editors" section citations only refer to digital editions, and that the section does not appear when there is no digital edition available
   - note: digital editions are only visible from the search results at the moment, as transcripts don't appear in the document detail view yet
- [x] Verify that "Editors" is correctly pluralized based on the number of people listed

## dev notes

- [x] Add "unpublished" to brief citations.
- [x] Only show brief citations for available digital editions.
- [x] Pluralize editor heading if there is more than one digital edition

<details>
<summary>docs with more than one digital edition</summary>

- Bodl. MS heb. f 29/69 + … (PGPID 2493)
- Bodl. MS heb. f 39/29 (PGPID 606)
- CUL Or.1080 J24 (PGPID 9121)
- CUL Or.1080 J38 (PGPID 4743)
- ENA 2727.7b (PGPID 4740)
- T-S 12.190 (PGPID 9089)
- CUL Or.1080 J23 (PGPID 5552)
- ENA NS 13.1 + … (PGPID 9804)
- T-S 8J5.1 (PGPID 2142)
- NLI 577.3/2 (PGPID 848)
- T-S 8J10.16 (PGPID 1849)
- ENA NS 21.14 (PGPID 2691)
- T-S K15.85 (PGPID 4585)
- T-S 8Ja1.3 (PGPID 4717)
- T-S 10J12.5 + … (PGPID 4721)
- T-S Misc.25.103 (PGPID 4738)
- ENA NS 21.7 (PGPID 5410)
- Bodl. MS heb. d 74/41 (PGPID 591)
- DK 333 (alt: 18) (PGPID 4495)

</details>

richmanrachel:
Looks perfect! Closing!

PR #395, opened 2021-12-06
#390 - Brief citations for digital editions

blms:
## What this PR does

Per #390:
- Only shows brief citations relating to digital editions on Document Detail page
- Pluralizes "Editor" label when there is more than one editor of digital edition(s)
- Shows "unpublished" in brief citations when applicable
- Updates the display of brief citations with "notes" to always end in a period
- (Finally!) disables phosphor "i" icon next to permalink

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -525,6 +525,23 @@ class Document(ModelIndexable):
             "content", "source"
         )
 
+    def digital_editions(self):
+        """All footnotes for this document where the document relation includes
+        edition AND the footnote has content."""
+        editions_with_content = (
+            self.footnotes.filter(doc_relation__contains=Footnote.EDITION)
+            .filter(content__isnull=False)
+            .order_by("content", "source")
+        )
+        display_strings = []
+        editions_with_unique_display = []
+        for f in editions_with_content:
+            if f.display() not in display_strings:
+                # Prevent redundant display of identical brief citations on detail page
+                editions_with_unique_display += [f.pk]
+                display_strings += [f.display()]
+        return editions_with_content.filter(pk__in=editions_with_unique_display)
+
     @classmethod
     def items_to_index(cls):
         """Custom logic for finding items to be indexed when indexing in
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -536,10 +536,7 @@ class Document(ModelIndexable):
 
     def editors(self):
         """All unique authors of digital editions for this document."""
-        editor_pks = []
-        for de in self.digital_editions():
-            editor_pks += [author.pk for author in de.source.authors.all()]
-        return Creator.objects.filter(pk__in=editor_pks)
+        return Creator.objects.filter(source__footnote__doc_relation__contains=Footnote.EDITION, content_isnull=False, source__footnote__document=self).distinct()
 
     @classmethod
     def items_to_index(cls):
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -528,19 +528,11 @@ class Document(ModelIndexable):
     def digital_editions(self):
         """All footnotes for this document where the document relation includes
         edition AND the footnote has content."""
-        editions_with_content = (
+        return (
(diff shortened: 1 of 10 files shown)


geniza/corpus/models.py
```diff
@@ -525,6 +525,23 @@ def editions(self):
...
+            .order_by("content", "source")
+        )
+        display_strings = []
+        editions_with_unique_display = []
+        for f in editions_with_content:
+            if f.display() not in display_strings:
+                # Prevent redundant display of identical brief citations on detail page
+                editions_with_unique_display += [f.pk]
+                display_strings += [f.display()]
+        return editions_with_content.filter(pk__in=editions_with_unique_display)
```
blms, line 543: @rlskoeser before I request your full review on this, I had a quick question. Does it make sense for me to leave this filtering of "duplicate" entries in the brief citations in this `Document` model function, or would it make more sense for this logic to be on the document detail view in `views.py` instead? It feels a bit strange as I have it here, since you might expect `digital_editions()` to return all digital editions, regardless of whether or not their brief citations would appear identical once rendered as strings.

blms: To see an example of why I did this, you can take a look at this document on the existing test site: https://test-geniza.cdh.princeton.edu/en/documents/2142/

rlskoeser: I agree, it feels strange here. I think the duplicates are likely a result of some footnote cleanup that needs to be done for the merged records, so it should hopefully be a temporary problem. I'd rather see the deduplication kept to the template — what about using `ifchanged` in the template to omit redundant citations?

blms: Thanks, didn't know about that template tag! Running into one problem with this, which is counting editors when there might be duplicates. For example, https://test-geniza.cdh.princeton.edu/en/documents/2493 will read "Editors" while displaying only one brief citation. Maybe just a quick view method like `editor_count` or something?

rlskoeser: right. editor(s) should be based on distinct number of edition authors anyway, right? so generating a distinct list that gives you access to count makes sense to me

geniza/corpus/models.py
```diff
@@ -525,6 +525,22 @@ def editions(self):
...
+            .filter(content__isnull=False)
+            .order_by("content", "source")
+        )
+
+    def editors(self):
+        """All unique authors of digital editions for this document."""
+        editor_pks = []
+        for de in self.digital_editions():
+            editor_pks += [author.pk for author in de.source.authors.all()]
+        return Creator.objects.filter(pk__in=editor_pks)
```
rlskoeser, line 542: Should be possible to get this into a single query, how about this:
```suggestion
        return Creator.objects.filter(source__footnote__doc_relation__contains=Footnote.EDITION, content_isnull=False, source__footnote__document=self).distinct()
```

geniza/corpus/models.py
```diff
@@ -525,6 +525,22 @@ def editions(self):
             "content", "source"
         )
 
+    def digital_editions(self):
+        """All footnotes for this document where the document relation includes
+        edition AND the footnote has content."""
+        return (
+            self.footnotes.filter(doc_relation__contains=Footnote.EDITION)
+            .filter(content__isnull=False)
+            .order_by("content", "source")
```
rlskoeser, line 534: I'm not sure how much sense this sorting makes, but not sure what other option to suggest.

geniza/corpus/templates/corpus/document_detail.html
```diff
@@ -30,11 +30,19 @@ <h2 class="sr-only">
...
+                    <dt>
+                        {# Translators: Editor label #}
+                        {% blocktranslate count counter=document.editors.count trimmed %}
+                            Editor
+                        {% plural %}
+                            Editors
+                        {% endblocktranslate %}
+                    </dt>
+                    {% for ed in document.digital_editions %}
+                        {% ifchanged %}
```
rlskoeser, line 43: suggest adding a template comment about why we're using `ifchanged` here

rlskoeser (approved):
Looks good to me. Made a couple of suggestions.

Issue #328, opened 2021-10-20
As a content editor, I want to create and edit content pages on the site so that I can update text on the site when information changes.


rlskoeser:
## testing notes

Visit https://test-geniza.cdh.princeton.edu/cms
- [x] Log in (if not already logged in)
- [x] Try to edit the existing content pages
- [x] Verify that the changes show up on https://test-geniza.cdh.princeton.edu/ by clicking the links in the header
- [x] On that same page (i.e. without returning to the Wagtail Admin interface), verify that the "Wagtail User Bar" appears (a [clickable circle](https://user-images.githubusercontent.com/7234006/145879209-a30b1976-ee3c-402a-91ab-20dee56b45aa.png) in the bottom left corner of the screen)
  - Note that this will only appear to logged in users and we can move/alter if it is intrusive, just comment here if that is the case
  - [x] Click the wagtail user bar and try to use the available functions

Visit https://test-geniza.cdh.princeton.edu/admin
- [x] Under "Content pages" --> "Contributors", add a few contributors to appear on the credits page
- [x] Verify that any contributors you add appear on https://test-geniza.cdh.princeton.edu/credits

## Deploy notes

- [x] Run `python manage.py bootstrap_content -H HOSTNAME -p PORT -l LANG` with appropriate values

## dev notes

basic setup for wagtail, including base template for content pages; one of example of each new page type should be added to lighthouse and percy builds

estimating as 5 points for now, with home page included; we may want to create a separate issue for the home page, but we don't have any designs for it yet, so let's just build out a basic version now that can be refined later

should be similar to work on S&co; see https://github.com/Princeton-CDH/mep-django/issues/157

- [x] add wagtail to python requirements and include in installed apps
- [x] bind wagtail admin interface urls to `/cms/`
- [x] create basic home page and content page models (or all content pages?) with basic page content (can adapt from mep-django)
- [x] setup templates and styles for content pages based on the designs in figma 
- [x] write a manage command to setup basic site content — should create placeholders for the pages we expect to include; will be used for bootstrapping content on the site and for ui/ux testing
- [x] add sample content page to lighthouse for testing (include home page also if different enough to merit)
- [x] add sample content page to percy for testing (include home page also if different enough to merit)
- [x] turn on the wagtailuserbar

richmanrachel:
[...]

PR #403, opened 2021-12-08
#328 - Add content management with Wagtail

blms:
## Visual review notes
[Percy build](https://percy.io/3201ecb4/geniza/builds/14486124)
- Note that I'm going off the old content page styling here, but I'll definitely update with the new designs when finalized and ready for me to start implementing!
- I also intentionally didn't use the Phosphor icons for the quotation marks, since those will change in the final design. I just put regular old quotation marks in there for the time being.

## What this PR does

- Per #328:
  - Installs Wagtail, allowing users to edit and manage content on the website's various content pages
  - Implements initial templates, styling, spacing, sizes for all content pages
  - Implements special models, logic, and styling for credits page
    - Users can add/edit `Contributors` in the Django admin panel (this was more feasible than doing it with Wagtail alone) to be displayed on the credits page
  - Adds management command `bootstrap_content` that allows for easy bootstrapping of needed content pages
    - `python manage.py bootstrap_content -H HOSTNAME -p PORT -l LANG`
     - If a Wagtail site is not present, creates one at hostname `HOSTNAME` (default localhost) and port `PORT` (default 8000)
     - Sets up home page and bootstraps subpages with `LANG` language code
     - Also includes optional `--fixtures` flag to add the test content page
  - Uses management commands in test site generation for Percy and Lighthouse


geniza/urls.py
```diff
@@ -15,24 +15,31 @@
...
 ]
 
 # urls that should be available in multiple languages
 urlpatterns += i18n_patterns(
     path("", include("geniza.corpus.urls", namespace="corpus")),
+    path("", include(wagtail_urls)),
 )
 
+# Media URLs for wagtail
+urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```
rlskoeser, line 42: this should be debug only, for the real deploy we serve out via nginx 

(I don't think this is related to the visual test problem, just thought I would flag)

geniza/templates/nav.html
```diff
@@ -14,7 +14,7 @@
                 {% translate "home" as aria_home %}
                 {# Translators: site name label for 'home' button in main navigation #}
                 {% translate "GENIZA" as site_name %}
-                <a class="home-link" href="#" aria-label="{{ aria_home }}" role="button">{{ site_name }}</a>
+                <a class="home-link" href="/" aria-label="{{ aria_home }}" role="button">{{ site_name }}</a>
```
rlskoeser, line 17: do we need a different link here for root of current multilingual site, or will this work as expected?

geniza/pages/image_formats.py
```diff
@@ -0,0 +1,26 @@
...
+        default_html = super().image_to_html(image, alt_text, extra_attributes)
+
+        return format_html(
+            "<figure>{}<figcaption>{}</figcaption></figure>", default_html, alt_text
+        )
+
+
+unregister_image_format("fullwidth")
+unregister_image_format("left")
+unregister_image_format("right")
```
rlskoeser, line 21: why do these need to be unregistered?

rlskoeser (approved):
This looks great. I had a couple of questions, but no blockers. Looking forward to getting this onto qa for testing soon!

I made a couple of adjustments:
- extended wagtail login form to add a PU cas login button (copied from another project)
- added a deploy note about running the bootstrap content script — please revise directions if they are not correct!

geniza/templates/nav.html
blms: It's automatically redirecting me to `/en`, but not sure we have a way of changing the language yet on the frontend so I can test with the others. But I couldn't find another way to link to the root page in Wagtail template tags—maybe I missed it!

geniza/pages/image_formats.py
blms: My thinking was that these should be unregistered until / unless we have specific styles for them from @gissoo, so that content authors can only pick from those that have been designed. On the designs so far I only see centered images with captions. @gissoo did you want to include styles for left and right aligned images, and/or centered without captions? Or was the intention that images in content pages should only be "centered with captions"?

blms:
@gissoo I did notice that about the images, and I wasn't sure if that was because those images were just smaller than the page width (in which case they won't be stretched here either)—or if it was because the max width of the images should be less than 100%. Did you have a specific max width (in pixels or percentage) in mind there?

geniza/pages/image_formats.py
rlskoeser: ok, makes sense! those are out of scope for MVP; I just didn't know if they already existed in wagtail (I thought we made them up when we implemented this before). maybe worth a comment?

geniza/templates/nav.html
rlskoeser: we don't have a way in the UI, I've just been editing the url when I wanted to test. this is probably good enough for now, we can revisit if needed!

geniza/pages/image_formats.py
gissoo: @blms I don't think we discovered a need for left and right images for these pages. I think I originally had them but when we decided to not have a editorial section we agreed to not have left and right images anymore, there doesn't seem to be a need.

gissoo:
> @gissoo I did notice that about the images, and I wasn't sure if that was because those images were just smaller than the body text width (in which case they won't be stretched here either)—or if it was because the max width of the images should be less than 100%. Did you have a specific max width (in pixels or percentage) in mind there?
> 

@blms Thanks for writing! I had defined a max width for the landscape images: 400px on desktop and 280px on mobile, and max height for portrait images: 300px on desktop and 200px on mobile – let me know if there are any issues

PR #405, opened 2021-12-09
Setup djiffy and add basic script to import IIIF manifests

rlskoeser:
I'm not certain this is the approach I want to use for the long-term, but for the MVP I'm going to go ahead with the approach I've used before. 

in this PR:
- add djiffy to requirements and installed apps
- local import manifest command to import iiif manifests based on urls associated with fragments
- notes about possible integration between manifest and fragment

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -173,6 +173,9 @@ class Fragment(TrackChangesModel):
 
     objects = FragmentManager()
 
+    # NOTE: may want to add optional ForeignKey to djiffy Manifest here
+    # (or property to find by URI if not an actual FK)
+
     class Meta:
         ordering = ["shelfmark"]
 
@@ -223,6 +226,9 @@ class Fragment(TrackChangesModel):
                 self.old_shelfmarks = ";".join(old_shelfmarks - {self.shelfmark})
             else:
                 self.old_shelfmarks = self.initial_value("shelfmark")
+
+        # NOTE: consider triggering manifest import here when iiif url changes
+
         super(Fragment, self).save(*args, **kwargs)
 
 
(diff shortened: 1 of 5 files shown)


geniza/settings/components/base.py
```diff
@@ -251,6 +252,11 @@
 # exclude admin and cms urls from csp directives since they're authenticated
 CSP_EXCLUDE_URL_PREFIXES = ("/admin", "/cms")
 
+# use jpg instead of png since some providers only support jpg
+DJIFFY_THUMBNAIL_FORMAT = "jpg"
+# disable djiffy import check, since we are not using djiffy views
+# DJIFFY_IMPORT_CHECK_SUPPORTED = False
```
blms, line 258: Should this be uncommented or is it a dev only thing?

geniza/corpus/models.py
```diff
@@ -173,6 +173,9 @@ class Fragment(TrackChangesModel):
 
     objects = FragmentManager()
 
+    # NOTE: may want to add optional ForeignKey to djiffy Manifest here
+    # (or property to find by URI if not an actual FK)
```
blms, line 177: I like the FK idea here because it links the data more tightly. Isn't it already possible to find by URI? (`iiif_url="uri"`)?

blms (approved):
Looks good, management command looks simple to use and tests are thorough. djiffy seems really cool! Just had a couple of small questions.

geniza/settings/components/base.py
rlskoeser: I fixed (well, changed) something in djiffy so it isn't needed right now — but I'm not sure the change I made in djiffy is actually the right one! which is why I left it. djiffy includes some views that assume book format paginated content, but we're not using them — so it might make sense to keep those supported checks as they were before and opt out of them in projects like this one where we don't care about those views.

geniza/corpus/models.py
rlskoeser: right — convenience property would be easy and might be good enough, but if we're going to cache them our db it would probably be better to make the connection explicit

Issue #413, opened 2021-12-14
Wagtail bootstrap script doesn't work as expected


blms:
## dev notes
- [x] If default site exists, set the fields to the provided values

rlskoeser:
tested this in qa after I loaded fresh data from production, and confirmed that it now works as expected

rlskoeser:
fyi @blms I don't consider something a bug unless the problem makes it into production; if we catch a problem before we make a release, then it's not a bug :-)

PR #414, opened 2021-12-14
#413 - Fix Wagtail bootstrap script

blms:
## What this PR does

- Per #413:
  - Fixes the bootstrap script to properly use the provided arguments when a default Wagtail site already exists (which is usually the case!)


rlskoeser (approved):
great!

PR #415, opened 2021-12-14
implement local iiif manifests & annotation lists to show images & transcriptions

rlskoeser:
initial PR for image & transcription functionality. Includes:
- add foreign key from `Fragment` to `Manifest`
- import `Manifest` on `Fragment` save, when appropriate
- revise `import_manifests` script to associate fragments and manifests; add a mode to _only_ associate a not import
- two new views to serve out a local IIIF manifest and local IIIF annotation list for documents with transcriptions; currently aggregates all available images from associated fragments and adds transcription annotation to the first canvas
- preliminary `iiif_utils` module to pull out some code from the view, probably more can move there
- revise mirador configuration to only expect a single manifest, and open annotation panel when there are transcriptions available
- move transcription styles out so they apply to mirador as well as search results (should they go somewhere else?)

doesn't handle all cases yet, and no unit tests


geniza/corpus/views.py
```diff
@@ -241,6 +240,147 @@ def get_context_data(self, **kwargs):
...
+        if document.has_transcription():
+            first_canvas["otherContent"] = {
+                "@context": "http://iiif.io/api/presentation/2/context.json",
+                "@id": absolutize_url(
+                    reverse("corpus:document-annotations", args=[document.pk]),
+                    request=request,  # request is needed to avoid getting https:// urls in dev
+                ),
+                "@type": "sc:AnnotationList",
+            }
+
```
blms, line 324: Could this make use of the new `iiif_utils` class?

geniza/corpus/views.py
```diff
@@ -241,6 +240,147 @@ def get_context_data(self, **kwargs):
...
+                # to track original source of this canvas
+                local_canvas["partOf"] = [
+                    {
+                        "@id": remote_manifest.id,
+                        "@type": "sc:Manifest",
+                        "label": {
+                            "en": ["original source: %s" % remote_manifest.label]
+                        },
+                    }
+                ]
```
blms, line 291: I started making a util for this too!

blms (commented):
All of this looks good so far! Made a couple of quick comments. I'll pass along my additional WIP work shortly.

geniza/corpus/views.py
blms: Too much variable here, so I'm doing away with this util 😆

blms:
Apologies if my WIP work makes no sense, brain feels a little fried like an egg at this time of day! Tried to start work on generating a new canvas in the case of transcriptions but no images. I didn't know if the whole annotation list (w/ resources) should be embedded there, or just the reference to its `@id`…?

geniza/corpus/views.py
blms: Actually scrapped this util, too many things variable here! 😆

Issue #396, opened 2021-12-06
Update sitewide type to use purchased fonts, new styles


blms:
## dev notes

- [x] Use the new typefaces provided by Gissoo
- [x] add a note to dev instructions in the readme
- [x] add to gitignore to ensure we don't accidentally add purchased fonts to the git repo
- [x] Use new type sizes and styles

gissoo:
Thank you !! A couple of points:
- I wonder if the two type styles below should be in fact 18 on desktop and 16 on mobile – the current active sort option looks too large, do you agree? 
<img width="653" alt="Screen Shot 2021-12-15 at 5 30 10 PM" src="https://user-images.githubusercontent.com/44615389/146274868-29e07d55-8d64-4fe4-bfa5-6d2f78d2d64e.png">

- The "Description" header on the document detail view needs to be bold and 20px on desktop 
- The "unpublished" should not be italicized 
- The "Transcription / Edition / Discussion" text in the grey box should be 18px on desktop and 16px on mobile 
- The page titles on content pages need to be bold and all page titles need to be 24px, I see some being 25.6px 

- Question: how are we calculating the type sizes? Is it by percentages? I see a lot of decimals, just wondering, it's not creating any significant changes except in the page title.

blms:
> I wonder if the two type styles below should be in fact 18 on desktop and 16 on mobile – the current active sort option looks too large, do you agree?

I think either is fine, so let's make it smaller!

> Question: how are we calculating the type sizes? Is it by percentages? I see a lot of decimals, just wondering, it's not creating any significant changes except in the page title.

I am using a harmonic type scale, which was implemented before I started working on the project, though I changed the scale factor from 1.25 to 1.125 to more closely match the `px` values you chose. It uses `rem` units:

```scss
// Type scale: generate sizes according to harmonic ratio. This scale is based
// on the "Major Second" scale, with a scale factor of 1.125. See:
// https://baseline.is/tools/type-scale-generator/
$text-size-xs: 0.79rem; //   ~ 13px
$text-size-sm: 0.89rem; //   ~ 14px
$text-size-md: 1rem; //      = 16px
$text-size-lg: 1.13rem; //   ~ 18px
$text-size-xl: 1.27rem; //   ~ 20px
$text-size-2xl: 1.42rem; //  ~ 23px
$text-size-3xl: 1.6rem; //   ~ 26px
$text-size-4xl: 1.8rem; //   ~ 29px
[...]

PR #416, opened 2021-12-14
#396 - New type styles

blms:
## What this PR does
- Per #396:
  - Uses new fonts and type styles from designs throughout site
  - Adds instructions to readme for local font usage


sitemedia/scss/components/_results.scss
```diff
@@ -103,6 +96,8 @@ section#document-list {
 
         // styles specific to transcription
         .transcription {
+            @include typography.transcription;
```
rlskoeser, line 99: I think we should apply the font style based on the script of the associated language, but I'm not indexing that in Solr yet and there are plenty of documents with transcription that don't have language metadata yet.

Is it a problem if we apply this to transcription with arabic text? There are only a handful, but there are some. Test cases: PGPID 4473, 19011, 1242, 8273, 8288

Maybe we can assume hebrew if script is hebrew or unset (once we add that), so that known-arabic documents won't get the wrong font.

rlskoeser (approved):
Documentation is clear; was able to install and run locally.  I have a question about the transcription fonts for Arabic, but otherwise this looks reasonable to me from what I can tell

I'll try to get to the font deploy/setup soon, so we can review this with percy!

sitemedia/scss/base/_fonts.scss
```diff
@@ -2,25 +2,68 @@
...
-// If not using Google Fonts, @font-face declarations would go here. Since we
-// use Google Fonts, we instead use the CSS2 API and request all the families,
-// weights, and variants we need using a single @import. See:
-// https://developers.google.com/fonts/docs/css2
+@charset "UTF-8";
 
-// Merriweather: light, light italic, regular, regular italic
-@import url("https://fonts.googleapis.com/css2?family=Merriweather:ital,wght@0,300;0,400;1,300;1,400&display=swap");
+// Usage counter and license for myfonts
+@import url("//hello.myfonts.net/count/4098e1");
```
rlskoeser, line 8: one more thought — in the past we've made the counter a production-only behavior, on the assumptions that dev/test requests shouldn't count; any thoughts?

PR #417, opened 2021-12-14
Allow webpack config and _fonts.scss to load fonts

blms:
## What this PR does

- Installs `resolve-url-loader` and `file-loader` with NPM to correctly load and resolve fonts in Webpacked CSS
- Updates webpack config, fonts CSS definitions for the above

Note: this will still fail in percy and lighthouse because they won't have access to the fonts. But this should work if the fonts are accessible by `/static/fonts` on QA for example.


webpack.config.js
```diff
@@ -31,6 +31,19 @@ module.exports = (env, options) => ({
...
         rules: [
+            // handle .woff and .woff2 fonts
+            {
+                test: /\.(woff(2)?|ttf|eot)(\?v=\d+\.\d+\.\d+)?$/,
+                use: [
+                    {
+                        loader: "file-loader",
+                        options: {
+                            name: "[name].[ext]",
+                            outputPath: "fonts/",
```
rlskoeser, line 42: does this mean fonts will still be included in the webpack bundle? what does that mean for restricting access by referrer?

blms: I think you're right. And actually, this part might not be needed—it's working fine without it for me!

rlskoeser:
going with a different approach, excluding urls in css loader

PR #418, opened 2021-12-15
Tweak fonts webpack config

rlskoeser:
- use `/static/fonts/` prefix for purchased fonts
- configure webpack css loader not to load urls


rlskoeser:
fonts aren't working for percy or lighthouse, but going to merge because we need to move forward; will revisit as able

Issue #429, opened 2021-12-16
As a user, I want to change site language so that I can switch languages when I don't want to use the browser-detected default.


rlskoeser:
Note that this is just an initial implementation so that we have the functionality available to help us with the functionality while we're developing it and for translators to review. We'll revise when there is a design available. 

## testing notes
- [x] confirm there is a language switcher form in the top right on every page
- [x] confirm that you can use the form to change the language used on the site
- [x] the language you choose persists as you navigate around the site

## dev notes
Implement a language-switching form; should be available on every page.

Should be able to adapt the form from the i18n prototype, which uses an existing django i18n view: https://github.com/Princeton-CDH/geniza/blob/experiment/i18n/geniza/templates/base.html#L19-L35

rlskoeser:
@gissoo (sorry to add to your list!)

Nick implemented a basic form when he built the i18n prototype; here's what it looks like:

![Screen Shot 2021-12-16 at 3 12 20 PM](https://user-images.githubusercontent.com/691231/146443118-af1d23b7-d25e-4d19-9632-f42ab9f7a754.png)
![Screen Shot 2021-12-16 at 3 12 16 PM](https://user-images.githubusercontent.com/691231/146443127-d30be7e2-9e00-4532-ae56-a01376984dd9.png)

Could you suggest where it should go on the page for desktop/mobile, and any other context/visuals that would be important to include? Hoping we can use the styles from the components you've already designed.

richmanrachel:
It works! So cool!

PR #447, opened 2021-12-17
#429 - Add language switcher, fix some multilingual issues

blms:
## What this PR does

- Per #429:
  - Adds a basic, unstyled language switcher (only enabled if more than one language present)
- Makes `Contributor` fields translateable
- Installs `wagtail_localize` to make it easier to translate Wagtail content pages
- Fixes slug URLs in navbar: `slugurl` breaks when using the language switcher, due to Django redirecting to the last URL visited, which in the case of `slugurl` will include the locale string! Better to let Django handle redirection to the appropriate language-coded slug URL based on the set language, and leave the URLs in the navbar as just the bare slug without the language code

## Deploy notes
- New pip dependency added for `wagtail_localize`
- Migration is required for `Contributor` and `wagtail_localize`


geniza/templates/nav.html
```diff
@@ -75,23 +75,23 @@
                     </li>
                     <li class="menu-item" role="none">
                         {# Translators: label for link to 'Credits' page in main navigation #}
-                        <a href="{% slugurl "credits" %}" role="menuitem">{% translate "Credits" %}</a>
+                        <a href="/credits" role="menuitem">{% translate "Credits" %}</a>
```
rlskoeser, line 78: add trailing slashes please — this will work, but no need to make django serve the redirect request

geniza/templates/snippets/language_switcher.html
```diff
@@ -0,0 +1,20 @@
+{% load i18n %}
+
+{% if LANGUAGES|length > 1 %}
+    <form id="language" action="{% url 'set_language' %}" method="post">{% csrf_token %}
+        <fieldset>
```
rlskoeser, line 5: I feel like we should add a label here; although I know this implementation is provisional. What do you think about "read this page in" [language] for now?

geniza/pages/admin.py
```diff
@@ -12,4 +13,10 @@
...
 admin.site.unregister(Image)
 admin.site.unregister(Document)
-admin.site.register(Contributor)
+
+# Set up tabbed translation admin for contributors
+@admin.register(Contributor)
+class ContributorAdmin(TabbedTranslationAdmin):
+    list_display = ("last_name", "first_name", "role")
+    search_fields = ("first_name", "last_name", "role")
+    fields = ("last_name", "first_name", "role")
```
rlskoeser, line 22: I wonder if this should be made editable in wagtail admin instead of django admin, but IDK if wagtail admin supports model translation! So let's leave it here for now.

rlskoeser (approved):
These changes look good to me! I pulled the code locally, reloaded the database, ran the bootstrap content, tried out the wagtail locales stuff. The translation stuff looks very powerful, but also could be a big learning curve.

The site was behaving strangely on the document details page with regards to translation and I finally figured out why (I think thanks largely to the work in this PR): the `permalink` method in the `Document` model is switching the active language to English! Would you mind fixing that before merging this PR? I think it should use [get_language](https://docs.djangoproject.com/en/3.1/ref/utils/#django.utils.translation.get_language) — either get the current language before and restore afterwards, or just use the current language to replace instead of keying on the `/en/`. (In hindsight this problem seems so obvious! 🤦‍♀️ )

rlskoeser:
question about the `slugurl` change: I presume things will work properly once we're able to use the wagtail page tree to generate the nav menu?

blms:
> question about the slugurl change: I presume things will work properly once we're able to use the wagtail page tree to generate the nav menu?

@rlskoeser I'm not 100% sure about this, because the django redirect behavior is to redirect to whatever was the last URL visited. However, I do know that Wagtail has a `page.localize.url` function that is different from `slugurl`, so that might be all that's needed to make it work. Either way, I think we can figure out a workaround once that's implemented. (Even if that means having two different forms for Django views and Wagtail pages! Thankfully `{% if page %}` can determine which one you're on.)

geniza/pages/admin.py
blms: Yeah, I did try this originally, but the Wagtail admin doesn't play nicely with non-Wagtail models. And it didn't seem to make sense to have Contributor as a subclass of Page or Document… not sure!

rlskoeser: There's a way to register models that aren't pages to be edited in wagtail — Nick set it up in our CDH website code base. We can look at that at some point and see whether the approach works with translatable models, but I think this is workable for now.

blms: Oh, great! Yes, that would be better in the future so that content editors don't have to go to multiple places.

Issue #325, opened 2021-10-18
As an admin, I want data from PGP v3 links database imported into the new database so that I can manage links from the main admin site.


rlskoeser:
## testing notes
- [x] check a handful of urls for each "link type" in the admin interface (document or footnote) and public interface (document scholarship details); confirm that the urls are present and work as expected; check that document relation is set correctly
    - [x] goitein note
    - [x] index card
    - [x] jewish traders
    - [x] india traders
- [x] check that log entries were created to document footnotes that were updated with links, new footnotes created (all index card notes), and sources that were created (all index cards; some new volumes of typed texts) — either from individual entry history page or from the log entries list
- [x] check a record that has multiple links of the same type to confirm that all urls are present: this is a good one: https://test-geniza.cdh.princeton.edu/en/documents/502/scholarship/

## dev notes
- [x] Implement logging
- [x] Implement testing
- [x] Handle multiple footnotes with the same source (India traders only)
- [x] Remove unnecessary headers in the header check

### revisions after testing
- [x] urlencode urls so they are valid and footnotes can be edited in django admin without error
- [x] for typed text, use file basename (up to but not including PGPID) for footnote location 
- [x] for index cards, use "Card ##" for footnote location (link target is card number)
- [x] for india traders, use numbers (e.g.  I-1-2, I-31a, II-13-15) for location
- [x] for jewish traders, use "document #" based on number in filename
- [ ] generate a csv report of documents not found for link types we care about, for later data work
- [x] do not set index cards site url when generating index cards source records

New manage command to import relevant data from a CSV export of the links table in the PGP v3 database.
Records should be imported differently based on the link type. Records with type image, iiif, cudl, and transcription, can all be ignored since we're already handling that content in a different way.

In all cases, we should match by PGPID, but if PGPID is not found search and match on old PGPID for attachments that belong to merged documents.

Script should create admin log entry documenting any sources & footnotes created and any footnote changed.

We only want to run the script once in production, but for developer testing and convenience we may want to make it idempotent.

## goitein_note
These links correspond to the typed texts footnotes already in the database. Find the footnote based on PGPID or old PGPID; filter by source title/author if necessary to find the Goitein typed texts source for the document. (If there are multiple candidates on merged documents we'll have to figure out how to resolve that.)
Update the footnote with the url to the scan: base url is https://commons.princeton.edu/media/geniza/ and `link_target` is the rest.

## indexcard
[...]

PR #454, opened 2021-12-17
#325 - Add links script revisions

blms:
## What this PR does
- Per #325, revises the add links script to add:
  - Escaped URLs to fix admin interface bugs
  - Footnote location generation / assignment
  - A report of documents not found

## notes
I just decided to save the "documents not found" report to the same directory as the original CSV. I also appended the original CSV's basename to the filename, in case multiple reports will be generated in the same directory for different CSVs.

Also, I reordered some of the logic in `test_add_link` to minimize the amount of times `link_target` is updated to reflect the kind of `link_target` that we expect from each document type.

### DIFF
--- a/geniza/corpus/management/commands/add_links.py
+++ b/geniza/corpus/management/commands/add_links.py
@@ -136,8 +136,16 @@ Created {footnotes_created:,} new footnotes; updated {footnotes_updated:,}.
 
         # generate CSV report of documents that could not be found
         if len(self.not_found_documents) > 0:
+            # basename without extension
+            original_csv = (
+                basename(csv_path)[: basename(csv_path).rindex(".")]
+                if "." in csv_path
+                else basename(csv_path)
+            )
             # save in same directory as original CSV
-            nf_report_path = dirname(csv_path) + "/not_found_report.csv"
+            nf_report_path = dirname(csv_path) + (
+                "/documents_not_found_in_%s.csv" % original_csv
+            )
             with open(nf_report_path, "w+") as f:
                 fieldnames = [
                     "linkID",
@@ -153,7 +161,7 @@ Created {footnotes_created:,} new footnotes; updated {footnotes_updated:,}.
                     csv_writer.writerow(not_found_row)
 
             self.stdout.write(
-                "Saved report of not found documents to %s." % nf_report_path
+                "Saved report of documents not found to %s." % nf_report_path
             )
 
     def log_action(self, obj, action=CHANGE):
--- a/geniza/corpus/management/commands/add_links.py
+++ b/geniza/corpus/management/commands/add_links.py
@@ -1,6 +1,6 @@
 import csv
 import re
-from os.path import basename, dirname
+from os.path import basename, dirname, join, splitext
 from urllib.parse import quote
 
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


geniza/corpus/management/commands/add_links.py
```diff
@@ -226,7 +242,8 @@ def get_india_book(self, title):
     # base url target for each supported link type
     base_url = {
         "goitein_note": "https://commons.princeton.edu/media/geniza/",
-        "indexcard": "https://geniza.princeton.edu/indexcards/index.php?a=card&id=",
+        "indexcard": "https://geniza.princeton.edu/indexcards/"
+        + quote("index.php?a=card&id="),
```
rlskoeser, line 246: these args should not be quoted
```suggestion
        "indexcard": "https://geniza.princeton.edu/indexcards/index.php?a=card&id=",
```

geniza/corpus/management/commands/add_links.py
```diff
@@ -255,32 +272,82 @@ def add_link(self, row):
...
             return
 
         # target url: combine base url for this link type with link target from csv
-        target_url = "".join([self.base_url[link_type], row["link_target"]])
+        target_url = "".join([self.base_url[link_type], quote(row["link_target"])])
         # get doc relation for this link type
         doc_relation = self.document_relation[link_type]
 
-        # get source based on the link type
+        # get source and location based on the link type
```
rlskoeser, line 287: all of these use the base filename without the extension; let's generate it once outside the if statements

you'll need to add another `os.path` import for `splitext`

```suggestion
        # get source and location based on the link type
        
        basename = splitext(row["link_target"])[0]
```

geniza/corpus/management/commands/add_links.py
```diff
@@ -255,32 +272,82 @@ def add_link(self, row):
...
         if link_type == "goitein_note":
             source = self.get_goitein_source(doc=doc, title="typed texts")
+            base_name = basename(row["link_target"])
+            try:  # use base name up to PGPID for goitein texts
+                end_index = base_name.index("(PGPID")
+            except ValueError:  # if PGPID not present, use base name up to extension
+                end_index = base_name.rindex(".") if "." in base_name else 0
+            # use string up to end_index and strip whitespace for location
+            location = base_name[:end_index].strip() if end_index else base_name.strip()
```
rlskoeser, line 296: let's simplify and just use `split`, which will work whether the text is there or not
```suggestion
            # use target location up to parenthetical pgpid
            location = basename.split("(PGPID")[0].strip()
```

geniza/corpus/management/commands/add_links.py
```diff
@@ -255,32 +272,82 @@ def add_link(self, row):
...
+            self.not_found_documents += [row]
             if self.verbosity > self.v_normal:
                 self.stdout.write(
                     "Document %s not found in database" % row["object_id"]
                 )
             return
 
         # target url: combine base url for this link type with link target from csv
-        target_url = "".join([self.base_url[link_type], row["link_target"]])
+        target_url = "".join([self.base_url[link_type], quote(row["link_target"])])
```
rlskoeser, line 283: this looks correct to me, do we have any tests that confirm ?

geniza/corpus/management/commands/add_links.py
```diff
@@ -255,32 +272,82 @@ def add_link(self, row):
...
+    def generate_csv_report(self, csv_path, csv_headers):
+        """Generate a CSV containing the rows of the original CSV for which a document
+        could not be found."""
+
+        # basename without extension
+        original_csv = (
+            basename(csv_path)[: basename(csv_path).rindex(".")]
+            if "." in csv_path
+            else basename(csv_path)
+        )
```
rlskoeser, line 340: should use `splitext` here too

geniza/corpus/management/commands/add_links.py
```diff
@@ -255,32 +272,82 @@ def add_link(self, row):
...
+        # save in same directory as original CSV
+        nf_report_path = dirname(csv_path) + (
+            "/documents_not_found_in_%s.csv" % original_csv
         )
```
rlskoeser, line 344: needs to handle input csv file in current working directory! (as I found when I ran it locally)

how about this?

(possibly we should use `os.path.join` while we're at it...)

```suggestion
        nf_report_path = "%s/documents_not_found_in_%s.csv" % (dirname(csv_path)  or '.', original_csv)
```

rlskoeser (changes requested):
looks good! some comments, only two things are not working correctly (quoting index card url, local path for missing data export); other comments are refactors to simplify code and improve readability that I hope will be quick to change

geniza/corpus/management/commands/add_links.py
blms: Quick question on this one, should it be that, or `splitext(basename(row["link_target"]))[0]`? I think it depends on whether we want:
```
"6E.1.5 Portrait of India trade pt 2/T-S 10J16.1_4 (PGPID 7892)"
"6E.1.5 Portrait of India trade pt 2/T-S 10J16.1_8 (PGPID 7892)"
"6E.1.5 Portrait of India trade pt 2/T-S 10J16.1_6 (PGPID 7892)"
```
or
```
"T-S 10J16.1_4 (PGPID 7892)"
"T-S 10J16.1_8 (PGPID 7892)"
"T-S 10J16.1_6 (PGPID 7892)"
```

rlskoeser: ooh, right ! — that's the one case where you shouldn't use the basename

rlskoeser: thanks for catching — probably worth a comment to note why this one is different than the others

blms: Just to clarify, using the `basename` function would give the second one, and not using it would give the first… which one do we want? If we want the first one, then the suggestion you wrote (without `basename`) is fine. I guess a better question is, are there examples where we _do_ want to use the `basename` function? It seems many Goitein notes are like this—should we just apply it to all except the Goitein notes?

rlskoeser: sorry for being ambiguous — we want the full location, and in fact the first part is more informative than the second

however, I looked at the links csv again and I don't think we need to use `basename` at all — right? these are the only ones that have a nested path, and we want to keep it

blms: Great! Then your original suggestion is correct as is, I'll go ahead and use it 😄

geniza/corpus/management/commands/add_links.py
blms: Do you think this is sufficient, or should it be tested more thoroughly?
https://github.com/Princeton-CDH/geniza/blob/d90f60f1ce79b7d244f130f2f285e1d74a1b02c0/geniza/corpus/tests/test_add_links.py#L316-L323

geniza/corpus/management/commands/add_links.py
rlskoeser: 😆 😂 ok! (the variable name may be slightly confusing!)

geniza/corpus/management/commands/add_links.py
rlskoeser: perfect, thank you! sorry I didn't look more closely and find this myself

blms: No problem! I actually amended it to manually use the URL encoding rather than use the `quote` function again, so I'm glad you brought it to my attention :)

rlskoeser (approved):
much cleaner!

PR #459, opened 2021-12-20
Tweak language switcher

rlskoeser:
When I saw the language switcher in qa I immediately saw we didn't have it quite right! It is probably still not great and will benefit from Gissoo's design input when she is able to get to it, but wanted to make a quick adjustment for now:

- display "read this page in ..." in the language that will be switched _to_, so it's readable to a user who can't read the site text in the current language
- add dir=rtl in css for hebrew & arabic options

### DIFF
--- a/DEPLOYNOTES.md
+++ b/DEPLOYNOTES.md
@@ -3,9 +3,10 @@
 ## 0.7
 
 -   Update settings to configure **TEI_TRANSCRIPTIONS_LOCAL_PATH** for the deploy environment.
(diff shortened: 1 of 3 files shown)


blms (approved):
Thanks for the quick fixes! Very funny that I didn't think to display "read this page in…" in the language of the choice 🤦 😆

rlskoeser:
> Thanks for the quick fixes! Very funny that I didn't think to display "read this page in…" in the language of the choice 🤦 😆

not obvious until you try to use it that it needs to differ from the translation pattern we're using everywhere else!

Issue #449, opened 2021-12-17
As a front-end user I want visual signifiers for the search form on the search page to improve my experience when I hover, click on, navigate through them with the keyboard, or know when an item is disabled


gissoo:
## testing notes (QA)

- [x] Visit the documents search page, https://test-geniza.cdh.princeton.edu/en/documents/, and confirm that the search styles match the designs
- [x] Test interactions such as click, hover, focus, select filter/sort options (note that the first time you click on a filter option, you will see the "focus" style, click a second option to see the "active" style on the first one you clicked)

---

Here are links to the designs: 

- [for desktop light mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A23976)
- [for mobile light mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A24196)
- [for desktop dark mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A26117)
- [for mobile dark mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A25898)

gissoo:
@blms Thank you!! The search functions look match the designs and function like the design intended – however the sort options allow you to see and choose "relevance" without inputting any search terms, I think we don't want that, correct @rlskoeser ?

- Also, we should eventually add a visual signifier to show when there are active filters for when the filter is collapsed right now (or _hidden_ as the original designs in Figma show)

blms:
@gissoo @rlskoeser Did we decide on JS to check for input into the search box for choosing "relevance"?

Or maybe it would make more sense to make the search box input a required field when "relevance" is selected?

The original implementation did not allow selecting "relevance" even if you had entered text, as it would only become enabled after you made a search with a different sort option.

blms:
@gissoo @rlskoeser Closing this, split the other two concerns off into two new issues:

https://github.com/Princeton-CDH/geniza/issues/523
https://github.com/Princeton-CDH/geniza/issues/524

@gissoo Thanks for your review!

PR #466, opened 2021-12-22
#449 - Implement new styles for search form

blms:
## Visual review notes
[Latest Percy build](https://percy.io/2cf28a24/geniza/builds/14911032)
- The scope for this PR is the search form.
- Since the light mode uses a dark drop shadow on the form elements, the dark mode uses a semi-transparent white for the same effect. On Figma they use dark drop shadows, but those wouldn't show up on the dark mode backdrop. Let me know if this should be changed. (Probably can only review this on the QA site since it involves dark mode.)
- There were some elements that were missing states such as active and focus, so I did my best to guess what these should look like. It's most noticeable in the filter and sort options. (Probably can only review this on the QA site since it involves interaction.)

## What this PR does
- Per #449:
  - Restyles all search form elements with new designs
  - Adjusts logic for search sort dropdown to allow design implementation
- Fixes issue with italicized highlights in search results

## dev notes
- I decided to make the sort dropdown a set of radio buttons nested inside a `<details>`/`<summary>` instead of a `<select>` element, since the dropdown menu of `select` elements cannot be styled. I actually started working on [something like this](https://css-tricks.com/striking-a-balance-between-native-and-custom-select-elements/), but after a while radio buttons seemed to make more sense. I would love feedback in particular on this implementation!
  - I added a bit of JS (`sort-select.js`) just to make it work much more like a standard `select`, but it's totally possible to use it without JS. I included comments there explaining what will happen when JS is disabled.
  - For the dropdown arrow, I used a `div` with `aria-role="button"` instead of a `button` or `input` element to avoid it stealing focus from the `summary` in which it resides. But maybe it's worth using one of those instead and figuring out a workaround.


rlskoeser (approved):
This all looks great to me. Yes, the inability style select elements is always a problem; I think your approach here makes sense.

The separate drop-down for the sort and search looks inconsistent from the document type filter, but I'm guessing that's because the document type will eventually be in a filter dialog with other search filter options.

This may be a small thing, but when I looked at Gissoo's designs it seems that the sort drop-down does not expand under the drop-down button — it lines up with the left side of the input, with the button alone. (I'm probably not describing this well).  Defer to Gissoo on the other visual specifics.

After reviewing these changes locally, I think we will need to add dark-mode screenshots to percy (maybe not necessary for every page, but at least a few of the major ones) — but I'm not sure the best way to do it! Do we need to implement a switcher and then set Percy to use it?

blms:
@rlskoeser 

> The separate drop-down for the sort and search looks inconsistent from the document type filter, but I'm guessing that's because the document type will eventually be in a filter dialog with other search filter options.

Right—I was thinking about the document type filter as part of a filter panel with collapsible and expandable menus that cause the page content to reflow. And meanwhile I thought of the sort as a dropdown (basically as close as possible to a `<select>` element without actually being one).

I did notice after coding the faux-select that it operates pretty differently via the keyboard (change focus using up and down arrows, then space or enter to choose an option) when compared to the doc type filter (where each checkbox gets focus by tabbing, and then chosen by pressing space). Again, I tried to match the dropdown as closely as possible to the select element, and I think it is fairly close. But it did make me wonder if we might want to go back and revisit the functionality of the doctype filter to match it slightly more closely.

> This may be a small thing, but when I looked at Gissoo's designs it seems that the sort drop-down does not expand under the drop-down button — it lines up with the left side of the input, with the button alone. (I'm probably not describing this well). Defer to Gissoo on the other visual specifics.

Oh, good catch! I missed that, thanks.

blms:
> After reviewing these changes locally, I think we will need to add dark-mode screenshots to percy (maybe not necessary for every page, but at least a few of the major ones) — but I'm not sure the best way to do it! Do we need to implement a switcher and then set Percy to use it?

I think we should be able to do this with `prefers-color-scheme` in the percy-specific CSS. I'll give it a shot!

rlskoeser:
> I did notice after coding the faux-select that it operates pretty differently via the keyboard (change focus using up and down arrows, then space or enter to choose an option) when compared to the doc type filter (where each checkbox gets focus by tabbing, and then chosen by pressing space). Again, I tried to match the dropdown as closely as possible to the select element, and I think it is fairly close. But it did make me wonder if we might want to go back and revisit the functionality of the doctype filter to match it slightly more closely.

This is tricky stuff! I think we should review the search form elements and interactions at some point; I think that once we have an MVP release I should request a review from our campus accessibility expert and see what he recommends.

blms:
@rlskoeser I could not figure out what is going on with Safari on Percy here. I can't replicate it locally. Other than that, I think this one is ready to go.

I'm opening a new issue for dark mode on Percy since my proposed solution doesn't work!

blms:
Safari bug:

Screen Shot 2022-01-04 at 10 53 34 AM

Chrome:

Screen Shot 2022-01-04 at 10 53 39 AM

rlskoeser:
@blms I'm on Safari 15 now and this looks fine. (I forget, did I have 14 before?)  Given that Apple prompts people to update and won't let us install older versions, maybe we ignore this. 

I wonder why the Percy builds are using such an old version? Project settings only let me turn Safari on/off, there are no controls to select a specific version that I can find.

blms:
That sounds fine to me. Not sure why Percy hasn't updated their version, and I don't know of any way to switch versions there. Looks like all their versions are a little out of date:

> **Browsers**
> Safari 14.0
> Firefox 88.0
> Chrome 91.0
> Edge 91.0
>

rlskoeser:
@blms I sent a support request to browserstack about browser versions in percy. I don't think we should make that a blocker for this issue, though.

blms:
@gissoo It is! I'm not sure why it would appear darker. Maybe an optical illusion because of the Figma background?

Here they are side by side. Percy on top, then Figma, then my local for comparison.

148095978-73f3ab73-607d-4cab-be03-5ba70933b3b8
Screen Shot 2022-01-04 at 1 52 38 PM
Screen Shot 2022-01-04 at 1 52 52 PM

gissoo:
@blms is the color of type (especially the "legal document") in your implementation 161616? 

@blms I don't think it's an optical illusion, I just compared them and added an image of the figma version in my previous comment above ....this is so weird

rlskoeser (approved):

blms:
> @blms I don't think it's an optical illusion, I just compared them and added an image of the figma version in my previous comment above ....this is so weird

@gissoo Interesting, that's not what my Figma looks like! Could it be a color profile mismatch? If you are on a Mac, try going to System Preferences --> Displays and switching your color profile to sRGB. You may also want to turn off "True Tone" if it's on. It seems like this is not as simple on Windows, but let me know if you're on Windows and we can try to figure it out.

Here's what my color meter reports when I switch my color profile to sRGB:

### Percy (and also what I see on figma, and my local): 122, 172, 123 (#7AAC7B)
Screen Shot 2022-01-04 at 2 59 56 PM

### Your figma screenshot: 108, 174, 118 (#6CAE76)
Screen Shot 2022-01-04 at 2 59 41 PM

gissoo:
@blms thank you!!!! This is WILD!!! You're awesome!!! 
Let's go with #7AAC7B and keep the second variation as back up in case users find it hard to read text on #7AAC7B – they are both accessible.

blms:
@gissoo We have dark mode on this one now, so ready for a re-review! 😄

gissoo:
@blms thank you!! These look great – do you know why in my designs the sort "scholarship Records Most – Least" fit in one line but not in the implementation? Also, can we vertically center the arrow in the sort drop down? 

Thanks!!

blms:
Approving this Percy build since Gissoo already approved the prior build, and this is just a merge commit.

Issue #453, opened 2021-12-17
As a front-end user I want to recognize links and visual signifiers for links to improve my experience when I hover, click on, or navigate through links with the keyboard


gissoo:
## Testing notes
- [x] Confirm that link underlines look right
- [x] On pages such as https://test-geniza.cdh.princeton.edu/en/documents/950/, confirm that link styles look right according to designs (excluding the tabbed navigation)
- [x] Test interactions (click, hover, focus) on these links

## dev notes
here are links to the designs:

- [for desktop light mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A25072)
- [for mobile light mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A25091)
- [for desktop dark mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A26447)
- [for mobile dark mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A26518)

note: links can be in both italics and regular (depends on the content, the underline is what remains the same in this component)

gissoo:
Everythink looks fine! 

- Except this one: Can we fix this so that it would look more like the design or the other link styles?
<img width="340" alt="Screen Shot 2022-01-12 at 6 09 29 PM" src="https://user-images.githubusercontent.com/44615389/149237511-15650591-893f-4f1a-81fe-694e7caa8443.png">
<img width="474" alt="Screen Shot 2022-01-12 at 6 05 02 PM" src="https://user-images.githubusercontent.com/44615389/149237076-d6d1e878-a6b1-426e-bec7-5ac3e3ea2b3e.png">

- unrelated to this issue, but is there a reason the "download as text" link is positioned the way it is? can we align it to the left? it looks odd right now.

rlskoeser:
@gissoo thank you for flagging, the interrupted underlines were bothering me on that one! I think it must be too close to the text.

The "download as text" was a preliminary implementation to test my proposed functionality for #434 — based on Rachel's feedback, I'm going to be moving the link so it will appear just above the mirador viewer (which displays the transcription text but doesn't allow selecting for copy/paste). The label will also change, it will be "Download author's edition". (LMK if you have any input!)

[...]

PR #470, opened 2021-12-27
#453 - Implement link styles

blms:
### [Percy build](https://percy.io/2cf28a24/geniza/builds/14883225)

**In scope:**
- All underlined links, excluding tabbed navigation (since that will be changed in #472)

**Additional notes:**
- I didn't see any focus styles, so it will use whatever the browser's native focus is, e.g.:
   Screen Shot 2021-12-27 at 11 27 44 AM
- The use of colors in dark mode is inconsistent with how they've been translated from light mode elsewhere, which necessitated creating a new set of colors for link styles. Normally light mode `primary` (`#436043`) would be swapped out for dark mode `primary` (`#c37f97`), but for links they've been flipped in the design so that light mode `primary` (`#436043`) is actually swapped out for dark mode `secondary` (`#b76680`). Because of the way color mapping between light/dark modes works, I've created new color names `link-primary` and `link-secondary` to get the swap to occur properly:
  Screen Shot 2021-12-27 at 11 22 56 AM
  - This is ultimately fine—it's just a couple of extra lines, but the inconsistency might be worth considering, in case it was unintended! Cleaner of course to only have one "primary" and one "secondary" universally 🙂

## What this PR does
- Per #453:
  - Implements link style and color according to the designs


gissoo:
Thanks @blms the link styles look fine!

Issue #410, opened 2021-12-13
As a content editor, I want to be able to manage pages and page order in the site navigation menu or about submenu, so that I can update the site as content changes.


blms:
### creating new pages
- Visit https://test-geniza.cdh.princeton.edu/cms
- Log in and click Pages, then click on the English home page ("The Princeton Geniza Project")
- Confirm that you can see a page called About and its type is a Container page
- Click the green right arrow to the right of the About page
- Confirm that several content pages are child pages of this About page
- Try adding a new child page with the "Add Child Page" link at the top. Give it a title and body, click "Promote" and check "Show in menus", then at the bottom next to "Save draft," click the arrow and click "Publish"
- Once published, return to https://test-geniza.cdh.princeton.edu
- Confirm that your new child page shows up in the About menu in the top navigation, and you can navigate to it

### managing page order
- Visit https://test-geniza.cdh.princeton.edu/cms, click Pages, then click on the English home page
- Click the green arrow to the right of the About page
- At the top, click "More" > "Sort menu order"
- Reorder the pages by dragging them around with the handles on the left
- Return to https://test-geniza.cdh.princeton.edu and confirm that the About menu order has changed accordingly

### managing translations
- Visit https://test-geniza.cdh.princeton.edu/cms, click Pages, then click on the Hebrew home page this time
- Click "edit" under some of the pages, such as Contact Us, About, and About's subpages
- Click "Translate this page", then "Submit"
- In the Content section, by each field, confirm you can see a button that says "Translate". Press that button on some fields  and try translating them to Hebrew
- Click "Publish in Hebrew" at the bottom
- You may see a warning "Published [page] in Hebrew with missing translations" if there are fields left untranslated, which includes the Slug field in the Promote section. Feel free to ignore or translate them
- Return to https://test-geniza.cdh.princeton.edu and switch to Hebrew using the language switcher in the top-right corner
- Confirm that your translations appear in the menu(s)

## dev notes 
- [x] Create ~"Main navigation" and~ "About" link pages and make appropriate content pages subpages
- [x] Update nav template to programmatically display the pages

## deploy notes
See `DEPLOYNOTES.md`. In addition:

- Content editors should be made aware that to edit pages in sub-menus, in the Wagtail admin, they need to click the green right arrow button next to a Container page. Editing the Container page itself will only allow them to change its title or slug.
[...]

PR #471, opened 2021-12-27
#410 - Retrieve navigation page names and slugs programmatically

blms:
## What this PR does

- Per #410:
  - Creates a `ContainerPage` class of which `ContentPage` instances may be subpages
  - Updates the bootstrap script to create an `ContainerPage` instance and make all `ContentPage` instances subpages of that instance, except for "Contact Us" and the test fixture page
  - Updates the nav template to use the new logic to display children of the Home page in the main nav, and children of Container pages in their sub-navs

## Deploy Notes

See `DEPLOYNOTES.md`. In addition:

- Content editors should be made aware that to edit pages in sub-menus, in the Wagtail admin, they need to click the green right arrow button next to a Container page. Editing the Container page itself will only allow them to change its title or slug.
- Content editors should also be made aware that new content pages must be of type "Content page", regardless of whether they are to appear in the main navigation or a sub-menu. If they want the new page to appear in a sub-menu, they should be made child pages of an existing "Container page" (like the others in that sub-menu).


geniza/pages/models.py
```diff
@@ -39,6 +40,19 @@ class Meta:
...
+    """An empty :class:`Page` type that has :class:`ContentPage` instances
+    as its subpages."""
+
+    # can only be child of HomePage
+    parent_page_types = [HomePage]
+    subpage_types = ["pages.ContentPage"]
+
+    # should not ever actually render
+    def serve(self, _):
+        raise Http404
```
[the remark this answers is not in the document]
rlskoeser: Maybe it should redirect to the parent page.

geniza/templates/nav.html
```diff
@@ -66,33 +68,21 @@
...
-                    </li>
-                    <li class="menu-item" role="none">
-                        {# Translators: label for link to 'FAQ' page in main navigation #}
-                        <a href="/faq/" role="menuitem">{% translate "FAQ" %}</a>
-                    </li>
+                    {% for child_page in about_page.localized.get_children.live.in_menu %}
+                        <li class="menu-item" role="none">
+                            {# Translators: label for link to 'Credits' page in main navigation #}
+                            <a href="/{{ about_page.localized.slug }}/{{ child_page.localized.slug }}/" role="menuitem">
+                                {{ child_page.localized.title }}
```
rlskoeser, line 82: We shouldn't be constructing the url like this.  Can't we use `page_url` instead?

Also, I think the translators comment is out of date now.

```suggestion
                            <a href="{% pageurl child_page.localized %}" role="menuitem">
                                {{ child_page.localized.title }}
```

geniza/templates/nav.html
```diff
@@ -42,18 +42,20 @@
...
                 <a id="open-about-menu"
                     href="#about-menu"
                     aria-haspopup="true"
                     aria-expanded="false"
                     role="button">
-                    {{ about_label }}
+                    {{ about_page.localized.title }}
                 </a>
                 {# Translators: accessibility label for 'about' menu in navigation #}
                 {% translate "about menu" as about_menu_label %}
```
rlskoeser, line 61: Would like to see this not be hard-coded to the about page set in context, but just iterate over pages and handle a container page with child pages accordingly.

rlskoeser (changes requested):
I would like to see `AboutPage` renamed to something more general and descriptive of how it functions, and see the nav logic reworked so it is based on page type and not hard-coded for current about page logic. (The about page should have `in_menus` set to true and nav should follow page order based on wagtail site.)

Not sure what to do about re-running the bootstrap content script — it doesn't look like they've added much content in production yet (just some preliminary credits), but it still might be easier to make the adjustments manually since we already have the locale mirroring configured. (Otherwise I think we have to delete all the sites, run the scripts, re-configure locales...)  Please add instructions to the deploy notes for the next version about what needs to be done to adjust for this.

geniza/pages/models.py
```diff
@@ -39,6 +40,19 @@ class Meta:
         verbose_name = "homepage"
 
 
+class AboutPage(Page):
```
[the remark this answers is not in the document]
blms: Done in a603fa0 👍

geniza/pages/models.py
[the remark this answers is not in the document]
blms: Done in a603fa0 👍

geniza/context_processors.py
```diff
@@ -1,12 +1,18 @@
...
 def template_globals(request):
     """Template context processor: add global includes (e.g.
     from django settings or site search form) for use on any page."""
 
+    site = Site.find_for_request(request)
     context_extras = {
         "SHOW_TEST_WARNING": getattr(settings, "SHOW_TEST_WARNING", False),
         "FONT_URL_PREFIX": getattr(settings, "FONT_URL_PREFIX", ""),
+        "site": site,
+        "about_page": site.root_page.get_children().exact_type(AboutPage).first(),
```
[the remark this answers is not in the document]
blms: Removed in 1269ac1 👍

geniza/templates/nav.html
blms: Done in 1269ac1 👍

blms:
@rlskoeser Last thing, I do still wonder if it's a bit confusing when you create a child page of the homepage, where it might appear as though to create a page that appears in the submenu, one should click "Sub-menu page" here:

Screen Shot 2022-01-05 at 12 35 12 PM

I went with "sub menu page" to explain its display purpose, but maybe "container page" _would_ be a better name to avoid the confusion. What do you think?

geniza/templates/nav.html
```diff
@@ -43,48 +43,49 @@
...
                     </li>
-                    <li class="menu-label" role="none">
-                        <span>{{ about_page.localized.title }}</span>
-                    </li>
-                    <li class="menu-button" role="none">
-                        <a href="#" role="button">
-                            <span class="sr-only">{{ close_button }}</span>
+                {% else %}
+                    <li class="menu-item" role="none">
+                        <a href="/{{ top_level_page.localized.slug }}/" role="menuitem">
```
rlskoeser, line 83: Let's use `pageurl` here too
```suggestion
                        <a href="{% pageurl child_page.localized %}" role="menuitem">
```

rlskoeser (approved):
Revisions & deploy notes look good to me. Just one change needed where we should use `pageurl`

geniza/templates/nav.html
blms: Done in d6c9c58 👍

blms:
Good catch! Also regenerated translations in 91428f6. Will merge when checks complete.

Issue #398, opened 2021-12-07
As a front-end user I want visual signifiers for tabs on the document detail view and scholarship records pages when I hover, click on, or navigate through them with the keyboard, or when they are disabled 


gissoo:
## testing notes (QA)

- [x] Confirm that hovering over the currently active tab doesn't change its color
- [x] Confirm that on mobile, the hover color doesn't appear when clicking tabs
- [x] Visit a document detail page, such as https://test-geniza.cdh.princeton.edu/en/documents/3504/ and confirm the tabs look right according to the designs
- [x] On desktop and mobile, test interactions such as click, hover, focus, and tap

---

Note: the designs include styles for the "external links" page as well, however since it's not in the MVP we can either not display it or display it like the disabled state in the designs 

the designs include styles for tabs in the following states!
- Active = selected
- Focus = keyboard nav
- Hover (only on desktop)
- Disabled – using the color #9E9E9E for this – 

Note: the active, focus or hover states do not apply to the disabled tab item AFAIK. 

[Screen Shot 2021-12-07 at 9.48.11 AM.png](https://images.zenhubusercontent.com/5bf307a69d595b459ec91bb1/369fc0e4-5144-43b0-a33d-825ba1bba897):

gissoo:
Here are links to the designs:

- [for desktop light mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A24600)
- [for mobile light mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A24630)
- [for desktop dark mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A26417)
- [for mobile dark mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A26807)

gissoo:
@blms thank you!!! These look great!! Two points below:

- Is there a way to not hover on something that's already active or active by default? For some reason it looks really odd to hover and get a hover response on an active tab but ignore me if I'm not making sense

- On mobile, on Safari as I'm only able to test on Safari right now, when you select a tab it's taking a few seconds to turn the selected tab green, it's first turning to the hover color. The hover color should not appear here. I appreciate it if we could fix this.

[...]

PR #472, opened 2021-12-27
#398 - Implement styles for tabbed navigation

blms:
## Visual review notes
[Percy build](https://percy.io/2cf28a24/geniza/builds/14882339)

In scope:
- Tabbed navigation in document detail/scholarship views

## What this PR does
- Per #398:
  - Implements styles for default, selected, hover/active, and disabled tabs in the document detail/scholarship records tabbed navigation


blms:
@rlskoeser Requesting your review because I wanted to ask: how do you feel about the use of elements (rather than pseudoelements) for the tab navigation underlines? Not semantically useful, but the best I could come up with to get it working consistently.

I really wanted to use `::after` pseudoelements for them, but the problem is that when relatively-positioned, their position is dependent on the preceding content—rather than the dimensions of the container. Since different browsers seem to render the text content at different heights, I couldn't get the individual hover/active underlines to align consistently with the gray persistent underline. 

This would be fine if we just positioned them absolutely, but in that case, their widths must be explicitly defined. And since the tabs have different widths, they need to inherit width from the tab `li`s! I also considered using `::after` on the `li` instead of the `a` and `span`, but then you can't depend on hover/active states to switch styles.

blms:
@gissoo This one is ready for visual review, notes posted in PR description above!

gissoo:
@blms there is no way for me to approve the build – what would you like me to do about the percy pages?

gissoo:
@blms I tested the tabs on the test site – the click works well however the tab style still has not been implemented, correct? The bar is missing and it looks like there is a link style on these, because they both have underlines instead of the one that's active (like in the percy pages) – otherwise it works well!

blms:
> @blms I tested the tabs on the test site – the click works well however the tab style still has not been implemented, correct? The bar is missing and it looks like there is a link style on these, because they both have underlines instead of the one that's active (like in the percy pages) – otherwise it works well!

@gissoo That's right, it's not ready in QA yet. Sorry for being unclear there! For now, please just review what you can see on Percy (see Slack for messages about access). I'll tag you in the issue #398 once it's ready on the QA site to test interactions.

blms:
@rlskoeser I hadn't thought of just using `border-bottom` on the `a` tags! That works, except it doesn't allow for the "extra bit" of gray outline on either side of the tabs (i.e. the width of the `border-bottom` in the designs is more like 80% of the width of the tab).

And the focus style is still a bit weird—let me know if you have any ideas for what to do about that.

Here's what things look like (hover, then focus, shown on left tab) using the `border-bottom` solution:

Screen Shot 2022-01-10 at 12 07 29 PM
Screen Shot 2022-01-10 at 12 07 36 PM
 
Also set the tabs to be equally sized. You're right, there's no reason they shouldn't be!

@gissoo Are you OK with this appearance? There's no "extra bit" on either side of the tabs, and the focus style doesn't exactly match the designs.

rlskoeser:
This looks good to me @blms ! And the focus style seems ok.

I like the idea of making the links & underlines full width within the tabs — it seemed like when I was testing it before there was a dead zone between tabs that wasn't clickable, I imagine that wouldn't be the case any more? 

Interested to hear what @gissoo thinks.

blms:
@rlskoeser Right, there'd be no "dead zone" with this design!

gissoo:
@rlskoeser @blms I'm ok with the tab decision, that is fine! I'm not really ok with the focus styles but it's not a deal breaker, wouldn't prioritize spending time on it right now.

sitemedia/scss/components/_tabs.scss
```diff
@@ -2,35 +2,58 @@
...
-    }
-    // Center tab padding boost
-    li:nth-child(2) a,
-    li:nth-child(2) span {
-        @include breakpoints.for-tablet-landscape-up {
-            padding: spacing.$spacing-2xs spacing.$spacing-2xl;
+        text-decoration: none;
+        &:focus-visible {
+            outline: 0.1rem solid var(--tabs-focus);
+            outline-offset: -0.33rem;
```
rlskoeser, line 35: @blms I had an idea about using a pseudo-element to get the focus outline closer to the designs. I tried this out locally, see what you think.


```suggestion
      position: relative;
       /* use ::after for focus outline */
        &::after {
            content: "";
            width: 80%;
            height: 80%;
            position: absolute;
        }
        &:focus-visible {
            outline: none;
        }
        &:focus-visible::after {
            outline: 0.1rem solid var(--tabs-focus);
        }
```

blms: This is great!! Excellent solution, Rebecca.

blms:
Thanks @gissoo and @rlskoeser! Went ahead and approved the Percy build after Rebecca's latest changes, since Gissoo already approved the prior build. Merging now.

Issue #476, opened 2021-12-27
Unable to delete documents due to lack of permissions to delete log entry


richmanrachel:
**Describe the bug**
Marina created a test document in the regular admin site and was unable to delete it, getting this error message:
![image](https://user-images.githubusercontent.com/68028846/147511565-2f55f3f8-4496-4c1e-9023-ca83f79ef724.png)

Rachel then tried, and got the same error message. Rachel went into the Users section of the admin site and specifically added permission to delete log entries to her account, but still received the same error. 

**To reproduce**
1. Go to PGPID 35058
2. Click on Delete
3. See error

**Expected behavior**
A warning message appears, but Super Users/Content Admin should be able to delete.

**Device information**
 - Rachel was on MacBook Air 2021, and Chrome browser 
 - Site version [0.7.0]

blms:
Duplicate of #371

Issue #371, opened 2021-11-18
permissions error trying to delete a document because it wants to delete the associated log entry


rlskoeser:
## testing notes
go to the document admin list on the test site and choose one to delete
- [x] confirm the delete review screen doesn't complain about permissions
- [x] confirm the delete review screen includes language about suppressing and merging documents in preference to deleting
- [x] confirm you are actually able to delete the object
- [x] go to the log entries admin list and confirm there is a new log entry for the deletion

--- 
Error has been reported by both Alan and Marina

<img width="1054" alt="01C2D919-22C0-4F62-888C-802C9B2D2F47" src="https://user-images.githubusercontent.com/691231/142446334-37a8e56b-bb74-4644-9732-54ca90f0254a.png">

Normally we'll want to suppress or merge documents rather than deleting, but it should be possible for content editors to delete them when the need arises.

richmanrachel:
Draft for custom language for admin delete: 
Are you sure this document should be deleted? Please only delete documents created in error. Non-documentary sources should be suppressed. Duplicates should be merged. Thanks for keeping our database safe and clean!

richmanrachel:
@rlskoeser - Perfect! And we also confirmed that it doesn't screw with the existence of the fragment (tested with both T-S 20.117 and BL OR 5566B.23).

PR #478, opened 2022-01-03
#371, #476 - Allow admins to delete documents

blms:
## What this PR does
- Per #371:
  - Fixes a bug preventing admins from deleting documents by changing `Document.log_entries` from a `GenericRelation` to a property that returns a `QuerySet`

## Questions
- Besides whether this is overall a good approach, I'm also wondering if other relations need to be changed, such as `Document.footnotes`. As a `GenericRelation`, it also gets `on_delete=models.CASCADE`. Take a look at what happens when deleting a document:
  Screen Shot 2022-01-03 at 12 06 51 PM
   This could potentially be dangerous! Should we change the relations to do nothing on delete, or restrict the ability to do this at all?


rlskoeser (changes requested):
I would like to keep the log entries relationship as a generic foreign key but not delete log entries on document deletion — that way the log entries will preserve the record of the document being deleted. It looks like we may have to adjust the footnote model to make the content object nullable and not cascade on delete. (I did some quick googling and found some stack overflow solutions for this, but IDK if they are for current versions of django.)

But to your other questions — delete _should_ warn about deleting all the related content, it's appropriate for those other items to be deleted, and delete should be used sparingly. I think in most cases content editors should suppress documents rather than deleting them.

blms:
@rlskoeser As I understand it, we would need to use `GenericRelation` to manage the reverse of the `GenericForeignKey` relation—and that is [forced to use `models.CASCADE` on delete](https://github.com/django/django/blob/027f4c4ceb93c7f7b45fecf94b7db1defc5c4064/django/contrib/contenttypes/fields.py#L272-L304). Is there another way to manage the reverse of this relation? 

If we do use the property, there will still be `LogEntry` objects with `GenericForeignKey` relations to `Document`—but they would just be accessible via the `QuerySet`-returning property rather than the relation. Is there required functionality that couldn't be achieved this way?

rlskoeser:
> @rlskoeser As I understand it, we would need to use `GenericRelation` to manage the reverse of the `GenericForeignKey` relation—and that is [forced to use `models.CASCADE` on delete](https://github.com/django/django/blob/027f4c4ceb93c7f7b45fecf94b7db1defc5c4064/django/contrib/contenttypes/fields.py#L272-L304). Is there another way to manage the reverse of this relation?

According to this, you can set the content object and object id to allow null: https://stackoverflow.com/a/27912261 
(but it is a fairly old answer)

> If we do use the property, there will still be `LogEntry` objects with `GenericForeignKey` relations to `Document`—but they would just be accessible via the `QuerySet`-returning property rather than the relation. Is there required functionality that couldn't be achieved this way?

You can't filter document querysets based on a property.

rlskoeser:
Sorry, brain still shifting out of vacation mode — I see from your link that it is the generic relation causing the problem, and the log entry model is already defined to allow null. If we can get the functionality we need from a queryset property, then let's go ahead and use that solution. Do we need to update any code where we're assuming it's a generic relation manager and calling `.all()` ? Or I guess that still works since it returns a queryset?

geniza/corpus/models.py
```diff
@@ -409,7 +409,13 @@ class Document(ModelIndexable):
     )
 
     footnotes = GenericRelation(Footnote, related_query_name="document")
-    log_entries = GenericRelation(LogEntry, related_query_name="document")
+
+    @property
+    def log_entries(self):
+        return LogEntry.objects.filter(
+            object_id=self.id,
+            content_type_id=ContentType.objects.get_for_model(Document),
```
rlskoeser, line 417: Could we get content type id once somewhere so we don't have to load it every time?

Or here at least we could filter on the content object name and app name, instead of looking up the id every time.

blms:
> Do we need to update any code where we're assuming it's a generic relation manager and calling `.all()` ? Or I guess that still works since it returns a queryset?

Right! That was the idea. The only time that it needed to be changed was `self.log_entries.add(log_entry)`. I changed this to involve saving the existing log entry with the new data, but maybe it actually makes more sense to just create a new log entry, to prevent actually altering logs.

geniza/corpus/models.py
```diff
@@ -580,7 +586,6 @@ def items_to_index(cls):
             "tags",
             "languages",
             "footnotes",
-            "log_entries",
```
rlskoeser, line 583: I don't think this is only place we're prefetching log entries? We can live with the additional time for bulk indexing, but I'm concerned about the impact for the admin csv download.

rlskoeser:
> Right! That was the idea. The only time that it needed to be changed was `self.log_entries.add(log_entry)`. I changed this to involve saving the existing log entry with the new data, but maybe it actually makes more sense to just create a new log entry, to prevent actually altering logs.

The only time we're altering logs is when we merge records — it feels a little strange to edit history, but we do need to reassociate those records with the merged document so we know who worked on it when. (And we do also create a new log entry to document the merge.)

geniza/corpus/models.py
blms: LogEntry actually doesn't have a `content_object` field, it only stores the object ID and content type ID, so I'll do the former!

geniza/corpus/models.py
blms: You're right, that's the other place where it's used. I imagine it's for this:

https://github.com/Princeton-CDH/geniza/blob/39e5be7335b8479f762455a22ce3b12e8cdc13b1/geniza/corpus/admin.py#L325-L329

I can see how this might result in a lot of unnecessarily hitting the DB without prefetching `log_entries`. I thought about just doing a simple queryset before the loop to get all matching log entries for all these documents, but that still would result in additional queries on filtering. I read through [this](https://www.mattduck.com/2021-01-django-orm-result-cache.html) but nothing there seemed particularly helpful. Should we consider this a blocker?

rlskoeser: It's a blocker if it makes the admin download unusable (which I suspect it does); we'll need to test the performance. There is probably a way to customize the prefetching, or maybe we can use a filtered relation https://docs.djangoproject.com/en/3.2/ref/models/querysets/#filteredrelation-objects

blms: I'm stuck on both of these options. I can't seem to get a way to use any of `prefetch_related`, `select_related`, `prefetch_related_objects`, `Prefetch`, or `FilteredRelation` without needing to pass a field name for the `Document` model that relates it directly to `LogEntry`.

The best I can come up with so far is doing the following before running the loop over documents in the CSV export:
```py
doc_ids = queryset.values_list("id", flat=True)
prefetched_log_entries = LogEntry.objects.filter(
    object_id__in=doc_ids,
    content_type__app_label="corpus",
    content_type__model="document",
)
```

But I'm still blocked there because any additional filtering on `prefetched_log_entries` will still result in the same additional DB queries as without this.

rlskoeser:
@blms I had an idea about a different approach to this problem — what if we keep the generic relation on document, but add a `pre_delete` signal handler for document deletion that sets log entry content object id to null?

blms:
> @blms I had an idea about a different approach to this problem — what if we keep the generic relation on document, but add a `pre_delete` signal handler for document deletion that sets related log entry content object id to null?

I love it! I'll give that a try.

blms:
@rlskoeser About to head off for the day, but after a bit of quick work it seems that `pre_delete` is not actually sent before those relationships are sent. Looks like there's a [long-running discussion](https://code.djangoproject.com/ticket/6870) about this too—someone even mentions admins and `LogEntry`! It should be possible if we attach the signal to the specific request that's raising the validation error, though. I'll give that a try tomorrow.

blms:
@rlskoeser Actually, after sleeping on this, I'm still at a bit of a loss. If those were set null when the original request were sent (basically, a request to display the "confirm delete" dialog), and the user decided to say "No, I don't want to delete this document actually," then the LogEntries would be unnecessarily orphaned. We could connect them back to the document if the user actually clicks "No", but they could also simply navigate away or close the window. Tricky stuff! 🤔

rlskoeser:
@blms I've been thinking about our log entry problem too! It seems crazy that this is so hard. I came to the same conclusion you did — we can't get by the permissions check problem with the solution I proposed.

I had another idea, though — we could stick with the queryset property on the actual document model, and then create a document proxy model with the generic relation and use the proxy model for the queries where we need prefetching. (I did test yesterday and downloading the full set of documents as csv is unusably slow without the prefetching.)  What do you think? Want to give it a try?

blms:
@rlskoeser I'll give that a try!

blms:
@rlskoeser I think you solved it with that one! Great idea.

Two questions:

1. Should this also be used in the bulk indexing? I was thinking I could try it by just changing `docs = Document.items_to_index().filter(**doc_filter)` to `DocumentPrefetchableProxy.items_to_index().filter(**doc_filter)` but not sure if that will work with parasolr, etc.
2. Testing it myself it seems the CSV export performance is back to what it was beforehand. Can you test it as well to make sure?

rlskoeser:
@blms hooray! amazing. 

Yes, it would be great to use it in the bulk indexing as well — I think it should work because the proxy model queryset returns the regular model (doesn't it?).

I'll test the csv download.

rlskoeser:
The CSV download looks like it's back to what it was before.

rlskoeser (approved):

PR #479, opened 2022-01-03
460 show shelfmarks in doc detail

kmcelwee:
Example of a multi-shelfmark doc: PGPID 444

### DIFF
--- a/geniza/corpus/templates/corpus/document_detail.html
+++ b/geniza/corpus/templates/corpus/document_detail.html
@@ -48,6 +48,10 @@
                         {# link ? #}
                     {% endfor %}
                 {% endif %}
+                {% if document.fragments.count > 1 %}
+                    <dt class="secondary">{% translate 'Shelfmark' %}</dt>
+                    <dd class="secondary">{{ document.shelfmark }}</dd>
+                {% endif %}
                 {# Translators: Date document was first added to the PGP #}
                 <dt class="secondary">{% translate 'Input date' %}</dt>
                 <dd class="secondary">{{ document.log_entries.last.action_time.year }}</dd>
--- a/geniza/corpus/templates/corpus/document_detail.html
+++ b/geniza/corpus/templates/corpus/document_detail.html
@@ -31,6 +31,10 @@
             </h3>
             {# metadata #}
             <dl class="metadata">
+                {% if document.fragments.count > 1 %}
+                    <dt>{% translate 'Shelfmark' %}</dt>
+                    <dd>{{ document.shelfmark }}</dd>
+                {% endif %}
                 {% if document.digital_editions %}
                     <dt>
                         {# Translators: Editor label #}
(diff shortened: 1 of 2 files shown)


geniza/corpus/templates/corpus/document_detail.html
```diff
@@ -48,6 +48,10 @@ <h3 class="sr-only">
                         {# link ? #}
                     {% endfor %}
                 {% endif %}
+                {% if document.fragments.count > 1 %}
+                    <dt class="secondary">{% translate 'Shelfmark' %}</dt>
+                    <dd class="secondary">{{ document.shelfmark }}</dd>
+                {% endif %}
```
rlskoeser, line 54: per the linked designs shelfmark should be displayed before editor and should not be in the lighter gray color

geniza/corpus/tests/test_corpus_templates.py
```diff
@@ -136,6 +136,16 @@ def test_editors(self, client, document, source, twoauthor_source):
         response = client.get(document.get_absolute_url())
         assertContains(response, "Editors")
 
+    def test_multiple_shelfmarks(self, client, document, join):
+        """Document detail template should show multiple shelfmarks"""
+        response = client.get(document.get_absolute_url())
+        # TODO: This is fragile, consider something better
+        assertNotContains(response, "Shelfmark", html=True)
```
rlskoeser, line 143: you're not actually using an html match here, but that would be more robust — why not include the full `<dt>` tag?

geniza/corpus/tests/test_corpus_templates.py
```diff
@@ -136,6 +136,16 @@ def test_editors(self, client, document, source, twoauthor_source):
         response = client.get(document.get_absolute_url())
         assertContains(response, "Editors")
 
+    def test_multiple_shelfmarks(self, client, document, join):
```
rlskoeser, line 139: please add comments or separate into two tests to clarify that one _should_ display the shelfmark as a field and one should not

rlskoeser (changes requested):
logic looks fine, but placement and styles are not quite right and the test should be fixed as you noted in your TODO

geniza/corpus/templates/corpus/document_detail.html
kmcelwee: 1a3f626 ✅

geniza/corpus/tests/test_corpus_templates.py
kmcelwee: ff40fa023ef5d9811b1622c51b2692d64423e05a ✅

geniza/corpus/tests/test_corpus_templates.py
kmcelwee: 9f8c3ee ✅

rlskoeser (approved):

PR #480, opened 2022-01-03
374 configure wagtail sitemaps

kmcelwee:
**Questions**
* The ticket mentions we can get “speed” from solr? I can't find that as a sitemaps variable.
* Why go through solr for `last_modified` instead of the `last_modified` attribute on the document object?
* For document scholarship, should `lastmod` be the last time the document was modified or last time the footnotes were updated?
* Should `DocumentScholarshipSitemap` be in the footnotes app? It may be more appropriate, but it feels like overkill.
* In Django 3.2, there's [a `languages` attribute](https://docs.djangoproject.com/en/4.0/ref/contrib/sitemaps/#django.contrib.sitemaps.Sitemap.languages) Is it hard to do that on our own?


rlskoeser:
Answers to your questions:

> * The ticket mentions we can get “speed” from solr? I can't find that as a sitemaps variable.

When I wrote that we should pull from solr "for speed" I meant it would be faster than hitting the db.

> * Why go through solr for `last_modified` instead of the `last_modified` attribute on the document object?

Last modified on the object only gets updated when the object itself is directly updated; it doesn't take into account all the other changes that could affect the content displayed on the document detail page. We reindex the object in Solr when those other things change, and also when we make other programmatic changes to the site, so we use it as a proxy for the "real" last modified date. (Better to err on the side of saying it's modified when it isn't than the reverse.)

> * For document scholarship, should `lastmod` be the last time the document was modified or last time the footnotes were updated?

It would need to be the last time any of the footnotes were modified _or_ when any of them were moved to a different document or deleted; we don't have an easy way to track that, so best to just use the last modified for the document in solr.

> * Should `DocumentScholarshipSitemap` be in the footnotes app? It may be more appropriate, but it feels like overkill.

It should be in the corpus app because that's where the view is.

> * In Django 3.2, there's [a `languages` attribute](https://docs.djangoproject.com/en/4.0/ref/contrib/sitemaps/#django.contrib.sitemaps.Sitemap.languages) Is it hard to do that on our own?

I'm not sure what you mean about doing this on our own. We have languages defined; do we just need to turn on i18n for the sitemap?

geniza/corpus/sitemaps.py
```diff
@@ -0,0 +1,22 @@
+from django.contrib.sitemaps import Sitemap
+
+from geniza.corpus.models import Document
+
+
+class DocumentSitemap(Sitemap):
+    def items(self):
+        return Document.objects.filter(status=Document.PUBLIC)
```
rlskoeser, line 8: this should use the solr queryset rather than the django model queryset

geniza/corpus/sitemaps.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def lastmod(self, obj):
+        return obj.last_modified
+
+
+class DocumentScholarshipSitemap(Sitemap):
+    def items(self):
+        return Document.objects.filter(status=Document.PUBLIC, footnotes__isnull=False)
```
rlskoeser, line 16: Do we 404 on document scholarship records pages that don't have any footnotes? They should only be excluded if the page doesn't exist. (But we could set the priority very low if they don't have footnotes.)

geniza/corpus/tests/test_sitemaps.py
```diff
@@ -0,0 +1,40 @@
+from geniza.corpus.models import Document
+from geniza.corpus.sitemaps import DocumentScholarshipSitemap, DocumentSitemap
+
+
+class TestDocumentSitemap:
+    def test_items(self, document):
+        # Ensure that documents are supressed if they aren't public
+        assert document.status == Document.PUBLIC
+        sitemap = DocumentSitemap()
+        assert len(sitemap.items()) == 1
```
rlskoeser, line 10: would prefer a more robust test (object / pk in list or not in list)

geniza/corpus/tests/test_sitemaps.py
```diff
@@ -0,0 +1,40 @@
...
+
+        document.status = Document.SUPPRESSED
+        document.save()
+        sitemap = DocumentSitemap()
+        assert len(sitemap.items()) == 0
+
```
rlskoeser, line 16: also test last modified function

geniza/corpus/tests/test_sitemaps.py
```diff
@@ -0,0 +1,40 @@
...
+        document.footnotes.all().delete()
+        sitemap = DocumentScholarshipSitemap()
+        assert len(sitemap.items()) == 0
+
+    def test_location(self, document, footnote):
+        document.footnotes.add(footnote)
+        document.save()
+        sitemap = DocumentScholarshipSitemap()
+        assert sitemap.location(sitemap.items()[0]).endswith("scholarship/")
```
rlskoeser, line 40: please test with `reverse` and check the actual url

also test the last modified function, and priority if you add it based on zero footnotes

rlskoeser (changes requested):

geniza/corpus/sitemaps.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def location(self, obj):
+        return obj.get_absolute_url() + "scholarship/"
```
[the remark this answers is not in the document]
kmcelwee: 05477454005b7a2c96069f510a98821603925bd9 ✅

geniza/urls.py
```diff
@@ -13,22 +13,51 @@
...
+    path(
+        "robots.txt",
+        TemplateView.as_view(template_name="robots.txt", content_type="text/plain"),
+    ),
     # redirect homepage to admin site for now
     path("admin/", admin.site.urls),
     path("accounts/", include("pucas.cas_urls")),
     path("i18n/", include("django.conf.urls.i18n")),
     path("taggit/", include("taggit_selectize.urls")),
+    # NOTE: Sitemap most occur before wagtail urls
```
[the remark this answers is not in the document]
kmcelwee: 95ac638e8b5e5af6efc5d7be3fdcd98c07d34105 ✅

geniza/urls.py
```diff
@@ -13,22 +13,51 @@
...
 from wagtail.admin import urls as wagtailadmin_urls
+from wagtail.contrib.sitemaps import Sitemap as WagtailSitemap
+from wagtail.contrib.sitemaps import views as sitemap_views
 from wagtail.core import urls as wagtail_urls
 
+from geniza.corpus.sitemaps import DocumentScholarshipSitemap, DocumentSitemap
+
+SITEMAPS = {
+    "pages": WagtailSitemap,
+    "document-detail": DocumentSitemap,
```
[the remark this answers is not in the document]
kmcelwee: 44b0797f6f8ffe7953ece0c8d98e63e5ea1c0865 ✅

geniza/corpus/tests/test_sitemaps.py
kmcelwee: 3ca83b2a66283d30c14141576b06030fd19aed4a ✅

geniza/corpus/tests/test_sitemaps.py
kmcelwee: 33f34caf4ae834c00486002fce209754a99ccee5 ✅

geniza/corpus/tests/test_sitemaps.py
kmcelwee: f37959c90b06931a1f954140fc1fbf2c34891ebd ✅

geniza/corpus/sitemaps.py
kmcelwee: Yes there is currently a 404 when there are no footnotes:
https://test-geniza.cdh.princeton.edu/en/documents/516/
https://test-geniza.cdh.princeton.edu/en/documents/516/scholarship/

rlskoeser: Ok, great — I missed where in the view logic it was excluding documents with no footnotes. So this logic is sound. (But add a comment to make it explicit)

kmcelwee: fedabc082b8e3ecaaa6b10a9667cb6adfeee28bf ✅

geniza/corpus/sitemaps.py
```diff
@@ -0,0 +1,49 @@
...
+
+class DocumentScholarshipSitemap(Sitemap):
+    def items(self):
+        # Only return documents with footnotes. A document scholarship page returns
+        #  a 404 if there are no footnotes.
+        return (
+            DocumentSolrQuerySet()
+            .filter(status="Public", footnotes__isnull=False)
+            .only("pgpid", "last_modified")
+        )
```
kmcelwee, line 43: @rlskoeser Tests are failing because `footnotes__isnull` is not available as a queryset filter option. What do you suggest we do?

rlskoeser: @kmcelwee you're in solr land now! we're indexing counts for scholarship records already, so you should be able to filter on `scholarship_count` ; parasolr has an existing range filter already, so might look like `scholarship_count__range=(1, None)` (the `None` indicates an open-ended range)

kmcelwee: ✅ fc93b90c198ec1c401292d0d23093e20f429bd69

geniza/urls.py
```diff
@@ -13,22 +13,51 @@
...
+from django.views.generic.base import RedirectView, TemplateView
 from wagtail.admin import urls as wagtailadmin_urls
+from wagtail.contrib.sitemaps import Sitemap as WagtailSitemap
+from wagtail.contrib.sitemaps import views as sitemap_views
 from wagtail.core import urls as wagtail_urls
 
+from geniza.corpus.sitemaps import DocumentScholarshipSitemap, DocumentSitemap
+
+SITEMAPS = {
+    "pages": WagtailSitemap,
```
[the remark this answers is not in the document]
kmcelwee: When the content is first created, it's `None`, but after restarting the server, the links appear as expected. I added more content through the wagtail UI, and the sitemap updated properly without having to restart. I think it may have to do with the automatic generation of web pages.

This is an issue that wagtail hasn't addressed for 6 years, looking at this issue: https://github.com/wagtail/wagtail/issues/1196#issuecomment-890365995. In `geniza/venv/lib/python3.7/site-packages/wagtail/contrib/sitemaps/sitemap_generator.py` if I replace:
```diff
def location(self, obj):
- return obj.get_full_url(self.request)
+ return obj.localized.get_full_url(self.request)
```
...we wouldn't need to restart the server. Let me know what you think!

kmcelwee:
@rlskoeser 

> do we just need to turn on i18n for the sitemap?

This looks like a longstanding issue https://github.com/wagtail/wagtail/issues/1196:

> Google requires a fairly specific output for the XML sitemap on multi-lingual sites https://support.google.com/webmasters/answer/2620865?hl=en
>
> Currently Wagtail isn't able to accommodate this.

A multi-lingual sitemap seems to be a bit involved, do we want to include it in the MVP? Or wait until the next release with Hebrew?

geniza/corpus/tests/test_sitemaps.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def test_lastmod(self):
+        assert DocumentSitemap().lastmod(
+            {"last_modified": "2020-05-12T15:46:20.341Z"}
+        ) == date(2020, 5, 12)
+
+
+class TestDocumentScholarshipSitemap:
+    def test_items(self, document, source, suppressed_document, fragment):
+        document_with_footnote = make_document(fragment, id=41)
```
kmcelwee, line 49: @rlskoeser 
I've been reading up on django testing, and structuring our tests with object factories seems to make the most sense because of all their dependencies. Have we tried [Factory Boy](https://factoryboy.readthedocs.io/en/stable/orms.html) before? You set defaults and then overwrite them for your specific use case. I know it's a bit late to introduce something like this, but I feel like half our tests are building objects.

rlskoeser: @kmcelwee I'm not familiar with this library, but I don't want to change approach right now, and don't want to introduce multiple new dependencies (I see there is a separate pytest plugin to connect pytest with factory boy).

rlskoeser:
> A multi-lingual sitemap seems to be a bit involved, do we want to include it in the MVP? Or wait until the next release with Hebrew?

Thanks for investigating. It makes sense to me to put that off until we add the Hebrew version of the site, as long as the sitemap urls are accurate for the english version of the site. It sounds like we may need to use the built in django sitemaps instead of wagtail. Please make an issue to document this so we don't lose track.

kmcelwee:
Great! Then we're all set. I addressed everything you asked, but there were quite a few changes though, so you'll probably need to do a proper review. I also brought in #492 to this PR as well.

geniza/corpus/sitemaps.py
```diff
@@ -0,0 +1,49 @@
...
+from geniza.corpus.models import Document
+from geniza.corpus.solr_queryset import DocumentSolrQuerySet
+
+
+def solr_timestamp_to_date(timestamp):
+    """Convert solr isoformat date time string to python date."""
+    # format: 2020-05-12T15:46:20.341Z
+    # django sitemap only includes date, so strip off time
+    yearmonthday = timestamp.split("T")[0]
+    return date(*[int(val) for val in yearmonthday.split("-")])
```
rlskoeser, line 15: parasolr includes a `solr_timestamp_to_datetime` utility method now, let's just use that and then convert to date (I'm guessing it wasn't available yet in whatever code you're adapting from)

geniza/corpus/sitemaps.py
```diff
@@ -0,0 +1,49 @@
...
+        )
+
+    def lastmod(self, obj):
+        return solr_timestamp_to_date(obj["last_modified"])
+
+    def location(self, obj):
+        return reverse("corpus:document", args=[obj["pgpid"]])
+
+
+class DocumentScholarshipSitemap(Sitemap):
```
rlskoeser, line 35: I would extend `DocumentSitemap` and then add the scholarship count filter to the parent queryset.

Location could even be inherited if you set the view name on the class.

geniza/corpus/tests/conftest.py
```diff
@@ -101,3 +108,21 @@ def document(db, fragment):
...
+        location="p.1",
+        doc_relation=Footnote.EDITION,
+    )
+
+
+@pytest.fixture
+def suppressed_document(fragment):
+    document = make_document(fragment, id=400)
+    document.status = Document.SUPPRESSED
+    document.save()
```
rlskoeser, line 127: Why do you need to use the fixture? couldn't you use an empty document?

geniza/corpus/tests/test_corpus_views.py
```diff
@@ -203,6 +203,24 @@ def test_pgp_metadata_for_old_site():
 
 
 class TestDocumentSearchView:
+    def test_document_suppression(self, document, suppressed_document):
+        SolrClient().update.index(
+            [
+                document.index_data(),  # no scholarship records
+                suppressed_document.index_data(),  # one scholarship record
+            ],
```
rlskoeser, line 211: You can actually just pass the documents, parasolr will check for and call the `index_data` method

geniza/corpus/tests/test_corpus_views.py
```diff
@@ -203,6 +203,24 @@ def test_pgp_metadata_for_old_site():
...
+        )
+        docsearch_view = DocumentSearchView()
+        docsearch_view.request = Mock()
+
+        # keyword search param
+        docsearch_view.request.GET = {"q": ""}
+        qs = docsearch_view.get_queryset()
+        assert qs.count() == 1
+        assert document.id in [obj["pgpid"] for obj in qs]
+        assert suppressed_document.id not in [obj["pgpid"] for obj in qs]
```
rlskoeser, line 222: When I'm using a list like this more that once, I find it more readable to generate the list first. Something like this:

```suggestion
        resulting_pgpids = [obj["pgpid"] for obj in qs]
        assert document.id in resulting_pgpids
        assert suppressed_document.id not in resulting_pgpids
```

geniza/corpus/tests/test_sitemaps.py
```diff
@@ -0,0 +1,81 @@
...
+            ],
+            commit=True,
+        )
+
+        # Ensure that only public documents with footnotes are returned
+        sitemap = DocumentScholarshipSitemap()
+        assert document_with_footnote.id in [obj["pgpid"] for obj in sitemap.items()]
+        assert suppressed_document.id not in [obj["pgpid"] for obj in sitemap.items()]
+        # TEST FAILURE: How do you filter by footnotes__isnull for a solr queryset?
+        assert document.id not in [obj["pgpid"] for obj in sitemap.items()]
```
rlskoeser, line 71: Is this still a problem?

geniza/corpus/views.py
```diff
@@ -65,7 +65,7 @@ def get_form_kwargs(self):
         return kwargs
 
     def get_queryset(self):
-        documents = DocumentSolrQuerySet()
+        documents = DocumentSolrQuerySet().filter(status="Public")
```
rlskoeser, line 68: We could make a `public` method on the local queryset similar to the wagtail `live` filter that applies this filter for us, since we're using it in multiple places. But I'm not sure how much it gets us.

geniza/corpus/sitemaps.py
```diff
@@ -0,0 +1,49 @@
...
+    return date(*[int(val) for val in yearmonthday.split("-")])
+
+
+class DocumentSitemap(Sitemap):
+    def items(self):
+        return (
+            DocumentSolrQuerySet()
+            .filter(
+                status="Public"
+            )  # ?: It's not saved as "Document.PUBLIC" which is "P"
```
rlskoeser, line 24: Please add a constant to `Document` for this value — maybe `Document.PUBLIC_LABEL` ? and use it everywhere you have the string value now (status field choices, here, and search view, ...). 

Unless you think we should just index the P/S codes instead of the labels; in which case, you should update the index data and use the existing constant here. (I don't _think_ we're currently using it anywhere else, so this is probably fine — but if we ever wanted to use them for facets it wouldn't be very meaningful!)

geniza/corpus/sitemaps.py
```diff
@@ -0,0 +1,49 @@
...
+
+def solr_timestamp_to_date(timestamp):
+    """Convert solr isoformat date time string to python date."""
+    # format: 2020-05-12T15:46:20.341Z
+    # django sitemap only includes date, so strip off time
+    yearmonthday = timestamp.split("T")[0]
+    return date(*[int(val) for val in yearmonthday.split("-")])
+
+
+class DocumentSitemap(Sitemap):
```
rlskoeser, line 18: Did you try setting `i18n = True` and `alternates = True` ? It would be nice if we didn't have to change anything with our document sitemaps when we're ready to enable other languages.

geniza/corpus/sitemaps.py
```diff
@@ -0,0 +1,49 @@
...
+    def items(self):
+        return (
+            DocumentSolrQuerySet()
+            .filter(
+                status="Public"
+            )  # ?: It's not saved as "Document.PUBLIC" which is "P"
+            .only("last_modified", "pgpid")
+        )
```
rlskoeser, line 26: I think you're not getting all of the results — did you check this?

Pretty sure this is  due to a quirk of solr / limitation of parasolr that I haven't figured out the right way to resolve yet (😬 ideas welcome💡 ). By default, Solr only returns 10 rows! When we're iterating over parasolr querysets the logic works fine, but if you just get the first / default set returned by the queryset you only get the default ten. The workaround I've used elsewhere is to just set a ridiculously large maximum number of rows - see e.g. https://github.com/Princeton-CDH/geniza/blob/main/geniza/corpus/admin.py#L275-L280

geniza/corpus/tests/conftest.py
```diff
@@ -32,10 +39,10 @@ def make_multifragment():
     )
 
 
-def make_document(fragment):
+def make_document(fragment, id=3951):
```
rlskoeser, line 42: please document the parameter if we keep it

geniza/urls.py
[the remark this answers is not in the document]
rlskoeser: is this resolved now ?

rlskoeser (changes requested):
The only thing that's not working correctly here is that you're not getting all of the documents.

I have several other suggestions & comments for you, but the rest of the logic looks reasonable.

geniza/urls.py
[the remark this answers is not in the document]
kmcelwee: Yes!

geniza/corpus/sitemaps.py
kmcelwee: 8cbc9204b6d6cf7f291826b228163520bbf78746 ✅

geniza/corpus/tests/test_corpus_views.py
kmcelwee: 48387c8ea314293018515303bc09478485505a38 ✅

kmcelwee: It looks like there's a problem with indexing the objects:
```
TypeError: Object of type Document is not JSON serializable
```
Let me know if you'd like me to open a ticket on parasolr

rlskoeser: oh, my mistake — you're using the update index method and not the index items on the indexable class; I wasn't reading closely enough because I just somehow assumed you were doing what I would do, sorry 🤦‍♀️ 

You can use this `Document.index_items([document, suppressed_document])`

geniza/corpus/views.py
kmcelwee: It used twice. Maybe on the third time?

geniza/corpus/sitemaps.py
kmcelwee: Both `sitemap-documents.xml` and `sitemap-document-scholarship.xml` are showing thousands of results for me locally. I assume when you tested locally you got something different? I'm using Solr 8, are you? Should I just add the `rows=100000` param regardless?

geniza/corpus/tests/test_corpus_views.py
kmcelwee: aaaaand I was unable to get this to work either 😅 querying solr after this returned an empty queryset.

geniza/corpus/sitemaps.py
kmcelwee: Looks like alternates isn’t implemented until Django 3.2? When I set it to true, I got a strange plain text file locally. [The sitemap framework | Django documentation | Django](https://docs.djangoproject.com/en/4.0/ref/contrib/sitemaps/#django.contrib.sitemaps.Sitemap.alternates)

I turned on `i18n`. As I explained elsewhere, I was getting thousands of results until I turned that on (?!) so after setting `i18n = True` I added the `.get_results(rows=100000)`

rlskoeser: We're on django 3.2; I don't know why the sitemap renders as text instead of xml when we turn on the alternates behavior, the content type still looks correct. I guess we should leave it off for now since it doesn't seem to work correctly.

PR #481, opened 2022-01-03
371 Create migration to update permissions

kmcelwee:
* I just realized that this is probably best fixed by changing "log entry" permissions by hand in the admin UI, correct?
* You said "content editors" but because content editors do not currently have document delete privileges, I assume you meant content admins
* I can't get this to work even through the admin UI. I think it has something to do with [Permissions Caching](https://docs.djangoproject.com/en/4.0/topics/auth/default/#permission-caching) but restarting the server and changing browsers isn't fixing it. I'm kinda stuck.


blms:
Hey Kevin, slacked you as well but my proposed solution for this is to change the relation of `LogEntry` on `Document`: https://github.com/Princeton-CDH/geniza/pull/478

It seems to work fine, though I raised a caveat that this may actually be something we want to restrict in the first place (or alter other relations, such as with footnotes).

kmcelwee:
Ugh thankfully I didn't spend too much time on this, thanks @blms 🤦

Issue #467, opened 2021-12-22
As a front-end user, I want to be able to switch between dark and light mode manually with a toggle or button.


blms:
## Testing/review notes:

- [x] Visit https://test-geniza.cdh.princeton.edu/ and confirm that your browser picks up your system's setting for light or dark mode.
- [x] In the top right corner, click on the switch with the sun and moon icon, and confirm that you can toggle back and forth between light and dark mode.
- [x] after changing the light/dark mode from the default site, click around the site a bit and confirm that your choice stays selected

This also is just a placeholder design for now, so design review may be additionally needed.

## dev notes

Attempted solutions (to add dark mode snapshots to Percy builds) that did not work:
- [x] Percy-specific CSS media query
- [x] Enable `force-dark-mode` argument in Selenium `ChromeOptions`

Solutions to try:
- [x] Dark/light mode switcher in UI manually clicked by Selenium

richmanrachel:
It all works and looks awesome! The only odd thing to me is that there are three different icons that all do the same thing: 
![image](https://user-images.githubusercontent.com/68028846/148982935-7fd32c5e-82ca-45d2-9261-fb2ca08f290a.png)
@rlskoeser / @gissoo is this how it's supposed to be?

blms:
@richmanrachel That's a good point, I see how that's confusing! It's all all meant to be one big toggle, but maybe it makes more sense if the sun always switches to light, and the moon always switches to dark.

rlskoeser:
It's a placeholder design because we needed the functionality for testing — @gissoo interested in your input on this. (I can't remember if we asked to put this on your list of design tasks. Is there an existing PU design system component we could adapt by any chance?)

richmanrachel:
@rlskoeser and @blms - great! Should I close then because functionality works, or keep it open for Gissoo?

rlskoeser:
@richmanrachel my preference would be to close it and revisit design later — is this usable enough for MVP?

richmanrachel:
@rlskoeser - okay yes. Thanks!

PR #484, opened 2022-01-04
#467 - Add toggle for light/dark mode

blms:
## What this PR does
- Per #467:
  - Converts the themes from SASS variables to pure CSS custom properties, so that they can be conditionally redefined depending on the body class (see "[using custom properties](https://css-tricks.com/a-complete-guide-to-dark-mode-on-the-web/#using-custom-properties)")
  - Creates dark and light mode body classes
  - Uses system setting for dark/light mode initially
  - Adds toggle switch (`<input type="checkbox">`) to UI for overriding system theme, using `localStorage` to store the choice, and body class to change display
  - Uses the toggle switch programmatically (setting `localStorage`) to create a set of dark mode Percy screenshots


geniza/common/management/commands/visual_tests.py
```diff
@@ -20,16 +20,22 @@ def get_browser(self):
         options.add_argument("--headless")
         return webdriver.Chrome(options=options)
 
-    def take_snapshots(self, browser):
+    def take_snapshots(self, browser, dark_mode=False):
```
rlskoeser, line 23: I noticed in the percy docs that [percy script](https://docs.percy.io/docs/percyscript) has been deprecated and we should switch to using the [snapshot command](https://docs.percy.io/docs/cli-snapshot) — we don't need to change it now necessarily, but you are making a lot of changes and I wonder if it would simplify anything here.

I'm ok with your approach, but seeing this makes me think we should switch sooner than later so we don't have to reimplement custom logic twice.

geniza/common/management/commands/visual_tests.py
```diff
@@ -40,50 +46,51 @@ def take_snapshots(self, browser):
...
-            "Mobile menu",
-            percy_css="ul#menu { left: 0 !important; transition: none !important; }",
+            browser, "Mobile menu%s" % dark_mode_str, percy_css=mobile_menu_css
         )
 
         # about submenu open on both desktop and mobile
         browser.get("http://localhost:8000/en/documents/2532/#menu")
         # open about menu
         browser.find_element_by_id("open-about-menu").send_keys(Keys.ENTER)
+        about_menu_css = "ul#about-menu { left: 0 !important; transition: none !important; } @media (min-width: 900px) { ul#about-menu { left: auto !important; } }"
```
rlskoeser, line 76: remind me, why do we need custom css for this?

geniza/common/management/commands/visual_tests.py
```diff
@@ -40,50 +46,51 @@ def take_snapshots(self, browser):
...
-        # percy_snapshot(browser, "500 Page")
+        # percy_snapshot(browser, "500 Page%s" % dark_mode_str)
 
     def handle(self, *args, **options):
         # spin up browser and take snapshots; shut down when finished
         browser = self.get_browser()
+        # light mode
         self.take_snapshots(browser)
+        # dark mode
+        self.take_snapshots(browser, dark_mode=True)
```
rlskoeser, line 95: do we need _every_ snapshot in both dark and light mode? if the layouts and fonts are the same, how much do we gain from having everything available in both? 

@gissoo maybe this is a question for you — do you want/need to review every page in both dark and light mode?

geniza/templates/snippets/theme_toggle.html
```diff
@@ -0,0 +1,12 @@
+{% load i18n static webpack_loader %}
+
+{% block extrascript %}
+    <script defer src="{% static 'js/theme-toggle.js' %}" nonce="{{ request.csp_nonce }}"></script>
+{% endblock %}
+
+<label id="theme-toggle" for="theme-toggle-checkbox" tabindex="0">
+    {# Translators: label for dark mode/light mode theme switcher #}
+    <span class="sr-only">{% translate "Enable dark mode" %}</span>
```
rlskoeser, line 9: that probably shouldn't be the default text, since depending on user preferences it could start in dark mode; maybe something more general, like switch between dark and light mode?

rlskoeser (approved):
I like the css variables solution.

Functionality seems good. I did get one flash of dark to light page after I toggled and then clicked into a second page, but couldn't get it to repeat after that.

geniza/common/management/commands/visual_tests.py
blms: Totally agree, I wasn't aware. I'll make a chore for it, since I'd like to get this one out first.

rlskoeser: That works for me.

geniza/common/management/commands/visual_tests.py
blms: Added some comments to explain this:

```python
# custom CSS to ensure that on mobile, the menu transition is disabled and the menu
# is in the correct position
mobile_menu_css = "ul#menu { left: 0 !important; transition: none !important; }"
```

```python
# custom CSS to ensure that on mobile, the about menu transition is disabled and the menu
# is in the correct position; and that on desktop, that override does not impact its position
about_menu_css = "ul#about-menu { left: 0 !important; transition: none !important; } @media (min-width: 900px) { ul#about-menu { left: auto !important; } }"
```

rlskoeser: oh right! the transitions. I don't love that you're overriding positioning as well, but I guess we can live with it. Thanks for adding comments.

geniza/templates/snippets/theme_toggle.html
blms: I was wondering about this, and that was actually similar to my original text for this. But thinking through the inputs, since it's a checkbox, I wondered if someone using a screen reader might understand "Checkbox labeled 'Enable dark mode,' checked" better than "Checkbox labeled 'Switch between light and dark mode,' checked". But I'm not sure if my thought process makes sense when you consider what the actual user input looks like in the UI.

rlskoeser: Ah, got it. Well, this seems ok for now — I think we'll get design input from Gissoo on it at some point, and can refine it then.

geniza/common/management/commands/visual_tests.py
blms: I don't love it either! 😆 I think the reason I did this is that the initial position for those menus is `+/-100vh` when you visit `#menu` or `#about-menu`, so they'll appear offscreen when Percy takes the screenshot. The transition moves them over, but since we disable the transition for Percy, we also have to move them manually. There's probably a better way to do this, though…

rlskoeser: ok. well, the comment should help for now until we can figure out a better solution.

PR #485, opened 2022-01-04
Implement django-taggit multi-word bugfix

kmcelwee:
Django-taggit allows for multi-word tags, but `TaggableManager` does not encourage users to input them correctly 🥴 Should I open a ticket on https://github.com/jazzband/django-taggit?

### DIFF
--- a/geniza/common/utils.py
+++ b/geniza/common/utils.py
@@ -46,5 +46,6 @@ def custom_tag_string(tag_string):
     This is configured in settings with TAGGIT_TAGS_FROM_STRING.
     """
     # Stack overflow solution: https://stackoverflow.com/questions/30513783/django-taggit-how-to-allow-multi-word-tags
+    # Our github issue for taggit: https://github.com/jazzband/django-taggit/issues/783
 
     return [t.strip(' "') for t in tag_string.split(",") if t.strip(' "')]
--- a/geniza/common/utils.py
+++ b/geniza/common/utils.py
@@ -36,15 +36,12 @@ def absolutize_url(local_url, request=None):
 
 def custom_tag_string(tag_string):
     """
-    A custom tag string parser for taggit so that we can have multi-word
-    tags. Django-taggit allows for multi-word tags, but TaggableManager does not
-    encourage users to input them correctly.
+    Custom tag parsing for taggit to better support multi-word tags.
 
-    TaggableManager does not have an intuitive string output. If there are
-    no tags, it will return a comma-delimited string. It will wrap existing
-    tags in quotes. So if a user wants to add a tag "Arabic script" to a document
-    with the existing tag "fiscal document" it will send the value:
-    `"fiscal document", Arabic script` to be parsed.
+    Expected parsing:
+    - 'legal document' -> ["legal document"]
+    - '"fiscal document", Arabic script' -> ["fiscal document", "Arabic script"]
+    - '"fiscal document", "Arabic script"' -> ["fiscal document", "Arabic script"]
 
     This is configured in settings with TAGGIT_TAGS_FROM_STRING.
     """
--- a/geniza/common/utils.py
+++ b/geniza/common/utils.py
@@ -48,8 +48,6 @@ def custom_tag_string(tag_string: str) -> list:
 
     This is configured in settings with TAGGIT_TAGS_FROM_STRING.
     """
-    if not tag_string:
-        return []
-    elif "," not in tag_string:
-        return [tag_string.strip('"')]
-    return [t.strip(' "') for t in tag_string.split(",")]
+    # Stack overflow solution: https://stackoverflow.com/questions/30513783/django-taggit-how-to-allow-multi-word-tags
+
+    return [t.strip(' "') for t in tag_string.split(",") if t.strip(' "')]
--- a/geniza/common/utils.py
+++ b/geniza/common/utils.py
@@ -32,3 +32,24 @@ def absolutize_url(local_url, request=None):
         root = root.rstrip("/")
 
     return root + local_url
+
+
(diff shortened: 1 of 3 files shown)


geniza/common/utils.py
```diff
@@ -32,3 +32,24 @@ def absolutize_url(local_url, request=None):
...
+    tags in quotes. So if a user wants to add a tag "Arabic script" to a document
+    with the existing tag "fiscal document" it will send the value:
+    `"fiscal document", Arabic script` to be parsed.
+
+    This is configured in settings with TAGGIT_TAGS_FROM_STRING.
+    """
+    if not tag_string:
+        return []
+    elif "," not in tag_string:
+        return [tag_string.strip('"')]
```
rlskoeser, line 54: doesn't need to be a separate case

geniza/common/utils.py
```diff
@@ -32,3 +32,24 @@ def absolutize_url(local_url, request=None):
...
+    return [t.strip(' "') for t in tag_string.split(",")]
```
rlskoeser, line 55: do the quotes need to be stripped? that isn't in the stack overflow example, and it looks like using quotes actually is a work-around solution that works without the custom parser

geniza/common/utils.py
```diff
@@ -32,3 +32,24 @@ def absolutize_url(local_url, request=None):
         root = root.rstrip("/")
 
     return root + local_url
+
+
+def custom_tag_string(tag_string: str) -> list:
```
rlskoeser, line 37: please link to the stack overflow answer if that's what this is based on

rlskoeser: also, no type annotations in this codebase (want to use them consistently where we do use them)

geniza/common/utils.py
```diff
@@ -32,3 +32,24 @@ def absolutize_url(local_url, request=None):
...
+    """
+    A custom tag string parser for taggit so that we can have multi-word
+    tags. Django-taggit allows for multi-word tags, but TaggableManager does not
+    encourage users to input them correctly.
+
+    TaggableManager does not have an intuitive string output. If there are
+    no tags, it will return a comma-delimited string. It will wrap existing
+    tags in quotes. So if a user wants to add a tag "Arabic script" to a document
+    with the existing tag "fiscal document" it will send the value:
+    `"fiscal document", Arabic script` to be parsed.
```
rlskoeser, line 47: please shorten this and just document what kinds of input we get and need to send

geniza/settings/components/base.py
```diff
@@ -291,3 +291,6 @@
 
 # default font base url
 FONT_URL_PREFIX = "/static/fonts/"
+
+# Taggit customization
+TAGGIT_TAGS_FROM_STRING = "geniza.common.utils.custom_tag_string"
```
rlskoeser, line 296: Please also turn on case-insensitive tag searching (I thought there was an issue for this but now I can't find it!)

```suggestion
TAGGIT_TAGS_FROM_STRING = "geniza.common.utils.custom_tag_string"
TAGGIT_CASE_INSENSITIVE = True 
```

rlskoeser (approved):
Solution looks good and can be merged when you address the minor cleanup I've requested. (No need for second review)

> Django-taggit allows for multi-word tags, but TaggableManager does not encourage users to input them correctly 🥴 Should I open a ticket on https://github.com/jazzband/django-taggit?

I'm not sure I understand what you mean about this. Am I understanding correctly that the javascript used in the admin form doesn't add quotes when you use a comma if it isn't an existing tag? That does sound like a bug to me, so please go ahead and open an issue if you can't find an existing one that documents it. If you find an existing issue with the same problem, or if you create one, then add a link to it in this code so we'll be able to check what the resolution is.

geniza/common/utils.py
kmcelwee: bf36e95f57c9d1989786354e407e57572a9a80e5 ✅

geniza/common/utils.py
kmcelwee: 472aa512ef32a24f40ff026c969a16ace4379ba0 ✅

geniza/common/utils.py
kmcelwee: 7978dd46a51f91d31bcc068f29ffc5ccdee61bf5 ✅

geniza/common/utils.py
kmcelwee: The quotes need to be stripped, otherwise the tag would include quotation marks. I'm pretty sure the SO solution is wrong. Quotation marks are removed in the [`_parse_tags` function we're overwriting.](https://github.com/jazzband/django-taggit/blob/7f0e159e01f69a7db61e3c6dcd92989d152d6a0d/taggit/utils.py#L6)

kmcelwee:
> I'm not sure I understand what you mean about this. Am I understanding correctly that the javascript used in the admin form doesn't add quotes when you use a comma if it isn't an existing tag? That does sound like a bug to me, so please go ahead and open an issue if you can't find an existing one that documents it. If you find an existing issue with the same problem, or if you create one, then add a link to it in this code so we'll be able to check what the resolution is.

@rlskoeser not exactly. This bug occurs in only one kind of case: when a user is inputting their first tag and it has multiple words. The JS renders it as one tag, but when processed, it's split up into multiple words. Once there's a comma in the input, it's handled fine. Here's how `django-taggit` handles input: https://django-taggit.readthedocs.io/en/latest/forms.html

Not sure if I'm explaining this well, happy to talk through it if easier.

geniza/common/utils.py
rlskoeser: Ok, fair enough. There's a lot of other logic in that function! Do we need any of it? Can we just add a comma when needed and pass to the parent method instead?

Issue #434, opened 2021-12-16
As a scholar, I want to get a copy of transcription text so that I can easily reference it and use it elsewhere.


rlskoeser:
## testing notes

all download tests should be done from document detail page, for documents with a digitized transcription associated
- [x] download link should appear just above the mirador viewer
- [x] download link text should include edition author's name; check display for sources with
  - [x] one author
  - [x] two authors
  - [ ] more than two authors
- [x] downloaded text filename should include pgpid, slugified shelfmark, and slugified author last name
- [ ] check file contents: I added rtl and ltr marks for content and line numbers, and I also added some whitespace justification; I'm also now including the labels
- [x] documents without a transcription available should not have a download link
- [ ] check the search results to make sure transcription display and transcription search is still ok (formatting has changed slightly; now will include labels and line numbers)

--- 

Limitation with displaying transcription as annotation in mirador  — not available to select for copy/paste.

... My only concern is that it doesn't seem like you can copy and paste the transcriptions in the viewer, so not only can you not see them next to the write image, but you can't cheat by copying them onto a word document and having two windows open at once. Are you able to enable copy-paste?

_Originally posted by @richmanrachel in https://github.com/Princeton-CDH/geniza/issues/322#issuecomment-996252389_

## dev notes
- [x] include slugified shelfmark in filename (first one only for join)
- [x] add line numbers to plain text version of transcription
- [x] move link to just above image viewer and label based on author, e.g. "Download Goitein edition"

rlskoeser:
@richmanrachel @mrustow would a plain text download of the transcription text be sufficient for now?  That seems like it would be the easiest solution. 

If that's acceptable, two followup questions:
1. Should the plain text file include line numbers? (My guess is yes; does formatting matter?)
2. Is it ok if search term highlighting of transcription text includes line numbers? (currently it does not; it's much easier if we use the same plain text version for both of these things)

rlskoeser:
Also: not specifically related to this issue, but how much would it help in the interim if the transcription text was displayed for every image? I can look into that again.

richmanrachel:
@rlskoeser - I think a plain text download would work. Can we test it?

1. Yes, I think line numbers would be helpful. And I think it'd be annoying, but okay as a temporary solution to have to add in line breaks myself. (I assume that's what you mean by formatting?)
2. Oooh, I don't think that would be good, because then if a phrase is broken across a line, a search wouldn't work.
3. Having the transcription text for every image would be great!

[...]

PR #491, opened 2022-01-05
Make transcription available for download as text

rlskoeser:
proof-of-concept implementation to allow downloading transcription content as plain text, to see if this will satisfy the functionality needed for #434


blms (approved):
This LGTM so far! Of course, will want to take the inline style out of the HTML when moving out of proof-of-concept phase. Otherwise everything else looks ready to finalize. Great call on `Content-Disposition`.

Issue #494, opened 2022-01-05
As a front-end user, I want to use fields in my keyword searches so I can make my searches more specific and targeted.


rlskoeser:
Enable field searching like we have on PPA based on the fields we already have indexed.

## testing notes
In the keyword search box on the public site, you will be able to use fields to search within specific content, using syntax like `field:text` or `field:"exact phrase"` — these can be combined with other search fields or booleans like AND, NOT (default is OR)

these are the fields that are currently available:
- pgpid
- old_pgpids
- shelfmark
- ~collection~ (will only match exact string as currently indexed)
- description
- transcription
- tags
- ~language_code~
- input_year

Note: the same logic should also work in the admin document search.

revisions after testing:
- [x] fix incorrect solr field for old pgpid
- [x] add singular aliases for tags and old pgpids
- [x] don't use fielded search terms for kwic highlighting

rlskoeser:
question: we use plurals on tags and old pgpids because it makes sense to us in the code (since they can be multiple), but would singular make more sense for someone searching?

richmanrachel:
@rlskoeser - I just searched "pgpid: 1259" and while the first result is indeed that document, there were two other results pulled that had a date of 1259 CE in the description. Is this expected behavior?

rlskoeser:
@richmanrachel you need to remove the space between the colon and the value you want to search on in that field

richmanrachel:
@rlskoeser - great! That field works now. I'll test the rest after lunch :)

richmanrachel:
- pgpid: Works
- **old_pgpids: Doesn't work as advanced search**. I used the example Alan thankfully captured in Slack a while ago (old: 6786, new: 11317), and the string "old_pgpid:6786" had no results. Luckily, it does come up when you just search 6786, though.
- shelfmark: Works when you add parentheses for precise term.
- **collection: does not work**. (I tried ENA, JTS, British_Library, Penn... which type of string am I supposed to use? I tried to do multiple kinds but none worked).
- description: works
- transcription: works
- **tags: doesn't work**
- **language_code: unsure.** I could test for Hebrew ("language_code:he") and Arabic "...:ar" but how do I look for Judaeo-Arabic? What are the codes for all of our non-standard languages? Just a heads up that it's pulling in the descriptions too, for example here with ar:
![image](https://user-images.githubusercontent.com/68028846/148590832-9f87c3b3-522d-496d-bb4b-a70881cee55d.png)

- input_year: works

[...]

PR #495, opened 2022-01-05
Convert solr alias fields in keyword search terms

rlskoeser:
tweaks the keyword search to convert aliased solr field names into actual solr fields so that we can allow users to do more powerful searches within the existing keyword search field; specifics are documented on #494 

(note that this also makes it possible to easily support tag searching)

### DIFF
--- a/geniza/corpus/solr_queryset.py
+++ b/geniza/corpus/solr_queryset.py
@@ -1,3 +1,5 @@
+import re
+
 from parasolr.django import AliasedSolrQuerySet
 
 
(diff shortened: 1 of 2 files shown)


geniza/corpus/solr_queryset.py
```diff
@@ -44,8 +46,23 @@ def admin_search(self, search_term):
 
     keyword_search_qf = "{!type=edismax qf=$keyword_qf pf=$keyword_pf v=$keyword_query}"
 
+    # regex to convert field aliases to actual solr fields
```
kmcelwee, line 49: NIT: Maybe because I haven't had caffeine yet, but this took me way too long to figure out what this would look like, could we maybe just add a comment that shows what this string looks like? 
```suggestion
    # regex to convert field aliases to actual solr fields: "(shelfmark|tags| ...)"
```

kmcelwee (approved):
Cool! Do we need to write up a content page that explains advanced search?

blms (approved):
LGTM!

PR #497, opened 2022-01-06
Undo condition that only shows join shelfmarks

kmcelwee:


kmcelwee:
@rlskoeser I'm merging this, I'm assuming that this is one of those instances where I don't need your review. Let me know if I shouldn't in the future!

PR #498, opened 2022-01-06
383 needs review list view

kmcelwee:
* We also have needs_review as an property on Fragments, should we implement that later?
* I double checked the changeform, and changes to one model are reflected in the proxy. Though I was surprised I still needed to make a migration for a proxy model (?)
* I tried adding needs_review to search but it doesn't seem to be working. It's indexed in solr though. Can you think of any reason why that might be?


geniza/corpus/admin.py
```diff
@@ -462,13 +463,48 @@ def get_urls(self):
...
+    search_fields = (
+        "fragments__shelfmark",
+        "tags__name",
+        "description",
+        "notes",
+        "needs_review",
+        "id",
+        "old_pgpids",
+        "needs_review",
+    )
```
rlskoeser, line 490: search fields are ignored because we're using solr instead of django

geniza/corpus/admin.py
```diff
@@ -462,13 +463,48 @@ def get_urls(self):
...
+
+    def get_queryset(self, request):
+        # ?: We often use `needs_review__is_empty` but I got this error. Is that expected?
+        #      Unsupported lookup 'isempty' for TextField or join on the field not permitted.
+        return super().get_queryset(request).exclude(needs_review="")
```
rlskoeser, line 495: I don't see `isempty` as an available queryset filter; maybe you meant `isnull`? This is the filter I would expect you to use, since we can't search on null for text fields.

geniza/corpus/admin.py
```diff
@@ -462,13 +463,48 @@ def get_urls(self):
...
 class DocumentPrefetchableProxyAdmin(admin.ModelAdmin):
     """Proxy model admin for :class:`DocumentPrefetchableProxy` that intercepts `get_queryset`
     in order to prefetch the :class:`GenericRelation` `log_entries`."""
 
     def get_queryset(self, request):
         return super().get_queryset(request)
 
+    def filter(self, request):
+        return super().filter(request)
+
```
rlskoeser, line 507: what does this do?

geniza/corpus/models.py
```diff
@@ -886,3 +886,11 @@ def __str__(self):
...
     def thumbnail(self):
         return self.fragment.iiif_thumbnails()
+
+
+class DocumentNeedsReview(Document):
+    """Proxy model for Document to give a separate view for documents that need review"""
+
+    class Meta:
+        proxy = True
+        verbose_name_plural = "Documents needing review"
```
rlskoeser, line 896: Will django let us put this in the admin instead of models? We really only want to use it there. I agree, it does seem annoying that we need a migration for a proxy model.

Wonder if we can shorten the name / label any. `ReviewDocuments` /  Documents for Review?

geniza/corpus/admin.py
```diff
@@ -104,7 +105,7 @@ def documents(self, obj):
 
     @admin.display(
         ordering="secondary_document__count",
-        description="# documents where this is a secondary langaug",
+        description="# documents where this is a secondary language",
```
rlskoeser, line 108: thank you 🙄

geniza/corpus/admin.py
```diff
@@ -462,13 +463,48 @@ def get_urls(self):
     actions = (export_to_csv,)
 
 
+@admin.register(DocumentNeedsReview)
+class DocumentNeedsReviewAdmin(DocumentAdmin):
+    ordering = ("needs_review",)
+
```
rlskoeser, line 469: Should probably turn off the needs review filter that we're inheriting from `DocumentAdmin`

rlskoeser (commented):
Glad to see that this is basically working!

Would really prefer not to have a separate edit form (even if it is the same form...). Can you think of anyway to avoid that? Can we make the pgpid a link to the real change form, or make it a redirect? Or am I thinking about this wrong, and it would be better to keep it separate so that when you're working through the review queue you stay in that context?

responses to your questions:
>     We also have needs_review as an property on Fragments, should we implement that later?

It's a problem for documents because there are so many documents and so many documents needing review; that's not a problem right now for fragments, so let's hold off.

>     I double checked the changeform, and changes to one model are reflected in the proxy. Though I was surprised I still needed to make a migration for a proxy model (?)

Right. That makes sense from what I understand of django proxy models. (I agree about the migration, seems silly to me but I guess it must be for versioning.)

>    I tried adding needs_review to search but it doesn't seem to be working. It's indexed in solr though. Can you think of any reason why that might be?

Admin search fields are ignored since we've implemented our own solr search — however, needs review is indexed and should be included in the admin search we have configured in solr. What do you mean that it isn't working?

Once we have the advanced search with this, we should also be able to use searches like `needs_review:...` — can you think of an easy way to put that as the default search text, or put help text near the search form? Might be worth extending the list view template if there isn't a better way.

geniza/corpus/admin.py
kmcelwee: Oh I was drawing from html query parameters and got confused: `?needs_review__isempty=0`
Sounds good!

geniza/corpus/admin.py
kmcelwee: I thought about it, but there's a lot to be copied. Are we okay having this twice, minus `needs_review`?

```
list_filter = (
        "doctype",
        (
            "footnotes__content",
            custom_empty_field_list_filter(
                "transcription", "Has transcription", "No transcription"
            ),
        ),
        (
            "textblock__fragment__iiif_url",
            custom_empty_field_list_filter("IIIF image", "Has image", "No image"),
        ),
        (
            "needs_review",
            custom_empty_field_list_filter("review status", "Needs review", "OK"),
        ),
        "status",
        ("textblock__fragment__collection", admin.RelatedOnlyFieldListFilter),
        ("languages", admin.RelatedOnlyFieldListFilter),
        ("secondary_languages", admin.RelatedOnlyFieldListFilter),
    )
```

geniza/corpus/models.py
kmcelwee: I was able to move it to `admin.py` and renamed the model to "Documents for Review"

kmcelwee:
> Would really prefer not to have a separate edit form (even if it is the same form...). Can you think of anyway to avoid that? Can we make the pgpid a link to the real change form, or make it a redirect? Or am I thinking about this wrong, and it would be better to keep it separate so that when you're working through the review queue you stay in that context?

I agre that after saving a document, I’d imagine you’d want to go back to the “Documents for Review” page. I don’t see a straightforward way of doing that with a shared edit form. Linking to the edit form seems straightforward, but getting back to the needs review page doesn’t. Should we ask the geniza team?

geniza/corpus/admin.py
kmcelwee: I removed these. Should we remove them from the other models as well?

rlskoeser: no, search fields needs to be defined for model admin to turn on search; I think there's a comment on document about this, and that's the only model that's affected

geniza/corpus/admin.py
rlskoeser: No, I don't want it copied. Maybe we could make a copy of the list and remove the filter we don't want, but perhaps  we should just leave it as is for first round testing and let the team tell us what filters they need.

rlskoeser:
I think this is probably ok to merge, but I'm now concerned that the new proxy model is going to cause problems with Solr indexing, because we've had a lot of problems with that related to the document prefetch proxy that we added.

Would it make be logical for parasolr to ignore proxy models that inherit from indexables? Does it seem safe to _always_ ignore them, or would it need to be configurable? @blms any opinion?

blms:
@rlskoeser My inclination is that it would be safe to always ignore them. I can't think of a scenario where you'd want to index the proxy rather than the original model.

blms:
@rlskoeser @kmcelwee My thoughts are:

1. If possible, it may be preferable to get rid of the proxy model, and just define this ModelAdmin without using the `@admin.register` decorator to register it in the admin interface, adding it manually by URL instead. I'm not sure if it will work, but I did find [this answer on StackOverflow](https://stackoverflow.com/a/36225723).
2. If that isn't possible, we need to think of what could be impacted by the proxy model's presence and account for that. So far I can think of: 
   1. LogEntries being associated with the proxy model instead of the original model. I noticed when editing them that we end up on `/admin/corpus/documentneedsreview/[id]/change/`. It may be preferable if all such things redirected to the original model
   2. solr indexing the proxy model unnecessarily

rlskoeser:
deciding not to pursue this approach, since the proxy model and additional admin urls are problematic

Issue #507, opened 2022-01-06
django admin document filter by "has transcription" reports inaccurate numbers


rlskoeser:
## testing notes

Use the transcription filter in the document admin list and check to see if the number of documents matches what you expect for both has transcription and no transcription.

**Describe the bug**
Filtering django admin document list by has transcription is giving different numbers in production than it was previously, which we don't think is due to changes in the data (but should confirm!). It also generates different results than filtering in python console. 

**Additional context**
Unclear what the django empty filter is doing; maybe it doesn't work properly for our json field; suggest using a custom filter that checks for the content we actually expect to indicate the presence of transcription content synced from TEI. We may also want to compare exports from the earlier version of the database to check for any problems or discrepancies.

richmanrachel:
Yes, this is looking much better! Here's a screenshot with current site on the right and test on the left:
![image](https://user-images.githubusercontent.com/68028846/149013075-37eff8db-ae2e-4015-a8fc-372a3d76ef00.png)

richmanrachel:
And for total transcriptions, test is now (i think correctly) reading 4,236 total docs with transcriptions (vs. 1,017 on current site). Closing!

rlskoeser:
reopening — this shouldn't have been closed, since even though the filter works, it introduced a performance problem when applying this filter in combination with others

rlskoeser:
please retest and confirm that the "has transcription" filter in the admin document list can be used with three or more other filters without the page becoming unusably slow

richmanrachel:
@rlskoeser - It works! Hallelujah!

PR #509, opened 2022-01-10
Custom admin filter for documents with transcription

rlskoeser:
ref #507

the empty admin list filter stopped working reliably after a recent update; this PR implements a custom filter that keys directly on the html content set by the `sync_transcription` script, and the numbers seem much better

### DIFF
--- a/geniza/corpus/admin.py
+++ b/geniza/corpus/admin.py
@@ -155,6 +155,25 @@ class DocumentForm(forms.ModelForm):
             )
 
 
+class HasTranscriptionListFilter(admin.SimpleListFilter):
+    """Custom list filter for documents with associated transcription content"""
+
+    title = "Transcription"
+    parameter_name = "transcription"
+
+    def lookups(self, request, model_admin):
+        return (
+            ("yes", "Has transcription"),
+            ("no", "No transcription"),
+        )
+
+    def queryset(self, request, queryset):
+        if self.value() == "yes":
+            return queryset.filter(footnotes__content__html__regex=r".*").distinct()
+        if self.value() == "no":
+            return queryset.exclude(footnotes__content__html__regex=r".*").distinct()
+
+
 @admin.register(Document)
 class DocumentAdmin(admin.ModelAdmin):
     form = DocumentForm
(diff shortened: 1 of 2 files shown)


geniza/corpus/admin.py
```diff
@@ -155,6 +155,25 @@ def clean(self):
...
+        return (
+            ("yes", "Has transcription"),
+            ("no", "No transcription"),
+        )
+
+    def queryset(self, request, queryset):
+        if self.value() == "yes":
+            return queryset.filter(footnotes__content__html__regex=r".*").distinct()
+        if self.value() == "no":
+            return queryset.exclude(footnotes__content__html__regex=r".*").distinct()
```
blms, line 174: This look good. Wondering about performance of `__regex=r".*"` vs `__exact=""`—likely insignificant but thought I'd mention. Though I've now read online that this regex should match an empty string, so would potentially have a different result from `__exact=""`. Is that true?

geniza/corpus/tests/test_corpus_admin.py
```diff
@@ -370,3 +371,45 @@ def test_collection_display(self, fragment):
...
+    @pytest.mark.django_db
+    def test_queryset(self, document, join, typed_texts):
+        filter = self.init_filter()
+
+        # no transcription: all documents should be returned
+        all_docs = Document.objects.all()
+        # no transcription: all documents should be returned
+        with patch.object(filter, "value", return_value="no"):
+            assert filter.queryset(Mock(), all_docs).count() == 2
+        # ahs transcription: no documents should be returned
```
blms, line 396: _Extremely_ minor, but both of these comments have a typo "ahs transcription".

blms (approved):
Looks great!

Because of the empty string question I mentioned, I wonder if it would be worth including a test for a document with `content={"html": ""}`, but not sure if that will ever actually happen via the transcription sync script.

geniza/corpus/tests/test_corpus_admin.py
rlskoeser: 🤦‍♀️ thank you

geniza/corpus/admin.py
rlskoeser: Thanks for pushing me on this, agree that regex should generally be a last resort option. It turns out there is a `has_key` filter for json fields!

And sync transcription script doesn't generate content with empty html (it checks for and skips empty TEI files).

blms: Perfect! :+1

PR #511, opened 2022-01-10
Override DocumentAdmin get_deleted_objects

rlskoeser:
@blms when I went to deliver the document deletion bug fix I discovered that we still hadn't solved the problem! the deletion view used the custom queryset, so it still had the log entries relation with the permissions problem. However, I figured out that once that queryset customization was in place, I could override just the `get_deleted_objects` method to make sure _that_ one uses the non-proxy `Document` model without the generic relation causing all our permission problems.


blms (approved):
Oops! 😱 Great catch and solution. Hooking into `get_deleted_objects` makes a lot of sense.

PR #517, opened 2022-01-12
Link tags to search documents by tag

rlskoeser:
Turns tags in search results & document detail page into links to a tag search

This will change the percy build, but I can't tell from the designs on Figma if the default link style is what's intended for tags.


blms (approved):
Looks good to me!

PR #518, opened 2022-01-12
Revise sync transcriptions plain text handling

rlskoeser:
updates to sync transcription plain text in support of #434 

- include line numbers & labels in plain text version of transcription text
- use unicode ltr and rtl marks in plain text for transcription and numbers
- revise transcription display in search results to make it work better with the revised plain text format

includes some other cleanup and improvements to the sync transcription script


geniza/corpus/management/commands/sync_transcriptions.py
```diff
@@ -63,7 +69,12 @@ def handle(self, *args, **options):
             # some files are stubs with no content
             # check if there is no text content; report and skip
             if tei.no_content():
-                self.stdout.write("%s has no text content, skipping" % xmlfile)
+                if self.verbosity >= self.v_normal:
```
blms, line 72: I'm not familiar with this pattern. Does it provide an advantage over just using `-v` for verbose? (Or I guess for this case it would be more like `-q` for quiet, since the default `v_normal` would write all of these outputs.)

geniza/corpus/tei_transcriptions.py
```diff
@@ -24,6 +24,10 @@ class GenizaTei(teimap.Tei):
     labels = xmlmap.NodeListField(
         "tei:text/tei:body/tei:div/tei:label", GenizaTeiLine
     )  # not really a line...
+    # source description sometimes contains reference to scholarship record
+    source = xmlmap.NodeListField(
+        "tei:teiHeader//tei:sourceDesc/tei:msDesc/tei:msContents/tei:p", GenizaTeiLine
```
blms, line 29: Wow! Does this then become a TEI `<source>` element?

blms (approved):
Looks good to me, just a minor question about the verbosity stuff in the sync script, and a question (simply curiosity) about how the source bit works!

geniza/corpus/management/commands/sync_transcriptions.py
rlskoeser: django manage command verbosity default parameter is a numeric argument so output can have varying levels of verbosity, and you can say `-v 0`, `-v 1`, `-v 2`, etc — this `v_normal` stuff is a convention I adopted ages ago as a more semantic way to control the output — although I don't think we're using the granularity much in this script, it could let us optionally make the output much more verbose if we wanted

blms: I see, makes sense! Didn't know manage commands had a built in verbosity parameter.

geniza/corpus/tei_transcriptions.py
rlskoeser: I'm trying to keep the tei handling as simple as I can, because I just want to easily access the parts of the content that I care about and don't need to manipulate the xml. (I don't remember why I used the `GenizaTeiLine` class, could probably be use more generic...) 

Some of the TEI files describe the source of the transcription in text inside `<p>` tags nested under the `<sourceDesc>` — this mapping gives me a list of lines that I can convert to string so I can easily check if e.g. "Goitein, typed texts" is referenced.

This mapping lets me access those paragraphs as an array on the tei object, so that for example I can check if Goitein and typed texts occur within the text content of `tei.source[0]`  (you'll see more on a subsequent pr!)

blms: I see! That makes a lot of sense, thanks.

PR #519, opened 2022-01-12
Improve solr indexing efficiency

rlskoeser:
The document prefetchable proxy we added was making parasolr think it had to index everything twice!

adjustments in this PR:
- use prefetchable proxy when we get items to index 
- add more prefetching (source related content needed for footnote/source string)
- implement class methods so parasolr will ignore the prefetchable class when indexing

I'm not sure how much the additional source prefetching helps... haven't measured or anything; but we are using all of those items to generate footnote strings for search and indexing, so seems like it would be help.


blms (approved):
I definitely think the more prefetching the better.

Parasolr fixes make sense, definitely the kind of thing I didn't catch due to not knowing this piece as deeply. Thanks!

blms:
Oh yes, and a good test case for `[skip percy]` 😀

Issue #452, opened 2021-12-17
As a front-end user I want to see the header and visual signifiers for the site's main menu to improve my experience when I hover, click on, navigate through menu items with the keyboard, or when an item is disabled


gissoo:
## testing notes (QA)

- [x] Visit the QA site https://test-geniza.cdh.princeton.edu/ and confirm that the new header design is present on both mobile and desktop, light and dark mode
- [x] On desktop, test hover interactions with both main and sub-menu on light and dark mode
- [x] Visit a page from the menu and sub-menu and confirm the "current page" indicator is correct
- [x] Test the mobile menu and confirm it works as expected

---

Here are the links to the designs:

- [for desktop light mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A24685)

- [for mobile light mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A24910)

- [for desktop dark mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A24796)

- [for mobile dark mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A24990)

blms:
Background color needs to be `var(--background)`, not transparent.

<img width="227" alt="Screen Shot 2022-01-12 at 12 43 35 PM" src="https://user-images.githubusercontent.com/7234006/149194239-a7fea5c8-475c-4640-9fda-307cd56db452.png">

rlskoeser:
The about menu opens on hover, but it is also clickable and when I click the page jumps down (presumably to the anchor link). Clicking seems to toggle the direction of the ^ but hovering does not.

blms:
> The about menu opens on hover, but it is also clickable and when I click the page jumps down (presumably to the anchor link).

Yes, I wasn't sure how to handle this on desktop, since this is also the link to open the submenu on mobile. It seems that making this not clickable, or using `preventDefault` in JS, would break the mobile functionality… 🤔 

> Clicking seems to toggle the direction of the ^ but hovering does not.

That's odd, it's working for me locally but not on the test site either. Maybe it has to do with its dependence on JS (to add the `.open` class/set the `aria-expanded` attribute). Maybe I could fix it with this change:
```diff
[...]

PR #520, opened 2022-01-13
#452 - Fixes to header, navigation, sub-navigation menu

blms:
## What this PR does
- Per #452:
  - Fixes issues with submenu background transparency by using two additional SVGs
  - Fixes issue with "about" anchor jumping around (by just disabling clicks for now)
  - Ensures that bold text always uses the correct font face
  - Fixes issue with the logo getting a hover state meant for nav menu
  - Improves transparency of gradients for header image
- Ensures that hover background images for long or short page titles are handled gracefully

## Questions
- Is it ok to use a `span` inside the `a`? It's a bit of an extraneous element, but it's extremely helpful for dealing with the `a::after` pseudoelement, so that the text piece of the `a` element can be handled independently of the `::after`. (Without this, I couldn't figure out how to get the `::after` background image positioned consistently underneath the text content.)
- I've configured the "About" menu to no longer respond to clicks on desktop—and to not receive a pointer cursor. It is a simple fix, and will only allow the dropdown menu to be active while hovering. Is this OK, or should I keep trying to work with the target anchor?


sitemedia/scss/base/_typography.scss
```diff
@@ -22,7 +22,7 @@ $text-size-5xl: 2rem; //     = 32px
 
 // H1 — page titles
 @mixin headline-1 {
-    font-family: fonts.$primary-bold;
+    font-family: fonts.$primary-bold !important;
```
rlskoeser, line 25: why do we need all the `!important` s ?

rlskoeser (approved):
Using a span and having about only responding to hover and not click sound fine to me.

Wondering why we need all the overrides on the headline fonts, but otherwise this looks okay to me.

Issue #397, opened 2021-12-07
As a front-end user I want to see the footer to view a list of website's menu items, licensing, accessibility, and links to social media 


gissoo:
here are links to the designs: 

- [for desktop light mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A25253)

- [for mobile light mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A25670)

- [for desktop dark mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A26983)

- [for mobile dark mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A26591)

blms:
Confirmed present in QA (implementation already reviewed in percy)

PR #528, opened 2022-01-14
#397 - Implement site footer

blms:
## Visual review notes
[Latest Percy build](https://percy.io/2cf28a24/geniza/builds/15028327)

In scope:
- Footer appearance on mobile and desktop, dark and light mode

We can ignore the weirdness with the mobile navigation menu screenshots for now, since I think that's an issue with Percy that won't show up on the actual site. If it does we'll revisit!

(Can also ignore the footer bg image not appearing in Percy on Safari.)

## Code review notes
- Looking especially for feedback on the templates for the footer and reusable nav—want to make sure they are semantically and logically sound.

## What this PR does
- Per #397:
  - Implements footer in light and dark mode, mobile and desktop, per designs
  - Makes site nav reusable so that it can be used in both header and footer
  - Edits images to be flexible with more or less content, larger screen sizes


rlskoeser (approved):
templates look good to me

rlskoeser:
happened to notice on percy:
- content width looks different in safari desktop and chrome desktop; is this due to different fallback fonts?
- footer background image is not displaying in safari

blms:
@rlskoeser 

> content width looks different in safari desktop and chrome desktop; is this due to different fallback fonts?
That's right! The width is defined as `60ch`, or 60 characters.

> footer background image is not displaying in safari
Good catch! I can't replicate locally, so I wonder if it is some Safari 14 or Percy quirk. Going to try running Percy again and see if it resolves.

PR #529, opened 2022-01-14
Correct document prefetch index type

rlskoeser:
When indexing document via prefetch proxy,
index data should be the same when indexed as a document


blms (approved):
Had to look up what `index_item_type` is from parasolr, but this PR is totally clear now and I see why it would have caused that problem. The surprises keep coming with the proxy model!

Issue #451, opened 2021-12-17
As a front-end user I want to see pagination and visual signifiers for when I hover or click on, use the keyboard to navigate the pages, or when the next/previous pages are disabled to help me navigate through search results


gissoo:
## testing notes (round 2)

For pagination controls on search pages:
- [x] check focus style on dark mode
- [x] confirm active page color does not change on hover
- [x] confirm no "jump" when hovering

---

## testing notes

- [x] Check hover, focus, and click styles for pagination controls on mobile and desktop, light and dark
- [x] Check search queries with hundreds of pages, and ones with 0-5 pages, and make sure the pagination controls look right (mobile and desktop)
- [x] Check the beginning, middle, and end of both long and short search queries (mobile and desktop)

---

here are links to the designs:

- [for desktop light mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A24411)
- [for mobile light mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A24487)
- [for desktop dark mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A26341)
- [for mobile dark mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A26858)

gissoo:
@blms thanks! Everything functions perfectly!

- Can we be consistent and make the focus style on the dark mode the same blue we have used for the other components?
- The active page is also hoverable and turns dark grey on hover, I think it should not be since we decided to fix it on the tabs as well
- Do you know why the page numbers slightly jump when you hover through them? There seems to be something about the underline that misplaces them when it appears. Could their position be fixed with or without the underline to avoid the jump?

Issue #493, opened 2022-01-05
Revise document title for joins


rlskoeser:
@rlskoeser and @gissoo - Marina and I just met and we agree that we need to have the full set of shelfmarks in the title (especially if the join is only 2, we might be flexible about 3+). 

Although I initially thought it would be repetitive to have the shelfmark in the title and listed again below, Marina has convinced me that this repetition is productive as it's "easier on the eyes" and will help scholars keep track of this most crucial data point.

_Originally posted by @richmanrachel in https://github.com/Princeton-CDH/geniza/issues/460#issuecomment-1006126981_

gissoo:
@rlskoeser @richmanrachel @mrustow I think this might make sense for two – I don't quite agree that it's easier on the eye, but having more specific examples and context would be helpful to see the issue – reading this, my perception is that a sweet spot for the 3+ shelfmarks is if they appeared as a subtitle right below the doc title in the search results – there really shouldn't be two lines of document title. Can you say more about "will help scholars keep track of this most crucial data point."? Are you saying this because it's not appearing on the search results? If so, then having them appear right below the title might make them more readable than as the document title

blms:
> Ideally long shelfmarks within joins like the above (however or wherever they end up appearing) should be broken not in the middle of the shelfmark but either before or after the + : 
> Bodl. MS heb. f 34/41
> + Bodl. MS heb. f 34/40
> + Bodl. MS heb. f 34/39
> 
> or 
> 
> Bodl. MS heb. f 34/41 +
> Bodl. MS heb. f 34/40 +
> Bodl. MS heb. f 34/39
> 
> @richmanrachel do you have a preference between those two options? I _think_ I prefer the first one.

_Originally posted by @mrustow in https://github.com/Princeton-CDH/geniza/issues/532#issuecomment-1021369240_

blms:
>  <img width="312" alt="Screen Shot 2022-01-19 at 6 42 20 PM" src="https://user-images.githubusercontent.com/44615389/150236029-33657467-3ea4-4d72-bbc4-cab7b550666b.png">

_Originally posted by @gissoo in https://github.com/Princeton-CDH/geniza/issues/532#issuecomment-1016972071_

blms:
Actually these might be two separate issues—the one referenced in the original description seems to be about the document detail page, and the rest is about the way a document title will appear in search results (e.g. the + indicator for joins, the above screenshot, etc).

[...]

PR #532, opened 2022-01-17
#451 - Implement new pagination designs

blms:
## Visual review notes

[Latest Percy build](https://percy.io/2cf28a24/geniza/builds/15062034)

In scope:
- Pagination controls, mobile and desktop, light and dark modes

## What this PR does

- Per #451:
  - Implements new pagination designs on mobile and desktop

### DIFF
--- a/geniza/corpus/templates/corpus/snippets/pagination.html
+++ b/geniza/corpus/templates/corpus/snippets/pagination.html
@@ -1,10 +1,10 @@
 {% load i18n corpus_extras %}
 {% spaceless %}
-    <nav class="pagination {% if page_obj.paginator.num_pages < 5 %} compressed {% endif %}">
+    <nav class="pagination">
         {# Translators: Label for "previous page" button in search results #}
         {% translate 'Previous' as previous_page %}
         {% if page_obj.has_previous %}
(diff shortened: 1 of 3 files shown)


gissoo:
@blms do you know why this is happening on safari? Everything else looks good. 

Screen Shot 2022-01-19 at 6 35 17 PM

gissoo:
also, unrelated but this looks scary: :D Screen Shot 2022-01-19 at 6 42 20 PM

blms:
> @blms do you know why this is happening on safari? Everything else looks good.

Thanks! I'm not sure why it's happening, first noticed it in #466. Rebecca and I weren't able to replicate in our own copies of Safari so we just chalked it up to Percy.

mrustow:
Ideally long shelfmarks within joins like the above (however or wherever they end up appearing) should be broken not in the middle of the shelfmark but either before or after the + : 
Bodl. MS heb. f 34/41
+ Bodl. MS heb. f 34/40
+ Bodl. MS heb. f 34/39

or 

Bodl. MS heb. f 34/41 +
Bodl. MS heb. f 34/40 +
Bodl. MS heb. f 34/39

@richmanrachel do you have a preference between those two options? I _think_ I prefer the first one.

blms:
@mrustow @richmanrachel @gissoo I think this pertains to #493 which is still open so we can move this discussion there. If a new issue needs to be created we can do that instead, though!

Issue #523, opened 2022-01-13
As a front-end user, I should not be able to sort by relevance without any search text, since relevance is not meaningful without search terms.


blms:
## testing notes
- [x] try making a search with no query text, and "relevance" selected for the sort option, and confirm that an error appears in the form
- [x] comment any thoughts on the design of the error display
- [x] compare with the following behavior and comment any thoughts:

> On the [prosody archive search](https://prosody.princeton.edu/archive/) we implemented js logic to disable/enable relevance, and automatically switch the sort to relevance when you start typing a search term, but preserve your sort if you change it — which I know is kind of complicated, and should not be MVP. Maybe we could have Gissoo do some usability testing on the two searches with regard to sort behavior to help figure out what would help users most.

_Originally posted by @rlskoeser in https://github.com/Princeton-CDH/geniza/pull/534#pullrequestreview-858429524_

## dev notes

Possible solutions:

- JS to check for input into the search box; if not present, disable "relevance"
- make the search box input a required field when "relevance" is selected

gissoo:
@blms @rlskoeser 
- the error message appears in the form
- the error message looks fine, but it doesn't have to display "Errors:" for those looking at the screen – you can just display the message. 
- you have written: "On the prosody archive search we implemented js logic to disable/enable relevance, and automatically switch the sort to relevance when you start typing a search term, **but preserve your sort if you change it** — which I know is kind of complicated, and should not be MVP. Maybe we could have Gissoo do some usability testing on the two searches with regard to sort behavior to help figure out what would help users most." – by "but preserve your sort if you change it" you mean preserve the "relevance" ? Let me know – Sure I can reach out to the ux office for a quick usability testing on it 
- I wish "Relevance" would not even be an option when there is no query text

blms:
@gissoo Thanks!

> the error message looks fine, but it doesn't have to display "Errors:" for those looking at the screen – you can just display the message.

Good call, I'll leave this issue open until implementing that.

> by "but preserve your sort if you change it" you mean preserve the "relevance" ?

No, it will automatically choose "relevance" when you start typing some text—but if you change it away from "relevance" after typing text, then it should preserve your choice.

It actually seems like the Prosody Archive search isn't working quite as described. It does have "relevance" disabled when there is no text. But then once you have entered text, it seems to always change back to "relevance" when you edit the search text, even if you changed it to something else before.

> I wish "Relevance" would not even be an option when there is no query text

[...]

PR #534, opened 2022-01-18
#523 - Require search query for "relevance" sort

blms:
## What this PR does

- Per #523:
  - Raises error when "relevance" sort is selected and no search query is entered


rlskoeser (approved):
Ok, tested and I see that this does work. I thought from your notes on the issue you had another suggestion that I liked, which is to use django form validation for combinations of fields: https://docs.djangoproject.com/en/4.0/ref/forms/validation/#validating-fields-with-clean   (but checking your notes again maybe I misread your intent)

I think the form validation would be easy to implement and useful because we could provide a more intelligible error message (something like 'relevance sort is not available without a keyword search term'). If we go with your js solution, that validation would almost never be triggered.  

I'm ok with merging but I'm not sure how much it gets us; it prevents the un-meaningful form submission, but it doesn't tell the user much about why. The bigger problem I'm concerned is that people won't know to switch to relevance sort and they won't understand why they're getting such seemingly bad search results. 

On the [prosody archive search](https://prosody.princeton.edu/archive/) we implemented js logic to disable/enable relevance, and automatically switch the sort to relevance when you start typing a search term, but preserve your sort if you change it — which I know is kind of complicated, and should not be MVP. Maybe we could have Gissoo do some usability testing on the two searches with regard to sort behavior to help figure out what would help users most.

Ok to merge, but would you also implement the form validation before closing the issue?

blms:
@rlskoeser I ended up just implementing that Django-side validation here, hope that's OK.

There is one piece I'm still unsure about, which is that you aren't able to get the document type filter facets when there's an empty queryset, which is the case when there's a form error. In that situation, `self.queryset.get_facets()` returns an empty dict and this line throws an error:

https://github.com/Princeton-CDH/geniza/blob/0502497ed759ac77f4c9346a55def73331babb65/geniza/corpus/views.py#L141

I added a conditional to check if `get_facets()` returned anything before evaluating that line. But in that case, when you are trying to search after receiving this error, the document type filter appears empty. 

Screen Shot 2022-01-20 at 12 29 51 PM

Should I just disable the document type filter in this situation, or is there a better solution here?

rlskoeser (approved):
Looks great!

And good catch on the unset facets. I just did a quick test, we can run the `facet_field` call on the empty solr queryset — if we just move it out of the conditional on the form being valid and apply it to the queryset in either case, well get a list of all types with zero counts. How does that sound for a simple solution?

Issue #501, opened 2022-01-06
As researcher, I want footnotes from the same source counted and displayed as a single scholarship record so that multiple links to parts of same document don't inflate the scholarship count and display.


richmanrachel:
## testing notes (round two—QA)

- [x] Check scholarship records pages for [PGPID 3504](https://test-geniza.cdh.princeton.edu/en/documents/3504/scholarship) and [PGPID 7665](https://test-geniza.cdh.princeton.edu/en/documents/7665/scholarship) to see a range of footnotes, and confirm the layout and text for the footnote links are satisfactory (feel free to check others as well)

---

## testing notes (QA)
- [x] Test document scholarship counts for accuracy, when sorting a search by scholarship records and when viewing individual documents
- [x] Confirm that previously-inflated documents like PGPID 7665 have accurate counts, and their scholarship records pages make sense

---

**Is your feature request related to a problem? Please describe.**
Currently PGPID 7665 looks like it has 14 footnotes, but really 7-8 are pages of the same PDF. So when sorting by scholarship records, the count is misleading. 

## dev notes
- [x] revise model index data so that all scholarship counts are totals of unique sources, not footnotes
- [x] revise document tabs so that scholarship count is reflected the same way (sources, not footnotes)
- [x] revise document scholarship template to display all footnotes for the same source together (use the  `regroup` if possible/appropriate)
- [x] on document scholarship template, display document relation type if set for any of the footnotes associated with that source

blms:
@rlskoeser Quick questions about this:

1. We are talking about `Footnote` URLs, not `Source` URLs, correct? If so, then it is Footnotes that should be merged?
2. In that case, does each URL also need an associated `location`? Currently, most of these duplicates have different URLs _and_ different locations. (PGPID 7665 is a good example)

blms:
After my last comment, I'm thinking through this with a function `footnotes_by_source` that will group footnotes on a particular document by source. Then these sources could only be displayed once on the Scholarship Records page. The footnote locations/URLs could appear as a comma-separated list, like "see X, Y, Z for" in the shaded box. And each source record would only be counted once in the scholarship counts.

However, this raises one question: are there any cases (or foreseeable cases) where there are multiple footnotes on a document with the same source, but different document relations?

rlskoeser:
[...]

PR #535, opened 2022-01-19
#501 - Resolve duplicate footnotes by grouping them by source

blms:
## What this PR does

- Per #501:
  - Indexes document scholarship records counts by unique source, rather than total footnotes
  - Indexes document relation type counts according to the following scheme: 1 per unique type per each unique source
  - Per each source, lists all footnote locations/URLs in scholarship records template, using `natsort` to ensure that the list is ordered naturally in cases where locations end with _1 and _10, such as in PGPID 7665
  - Per each source, lists first footnote's document relations in scholarship records template


geniza/corpus/templatetags/corpus_extras.py
```diff
@@ -44,3 +49,51 @@ def querystring_replace(context, **kwargs):
...
+
+@register.filter
+def unquote(url):
+    """Template filter to parse URL-encoded URLs.
+
+    Example use::
+        <a href="{{ url|unquote }}">
+    """
+
+    return parse.unquote(url)
```
rlskoeser, line 62: why do we need this? are urls encoded incorrectly?

geniza/corpus/templates/corpus/document_scholarship.html
```diff
@@ -20,33 +20,43 @@ <h1 class="sr-only">{{ page_title }}</h1>
...
+                                                    {% include "corpus/snippets/footnote_location.html" %}
+                                                </li>
+                                            {% endfor %}
+                                        </ul>
+                                        for
+                                    {% endif %}
+                                </dt>
+                                <dd class="relation">
+                                    {% if footnotes|length == 1 %}
+                                        {{ footnotes.0.doc_relation }}
```
rlskoeser, line 54: for now let's just assume we can rely on the document relationship from the first footnote — they way our data is now, and certainly for the specific problem we're trying to solve, the document relationships will all be the same for footnotes from the same source

geniza/corpus/templates/corpus/document_scholarship.html
```diff
@@ -20,33 +20,43 @@ <h1 class="sr-only">{{ page_title }}</h1>
...
+                                        see
+                                        <ul>
+                                            {% for fn in footnotes %}
+                                                <li class="location">
+                                                    {% include "corpus/snippets/footnote_location.html" %}
+                                                </li>
+                                            {% endfor %}
+                                        </ul>
+                                        for
+                                    {% endif %}
```
rlskoeser, line 50: suggestion: always display the first footnote location/url as 
``` see [location] for ```

Then, for any additional links, display them but indented so they line up visually with the first one; maybe you can use an nth child selector on the links, or maybe you need to wrap them in something. I'm imagining it would look something like this:
```
see [location 1] for 
    [location 2]
    [location 3]
    [location 4]
```

Will this also simplifies the translation problem you flagged, so we only need to translate the first line?

**Review by rlskoeser — CHANGES_REQUESTED** [review]
You have a lot of logic in custom template tags right now — I'd like to see this using `regroup`; you can make `natsort` available as a template tag on its own. I also think we can simplify the display so that the first footnote is always the displayed the same way, and rely on document relation from the first footnote.

File: geniza/corpus/models.py [hunk 4]
```diff
@@ -636,27 +640,34 @@ def index_data(self):
...
-            if Footnote.EDITION in fn.doc_relation and fn.content:
-                plaintext = fn.content_text()
-                if plaintext:
-                    transcription_texts.append(plaintext)
+        for source in unique_sources:
+            source_relations = []
+            footnotes = source.footnote_set.filter(document=self)
+            for fn in footnotes:
+                for val in fn.doc_relation:
+                    source_relations.append(val)
```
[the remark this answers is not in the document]
blms: Then, each relation would only be counted once for the entire document, right? I think we need to count once for each source. Does something like this make sense?

```py
# dict of sets of relations; keys are each source attached to any footnote on this document
source_relations = defaultdict(set)

for fn in footnotes:
    ... (transcription text indexing)
    # add any doc relations not already present for this source to its key in source_relations
    source_relations[fn.source] = source_relations[fn.source].union(fn.doc_relation)

# flatten sets of relations by source into a list of relations
for relation in list(chain(*source_relations.values())):
    # add one for each relation in the flattened list
    counts[relation] += 1
```

geniza/corpus/templatetags/corpus_extras.py
blms: @rlskoeser Huh, I was having this problem with index card URLs but I think it was a result of the outdated data I was working with at the time, and I can no longer replicate. So I'm removing this. Thanks for flagging!

rlskoeser: oh, yes, that's entirely possible —  we weren't encoding them properly at first, which Rachel caught in testing — when she tried to edit one, django admin validation reported that it wasn't a valid url! Glad we can remove this

blms: Done in 4636049 👍

geniza/corpus/models.py
```diff
@@ -579,6 +579,10 @@ def editors(self):
             source__footnote__document=self,
         ).distinct()
 
+    def unique_sources(self):
```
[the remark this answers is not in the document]
blms: Done in e0b997a 👍

geniza/corpus/models.py
```diff
@@ -636,27 +640,34 @@ def index_data(self):
...
-            for val in fn.doc_relation:
-                counts[val] += 1
-            # if this is an edition/transcription, try to get plain text for indexing
-            if Footnote.EDITION in fn.doc_relation and fn.content:
-                plaintext = fn.content_text()
-                if plaintext:
-                    transcription_texts.append(plaintext)
+        for source in unique_sources:
+            source_relations = []
+            footnotes = source.footnote_set.filter(document=self)
```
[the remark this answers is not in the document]
blms: Done in 81f2aca 👍

geniza/corpus/models.py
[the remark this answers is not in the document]
blms: Implemented this in 81f2aca, I'll let you re-review

geniza/corpus/templates/corpus/document_scholarship.html
blms: Done in 8e83924 👍

geniza/corpus/templates/corpus/document_scholarship.html
blms: Done in e0b997a7e7c2ef999a88c6c00b2194d0d9e69071 👍

rlskoeser (approved):
This looks so much better!

Issue #537, opened 2022-01-20
As an admin, if I change a document I want the log entries to be correctly associated with the document.


blms:
## testing notes (QA)

- [x] In the admin panel: make changes to a document, save those changes, and check the [Log Entries](https://test-geniza.cdh.princeton.edu/admin/admin/logentry/). Confirm that your change is there, then click on the link in the log entry and confirm it takes you to the correct document.
- [x] Test that you can delete a document or multiple documents from the admin panel without receiving a permissions error

Previously, because of some workarounds we did, log entries were being associated with a model called "documentprefetchableproxy" instead of Document, and they were not correctly linking to the document that was changed.

---

## dev notes

- [x] ~~Remove `DocumentPrefetchableProxy` model~~
- [x] ~~Replace with `GenericRelation` reimplementation that overrides `__init__`~~
- [x] ~~Inherit everything else from [existing `GenericRelation`](https://github.com/django/django/blob/2c76c27a951ee76c6f763a05c7e119305a21c78d/django/contrib/contenttypes/fields.py#L275)~~
- [x] add a unit test to check the admin delete page — it shouldn't display the text about not having permissions
- [x] Data migration to change all `LogEntries` that were attached to `DocumentPrefetchableProxy` instances to go to the correct `Documents`
- [x] Consider implications for #498

richmanrachel:
Still getting errors when trying to delete :( 
![image](https://user-images.githubusercontent.com/68028846/151249673-dcda7bcd-2123-4cac-ba13-43cec5c38b69.png)

richmanrachel:
Oh, but it's actually the whole site for me right now!

richmanrachel:
Success on both counts! Closing :)

PR #539, opened 2022-01-20
Remove log_entries from object deletion logic, detach documents from them on delete

blms:
## What this PR does

- Overrides admin `get_deleted_objects` so that log entries are removed from the pre-delete check
- Detaches documents from log entries before a document is deleted
- Shows delete confirmation message on "delete selected" in addition to "delete a single document"
- Per #537:
  - Deletes `DocumentPrefetchableProxy` model
  - In a migration, reassigns all `LogEntry` objects from the proxy model to the original `Document` model


geniza/corpus/migrations/0026_delete_documentprefetchableproxy.py
```diff
@@ -0,0 +1,33 @@
...
+    ]
+
+    operations = [
+        # delete DocumentPrefetchableProxy model
+        migrations.DeleteModel(
+            name="DocumentPrefetchableProxy",
+        ),
+        # reassign all log entries associated with DocumentPrefetchableProxy
+        # instances to Documents instead
+        migrations.RunPython(reassign_log_entries, migrations.RunPython.noop),
```
rlskoeser, line 32: does this order work fine? the prefetchable proxy content type isn't deleted when the model is deleted, right?

geniza/corpus/templates/admin/corpus/document/delete_selected_confirmation.html
```diff
@@ -0,0 +1,6 @@
+{% extends "admin/delete_selected_confirmation.html" %}
```
rlskoeser, line 1: is this one for deleting multiple documents? good catch, thanks for generalizing the message

geniza/corpus/tests/test_corpus_models.py
```diff
@@ -1,4 +1,5 @@
 from datetime import datetime
+from multiprocessing.dummy import freeze_support
```
rlskoeser, line 2: what's this?

rlskoeser (approved):
Looks good. Just a couple of small questions about migration order and a test, and one import that looks unnecessary.

geniza/corpus/tests/test_corpus_models.py
blms: Oops, looks like an intellisense mistake that I didn't catch, thanks!

geniza/corpus/templates/admin/corpus/document/delete_selected_confirmation.html
blms: Exactly!

geniza/corpus/migrations/0026_delete_documentprefetchableproxy.py
blms: That's right, but it doesn't hurt to change the order and it makes more logical sense the other way.

geniza/corpus/tests/test_corpus_models.py
blms: Done in 476afbd 👍

geniza/corpus/tests/test_corpus_migrations.py
```diff
@@ -0,0 +1,38 @@
...
+            change_message="created %s" % d,
+            action_flag=1,
+        )
+
+    def test_log_entries_reassigned(self):
+        # LogEntry should be reassigned so that its ContentType is corpus.Document
+        LogEntry = self.apps.get_model("admin", "LogEntry")
+        ContentType = self.apps.get_model("contenttypes", "ContentType")
+        document_type = ContentType.objects.get(app_label="corpus", model="document")
+        log_entry = LogEntry.objects.first()
```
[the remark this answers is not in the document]
blms: Done in 6182d14 👍

geniza/corpus/migrations/0026_delete_documentprefetchableproxy.py
blms: Done in 92e4487 👍

PR #541, opened 2022-01-20
improve IIIF handling so it works with PUL/JTS content

rlskoeser:
- attribution is optional
- use images from cached manifest when available


rlskoeser:
thanks @blms for adding the tests for these changes, looks good to me

Issue #512, opened 2022-01-11
As a front end user, I want Mirador's theme (light/dark mode) to match my user preference.


blms:
## dev notes

- [x] Mirador theme, relevant doc: https://github.com/ProjectMirador/mirador/wiki/M3-Theming-Mirador
- [ ] Also check why HTML transcription isn't showing up in Percy

blms:
Closing, can reopen if the project switches back to Mirador.

Issue #401, opened 2021-12-08
As a front end user, I want a transcription and image display that works on mobile devices, and allows me to zoom in and out on images.


gissoo:
## testing notes (QA)

Check transcription and image display/interactions for the following kinds of documents, on the document detail page, and confirm that all are working as expected:
- [x] Image(s) and transcription
- [x] Only image(s)
- [x] Only transcription

---

Issue for Mirador edits, as well as implementation of Mirador replacement.

rlskoeser:
@blms I think this may already be done (unless we want to add script indexing & classes to improve transcription display) — however, would appreciate if you could adjust the transcription font display when it is rendered in mirador.

I thought maybe it was too large; Gissoo responded as follows:
> don’t think the issue is the size – I think it’s the spacing between each line # – maybe increase that by 2-4px?
if not, please decrease the type size to 18px on desktop and 16px on mobile only for the mirador view

Also, we need to adjust the `dir: rtl` so it doesn't apply to the section headings that are rendered in the full version of the transcriptions we use in mirador (labels like recto, verso, angled text in margin).

blms:
@rlskoeser @gissoo  Is this more or less what you guys were thinking?

<details>
<summary>Original</summary>

<img width="525" alt="Screen Shot 2022-01-18 at 10 55 03 AM" src="https://user-images.githubusercontent.com/7234006/149972104-d9676b3b-39f9-40af-be92-9c867b6be413.png">
</details>


<details>
<summary>New</summary>

<img width="527" alt="Screen Shot 2022-01-18 at 10 53 01 AM" src="https://user-images.githubusercontent.com/7234006/149971549-c77d5a95-ef3c-4f19-b2dd-fce408385561.png">

</details>

Couple of additional questions on this:

1. Is mobile view for Mirador in MVP (and is it meant to be supported out of the box)? The annotation view is pretty much unusable on mobile right now 😬 Not sure if that's because of the way we're using it, or just how Mirador works with annotations.
2. Are recto and verso transcriptions meant to be annotations on their appropriate page image? I noticed that sometimes (maybe always?) the annotation just appears on the first page in the set, and therefore contains both recto and verso. And then, when you for example change the active image from recto to verso, the transcription disappears altogether. Example: https://test-geniza.cdh.princeton.edu/en/documents/1809/

rlskoeser:
[...]

PR #544, opened 2022-01-21
#401 - Improve mirador display

blms:
## review notes

- Not much code review needed here, just not sure how to handle comparing this with PR #546, the Mirador replacement design. Leave both unmerged?

## What this PR does

- Regenerates UI/UX test fixtures with updated data (especially transcriptions!)
- Per #401:
  - Improves spacing for Mirador on desktop and mobile
  - Resizes some Mirador components on mobile such that it is usable
- Per #512:
  - Enables Mirador dark mode when dark mode is active on the site

## dev notes
- Still wasn't able to resolve Mirador not showing transcriptions in Percy from #512. Thought it might be the test fixtures but still not working!


rlskoeser (commented):
I tried this locally and couldn't didn't see any difference in mirador when I switched to dark mode — am I missing something?

The updated styles do improve things some. 

I think if we stick with mirador we should move the thumbnails to the bottom instead of the side, they take up too much room on mobile.  (Or maybe suppress them on mobile?)

blms:
@rlskoeser 
> I tried this locally and couldn't didn't see any difference in mirador when I switched to dark mode — am I missing something?

That's odd—check out [Percy](https://percy.io/2cf28a24/geniza/builds/15290606), it's even working there! Maybe a JS cache issue?

> I think if we stick with mirador we should move the thumbnails to the bottom instead of the side, they take up too much room on mobile. (Or maybe suppress them on mobile?)

When they're set to bottom by default they cover up a big chunk of the transcription, which is also bad on mobile! I set them to be off by default on mobile, just based on the screen size on page load. (So you'd have to reload when testing this in the browser)

blms:
@rlskoeser Maybe it makes sense to merge this one since it's just updates to Mirador, which is already present. Then we can leave the other one open for decision-making purposes. What do you think?

rlskoeser:
Tested again — you're right, I must have had something cached. I see the dark theme now. And now that I know to reload for mobile, I see the adjustments you made (and makes sense to configure it on page load). 

I agree, we may as well merge these improvements in.

Issue #483, opened 2022-01-04
As a user I want to know when there is a 500 / internal server error on the site so that I know what it is and what my options are


gissoo:
## testing notes (qa)

- [x] Visit https://test-geniza.cdh.princeton.edu/_500/ to force a 500 error, and confirm font sizes, spacing, and images look right (these were not possible to confirm on Percy)

## dev notes
add an error page template for 500 errors based on the design

gissoo:
@blms on this page only the "Search" is showing up in the menu, is this normal?

rlskoeser:
> @blms on this page only the "Search" is showing up in the menu, is this normal?

@gissoo including the menu requires adding more information into the context, and could trigger an error — default django behavior is to try to minimize the possibility of generating errors when you're generating the 500 error page, so I decided we should omit links to the dynamic navigation pages. Hopefully people won't see this page often, and what nav is present is enough to get away from the error page, so it should be sufficient.  Sorry I didn't think about this when you were working on the designs for the error pages — 500 is the only one that is like this. If you have thoughts on how you'd like us to revise this later on please let me know.

Issue #482, opened 2022-01-04
As a user I want to know when there is a 404 / not found error on the site so that I know what it is and what my options are


gissoo:
## testing notes (qa)

- [x] visit a page that will give a 404, such as https://test-geniza.cdh.princeton.edu/bad_url/, and confirm font sizes, spacing, and image look right (since these were not checkable on Percy)

## dev notes
404 logic is already in place for the views that need it; just add a custom 404 template so it will display as designed

PR #545, opened 2022-01-24
#482, #483 - Implement 500 and 404 error pages

blms:
## visual review notes

[Latest Percy build](https://percy.io/2cf28a24/geniza/builds/15290198)

In scope:
- images and vertical spacing for error pages (mobile and desktop, light and dark mode)

Can ignore fonts, horizontal spacing (due to fonts), and as usual the lack of the footer in Safari.

## review questions

- Is the view/template logic ok for the error pages?

## What this PR does
- Per #482:
  - Implements 404 error page per designs
- Per #483:
  - Implements 500 error page per designs
- Uses custom views and templates for both of these in order to pass some additional context, and in order to render them at paths `/500/` and `/404/`
- Replaces images with slightly higher file size—but much higher res—versions
- Cleans up some errant `print` statements


geniza/urls.py
```diff
@@ -18,18 +18,18 @@
...
+from geniza.common.views import error_500
+
 urlpatterns = [
-    # redirect homepage to admin site for now
     path("admin/", admin.site.urls),
     path("accounts/", include("pucas.cas_urls")),
     path("i18n/", include("django.conf.urls.i18n")),
     path("taggit/", include("taggit_selectize.urls")),
     path("cms/", include(wagtailadmin_urls)),
+    path("500/", error_500, name="error_500"),
```
rlskoeser, line 32: can we make this `_500/` ?

here's what we've used in other projects to trigger a real 500:
```suggestion
     path("_500", lambda _: 1 / 0),
```

geniza/common/views.py
```diff
@@ -1,3 +1,26 @@
...
+def error_500(request):
+    """Render a 500 error with the correct status, context, and template."""
+    context = {
+        # Translators: title for Internal Server Error (500) error page
+        "page_title": _("Server Error"),
+        # Translators: description for Internal Server Error (500) error page
+        "page_description": _("Something went wrong putting this page together."),
+        "page_type": "error",
+    }
+    return render(request, "500.html", context, status=500)
```
rlskoeser, line 26: why are these needed? I don't see anything here that couldn't be done in the template directly

geniza/templates/404.html
```diff
@@ -0,0 +1,10 @@
+{% extends "base.html" %}
```
rlskoeser, line 1: is there enough common between 404/500 to merit a base error template?

rlskoeser (commented):
Looks good generally, but I don't see why we need custom error handlers. Can't everything be done in the templates? Or in a base error template if necessary?

Not sure how to respond to your question about the image sizes, is that a question for Gissoo?

geniza/urls.py
blms: Love this solution!

geniza/common/views.py
blms: I'll try again, but I believe the first time I tried without the views, the `site` context wasn't coming through via `context_processors`.

rlskoeser: @blms at least in the 500 error it's possible that it's intentionally not available — you don't want to trigger a new error when you're trying to render an error! we can check the django docs and look at other of our apps to see how we approached this; we need `site` to render menus/nav?

blms: Yes, that's why we need it. Good idea to check other projects!

blms: It breaks everywhere that `site.root_page` is used, which means:
1. Places where we loop through its children (nav menus in header and footer)
2. Places where we reference it directly to get its `pageurl` (logo in header, "Home" link in footer)

rlskoeser: from the documentation: https://docs.djangoproject.com/en/3.2/ref/views/#the-500-server-error-view 

> The default 500 view passes no variables to the 500.html template and is rendered with an empty Context to lessen the chance of additional errors.

Would it make sense to skip the nav on the 500 page to keep it as simple as possible? I think you should have access to normal context on the 400

geniza/urls.py
blms: Done in ab7d30f 👍

geniza/common/views.py
blms: Done in ab7d30f 👍

geniza/templates/404.html
blms: There's a lot in common, but I'm not sure a base template would actually reduce more than one or two lines of repetition. Let me know what you think after these new revisions!

geniza/templates/footer.html
```diff
@@ -5,9 +5,11 @@
 <nav id="footer-nav">
     <ul role="menu">
         <li class="menu-item" role="none">
-            {% pageurl site.root_page.localized as root_url %}
+            {% if skip_nav != "true" %}
```
rlskoeser, line 8: minor, but it looks a little strange to check for `"true"`. Could we do `skip_nav=1` instead when calling the template and then this could just check `if not skip_nav` ? Or will that not work if skip nav is undefined?

rlskoeser (approved):
The code looks much cleaner to me, thank you. 

Not sure how Gissoo will feel about such limited nav for the 500 — but it should hopefully be a pretty rare occurrence! And it's sufficient to get people somewhere else.

geniza/templates/404.html
rlskoeser: I agree, doesn't seem worth it to create a base error template.

geniza/templates/footer.html
blms: You're right! It works fine like that, done in d74e959 👍

gissoo:
@blms thank you!! Everything looks good, I can't test the image measurements on Percy though, but I think they look fine. (except there is an issue with font sizes for the page title of the server error on mobile, and the page title of the 404 error on desktop on light mode.) but I see it's not included in this issue.

PR #546, opened 2022-01-25
#401 - Alternative transcription view

blms:
## What this PR does
- Per #401:
  - Provides an alternative to #544 by implementing a stripped-down version of Gissoo's designs
- Pulls new test fixtures from #544

## Todo
- [x] Double check what it should look like when there is only images but no transcription, and vice versa
- [x] Click to view larger image
- [x] Template tag
- ~~Slideshow in html/css~~

## Code review questions
- What functionality is missing from this for MVP? I've been working on a sort of slideshow to page through the image list, but not sure if that's MVP. Not sure about zoom/etc either.
- Does the logic make sense so far? I was going to use the `iiif_thumbnails` function from `Fragment`, but I wanted to be able to use a larger image size (500px) and wrap them in `<li>`s, so I defined a new `Document.iiif_images` function.
  - The `<li>`s will be necessary for a slider, but are already semantically useful without it.


geniza/corpus/models.py
```diff
@@ -527,6 +527,20 @@ def iiif_urls(self):
             )
         )
 
+    def iiif_images(self):
```
rlskoeser, line 530: this is more display logic than I'd prefer in the models; we should think about adapting this template tag from mep (or maybe it could be added to `djiffy` for re-use):

https://github.com/Princeton-CDH/mep-django/blob/main/mep/common/templatetags/mep_tags.py#L149-L172

rlskoeser (commented):
Wow, I find this so much more appealing and usable than what we're getting from mirador right now!

geniza/corpus/templates/corpus/snippets/document_transcription.html
```diff
@@ -1,15 +1,23 @@
...
+                {% for image_dict in document.iiif_images %}
+                    <li id="image-{{ forloop.counter }}">
+                        <a href="{{ image_dict.image|iiif_image:"size:width=2880" }}">
+                            <img src="{{ image_dict.image|iiif_image:"size:width=500" }}" alt="{{ image_dict.label }}" title="{{ image_dict.label }}" loading="lazy">
+                        </a>
+                    </li>
+                {% endfor %}
+            </ol>
+        </noscript>
+        <div id="iiif-images" data-iiif-urls="{{ document.iiif_image_ids|iiif_info_json }}">
```
rlskoeser, line 19: if we reference the piffle image objects, we could use the `info()` method

(I don't know if we have easy access to them here; this seems clunky to me but I think acceptable)

rlskoeser (commented):
Tested this locally and it looks really nice — I like the easy image rotation (which I don't think we had in mirador yet — I think it's possible, I just don't know how to configure it).

The noscript fallback looks good too.

There's plenty more we could do to improve this, but I don't think we should invest in any of it now:
- would be nice to have image labels from the manifests (recto/verso)
- the seadragon buttons don't fit that well with the site visuals (🤔 how hard to use equivalent icons from phosphor? don't want to add design work, but maybe that would be an easy solution)

I do think we need to display attribution (and possibly logo?) from the IIIF manifests for the images we're displaying, so that we make the source and permissions clear. (I know the CUDL manifests have an attribution statement and some limitations on use.)

geniza/corpus/templates/corpus/snippets/document_transcription.html
blms: Done in 9b3b1d6, much simpler, thanks!

blms:
Thanks @rlskoeser, implemented the above suggestions. Probably good to do another quick review and then can deploy to QA for testing. 

If approved, I have this checklist of issues to create post-merge:
- [ ] Percy still [not showing](https://percy.io/2cf28a24/geniza/builds/15348958?utm_campaign=2cf28a24&utm_content=geniza&utm_source=github_status_public) document detail page
- [ ] Adding labels to images on OpenSeaDragon
- [ ] Using Phosphor icons to replace OpenSeaDragon defaults

geniza/corpus/templatetags/corpus_extras.py
```diff
@@ -93,9 +93,20 @@ def iiif_image(img, args):
...
+
+@register.filter
+def format_attribution(attribution):
+    (attribution, additional_restrictions, extra_attrs_set) = attribution
+    extra_attrs = "\n".join("<p>%s</p>" % attr for attr in extra_attrs_set)
+    return '<div class="attribution"><p>%s</p><p>%s</p>%s</div>' % (
+        attribution,
+        additional_restrictions,
+        extra_attrs,
+    )
```
rlskoeser, line 112: this is awfully specific for a template tag; any reason this couldn't just be done in the template?

geniza/corpus/models.py
```diff
@@ -622,6 +607,36 @@ def sources(self):
         """All unique sources attached to footnotes on this document."""
         return Source.objects.filter(footnote__document=self).distinct()
 
+    def attribution(self):
```
rlskoeser, line 610: wow, thanks for generalizing this.

Now that I see it on the document — so much clearer than wherever it gets displayed in mirador — I think we need to figure out how to clarify what the Cambridge rights statement applies to (especially since we're not! displaying their metadata). Maybe display per shelfmark? Or group of shelfmark if they're the same, but sometimes there is small variation (IIRC some of the Cambridge attribution text had links but some did not; otherwise they were the same.)  

I don't think we need to do anything about this right away, though! Let's get feedback and go-ahead on the approach first.

rlskoeser (commented):
your checklist looks reasonable to me; we may want to add something to tweak how we're displaying attribution, but let's get input on it first

geniza/corpus/templatetags/corpus_extras.py
blms: I thought it might be nice to have that formatting function available to the ManifestView as well, but not sure what we're doing with that atm, so maybe it makes sense to just move it into the template.

blms:
@rlskoeser Are you satisfied to merge this now and create the issues from the checklist, or is there anything else we are waiting on / code things I should address first?

rlskoeser:
@blms lets merge it! We can work on improvements as we get to them, I think it will help to get this initial version merged in sooner than later so we can start working with it.

PR #548, opened 2022-01-27
needs review alternate view for document admin list display

rlskoeser:
Had an idea about a different way to solve the "needs review" queue display problem — add a toggle to the template and control the display with css. What do you think?

### DIFF
--- a/geniza/corpus/admin.py
+++ b/geniza/corpus/admin.py
@@ -104,7 +104,7 @@ class LanguageScriptAdmin(admin.ModelAdmin):
 
     @admin.display(
         ordering="secondary_document__count",
-        description="# documents where this is a secondary langaug",
+        description="# documents where this is a secondary language",
     )
     def secondary_documents(self, obj):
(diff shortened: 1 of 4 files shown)


blms (approved):
I like this approach! Very simple and clean. One thing I noticed is that when you try to sort by a column after selecting "needs review" mode, the needs review column disappears again. Not sure if there's an easy way to fix that, though.

rlskoeser:
> I like this approach! Very simple and clean. One thing I noticed is that when you try to sort by a column after selecting "needs review" mode, the needs review column disappears again. Not sure if there's an easy way to fix that, though.

You're right! And it also doesn't persist if you choose another page of results. I didn't really play around with it very much. Probably most changes won't persist it, which could be annoying. Might need to use js to persist the display mode in local storage, which wouldn't be too hard (and could also make the column display logic more semantic). I think I'd like to have the team test this version first before investing more in it, though.

Issue #547, opened 2022-01-27
Clean up Document Details page


blms:
## dev notes
- [x] Revise document title in detail view to show all shelfmarks
- [x] Remove footnote `location` from doc detail page
- [x] Handle unknown source for download edition text

blms:
Confirmed all changes are present in QA site.

PR #551, opened 2022-01-28
#547 - Document detail view tweaks

blms:
## What this PR does

- Per #547:
  - Removes `shelfmark_display` method, and instead just uses shelfmark string method to display all shelfmarks in document titles
  - Removes footnote location from detail view; moves test for footnote location presence to scholarship records template
  - Fixes "download edition" link for sources with no authorships

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -688,7 +688,7 @@ class Document(ModelIndexable):
                 "description_t": self.description,
                 "notes_t": self.notes or None,
                 "needs_review_t": self.needs_review or None,
-                "shelfmark_ss": [f.shelfmark for f in fragments],
+                "shelfmark_ss": self.certain_join_shelfmarks,
                 # library/collection possibly redundant?
                 "collection_ss": [str(f.collection) for f in fragments],
                 "tags_ss_lower": [t.name for t in self.tags.all()],
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -461,18 +461,22 @@ class Document(ModelIndexable):
             )
         )
 
+    @property
+    def certain_join_shelfmarks(self):
+        return list(
+            dict.fromkeys(
+                block.fragment.shelfmark
+                for block in self.textblock_set.filter(certain=True)
+            ).keys()
+        )
+
     # NOTE: not currently used; remove or revise if this remains unused
     @property
     def shelfmark_display(self):
         """First shelfmark plus join indicator for shorter display."""
         # NOTE preliminary pending more discussion and implementation of #154:
         # https://github.com/Princeton-CDH/geniza/issues/154
-        certain = list(
-            dict.fromkeys(
-                block.fragment.shelfmark
-                for block in self.textblock_set.filter(certain=True)
-            ).keys()
-        )
+        certain = self.certain_join_shelfmarks
         if not certain:
             return None
         return certain[0] + (" + …" if len(certain) > 1 else "")
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -437,7 +437,7 @@ class Document(ModelIndexable):
         # ordering = [Least('textblock__fragment__shelfmark')]
 
     def __str__(self):
-        return f"{self.shelfmark_display or '??'} (PGPID {self.id or '??'})"
+        return f"{self.shelfmark or '??'} (PGPID {self.id or '??'})"
 
     @staticmethod
     def get_by_any_pgpid(pgpid):
@@ -458,21 +458,6 @@ class Document(ModelIndexable):
                 if block.certain  # filter locally instead of in the db
             )
         )
-
-    @property
-    def shelfmark_display(self):
-        """First shelfmark plus join indicator for shorter display."""
-        # NOTE preliminary pending more discussion and implementation of #154:
-        # https://github.com/Princeton-CDH/geniza/issues/154
-        certain = list(
-            dict.fromkeys(
-                block.fragment.shelfmark
-                for block in self.textblock_set.filter(certain=True)
-            ).keys()
-        )
-        if not certain:
-            return None
-        return certain[0] + (" + …" if len(certain) > 1 else "")
 
     @property
     def collection(self):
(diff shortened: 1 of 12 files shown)


geniza/corpus/templates/corpus/document_detail.html
```diff
@@ -84,7 +84,7 @@ <h3>
             <section class="transcription-link">
                 {% for ed in document.digital_editions %}
                     {% if ed.content.text %}
-                        <p><a href="{% url "corpus:document-transcription-text" document.pk ed.pk %}">Download {% for auth in ed.source.authorship_set.all %}{% include "snippets/comma.html" %}{{ auth.creator.last_name }}{% endfor %}'s edition.</a></p>
+                        <p><a href="{% url "corpus:document-transcription-text" document.pk ed.pk %}">Download {% if ed.source.authorship_set.count > 0 %}{% for auth in ed.source.authorship_set.all %}{% include "snippets/comma.html" %}{{ auth.creator.last_name }}{% endfor %}'s {% endif %}edition.</a></p>
```
rlskoeser, line 87: is `for` / `empty` useful at all here? admittedly "[unknown]'s transcription" doesn't look great. But shouldn't we indicate that the transcription is from an unknown source?

geniza/corpus/models.py
```diff
@@ -459,21 +459,6 @@ def shelfmark(self):
...
-            dict.fromkeys(
-                block.fragment.shelfmark
-                for block in self.textblock_set.filter(certain=True)
-            ).keys()
-        )
-        if not certain:
-            return None
-        return certain[0] + (" + …" if len(certain) > 1 else "")
-
     @property
```
rlskoeser, line 462: Are we not using this anywhere anymore?

I am hesitant to remove it entirely because I suspect we'll still end up using some version of this (maybe with a minimum number of shelfmarks that is higher than 1). Would prefer to keep the method and test with a note in the comments that it isn't currently used and should either be revised or removed if we don't end up using it.

geniza/corpus/templates/corpus/snippets/document_header.html
```diff
@@ -16,5 +16,5 @@
     {# Translators: Default label when document does not have a type #}
     {% translate 'Unknown type' as unknown_type %}
     <span class="doctype">{{ document.doctype|default:unknown_type }}</span>
-    <span class="shelfmark">{{ document.shelfmark_display }}</span>
+    <span class="shelfmark">{{ document.shelfmark }}</span>
```
rlskoeser, line 19: I think we need a formatted version so you can wrap individual shelfmarks with a tag and set white space not to wrap in the styles.

not sure we need a new method, though — maybe just iterate and generate here in the template? but should probably do something similar for search results; maybe a template snippet would work for both cases

rlskoeser (changes requested):
looks good, just noticed a couple of small things

geniza/corpus/templates/corpus/document_detail.html
blms: I never knew about `{% empty %}` before, that's so useful! I see now it's used on the needs review template as well 😄 

Done in 3683dc1 👍

geniza/corpus/models.py
blms: Done in 15fde44 👍

geniza/corpus/templates/corpus/snippets/document_header.html
blms: Done in f3aacc7 👍

geniza/corpus/templates/corpus/snippets/document_shelfmarks.html
```diff
@@ -0,0 +1,10 @@
+{% for block in document.textblock_set.all %}
+    {% if block.certain %}
+        <span>{{ block.fragment.shelfmark }}</span>
+    {% endif %}
+{% empty %}
+    {# search results page uses list "shelfmark" instead of textblocK_set #}
```
rlskoeser, line 6: nice solution for handling both cases

geniza/corpus/templates/corpus/snippets/document_shelfmarks.html
```diff
@@ -0,0 +1,10 @@
+{% for block in document.textblock_set.all %}
+    {% if block.certain %}
+        <span>{{ block.fragment.shelfmark }}</span>
```
rlskoeser, line 3: I think we need to add the ` + ` in the template rather than with css — when I tested locally and tried to cut and paste the shelfmark, it didn't include the ` + `; I suppose that will require extra logic to work with uncertain shelfmarks, though ...

rlskoeser (changes requested):
this looks good — my only concern is that adding the `+` with css makes it not quite regular text and not selectable for copy/paste; I think we should generate it in the template, but that might mean we need a small model method to generate a list of certain shelfmarks

rlskoeser:
one more thing, not sure relevant for this task — document title on mobile doesn't look like it has the right container margins, it goes almost all the way to the edges (noticed when I was looking at a doc with lots of shelfmarks)

geniza/corpus/templates/corpus/snippets/document_shelfmarks.html
blms: Done in e14654b 👍

geniza/corpus/templates/corpus/snippets/document_shelfmarks.html
```diff
@@ -1,10 +1,10 @@
 {% for block in document.textblock_set.all %}
     {% if block.certain %}
-        <span>{{ block.fragment.shelfmark }}</span>
+        <span>{{ block.fragment.shelfmark }}{% if not forloop.last %} + {% endif %}</span>
```
rlskoeser, line 3: This logic won't display correctly for uncertain joins if there are multiple or the uncertain join doesn't come last, but I just checked — there are only seven, they each only have two associated shelfmarks, and only one of them shows the problem (an extra +) because of the order of the certain / uncertain. I think we can live with that for now, until we address the uncertain join display problem properly.

rlskoeser: (alternative would be to make a `certain_shelfmarks` method that lists only those so the `forloop.last` is accurate)

rlskoeser (approved):

rlskoeser (approved):

Issue #486, opened 2022-01-04
Display Typed Text Volume Name in Footnote Detail Source Box


richmanrachel:
## testing notes (QA)

- [x] In the admin interface, confirm the volume names appear for unpublished sources (see below for more detail)
- [x] in the public document detail view, confirm that the volume for unpublished sources is not displayed

---

**Is your feature request related to a problem? Please describe.**
When trying to clean up Goitein's typed texts into volumes, we discovered that you can't easily change the source from the footnote detail view. In contrast to published works in volumes (such as Gil, Kingdom of Ishmael, vol. 3), Goitein's typed texts don't display their volume number in the drop down menu and therefore are hard to sort.

Here's a screenshot for reference:

![image](https://user-images.githubusercontent.com/68028846/148111987-7b46ae3b-21ec-469d-842d-7a65e6765671.png)

**Describe the solution you'd like**
Hopefully this is just an easy fix, updating the code to pull volume number/letters wherever they exist into the source drop down menu.

rlskoeser:
@blms I think this is a fairly easy fix — we need to add the volume back to the string method for unpublished sources; my only concern is that I'm not sure where else we're using the string method, it might require changing the method we use to display sources elsewhere

richmanrachel:
It works! Such an easy but useful improvement - thank you!

PR #552, opened 2022-01-28
#486 - Add volume back to unpublished source string method

blms:
## What this PR does
- Per #486:
  - Adds source volume for unpublished sources back to the `__str__` method; adds a new method `formatted_stripped` to allow displaying without it

## dev notes
- I think the only other place the source string method was called, besides the admin interface, was `footnote.display()`. Otherwise the frontend uses `source.formatted_display()`. Running Percy just in case! (edit: confirmed in Percy)

### DIFF
--- a/geniza/footnotes/models.py
+++ b/geniza/footnotes/models.py
@@ -132,12 +132,14 @@ class Source(models.Model):
         ordering = ["title", "year"]
 
     def __str__(self):
+        """Method used for for internal/data admin use.
+        Please use the `display` or `formatted_display` methods for public display."""
         # Append volume for unpublished (e.g. typed texts)
         if self.source_type.type == "Unpublished" and self.volume:
-            return self.formatted_stripped() + ", " + self.volume
+            return "%s (%s)" % (self.display(), self.volume)
         # Otherwise return formatted display without html tags
         else:
-            return self.formatted_stripped()
+            return self.display()
 
     def all_authors(self):
         """semi-colon delimited list of authors in order"""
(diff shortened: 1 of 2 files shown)


geniza/footnotes/models.py
```diff
@@ -132,8 +132,12 @@ class Meta:
         ordering = ["title", "year"]
 
     def __str__(self):
-        """strip HTML tags and trailing period from formatted display"""
-        return strip_tags(self.formatted_display(extra_fields=False))[:-1]
+        # Append volume for unpublished (e.g. typed texts)
+        if self.source_type.type == "Unpublished" and self.volume:
+            return self.formatted_stripped() + ", " + self.volume
```
rlskoeser, line 137: I think previously volume was in parentheses, but IDK how much that matters.

Might be good to add a comment here that the string method is for internal/data admin use and the display method should be used for public display?

geniza/footnotes/models.py
```diff
@@ -335,6 +339,10 @@ def formatted_display(self, extra_fields=True):
     all_authors.short_description = "Authors"
     all_authors.admin_order_field = "first_author"  # set in admin queryset
 
+    def formatted_stripped(self):
```
rlskoeser, line 342: don't love the method name (not very intelligible); could we just call it display for parallel with `formatted_display` ?

geniza/footnotes/models.py
```diff
@@ -335,6 +339,10 @@ def formatted_display(self, extra_fields=True):
...
+        """strip HTML tags and trailing period from formatted display"""
+        return strip_tags(self.formatted_display(extra_fields=False))[:-1]
```
rlskoeser, line 344: seems like it would be more semantic and less brittle to use `.rstrip(".")`

why do we strip the trailing period from this version?

blms: We'd stripped it for a purpose will be made obsolete by #551 😆  (using a trailing comma before footnote location in doc detail page). Removed in af1b073

geniza/footnotes/models.py
blms: Done in af1b073 👍

geniza/footnotes/models.py
blms: Done in af1b073 👍

rlskoeser (approved):

PR #553, opened 2022-01-31
225 admin document join

kmcelwee:
@rlskoeser I'd appreciate a quick look-through to double check that I haven't missed anything significant. I'll be implementing Rachel's request tomorrow morning. I still haven't double-checked everything on my own, but I'd love to know if I'm on the right track


geniza/corpus/forms.py
```diff
@@ -168,3 +171,41 @@ def clean(self):
...
+        help_text="Provide a brief explanation for merging these records.",
+    )
+
+    def __init__(self, *args, **kwargs):
+        document_ids = kwargs.get("document_ids", [])
+        # ? : I don't understand what's going on here.
+        try:
+            del kwargs["document_ids"]
+        except KeyError:
+            pass
```
rlskoeser, line 206: looks like just deleting it from kwargs before we call the parent init method, because it otherwise would cause problems; you could uncomment to confirm and then add a better comment

rlskoeser (commented):
I didn't pull it to run locally, but it looks reasonable to me so far.  Added comments on the things I noticed (but recognize it's WIP)

I would make sure the functionality is correct first before spending much time on the document display on the form.

geniza/corpus/models.py
```diff
@@ -728,15 +728,14 @@ def merge_with(self, merge_docs, rationale, user=None):
...
-
         # initialize old pgpid list if previously unset
         if self.old_pgpids is None:
             self.old_pgpids = []
 
         # if user is not specified, log entry will be associated with
         # script and document will be flagged for review
         script = False
-        if user is None:
+        if user is None or user.is_anonymous:
```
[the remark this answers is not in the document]
kmcelwee: Right. This was a short term fix for failing tests. How would I create a user with the correct permissions? I try `admin_client.force_login(user)` where I've defined a user with the expected permissions as we did in S&Co, but I keep getting redirected to the CAS login page. If you have time, this is something I'd appreciate us working through together

geniza/corpus/urls.py
```diff
@@ -34,4 +34,9 @@
         name="document-annotations",
     ),
     path("export/pgp-metadata-old/", corpus_views.pgp_metadata_for_old_site),
+    path(
+        "documents/merge/",
```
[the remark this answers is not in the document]
kmcelwee: The name of the link is still “corpus:document-merge” is that ok? Searching around, I'm not seeing obvious ways to add an admin route. Also, It now prepends “/en/“, and I don’t think that’s what we want.

geniza/corpus/admin.py
```diff
@@ -467,6 +468,21 @@ def tabulate_queryset(self, queryset):
...
 
+    @admin.display(description="Merge selected documents")
+    def merge_documents(self, request, queryset=None):
+        """Merge selected documents together"""
+        # Functionality drawn from https://github.com/Princeton-CDH/mep-django/blob/main/mep/people/admin.py
+        # ? : Should we raise an error if there are no selected documents?
+
+        # NOTE: using selected ids from form and ignoring queryset
+        # because we can't pass the queryset via redirect
+        # ? : Is this correct? Is there a way to just pass the queryset?
```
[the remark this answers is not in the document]
kmcelwee: The django docs for our version recommend exactly this (doesn't change in later versions), so I think this is the best bet.

geniza/corpus/models.py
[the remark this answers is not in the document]
rlskoeser: pytest-django has an admin client fixture — https://pytest-django.readthedocs.io/en/latest/helpers.html#admin-client-django-test-client-logged-in-as-admin 

would you see if that works here? I think I've used it successfully before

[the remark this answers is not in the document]
kmcelwee: right! that's what I'm using. But it has a null user id, so the DB fails because it tries to create a log entry with a null user id.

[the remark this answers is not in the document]
rlskoeser: wow! did not know the fixture had a null user id, that seems weird! we might need a local admin user fixture, in that case

geniza/corpus/admin.py
```diff
@@ -467,6 +468,21 @@ def tabulate_queryset(self, queryset):
         "collection",
     ]
 
+    @admin.display(description="Merge selected documents")
+    def merge_documents(self, request, queryset=None):
+        """Merge selected documents together"""
+        # Functionality drawn from https://github.com/Princeton-CDH/mep-django/blob/main/mep/people/admin.py
+        # ? : Should we raise an error if there are no selected documents?
```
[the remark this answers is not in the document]
kmcelwee: I think I'll try showing an error in the template if there are fewer than two documents selected. (You can't perform an action with no documents selected, but you can select one. Also the route is still accessible with no documents selected.)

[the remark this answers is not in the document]
rlskoeser: Ok, thanks for checking.  If it's simpler (and seems like it would be?), you could set a message and redirect to the document list view.

geniza/corpus/templates/admin/corpus/document/merge.html
```diff
@@ -0,0 +1,29 @@
...
+{% block breadcrumbs %}
+    {% if not is_popup %}
+        <ul>
+            <li><a href="{% url 'admin:index' %}">Home</a></li>
+            <li><a href="{% url 'admin:app_list' app_label='corpus' %}">Corpus</a></li>
+            <li><a href="{% url 'admin:corpus_document_changelist'%}">Document</a></li>
+            <li>Merge selected documents</li>
+        </ul>
+    {% endif %}
+{% endblock %}
```
rlskoeser, line 13: These breadcrumbs don't match the styles for the rest of the admin site; is this still based on the mep/grappelli version?

See if there's a base template that's closer to what we need, or else adjust to be closer to the format django uses for displaying breadcrumbs.

geniza/corpus/templates/admin/corpus/document/merge.html
```diff
@@ -0,0 +1,29 @@
...
+
+{% block content_title %}
+    <h1>Merge selected documents</h1>
+{% endblock %}
+
+{% block content %}
```
rlskoeser, line 19: Let's add a warning here, maybe something like:

> Note: there is no automated way to unmerge documents! Please review to make sure these documents should be merged before submitting the form.

geniza/corpus/templates/admin/corpus/document/merge.html
```diff
@@ -0,0 +1,29 @@
...
+    <form method="post">
+        {% csrf_token %}
+        {{ form.as_p }}
```
rlskoeser, line 22: Please add some light override css to adjust the form display to make it easier to read; you can probably just add these to our existing local admin css. 

I would like to see the bullets for the list suppressed and the radio input floated to the left of the document summary block.

geniza/corpus/templates/corpus/snippets/document_option_label.html
```diff
@@ -0,0 +1,25 @@
+{# template snippet for displaying document label on a form #}
+{# used on document merge form to provide enough information to merge accurately #}
+<div class="merge-document-label">
+    <strong><a target="_blank" href="{% url 'admin:corpus_document_change' document.id %}">{{ document }}</a></strong>
```
rlskoeser, line 4: I wanted to click the document title to select the primary, so I didn't like the fact that it was linked (although linking to the change form is helpful!). Let's follow django admin convention and display a pencil edit icon to the left of the document id as the link to the document change form.

rlskoeser, line 4: please use an appropriate heading tag instead of `<strong>`

geniza/corpus/templates/corpus/snippets/document_option_label.html
```diff
@@ -0,0 +1,25 @@
...
+    <p>PGPID {{ document.id }}</p>
+    {% if document.description %}
+        <p>
+            <dt>Description</dt>
+            <dd>{{ document.description }}</dd>
+        </p>
+    {% endif %}
+
+    {% if document.footnotes.count %}<strong>Footnotes</strong>{% endif %}
```
rlskoeser, line 13: here also, use an appropriate heading tag instead of `<strong>`

and move the `endif` to include the `<ul>` so we don't get an empty list if there are no footnotes

rlskoeser, line 13: Let's rename to scholarship records instead of footnotes (not sure what h-level this should be)

```suggestion
    {% if document.footnotes.count %}<strong>Scholarship Records</strong>{% endif %}
```

geniza/corpus/templates/corpus/snippets/document_option_label.html
```diff
@@ -0,0 +1,25 @@
...
+    <ul>
+        {% for footnote in document.footnotes.all %}
+            <li><a target="_blank" href="{% url 'admin:footnotes_footnote_change' footnote.id %}">{{ footnote }}</a>
+                {% if footnote.source %}<dt>Source</dt>
+                    <dd>{{ footnote.source }}</dd>{% endif %}
+                {% if footnote.url %}<dt>URL</dt>
+                    <dd><a target="_blank" href="{{ footnote.url }}">{{ footnote.url }}</a></dd>{% endif %}
+            </li>
+
```
rlskoeser, line 22: This display is not very usable. So that we don't have to reinvent the wheel here, can we adapt existing designs and use the pubic view format in `document_scholarship.html`, and add in the pencil icon with links to the admin change form?  (I don't know if there may be display problems because we don't have the styles here — I don't propose we add them, it doesn't need to look the same — just suggesting we should adapt that display if possible.)

rlskoeser: Rachel said that it's really important for them to know which footnotes have content attached; you're showing the links (although not the labels), but you're not showing any transcription content. 

If you look at the admin footnote change form, you'll see how we're displaying it there — I think you should include the same template snippet here (`footnotes/transcription.html`). However, because transcriptions can be very long, I suggest wrapping the transcription in a disclosure element — otherwise I think it may be hard for them to view the documents together and compare.

geniza/corpus/templates/corpus/snippets/document_option_label.html
```diff
@@ -0,0 +1,25 @@
+{# template snippet for displaying document label on a form #}
+{# used on document merge form to provide enough information to merge accurately #}
+<div class="merge-document-label">
+    <strong><a target="_blank" href="{% url 'admin:corpus_document_change' document.id %}">{{ document }}</a></strong>
+    <p>PGPID {{ document.id }}</p>
+    {% if document.description %}
```
rlskoeser, line 6: I think we should always have a description — and if not, it would be useful for them to know it's empty.

geniza/corpus/templates/corpus/snippets/document_option_label.html
```diff
@@ -0,0 +1,25 @@
...
+        <p>
+            <dt>Description</dt>
+            <dd>{{ document.description }}</dd>
+        </p>
+    {% endif %}
```
rlskoeser, line 11: I know Rachel didn't request it, but the merge function includes languages, tags, notes, and needs review — and I suspect some of them at least would be helpful here; in many cases there are documents now that have notes indicating which records they should be merged with.  Should we check with her about which of those should be added to the display?

geniza/corpus/templates/admin/corpus/document/merge.html
```diff
@@ -0,0 +1,29 @@
...
+
+        <div>
+            <input type="submit" value="Submit">
+        </div>
+    </form>
+
```
rlskoeser, line 28: Let's add another note here (we may want to get input on wording from Rachel):

> Please note: if this merge results in a document with multiple TEI transcriptions, it is likely that the TEI files will need to be updated or merged in order for the transcription synchronization to work properly.

geniza/corpus/views.py
```diff
@@ -459,6 +463,67 @@ def get(self, request, *args, **kwargs):
...
+                "admin:corpus_document_change", args=[primary_doc.id]
+            )
+            messages.success(
+                self.request,
+                mark_safe(
+                    f"Successfully merged document(s) {secondary_doc_str} with {primary_doc_str}."
+                ),
+            )
+
+        # error if person has more than one account, no account
```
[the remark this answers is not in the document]
rlskoeser: codecov flags this as not being covered — IDK if because you don't have a test case that triggers it, or if it's not possible to happen with geniza data

geniza/corpus/views.py
```diff
@@ -459,6 +463,67 @@ def get(self, request, *args, **kwargs):
...
+
+    def get_initial(self):
+        # Default to first document selected
+        document_ids = self.request.GET.get("ids", None)
+        if document_ids:
+            self.document_ids = [int(pid) for pid in document_ids.split(",")]
+            # by default, prefer the first record created
+            return {"primary_document": sorted(self.document_ids)[0]}
+        else:
+            self.document_ids = []
```
rlskoeser, line 487: codecov says this isn't covered

rlskoeser: (what happens if there are no documents selected?)

rlskoeser (changes requested):
This is looking really good and works well!

I noticed a few small things, and the review display and formatting needs some work.

Also needs a few more test cases (and/or remove the catch for an error that's not possible with our data).

I wondered if there is a practical/actual maximum on how many documents can be merged at once — IDK if it would come up in a real world scenario, but might be good to do a little testing to check — it would be nice to know what happens if someone tries to merge more than a few documents.

geniza/corpus/views.py
```diff
@@ -459,6 +463,67 @@ def get(self, request, *args, **kwargs):
         return JsonResponse(dict(annotation_list), encoder=iiif_utils.AttrDictEncoder)
 
 
+class DocumentMerge(FormView, PermissionRequiredMixin):
+    permission_required = ("corpus.change_document", "corpus.delete_document")
```
kmcelwee, line 467: @rlskoeser This view is open to the public, despite permissions being set. I tried updating to `permission_required = ("change_document", "delete_document")` after seeing the codenames related to `Permissions.objects.filter(codename__contains="document")` but that too didn't work. What's worse, incorrect permissions don't raise errors.  Any ideas?

rlskoeser: It's the order!  (One of the checks must get skipped if the permission mixin is not first...)

You need to switch them:
`class DocumentMerge(PermissionRequiredMixin, FormView):`

geniza/corpus/urls.py
```diff
@@ -34,4 +34,9 @@
         name="document-annotations",
     ),
     path("export/pgp-metadata-old/", corpus_views.pgp_metadata_for_old_site),
+    path(
+        "admin/corpus/document/merge/",
+        corpus_views.DocumentMerge.as_view(),
+        name="document-merge",
+    ),
```
rlskoeser, line 41: Let's configure the url for this on the admin model instead of doing it here in the main corpus config. Here's the relevant section of the docs:
https://docs.djangoproject.com/en/4.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_urls

geniza/corpus/views.py
kmcelwee: Really?!? Yeah that was it 😄thanks for the help

geniza/corpus/views.py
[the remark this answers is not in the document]
kmcelwee: I was unable to trigger this error. I assumed it would show up if a bad ID was provided. I've removed the try/except block entirely. Stop me if necessary.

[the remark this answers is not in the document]
rlskoeser: I was pretty sure from the comment that it wouldn't come up here (looked like it was related to merging people with one or more accounts in mep data model), so if you can't trigger it I take that as confirmation! Glad we were able to remove it and simplify.

geniza/corpus/templates/corpus/snippets/document_option_label.html
kmcelwee: Ok! I made this true for all fields, e.g. tags, notes, etc. I assumed the same would be true for those fields as well

kmcelwee:
The two things that seem to be left to do are:

1. Styling the document summary: All the information is there, but since we're pulling in existing templates, we're without the correct styling, making it look weird.
2. Placing the merge document view as an admin URL

I've added these todos to the ticket.

geniza/corpus/templates/admin/corpus/document/merge.html
```diff
@@ -0,0 +1,29 @@
...
+{% block title %} Merge selected documents {% endblock %}
+
+{% block breadcrumbs %}
+    {% if not is_popup %}
+        <ul>
+            <li><a href="{% url 'admin:index' %}">Home</a></li>
+            <li><a href="{% url 'admin:app_list' app_label='corpus' %}">Corpus</a></li>
+            <li><a href="{% url 'admin:corpus_document_changelist'%}">Document</a></li>
+            <li>Merge selected documents</li>
+        </ul>
```
rlskoeser, line 11: these tags don't match what other django admin templates output, looks like we need a `div` with `a` tags

geniza/corpus/templates/admin/corpus/document/merge.html
```diff
@@ -0,0 +1,29 @@
+{% extends 'admin/base_site.html' %}
```
rlskoeser, line 1: I wonder if there is a template closer to what we need? maybe the change form? (but possibly not, could end up overwriting a lot more)

geniza/corpus/templates/corpus/snippets/document_option_label.html
```diff
@@ -0,0 +1,68 @@
...
+        <dt>Description</dt>
+        <dd>{{ document.description }}</dd>
+        <dt>Languages</dt>
+        <dd>{{ document.all_languages }}</dd>
+        <dt>Tags</dt>
+        <dd>{{ document.all_tags }}</dd>
+        <dt>Notes</dt>
+        <dd>{{ document.notes }}</dd>
+        <dt>Needs Review</dt>
+        <dd>{{ document.needs_review }}</dd>
```
rlskoeser, line 30: I don't think we should display these fields if they are empty. And I think we might want to put the labels and text on the same line so the display is more consolidated (at least for non-note / textarea fields)

geniza/corpus/templates/corpus/snippets/document_option_label.html
```diff
@@ -1,25 +1,68 @@
...
+    .transcription-container {
+        max-height: 350px;
+        overflow: scroll;
+    }
+
+    ul>li {
+        list-style-type: none;
+    }
+</style>
+
```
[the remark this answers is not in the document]
blms: Done in ffe557a 👍

geniza/corpus/templates/admin/corpus/document/merge.html
blms: Done in ffe557a 👍

geniza/corpus/templates/corpus/snippets/document_option_label.html
blms: Done in ffe557a 👍

geniza/corpus/templates/admin/corpus/document/merge.html
blms: This template provides about the same amount of what's needed as the change_form.

geniza/corpus/urls.py
blms: Done in 211b59d 👍

rlskoeser (approved):
Everything looks great!

One question: should we set a max number of documents we allow for the merge, or just trust content editors to be reasonable when they use the feature? The form loads very slowly if you select e.g. a page worth of documents, but I think that would only happen by mistake and is probably not a real-world scenario.

blms:
> Everything looks great!
> 
> One question: should we set a max number of documents we allow for the merge, or just trust content editors to be reasonable when they use the feature? The form loads very slowly if you select e.g. a page worth of documents, but I think that would only happen by mistake and is probably not a real-world scenario.

@rlskoeser I agree. I think we can let content editors make case by case decisions, and trust them not to e.g. select a whole page!

PR #554, opened 2022-02-01
Update solr_conf copy command

kmcelwee:
Can I just make this small change? I know I won't make this mistake again, but it took more of my time than it should have

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -73,7 +73,7 @@ Remember to add a ``SECRET_KEY`` setting!
 
 - Copy Solr configset into your solr server configset directory. For a local install::
 
-    cp -r solr_conf /path/to/solr/server/solr/configsets/geniza
+    cp -r solr_conf/conf /path/to/solr/server/solr/configsets/geniza
     chown solr:solr -R /path/to/solr/server/solr/configsets/geniza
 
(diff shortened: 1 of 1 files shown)


rlskoeser (changes requested):
I think the command depends on whether you have copied it once or not — and I'm pretty sure this one won't work for the initial setup. I couldn't figure out one docker command that worked for both the initial copy and for updating (although I didn't put a ton of time into it).

I agree we should update the notes, but please either figure out syntax that will work for both cases, or document them separately.

rlskoeser:
My comments earlier were for the `docker cp` command, which I haven't been able to figure out a way to use the same syntax for both new and updates — but looking again I see this is a local command, and we don't include the docker examples! There should be a way to write one command that works for both cases. Trailing slash, I think?

kmcelwee:
I'm 98 percent certain this is the idempotent way of copying either `conf` to the directory or the contents of `conf` to the existing `conf` directory. Did you have reason to think otherwise?

rlskoeser:
> I'm 98 percent certain this is the idempotent way of copying either `conf` to the directory or the contents of `conf` to the existing `conf` directory. Did you have reason to think otherwise?

When you're doing the initial setup the configset doesn't exist yet.

PR #557, opened 2022-02-01
522 show many tags

kmcelwee:
Two questions:
* Do we want to make a `max_tag_count` for the template? Because it shows up in different forms each times it's used (e.g. as a string in a slice, as an int for a boolean, and as a negative string for the `add` function), I'm not sure how much value it would add.
* Please confirm that the image below is the wording and appearance you need. The `#` at the start is controlled by a CSS `:before` element, which I don't think I can change in-line. Before I start on that, would you mind confirming that this is what we were generally shooting for?

Screen Shot 2022-02-01 at 1 51 52 PM

### DIFF
--- a/geniza/corpus/templates/corpus/snippets/document_result.html
+++ b/geniza/corpus/templates/corpus/snippets/document_result.html
@@ -95,6 +95,9 @@
                     {% for tag in document.tags|alphabetize|slice:":5" %}
                         <li><a href='{% url "corpus:document-search" %}?q=tag:"{{ tag }}"'>{{ tag }}</a></li>
                     {% endfor %}
+                    {% if document.tags|length > 5 %}
+                        <li>{{ document.tags|length|add:"-5" }} {% translate 'more...' %}</li>
+                    {% endif %}
                 </ul>
             {% endif %}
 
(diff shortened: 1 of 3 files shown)


rlskoeser (commented):
I agree, not sure it's worth making the 5 into a template variable at the moment (would that even work with the -5?).

We don't have designs for this, but I was thinking of this text: `(+ N more)`. I would like it styled using secondary metadata styles (hopefully there's something appropriate to adapt) so it isn't as prominent and obviously remove the `#`. Using an `<li>` for this seems a little non-semantic, but I don't see an easy/obvious alternative.

kmcelwee:
@rlskoeser I updated the language. Secondary metadata use the `.secondary` class in the document detail page. I can port that class over to the list view, but I think it just changes the color to grey, not the size.

So smaller and gray? Here's the comparison:

Screen Shot 2022-02-01 at 2 32 43 PM
Screen Shot 2022-02-01 at 2 42 02 PM

rlskoeser:
@kmcelwee I think it looks better smaller, but we shouldn't introduce new text sizes/styles that aren't used elsewhere; use existing secondary color for now and Gissoo can revisit later on if necessary when she's working on other updates.

rlskoeser (approved):

Issue #542, opened 2022-01-21
As a front end user, I want to see a homepage when I first visit the website so I can learn context for its contents.


blms:
## dev notes
- [x] Implement Gissoo's designs for background images, content flow
- [x] Create a new directory for shared light/dark mode images (done in #545)
- [x] Ensure content can be edited from the Wagtail CMS

blms:
Confirmed present on QA site!

PR #561, opened 2022-02-01
#542 - Implement home page designs

blms:
## visual review notes

[latest Percy build](https://percy.io/2cf28a24/geniza/builds/15368485)

In scope:
- Home page, dark and light, mobile and desktop
  - Is this homepage body text width ok? It's the same width used for content pages, and I think it's based on best practices for readability (60 characters across). We can widen it according to the design, but wanted to double check first.
  - I noticed in the designs the header text was 24px on both mobile and desktop. Was that intended? I preserved it in the implementation but can change if not.
- Reduced opacity for header drop shadow
  - I included content page (light) and mobile menu (dark) so you can check this independently of the home page!
- Can you also check the spacing for the content page (non-homepage) on mobile? The header looks a little low to me, maybe I should move it up?

As usual, fonts and horizontal spacing are not in scope.

## dev review notes
- Is it ok to hide the language switcher on mobile? IIRC it's not going to show up at all in MVP and it's only for admin use right now. Otherwise, I couldn't find a good way to implement it so that it's out of the way on mobile.

(The rest of this PR is all CSS and images, except for bootstrapping the homepage and adding the Percy snapshot)

## What this PR does

- Per #542:
  - Implements design for homepage (background, header, and body text)
  - Bootstraps homepage content for use with Percy
- Adjusts opacity on header image drop shadows
- Moves the light/dark control out of the way on mobile


rlskoeser (approved):
👍 for dev review

Agree to suppressing language switcher on mobile for now (and I will update production language settings at some point so it shouldn't show up there until we're ready for it)

gissoo:
@blms thank you!! 
- Apologies about the homepage header on mobile – I agree it should be the 22px with a 34px leading. 
- I totally agree with your decision on the line length on desktop – following the content page is a good idea, I think I forgot to update it and it wasn't caught during the design review either. 
- I agree that you could slightly move up the page title on the content page. 
- Reduced opacity on the header looks fine, but something looks weird here: Can you move up/extend the overlay so we wouldn't see the line cut there?
darkmode
lightmode

Issue #488, opened 2022-01-05
Fix lighthouse failures


kmcelwee:
Failing Lighthouse tests have piled up (Lighthouse may have updated recently, but errors may have also been introduced).

## dev notes
- [x] get things passing
- [x] configure to ignore if there are things we can't get passing that we can live with
- [x] revisit the font preloading

rlskoeser:
✅  All checks are now passing! Let's keep them that way — PRs should now address any failing lighthouse tests before merging.

PR #565, opened 2022-02-02
#488 - Address lighthouse errors

blms:
## What this PR does

- Per #488:
  - Pins Lighthouse CLI to 0.8.x, as 0.9 drops support for Node 12
  - Disables `unused-javascript` and `unsized-images` warnings that are only broken by OpenSeaDragon
  - Increases tap target sizes for tags on document detail view
  - Fixes invalid heading hierarchy
  - Uses `aria-disabled` and `disabled` attributes to ignore contrast on disabled elements
  - Uses `link rel="preconnect"` for unpkg
- Adds home page to Lighthouse tests

### DIFF
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -54,7 +54,7 @@ jobs:
           restore-keys: |
             npm-${{ hashFiles('package-lock.json') }}
             npm-
-      - run: npm install -g @lhci/cli
+      - run: npm install -g @lhci/cli@0.8
       - run: npm ci
       - run: npm run build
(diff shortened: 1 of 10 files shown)


blms:
> All the changes look great! Such a relief to have all of those lighthouse checks passing again, we should work to keep them passing in future.

Agree—which might eventually mean forking OSD or opening issues on [their repo](https://github.com/openseadragon/openseadragon) so that we can re-enable those two audits.

> Is the node version 12 limitation a concern we should plan to address at some point?

I would say so! Seems like Node 12 is in LTS and goes EOL on April 30.

Issue #468, opened 2021-12-23
search for partial shelfmarks doesn't yield the expected results


alanelbaum:
## testing notes

Searching on partial shelfmarks (especially parts of dot-separated text) should now result in much better results. 

- [x] test public search with a partial shelfmark like `T-S+Ar.35` and relevance sort, e.g. https://test-geniza.cdh.princeton.edu/en/documents/?q=+T-S+Ar.35&sort=relevance
- [x] test admin search with a partial shelfmark like `AIU II.A.`

---- 

**Describe the bug**
Original Slack message: How does one pull up all the documents (say) in the T-S Misc.29 folder? They all begin with T-S Misc.29, but typing that string yields only 1 document (from another folder). Searching T-S Misc.29.* also leads to nothing. Searching T-S Misc. 29.* (not sure why adding a space helps) yields the documents I’m looking for, but also a whole bunch of irrelevant documents from other folders. Should I be going about this differently? Would it be difficult to adjust the search settings such that searching T-S Misc.29 or “T-S Misc.29” would only yield records that contain that string?

**To reproduce**
1. Go to the Django backend.
2. Try searching the opening string of whichever folder you are interested in, e.g., T-S Misc.29, or ENA NS I, or T-S Ar.35.
3. The results will be a bit chaotic.

**Expected behavior**
The JTS Viewer is very intuitive for me. E.g., if you type in ENA NS 77 and hit search, you get all the records that contain the string ENA NS 77, which in practice means you get the entire ENA NS 77 folder.

**Screenshots**

**Device information**
 - Device: MacBook Air
 - OS and version: OS X 10.15.7
 - Browser and version: Firefox and Chrome
 - Site version: v. 0.7.0

**Additional context**

rlskoeser:
confirmed working as expected in qa

PR #572, opened 2022-02-03
Improve shelfmark indexing

rlskoeser:
ref #468

add another solr configuration and copy field for shelfmarks to index letters and numbers only, ignoring punctuation and so that portions of shelfmarks like `20.169` will be split apart and made available for partial matching.

### DIFF
--- a/geniza/corpus/tests/test_corpus_views.py
+++ b/geniza/corpus/tests/test_corpus_views.py
@@ -544,7 +544,7 @@ class TestDocumentSearchView:
         doc2 = Document.objects.create()
         TextBlock.objects.create(document=doc2, fragment=folder_fragment)
 
-        # ensure solr index is updated with all three test documents
+        # ensure solr index is updated with the two test documents
         SolrClient().update.index(
             [
                 doc1.index_data(),
--- a/geniza/corpus/tests/test_corpus_views.py
+++ b/geniza/corpus/tests/test_corpus_views.py
@@ -532,6 +532,38 @@ class TestDocumentSearchView:
         ), "document with shelfmark in description returned second"
         # (document with similar shelfmark is third)
 
+    def test_shelfmark_partialmatch(self, empty_solr, multifragment):
+        # integration test for shelfmark indexing with partial matching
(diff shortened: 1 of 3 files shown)


solr_conf/conf/solrconfig.xml
```diff
@@ -749,6 +750,7 @@
       <str name="keyword_pf">
         description_t^50
         shelfmark_t^100
+        shelfmark_textnum^140
```
kmcelwee, line 753: Is there ever a science to our boosting? Would it be easier to sort these by their boosting levels?

geniza/corpus/tests/test_corpus_views.py
```diff
@@ -532,6 +532,38 @@ def test_shelfmark_boost(self, empty_solr, document, multifragment):
...
+
+        # multifragment shelfmark can test for this problem: T-S 16.377
+        doc1 = Document.objects.create()
+        TextBlock.objects.create(document=doc1, fragment=multifragment)
+        # create an arbitrary fragment with similar numeric shelfmark
+        folder_fragment = Fragment.objects.create(shelfmark="T-S 16.378")
+        doc2 = Document.objects.create()
+        TextBlock.objects.create(document=doc2, fragment=folder_fragment)
+
+        # ensure solr index is updated with all three test documents
```
kmcelwee, line 547:
```suggestion
        # ensure solr index is updated with the two test documents
```

kmcelwee (approved):
Looks good! (but I didn't pull down locally, let me know if you'd feel better if I did)

Is this the last time something like this will come up? If not, should we recognize that shelfmarks have this structure in some kind of shelfmark class?

geniza/corpus/tests/test_corpus_views.py
rlskoeser: whoops, thank you!

solr_conf/conf/solrconfig.xml
rlskoeser: maybe? no, it's not very scientific — open to suggestions if you have thoughts on how to improve it

Issue #566, opened 2022-02-03
footer and header display problems on wide screens


rlskoeser:
the header and footer don't extend the full width on large screens (noticed this on my desktop). It seems fine to not have  the image at full width (we don't want lower quality/degradation), as long as we have some kind of background or fallback so it doesn't just cut off the way it does now. (It looks like this may be easier with the header, which is already transitioning to a solid color before it cuts off; not sure how to handle the footer.)

![Screen Shot 2022-02-03 at 10.59.26 AM.png](https://images.zenhubusercontent.com/5931d36351aa2c3057804a51/c7876356-e3df-49e2-a41e-ee5837ee99e5)![Screen Shot 2022-02-03 at 10.59.33 AM.png](https://images.zenhubusercontent.com/5931d36351aa2c3057804a51/ffd261d3-8ffd-426f-ad9d-74460cc6d719)

rlskoeser:
this looks good locally and in qa; I thought I merged the changes into the release branch before I tagged the 4.0, but it doesn't display correctly in production

rlskoeser:
I checked again in production this morning and it's fine now — maybe it was a caching issue between the load balancer and the server.

PR #575, opened 2022-02-03
#566 - Extend header and footer on desktop

blms:
## What this PR does

- Per #566:
  - Adds new extender ("filler") SVGs for header and footer, light and dark mode
  - Slightly shortens footer image and uses gradient to fade it out (only visible on very large displays)
  - Clips header images slightly early on the right to fix drop shadow


blms:
@rlskoeser Is this something you'd be able to test locally, or should we merge first and test on QA site?

rlskoeser (approved):
tested locally, looks great!

PR #579, opened 2022-02-03
Add disabled attributes for empty scholarship tab

rlskoeser:
@blms I think maybe this got missed when you fixed the lighthouse checks — please confirm


blms (approved):
Yes, I missed this! 🤦 Thanks for catching.

Issue #580, opened 2022-02-04
As a content editor, I want to be able to underline text in the Wagtail CMS.


blms:
## testing notes (QA)
From the [QA Wagtail admin](https://test-geniza.cdh.princeton.edu/cms):
- [x] edit a page and try adding an underline with the toolbar button and with the keyboard shortcut
- [x] edit a page and try adding two blockquotes next to each other. click "preview" and confirm that the close quote and open quote are separated
- [x] create a new page, go to the "promote" tab and confirm it is set to appear in menus

## dev notes

- [x] Add underline style using `mark` elements

Can also roll in the following tweaks:
- [x] Remove embed option until/unless we have a design for it
- [x] Set new content pages to appear in menu by default
- [x] Fix bug with back-to-back blockquotes (edge case but easy enough to fix)

richmanrachel:
@blms - looks perfect!

Issue #581, opened 2022-02-04
As a content editor, I want to be able to separately define alt text and caption for images in Wagtail.


blms:
## testing notes (qa)
- [x] From the [QA Wagtail admin](https://test-geniza.cdh.princeton.edu/cms), edit a page and confirm you can add an image block with captions and alt text

richmanrachel:
@blms - it may just be my image type/file size, but it didn't work :(

blms:
@richmanrachel What happened when you tried? (A screenshot or video would help if you can't easily describe!) And @rlskoeser I'm wondering if this is a config or code issue… it might be the former if it's just the upload that's failing.

rlskoeser:
@blms ooh good call, I remember now that we had problems with file sizes on our other app because the defaults are set pretty low. I'll look into it.

richmanrachel:
@blms and @rlskoeser - would you like me to add screenshots/a more detailed description, or would you like to check on file size first?

The process did seem to work (in terms of being able to select jpgs from my desktop) but the broken image icon was present once I uploaded and again on the site page.
![image](https://user-images.githubusercontent.com/68028846/153266349-56343333-aac8-408f-bb41-ccc7f9160b3e.png)

rlskoeser:
@richmanrachel oh, sorry! I'm looking into this — I think there's a server configuration missing, which is why you aren't seeing the images. I'm working on the change

blms:
@richmanrachel Thanks for the screenshot anyway! It also exposed the "{'paragraph': '" there, which shouldn't be present—at least not visible to the user. So we may have to do some troubleshooting on that too. I assume it's not visible in the text while you're editing it?

richmanrachel:
@blms - It is actually in the text:
![image](https://user-images.githubusercontent.com/68028846/153267378-12ff06b1-1b65-4e69-9660-4550cf7f0f96.png)
I'll try deleting just to make sure

richmanrachel:
Yes - confirmed that the image still isn't working but wonky paragraph text was deletable :)

blms:
@richmanrachel Ah! Okay, that's a relief 😄  It may be a side effect of the migration we did.

rlskoeser:
@richmanrachel I think I've configured the upload directory properly now on the test site (although I have not increased the default file size limits); please test the image upload and image caption again when you can.

richmanrachel:
[...]

PR #583, opened 2022-02-04
#580 - Wagtail tweaks (underline, etc.)

blms:
## What this PR does

- Per #580:
  - Adds underline using `mark` elements to Wagtail editor toolbar
  - Removes embed option (until/unless we have a design for it)
  - Sets new content pages to appear in menu by default
  - Fixes bug with back-to-back blockquotes

## dev review notes

- My main question: how should I test `wagtail_hooks.py`? I put in a preliminary test just to get some coverage but I'm not really sure how to approach such a test. Ideally I'd like to test that it actually outputs a `<mark class="underline">` element, but I don't know if that's possible given all the Wagtail library code.


geniza/pages/models.py
```diff
@@ -16,11 +16,11 @@ class HomePage(Page):
             "h3",
             "bold",
             "italic",
+            "underline",
```
rlskoeser, line 19: Are these two lists of features the same (for home page and content page)? can we define them once and reuse?

rlskoeser (approved):
Looks good. Cool to see the custom wagtail/draftail feature, I don't think we've used that much (if at all) yet on other projects where we're using wagtail.

geniza/pages/models.py
blms: Yes! Actually will be doing that in work on #581, so I'll go ahead and merge this first.

rlskoeser:
> * My main question: how should I test `wagtail_hooks.py`? I put in a preliminary test just to get some coverage but I'm not really sure how to approach such a test. Ideally I'd like to test that it actually outputs a `<mark class="underline">` element, but I don't know if that's possible given all the Wagtail library code.

My approach is that we only need to test our own code, not third-party code (unless we have reason to, e.g. if we know it doesn't behave correctly in certain scenarios). It's not obvious that there's a very meaningful way to test the functionality related to this hook... Any idea, would your current test alert us if this hook didn't work with a newer version of wagtail?

blms:
> Any idea, would your current test alert us if this hook didn't work with a newer version of wagtail?

@rlskoeser It might! At the very least, it will alert us if the API changes to the point that this function produces any exceptions.

rlskoeser:
> > Any idea, would your current test alert us if this hook didn't work with a newer version of wagtail?
> 
> @rlskoeser It might! At the very least, it will alert us if the API changes to the point that this function produces any exceptions.

yes, I was thinking that too — if it registers successfully then that might be enough for what we need in terms of testing/maintenance

PR #586, opened 2022-02-07
#581 - Replace RichTextField with StreamField, add Image stream block

blms:
## What this PR does

- Per #581:
  - Converts all `ContentPage` and `HomePage` models' `body` attributes from `RichTextField` to `StreamField` in migration
  - Provides reverse migration `StreamField` back to `RichTextField`
  - Replaces `CaptionedImageFormat` with `CaptionedImageBlock`
  - Provides `BodyContentBlock`, containing both `RichTextBlock` and `CaptionedImageBlock`, that can be used to fill the `StreamField`
  - Displays images responsively according to Gissoo's designs


rlskoeser (approved):
Looks great! Tested locally, was able to confirm the round-trip migration works

Issue #225, opened 2021-06-02
As a content editor, I want to merge document records without losing data so that I can combine records when I've identified duplicates or joins.


thatbudakguy:
## testing notes (round 2)

When merging documents in the admin backend:

- [x] Confirm that it now reads "Select primary document"
- [x] Confirm that you can see all scholarship records for all documents in the merge
- [x] Choose a rationale from the radio buttons; confirm that "duplicate", "join" and "other" are there
- [x] Confirm that you can't enter "other" without also entering a note below it; if you do, there should be an error after you try to submit

Note that the other requests mentioned (transcriptions labeled with footnote/citation in admin backend; all transcriptions expanded on the public site) have been spun off into separate issues.

---

## Testing Notes
- [x] Go into the admin interface and use the admin action to begin merging two or more documents. To get the same PGPIDs in the search results, you can use the "OR" syntax with their PGPIDs, e.g. "35058 OR 34031".
- [x] View the intermediate page. 
    - [ ] Is it clear what the primary document is?
    - [x] Is it clear what the "rationale" section is?
    - [ ] Do you have enough information in the "primary document" section to choose the correct document?
- [x] Click "Submit" on the intermediate page (note that this cannot be undone). Does the page redirect to the new primary document? Did the action preserve and merge all the information you expected?
- [x] Check the document history: does it include the merge with your name and the rationale you entered in the form?

## dev notes
- [x] Style `document_option_label.html`
- [x] Place the merge document view as an admin URL: https://docs.djangoproject.com/en/4.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_urls

---

**Is your feature request related to a problem? Please describe.**
See #170 for the management command that automatically identifies some candidates for merging and implements the logic for merging, but doesn't allow users to manually execute a merge.

See also https://github.com/Princeton-CDH/mep-django/issues/72 which provides the reference implementation for merging people, from shakespeare & company.

**Test case**
Alan's [recently identified join, `T-S 8J25.7 + T-S 8J16.15`](https://geniza.cdh.princeton.edu/admin/corpus/document/?q=T-S+8J16.15), which should be merged into a single record.

## dev notes
- [ ] create [new admin action that provides an intermediate page](https://docs.djangoproject.com/en/3.2/ref/contrib/admin/actions/#actions-that-provide-intermediate-pages)
- [ ] create a merge form & view that takes the documents to be merged, renders them with enough detail for the user to confirm the merge and choose a primary document
[...]

PR #595, opened 2022-02-09
#225 - Admin document merge tweaks

blms:
## What this PR does

- Per #225:
  - Changes scholarship records from `details` back to `div` so that all are visible in document merge
  - Changes verbiage "primary document" to "select primary document"
  - For rationale: uses radio buttons with "other" and an optional notes field, which becomes required if "other" is selected
    - Sets "duplicate" as default choice
  
## review notes

- I didn't include labeling transcriptions in the admin, or expanding all trancsriptions on the front end, in this PR. I was thinking it's best to get this stuff reviewed first and do that separately, since it's a different concern.

### DIFF
--- a/geniza/corpus/forms.py
+++ b/geniza/corpus/forms.py
@@ -1,5 +1,7 @@
 from django import forms
+from django.forms import renderers
 from django.template.loader import get_template
+from django.utils.encoding import force_text
 from django.utils.safestring import mark_safe
 from django.utils.translation import gettext as _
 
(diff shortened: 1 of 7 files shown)


rlskoeser (approved):
Looks good!

rlskoeser:
> * I didn't include labeling transcriptions in the admin, or expanding all transcriptions on the front end, in this PR. I was thinking it's best to get this stuff reviewed first and do that separately, since it's a different concern.

agreed about those being separate concerns; should we create issues to track those requests?

blms:
> agreed about those being separate concerns; should we create issues to track those requests?

@rlskoeser  Yes, I'll do that!

PR #549, opened 2022-01-27
index iiif images

rlskoeser:
initial indexing to support images in search results, with proof-of-concept image display

Issue #28, opened 2020-11-03
As a user, I want to see image thumbnails with search results when available, so that I can quickly see which records have images and what they look like.


rlskoeser:
## testing notes

- [x] On the [QA site](https://test-geniza.cdh.princeton.edu/documents/), do some searches and confirm that you can see image thumbnails for each result
  - [x] Check documents with no images, one image, two image, and 3+ images
  - [x] Comment feedback here about the design and implementation

richmanrachel:
@rlskoeser - did we decide that we'll keep this as a priority for the frontend but not the backend? While it would be lovely on the backend, the existing boolean is sufficient to assist the researcher.

rlskoeser:
@richmanrachel yes, this is intended for the frontend search. We don't have that much control or space for that in the backend search.  (I wonder if it would be helpful to add and use tags to help us distinguish frontend features from backend/admin features?)

richmanrachel:
@rlskoeser - Yes, adding tags would be extremely helpful. Thanks!

blms:
@richmanrachel @rlskoeser @mrustow @gissoo 

Working on implementing this, and due to the different dimensions of images there are some edge cases that we have to account for. Here's a bunch of examples the current implementation I've come up with so far. Are they acceptable? Is it clear when there are one, two, or more images? Is the sizing appropriate? Looking for any/all feedback!

<details>
<summary>A bunch of example screenshots</summary>

<img width="956" alt="Screen Shot 2022-02-09 at 11 17 32 AM" src="https://user-images.githubusercontent.com/7234006/153243509-4b9e70a4-d1f7-4fe8-91ec-5a4a41bfdcdf.png">
<img width="915" alt="Screen Shot 2022-02-09 at 11 18 11 AM" src="https://user-images.githubusercontent.com/7234006/153243513-efedaf4a-6981-4d98-b1db-66f772515f5f.png">
<img width="930" alt="Screen Shot 2022-02-09 at 11 16 17 AM" src="https://user-images.githubusercontent.com/7234006/153243520-2a3c0c14-25c4-456b-8f21-e9601677e4db.png">
<img width="953" alt="Screen Shot 2022-02-09 at 11 16 26 AM" src="https://user-images.githubusercontent.com/7234006/153243524-6160ab06-aedc-45b0-bee7-7187fedf8a05.png">
[...]

PR #598, opened 2022-02-09
#28 - Index IIIF images and display them as search result thumbnails

blms:
## What this PR does

- Per #28:
  - Includes all of @rlskoeser's work from #549
  - Displays IIIF images in search results as thumbnails according to designs


rlskoeser (changes requested):
Please add a line in deploy notes reminding to reindex to get images so I don't forget when I deploy!

I think we need to add a couple of small unit tests and am not sure what we should do about image alt text, but it looks and works great! Exciting to see this

geniza/corpus/templates/corpus/snippets/document_result.html
```diff
@@ -100,8 +100,19 @@ <h3 class="sr-only">{% translate 'Tags' %}</h3>
...
             {% endif %}
-
         </section>
+
+        {% if document.iiif_images %}
+            <ul class="images">
+                {% for image in document.iiif_images|slice:":3" %}
+                    <li class="image-{{ forloop.counter }}">
+                        <a href="{% url 'corpus:document' document.pgpid %}">
+                            <img src="{{ image|iiif_image:"size:width=250" }}" loading="lazy">
```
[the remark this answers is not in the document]
blms: I was thinking about that too! The ordering should be done in the `index_data` method, right? (And when you say "check" do you mean in tests?)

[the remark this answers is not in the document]
blms: I think we should either index labels, or else use some placeholder like "`shelfmark` facsimile `forloop.counter`" for alt text.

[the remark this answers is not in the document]
rlskoeser: > I was thinking about that too! The ordering should be done in the `index_data` method, right? (And when you say "check" do you mean in tests?)

by check I mean, we should test manually that solr returns list values in the same order that we send them when we index the data — I think order might not be guaranteed; we might need to use a different field type/structure if we care about order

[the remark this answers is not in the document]
rlskoeser: I'm inclined to indexing labels as well, we have access to them and it will be more meaningful; we might need to include shelfmark though...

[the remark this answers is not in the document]
blms: Done in 560169e 👍

geniza/corpus/models.py
```diff
@@ -695,6 +695,13 @@ def index_data(self):
                 "status_s": self.get_status_display(),
                 "old_pgpids_is": self.old_pgpids,
                 "language_code_ss": [lang.iso_code for lang in self.languages.all()],
+                # use image info link without trailing info.json to easily convert back to iiif image client
+                "iiif_images_ss": [
+                    img.info()[:-10]  # i.e., remove /info.json
+                    for f in fragments
+                    if f.manifest
+                    for img in f.iiif_images()[0]
+                ],
```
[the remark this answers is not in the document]
blms: Done in 3550399 👍

geniza/corpus/models.py
```diff
@@ -702,6 +702,12 @@ def index_data(self):
                     if f.manifest
                     for img in f.iiif_images()[0]
                 ],
+                "iiif_labels_ss": [
+                    label
+                    for f in fragments
+                    if f.manifest
+                    for label in f.iiif_images()[1]
```
rlskoeser, line 709: we should only call `iiif_images` once for efficiency; I think it will make the code more readable too...

geniza/corpus/solr_queryset.py
```diff
@@ -90,7 +91,10 @@ def get_result_document(self, doc):
         # default implementation converts from attrdict to dict
         doc = super().get_result_document(doc)
         # convert indexed iiif image paths to IIIFImageClient objects
```
rlskoeser, line 93: would be helpful to update the comment here

geniza/corpus/solr_queryset.py
```diff
@@ -90,7 +91,10 @@ def get_result_document(self, doc):
...
+        images = doc.get("iiif_images", [])
+        labels = doc.get("iiif_labels", [])
         doc["iiif_images"] = [
-            IIIFImageClient(*img.rsplit("/", 1)) for img in doc.get("iiif_images", [])
+            (IIIFImageClient(*img.rsplit("/", 1)), labels[i])
+            for i, img in enumerate(images)
```
rlskoeser, line 98: could we use `zip` for this?

geniza/corpus/templates/corpus/snippets/document_result.html
```diff
@@ -107,7 +107,7 @@ <h3 class="sr-only">{% translate 'Tags' %}</h3>
                 {% for image in document.iiif_images|slice:":3" %}
                     <li class="image-{{ forloop.counter }}">
                         <a href="{% url 'corpus:document' document.pgpid %}">
-                            <img src="{{ image|iiif_image:"size:width=250" }}" loading="lazy">
+                            <img src="{{ image.0|iiif_image:"size:width=250" }}" loading="lazy" alt="{{ image.1 }}">
```
rlskoeser, line 110: maybe a comment here also to make clear it's a list of tuples of iiif image and label

geniza/corpus/models.py
rlskoeser: wonder how this is going to impact indexing time, actually... can/should we add prefetching on manifests and canvases? 😬

rlskoeser: here's my suggestion for revising:

```python
        images = self.iiif_images()  
        index_data.update(
              # ...
               "iiif_images_ss": [
                    img['image'].info()[:-10]  # i.e., remove /info.json
                    for img in images
                ],
                "iiif_labels_ss": [img['label'] for img in images]
             # ...
````

rlskoeser (approved):
Looks great! One slight concern about index data that I think does impact indexing speed, and a couple of comments could be improved, but otherwise good — can be merged once you address those items.

geniza/corpus/models.py
blms: Done in fa601a7, I think—unless we still need to add prefetching?

rlskoeser: let's try without for now

geniza/corpus/solr_queryset.py
blms: Done in 46dd6fb :+1:

geniza/corpus/templates/corpus/snippets/document_result.html
blms: Done in 4ae76ee :+1:

Issue #596, opened 2022-02-09
As a content admin, I want to be able to see which transcriptions belong with which footnote.


blms:
## Testing notes

In the admin interface, navigate to a document that has multiple transcriptions (such as PGPID 5555) and confirm:
  - [x] Both transcriptions are preceded by source citations
  - [x] It is clear which source belongs to which transcription

---

@rlskoeser - Labeling all transcriptions would be a huge help. I think we've discussed this before, but knowing which footnotes are "empty" vs. "linked to content" would certainly prevent many future mistakes!

_Originally posted by @richmanrachel in https://github.com/Princeton-CDH/geniza/issues/225#issuecomment-1034082821_

I did a test of merging two records with two different transcriptions (PGPID 516 and 4414) to see what happened. I see that both transcriptions were successfully preserved in the "Transcription" section, but on the resulting merged document record it's not clear which transcription belongs with which footnote. Maybe the editor information (in this case Gil for the Arabic-script one and Friedman for the Hebrew-script one) should be more closely linked to the transcription itself, and displayed together with it? 

_Originally posted by @alanelbaum in https://github.com/Princeton-CDH/geniza/issues/225#issuecomment-1033081856_

richmanrachel:
@blms and @rlskoeser - this works properly, but is it possible to display the transcriptions side-by-side?

blms:
@richmanrachel Yes, that should be possible! I'll make a new issue for it.

PR #600, opened 2022-02-09
#596 - Show transcription source labels in admin

blms:
## What this PR does

- Per #596:
  - When displaying multiple transcriptions in the document admin: groups transcriptions by source, labels them each with source citation

### DIFF
--- a/geniza/corpus/templates/admin/corpus/document/change_form.html
+++ b/geniza/corpus/templates/admin/corpus/document/change_form.html
@@ -17,20 +17,13 @@
             </div>
         </fieldset>
     {% endif %}
-    {% if original.has_transcription %}
+    {% if original.digital_editions %}
         <fieldset class="module aligned transcriptions-field">
             <div class="form-row">
-                <label>Transcription(s)</label>
-                {% regroup original.footnotes.all by source as source_list %}
-                {% for source in source_list %}
-                    {% if source_list|length > 1 %}<h3 class="source">{{ source.grouper|safe }}</h3>{% endif %}
-                    {% for footnote in source.list %}
-                        {% if footnote.content %}
-                            {% include "footnotes/transcription.html" with transcription=footnote.content %}
-                        {% else %}
-                            <div class="transcription">No transcription</div>
-                        {% endif %}
-                    {% endfor %}
+                <label>Transcription{{ original.digital_editions.count|pluralize }}</label>
+                {% for footnote in original.digital_editions %}
+                    {% if original.digital_editions|length > 1 %}<h3 class="source">{{ footnote.source|safe }}</h3>{% endif %}
+                    {% include "footnotes/transcription.html" with transcription=footnote.content %}
                 {% endfor %}
             </div>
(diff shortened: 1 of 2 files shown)


geniza/corpus/templates/admin/corpus/document/change_form.html
```diff
@@ -18,11 +18,21 @@
...
     {% if original.has_transcription %}
-        <h2>Transcription</h2>
-        {% for footnote in original.footnotes.all %}
-            {% if footnote.content %}
-                {% include "footnotes/transcription.html" with transcription=footnote.content %}
-            {% endif %}
-        {% endfor %}
+        <fieldset class="module aligned transcriptions-field">
+            <div class="form-row">
+                <label>Transcription(s)</label>
```
rlskoeser, line 23: let's use `pluralize` here

geniza/corpus/templates/admin/corpus/document/change_form.html
```diff
@@ -18,11 +18,21 @@
...
+                {% regroup original.footnotes.all by source as source_list %}
```
rlskoeser, line 24: can't we just iterate over our `digital_editions` method? and I don't think we should regroup (don't need to, but also better to be explicit about what is attached where e.g. when transcription content needs to be consolidated in TEI after merging documents)

geniza/corpus/templates/admin/corpus/document/change_form.html
```diff
@@ -18,11 +18,21 @@
...
+                {% for source in source_list %}
+                    {% if source_list|length > 1 %}<h3 class="source">{{ source.grouper|safe }}</h3>{% endif %}
+                    {% for footnote in source.list %}
+                        {% if footnote.content %}
+                            {% include "footnotes/transcription.html" with transcription=footnote.content %}
+                        {% else %}
+                            <div class="transcription">No transcription</div>
```
rlskoeser, line 31: I don't think we should display this here — it's redundant and takes up space.

There's already a digitized transcription red/green boolean indicator in the footnote inline display; if that isn't enough for the team (in combination with labels for the transcriptions), then we should figure out what would make it more readable for them.

rlskoeser (changes requested):
looks good, just requesting a couple adjustments

geniza/corpus/templates/admin/corpus/document/change_form.html
blms: I always forget about that! Done in c62ff7a 👍

geniza/corpus/templates/admin/corpus/document/change_form.html
blms: Done in c62ff7a 👍

geniza/corpus/templates/admin/corpus/document/change_form.html
blms: Done in c62ff7a 👍

rlskoeser (approved):

Issue #589, opened 2022-02-08
As a researcher, I want to see Goitein's unpublished editions labeled more clearly, so I'm not confused by the ambiguous title "typed texts".


rlskoeser:
## testing notes (round 2 - QA)

Navigate to a scholarship records page for a document with unpublished sources, and confirm that:

- [x] The "unpublished" label is gone (and that both RR and MR sign off on this)
- [x] The labels such as "for edition see" are all lowercase

---

## testing notes (QA)
Check the following areas in the QA site for Goitein (and Motzkin and Udovitch) "unpublished editions", and comment how you feel about it in context:
- [x] Admin view, when selecting Source from a dropdown menu
- [x] Document detail page for a document with a Goitein unpublished edition, under "Editor(s)" in metadata
- [x] Document detail page for a document with multiple transcriptions including one Goitein unpublished edition, above the transcription text from Goitein (see #597 to give feedback on the transcription view overall)
- [x] Document scholarship records page for a document with a Goitein unpublished edition

## dev notes
- [x] one-time data migration to change the title on all Goitein typed text sources
   - [x] should make log entries to document the change
- [x] revise sync transcription script to account for this change when searching for sources in the db

blms:
@richmanrachel @mrustow Should this also apply to “Motzkin, typed texts” and “A. L. Udovitch, typed texts”? If so, should they also get renamed to "unpublished editions", or something else?

richmanrachel:
@blms - yes, I think it should apply to all of them.

blms:
@richmanrachel Your item 1 actually raises the question about whether we should have that "unpublished" label at all! Would love to hear your and @mrustow's thoughts on this. 

Looking at the data, we have:

<details>
<summary>164 unpublished sources</summary>

```
Mark Cohen, [digital geniza document edition] (n.p., 1983).
Grace Masback, [digital geniza document edition] (n.p., 2020).
Stephanie Luescher, [digital geniza document edition].
Werner Diem, [digital geniza document edition] (in German).
Geoffrey Khan and Marina Rustow, [digital geniza document edition].
Marina Rustow and Anna Bailey, [digital geniza document edition].
Naïm Vanthieghem and Marina Rustow, [digital geniza document edition].
Yaara Perlman and Joshua Picard, [digital geniza document edition].
Motzkin, [digital geniza document edition].
Ṣabīḥ ʿAodeh, [digital geniza document edition].
[...]

PR #603, opened 2022-02-10
#589 - Rename "typed texts" to "unpublished editions"

blms:
## What this PR does

- Per #589:
  - Creates a one-time data migration to update all "typed texts" with the new title "unpublished editions" and log each change
  - Updates sync transcriptions script to handle this change


geniza/footnotes/migrations/0016_rename_typed_texts.py
```diff
@@ -0,0 +1,46 @@
...
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    LogEntry = apps.get_model("admin", "LogEntry")
+    source_contenttype = ContentType.objects.get(app_label="footnotes", model="source")
+    User = apps.get_model("auth", "User")
+    (user, _) = User.objects.get_or_create(
+        username=settings.SCRIPT_USERNAME, is_staff=False, is_active=False
+    )  # generic user for script execution (should already be created)
+
+    # get all sources called "typed texts"
+    typed_texts = Source.objects.filter(title="typed texts")
```
rlskoeser, line 18: any reason not to do a bulk / queryset update? is it because of the log action?

geniza/footnotes/migrations/0016_rename_typed_texts.py
```diff
@@ -0,0 +1,46 @@
...
+        source.title_en = "unpublished editions"
+        source.save()
+        # log action
+        LogEntry.objects.log_action(
+            user_id=user.id,
+            content_type_id=source_contenttype.pk,
+            object_id=source.pk,
+            object_repr=str(source),
+            change_message='changed title "typed texts" to "unpublished editions"',
+            action_flag=2,  # CHANGE
```
rlskoeser, line 31: can we not import the constant not available in migrations?

rlskoeser (approved):
Looks fine. Only in reviewing this did I remember that we have a "typed text" test fixture; wondering if should we update it to avoid confusion or if it's fine to leave for since it's only for dev... I'm guessing updating it could require changing a lot of tests, so I'm inclined to leave it, but might be worth adding a note that it isn't the official terminology anymore.

geniza/footnotes/migrations/0016_rename_typed_texts.py
blms: That was the idea—maybe it makes more sense to do that and have a separate loop for the log entries since all of them will be identical except for the PK?

rlskoeser: yeah, I think I've done something like that before... 

this is the example I found, not exactly analogous https://github.com/Princeton-CDH/mep-django/blob/main/mep/footnotes/migrations/0007_rename_pulrbsc.py

blms: Thanks, that was extremely helpful! It's nearly identical. Done in 56396f0 👍

geniza/footnotes/migrations/0016_rename_typed_texts.py
blms: I wasn't sure but it worked fine! Done in a80faf6 👍

rlskoeser (approved):
Great!

PR #604, opened 2022-02-10
#589 - Use appropriate modeltranslation fields for "typed texts" source migration

blms:
## review notes

- Should the migration query on and update just `title_en`, or `title` as well (as I currently have it)?
- I guess it wasn't caught by tests because I do set the `title` field directly when instantiating the Source records. I guess that doesn't automatically populate the `title_en` field as well. Should I be setting `title_en` there instead as well?

### DIFF
--- a/geniza/footnotes/migrations/0016_rename_typed_texts.py
+++ b/geniza/footnotes/migrations/0016_rename_typed_texts.py
@@ -3,6 +3,7 @@
 from django.conf import settings
 from django.contrib.admin.models import CHANGE
 from django.db import migrations
+from django.db.models import Q
 
 
 def rename_typed_texts(apps, schema_editor):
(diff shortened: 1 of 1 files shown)


rlskoeser:
> ## review notes
> 
>     * Should the migration query on and update just `title_en`, or `title` as well (as I currently have it)?
> 
>     * I guess it wasn't caught by tests because I do set the `title` field directly when instantiating the Source records. I guess that doesn't automatically populate the `title_en` field as well. Should I be setting `title_en` there instead as well?

Sorry I didn't catch this, I'm not as familiar with django model translate as I should be!

Based on this, it seems like either way should work:
https://django-modeltranslation.readthedocs.io/en/latest/usage.html#rules-for-translated-field-access

I wonder if default languages is set in a migration context?

However, it seems more reliable and more explicit to set and filter on `title_en` (I don't think it's worth updating your filter query, though; I don't think it matters). We'll have to keep this in mind for future data migrations on model-translation content!

rlskoeser (approved):

Issue #605, opened 2022-02-10
Indexing breaks with a bad IIIF URL or when the request otherwise fails


blms:
## dev notes

- [ ] Use a try/catch in `iiif_images`

rlskoeser:
Tried to test this by adding a bad iiif url to a fragment, but discovered that I _can't_ do that — both in admin or python console, setting a value for iiif url triggers the djiffy manifest importer logic, which causes an error. This is good in the sense that we won't save records with bad iiif urls but bad because it's causing a 500 error right now in the admin!

@blms what do you think, track that as a new bug or combine with this one?

rlskoeser:
thanks for investigating! I knew it was the manifest import failing but hadn't looked at the specifics. So is there a change needed in djiffy?

blms:
@rlskoeser Sure! 

Such a config might be the right approach, since we might want different behavior for bulk import (skip and log an error) vs. just saving a single record from the admin interface (perhaps actually raise the `IIIFException`—need to look into error handling when saving from Django admin).

An alternative could be to continue to *always* skip and move on in case of a `IIIFException` in the djiffy importer, and rather catch a bad URL in the form validation on the PGP side rather than relying on djiffy to report it.

rlskoeser:
tested in admin, no longer get a 500 error when I enter a bad iiif url and I do get a warning that it coudn't be loaded or cached (tested both an invalid url on a real domain and an invalid domain)

indexed the document with the invalid url and it didn't blow up

PR #606, opened 2022-02-11
#605 - Catch IIIFException during indexing

blms:
## What this PR does
- Per #605:
  - Uses a try/except block to catch and ignore `IIIFExceptions` raised when calling `from_url` on IIIF manuscripts

## Questions
- This will fix the problem during indexing, but there are a couple of views where `from_url` is also called. Should those get try/except blocks as well?
- Should there be some kind of error logging when this occurs? I don't have a good idea for how to hook into stdout during indexing in a way that would make sense here.

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -24,7 +24,7 @@ from djiffy.importer import ManifestImporter
 from djiffy.models import Manifest
 from parasolr.django.indexing import ModelIndexable
 from piffle.image import IIIFImageClient
-from piffle.presentation import IIIFPresentation
+from piffle.presentation import IIIFException, IIIFPresentation
 from taggit.models import Tag
 from taggit_selectize.managers import TaggableManager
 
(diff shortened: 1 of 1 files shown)


geniza/corpus/models.py
```diff
@@ -220,6 +220,9 @@ def iiif_images(self):
                     # label provides library's recto/verso designation
                     labels.append(canvas.label)
             except IIIFException:
+                logger.warning(
+                    "IIIFException raised by from_url with URL: %s" % self.iiif_url
```
rlskoeser, line 224: this is the log message I would write — do you think it gives enough context?

```suggestion
                    "Error loading IIIF manifest: %s" % self.iiif_url
```

geniza/corpus/tests/test_corpus_models.py
```diff
@@ -169,6 +200,22 @@ def test_iiif_images_locally_cached_manifest(self, mock_manifestimporter):
...
+            mock_iiifpresentation.from_url = Mock()
+            mock_iiifpresentation.from_url.side_effect = IIIFException
+            frag = Fragment(shelfmark="TS 1")
+            frag.iiif_url = "http://example.io/manifests/1"
+            frag.save()
+            # should raise the exception
+            with self.assertRaises(IIIFException):
+                # should log at level WARN
+                with self.assertLogs(level="WARN"):
+                    frag.iiif_images()
```
rlskoeser, line 217: nice

geniza/corpus/models.py
```diff
@@ -621,6 +624,7 @@ def attribution(self):
         # keep track of unique attributions so we can include them all
         extra_attrs_set = set()
         for url in self.iiif_urls():
+            # NOTE: If this url fails, may raise IIIFException
```
rlskoeser, line 627: is attribution only being loaded from remote manifests and not using the cached versions?

rlskoeser (approved):
Tests look great. I suggested alternate wording for the log message.

Thanks for flagging the other places we're loading manifests — I think probably those should be revisited and we should use cached manifests as much as possible, so maybe a follow-up issue to figure out the right approach.

geniza/corpus/models.py
blms: Yes!

geniza/corpus/models.py
blms: Looks like it! Good call on using cached manifests where possible. I'll make a new issue.

PR #610, opened 2022-02-14
Update Hebrew fonts, Percy/fixture configs

blms:
## What this PR does

- Adds new Hebrew fonts provided by designer (smaller file size)
- Uses new production URL to allow fonts to load in Percy
- Uses new Wagtail blocks format for content page fixtures, and uses a new approach to inject static images
- Enables `crossOriginPolicy: "Anonymous"` on OpenSeadragon to allow document detail pages to load in Percy

## review notes

- The only thing I couldn't get to work is the weird stretched out IIIF images in OpenSeadragon. I looked into it and I don't see a way to alter images that are a part of a canvas. Widths should be picked up correctly from the IIIF manifests.
- New Hebrew fonts won't be picked up by Percy until they appear on the production server.


blms:
@rlskoeser Interesting stuff going on here! Turning on debug for Percy confirmed that the last remaining problem for the content page fixtures is how to serve Wagtail-generated images from the `/media` directory. We originally resolved that by just injecting the image path in `/static` into the html, but that’s no longer possible with StreamField, as the image stream block expects a real Wagtail `Image`. So we need a way to store and view those Wagtail generated images in Percy.

It also, expectedly, broke the Percy tests for the 404 and 500 error pages by presenting the debug Django pages instead of the actual production error pages.

But, unexpectedly, it also suddenly started capturing document detail pages again! Albeit, once again without IIIF images inside of OpenSeadragon. More to investigate, it seems!

geniza/common/management/commands/visual_tests.py
```diff
@@ -38,6 +38,29 @@ def take_snapshots(self, browser, dark_mode=False):
 
         # content page
         browser.get("http://localhost:8000/en/content/")
+        # disable srcset and swap out content page images for ones from /static that Percy can see
```
rlskoeser, line 41: ah — is this your workaround for the uploaded user-content images?

let's flag it a little more loudly and clearly here in the comments with a NOTE and explain why we're doing it

rlskoeser (approved):
looks ok to me; will be so nice to have the fonts in percy finally

I'll add the new fonts to the deploy now so I don't forget — thanks for documenting it in the deploy notes.

Issue #594, opened 2022-02-08
As an admin, I want to configure which languages are available on the site without disabling them in the admin site, to avoid people accidentally receiving a partially-translated version of the site that isn't ready.


rlskoeser:
## testing notes (qa - round 2)

- [x] Log out using the admin interface, return to the public site, and confirm that the empty dropdown does not appear in the top right corner

## testing notes (qa)

- [x] Navigate to the admin interface, start editing a source, and confirm you can edit fields in all three languages: English, Hebrew, and Arabic
- [x] Return to the public site and confirm you can switch between those languages with the dropdown in the corner
- [x] Return to the admin site, log out, and then navigate to the public site and confirm you can no longer switch languages

---

We want to be able to do model translation data work in the admin site without giving people access to a semi-translated version of the public site. AFAICT django doesn't support this by default.

Ideally we want a solution that will work for other languages going forward, so that languages could be iteratively added to the site without making them available until the team is ready.

## dev notes
I did a little looking into this, and I think we should add a new configuration setting — something like **PUBLIC_SITE_LANGUAGES**, which should either be a subset of the official **LANGUAGES** configuration or not defined at all if no override is needed. 

And then we can write a custom middleware (inspired by this stack overflow answer https://stackoverflow.com/a/63203148) to redirect to the primary language version of the page if a language is defined but not in the public site languages list. (Or alternately we could define the list of languages to suppress — that would probably be simpler to implement; suppressed languages?). 

I think the middleware will need to use one of the i18n views or utility methods to set the language so it persists:
https://docs.djangoproject.com/en/4.0/topics/i18n/translation/#django.views.i18n.set_language

We may want to implement the middleware so the redirect only happens for anonymous users, so that logged in users can preview and work with the translated version of site before we make it public. 

### steps:
- [x] define and document new setting; add example/placeholder in local settings and note in deploy notes
- [x] implement middleware to redirect non-public languages to the default language for anonymous users
- [x] add the public languages config to `template_globals` context processors
- [x] adjust logic for language switcher to use public languages for anonymous users but full language list for logged in users

blms:
This suggestion makes sense to me.

> (Or alternately we could define the list of languages to suppress — that would probably be simpler to implement; suppressed languages?).

I feel like this is slightly less intuitive, but that's not a big deal—how were you thinking it would be simpler to implement?

> We may want to implement the middleware so the redirect only happens for anonymous users, so that logged in users can preview and work with the translated version of site before we make it public.

[...]

Issue #591, opened 2022-02-08
As a content editor, I want to a way to add hebrew descriptions of documents to the document record, so that available information can be managed in the same place.


rlskoeser:
## dev notes
- [ ] enable model translation for document description field
- [ ] update document detail page to display language-specific description with fallback to default language when unavailable 

(updating search to work with description translations will be handled separately)

blms:
Bug found:


My only question is about these drop downs for languages:
![image](https://user-images.githubusercontent.com/68028846/154361480-a2f38f60-76dd-4fa3-bd42-c22e90706731.png)
@blms or @rlskoeser - are these part of another issue?

_Originally posted by @richmanrachel in https://github.com/Princeton-CDH/geniza/issues/608#issuecomment-1042334731_

richmanrachel:
@blms and @rlskoeser - when I try to load a page that has no translation elements, the Hebrew and Arabic versions redirect to English (so you don't get the headers in the proper language).

richmanrachel:
It also changed the language of the admin interface as I switched between tabs, fyi (so I had to go back to the public site to change back to English).

blms:
> when I try to load a page that has no translation elements, the Hebrew and Arabic versions redirect to English (so you don't get the headers in the proper language).

@richmanrachel Do you mean content/About pages, or also pages like document detail and scholarship records? 

The former is intended! But the latter should always give you the Hebrew or Arabic version.

richmanrachel:
@blms - it was the document detail view that still didn't show any Arabic/Hebrew elements.

richmanrachel:
Everything else works! But this is certainly a bug that when I translated just a source name that this much showed up:
![image](https://user-images.githubusercontent.com/68028846/154554679-93096b50-3218-4533-96e2-3c4d8f44417d.png)
And a page where no detail has been translated is stuck with the headers in English.

PR #625, opened 2022-02-17
Feature/591 translation issues

blms:
## What this PR does

- Per #591:
  - Fixes issue with solr indexing of descriptions
  - Fixes issue with admin document change form by replacing all instances of h1 with h3 in transcriptions. This is more consistent with our other transcription formatting, and it prevents modeltranslation from automatically inserting its dropdown into any h1 on the page (annoying!)
  - Updates all tests to use the English `Source.title_en`, `Creator.first_name_en` and `last_name_en`, and `Document.description_en`
  
- Per #594:
  - Fixes issue with fallback to `settings.LANGUAGES` where it was using a list of tuples instead of the expected list of strings

## review notes

- The one test that I couldn't get to work with the `_en` version of a field is the test for the migration that merges India Books (`test_footnotes_migrations.MergeIndiaBookSources.test_sources_merged`). I guess those migrations only test the `title` field. But it's probably fine since it's an old migration.
- I noticed that the dropdown behavior seen by Rachel only happens when `PUBLIC_SITE_LANGUAGES` is not set properly in the settings. `PUBLIC_SITE_LANGUAGES` logic expects a list of language codes, not tuples, so we might want to make sure `PUBLIC_SITE_LANGUAGES` is set to just `["en"]` in the settings. If it's not set at all, it reverts back to `LANGUAGES`, which introduced a bug where it was reverting back to the list of tuples instead of the list of language codes (fixed here).
  - I could change this to make it accept tuples instead, but I thought it might be unnecessary to redefine "English", "Hebrew", and "Arabic" label strings when they're not being used.

### DIFF
--- a/geniza/context_processors.py
+++ b/geniza/context_processors.py
@@ -7,11 +7,12 @@ def template_globals(request):
     from django settings or site search form) for use on any page."""
 
     site = Site.find_for_request(request)
+    admin_language_codes = [lang[0] for lang in settings.LANGUAGES]
     context_extras = {
         "SHOW_TEST_WARNING": getattr(settings, "SHOW_TEST_WARNING", False),
         "FONT_URL_PREFIX": getattr(settings, "FONT_URL_PREFIX", ""),
         "PUBLIC_SITE_LANGUAGES": getattr(
-            settings, "PUBLIC_SITE_LANGUAGES", settings.LANGUAGES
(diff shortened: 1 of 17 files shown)


rlskoeser (approved):
updates all look fine to me, thanks for updating all the fixtures for translated fields

Issue #550, opened 2022-01-28
As a front-end user, I want keyword searches automatically sorted by relevance, so that I see the most useful results first.


blms:
## testing notes (qa)

- [x] Visit the search page and do not enter any text in the search box
- [x] Confirm that under "Sort by", "relevance" is clearly disabled and cannot be chosen
- [x] Start typing into the search box and confirm that "relevance" is automatically chosen
- [x] Make a search and confirm that the search results are sorted properly by relevance
- [x] Delete the text in the search box and confirm that it reverts back to ~"Scholarship Records (Most - Least)"~ Random (new default)
- [x] Add a keyword search term, change the sort to something besides relevance, submit the form, and confirm that the sort works properly 

---

> On the [prosody archive search](https://prosody.princeton.edu/archive/) we implemented js logic to disable/enable relevance, and automatically switch the sort to relevance when you start typing a search term, but preserve your sort if you change it — which I know is kind of complicated, and should not be MVP. Maybe we could have Gissoo do some usability testing on the two searches with regard to sort behavior to help figure out what would help users most.

_Originally posted by @rlskoeser in https://github.com/Princeton-CDH/geniza/pull/534#pullrequestreview-858429524_

> - by "but preserve your sort if you change it" you mean preserve the "relevance" ? Let me know – Sure I can reach out to the ux office for a quick usability testing on it 
> - I wish "Relevance" would not even be an option when there is no query text

_Originally posted by @gissoo in https://github.com/Princeton-CDH/geniza/issues/523#issuecomment-1018652219_

> That's how the Prosody Archive does it, and then it automatically chooses "relevance" when you enter text because it might be confusing to users that they would have to change it once they entered text. This strongly assumes that "relevance" is the desired sort option, but it probably is when you're entering a query, at least at first!

_Originally posted by @blms in https://github.com/Princeton-CDH/geniza/issues/523#issuecomment-1018658673_

> IIRC our assumption was that if you _changed_ your keyword search terms you would still (again) want to see the most relevant results first. But I do remember it being something we discussed a fair bit, and we should have probably thought about usability testing on it before.

_Originally posted by @rlskoeser in https://github.com/Princeton-CDH/geniza/issues/523#issuecomment-1022596677_

PR #629, opened 2022-02-17
#550 - Enable/disable/auto-select relevance in search form

blms:
## What this PR does

- Per #550:
  - Adds JS event handling to automatically select "relevance" sort when text is entered in the search field
  - Disables the "relevance" sort when no text is entered, reverting back to "scholarship descending" if relevance was previously selected
  - Updates lighthouse config to allow 1 unminified JS per page (not sure how this wasn't triggering before)
- Increases size of tap targets in footer to get lighthouse passing

## review notes

-  I chose to handle this entirely in JS rather than restoring the previous functionality of relevance being disabled by default via the Django form. I think if we were to disable that in Django but it required JS to be re-enabled, then it would be a problem for noscript users!


geniza/corpus/templates/corpus/document_detail.html
```diff
@@ -11,7 +11,7 @@
     {% if document.has_images or document.iiif_urls %}
         {# NOTE: referencing specific version avoids an extra render-blocking lookup #}
         <script defer src="https://cdnjs.cloudflare.com/ajax/libs/openseadragon/3.0.0/openseadragon.min.js" nonce="{{ request.csp_nonce}}"></script>
```
rlskoeser, line 13: should we include seadragon in our npm install & bundle? (maybe to be addressed in the stimulus refactor chore)

geniza/corpus/templates/corpus/document_list.html
```diff
@@ -1,17 +1,17 @@
...
+    {% render_bundle_csp "search" "js" attrs='defer' %}
 {% endblock %}
 
 {% block main %}
     <h1 class="sr-only">{{ page_title }}</h1>
-    <form>
+    <form data-controller="search">
         <fieldset>
-            {{ form.q }}
+            {% render_field form.q data-search-target="query" data-action="input->search#updateSort change->search#submit" %}
```
rlskoeser, line 14: I made a new issue for the search submitting automatically — and I don't think we really want that until we have turbo working; let's disable the `change` handlers for now until we integrate turbo and can apply it to all of the search form fields that require a reload.

rlskoeser (approved):
Tested locally, seems like this is working really well.

Only change I want to see before merging is disabling the form submit on change (until we get turbo setup).

geniza/corpus/templates/corpus/document_detail.html
blms: It makes sense for now since we just added the `"openseadragon"` webpack entry point, and I think importing there includes it in the bundle.

PR #643, opened 2022-02-22
implement random document sort

rlskoeser:
- adds random sort option to document search form
- set random as default sort in view when there is no keyword search term
- update stimulus search controller to change non-relevance default sort to random 
- new method to get solr sort option based on submitted sort chosen in the search form; generates a random dynamic solr field to sort randomly
- adds a parameter to context & page links to preserve current random sort when paging with within results for a randomly sorted set of documents

notes:
- might make more sense for stimulus to remember and restore the previous sort option when disabling relevance
- don't love the random seed handling in the template (feels redundant), but didn't see an obvious alternate solution

### DIFF
--- a/geniza/corpus/forms.py
+++ b/geniza/corpus/forms.py
@@ -122,6 +122,8 @@ class DocumentSearchForm(forms.Form):
     SORT_CHOICES = [
         # Translators: label for sort by relevance
         ("relevance", _("Relevance")),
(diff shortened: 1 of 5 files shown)


blms (approved):
All looks good to me, and I think the random seed idea is great! It's not any more redundant than that template already was, imo. Tested locally as well and works great—good call setting it as the default sort option.

Issue #640, opened 2022-02-21
Scholarship search filters implementation


blms:
Epic to keep track of both UI and backend implementation

Issue #637, opened 2022-02-21
As a front end user, I want to filter search results to records with discussion available, so that I can find documents with existing scholarship notes.


rlskoeser:
## testing notes (qa)

- Navigate to the document search page on the QA site, type in a search query and press the submit button
- [x] Click the "Filters" button below the search bar and confirm "Has Discussion" appears as a filter with a count of documents matching your search terms that have discussions
- [x] Enable that filter and click "Apply", then confirm the results all have discussions

richmanrachel:
Works with the same caveats as #630 and #636. One style thing I noticed though, @gissoo, is that the checkmark and down arrow look extremely similar in a confusing way:
![image](https://user-images.githubusercontent.com/68028846/155752260-2ac8db31-ede9-460d-aca6-0add311eb0cb.png)

richmanrachel:
Should I be making a new design issue for this?

blms:
@richmanrachel Yes, please!

rlskoeser:
functionality testing passed, design concerns addressed separately

Issue #636, opened 2022-02-21
As a front end user, I want to filter search results to records with translations available, so that I can find documents that are easier for me to work on.


rlskoeser:
## testing notes (qa)

- Navigate to the document search page on the QA site, type in a search query and press the submit button
- [x] Click the "Filters" button below the search bar and confirm "Has Translation" appears as a filter with a count of documents matching your search terms that have translations
- [x] Enable that filter and click "Apply", then confirm the results all have translations

rlskoeser:
@richmanrachel we have a separate issue about making the form submit automatically when you make changes — if submitting the form (either by hitting enter or clicking the search icon) resolves the problems you noted, then yes, please close.

Issue #633, opened 2022-02-21
As a front-end user, I want visual indicators for filtering search results, in a separate panel from the main search functions, so that I know where they are and can easily ignore them if I do not want to filter.


blms:
## testing notes (qa - round two)

On the document search page, click the "Filters" button and confirm:
- [x] the "close" button has the correct focus style (on mobile)
- [x] either the individual filter option focus styles are OK (per [comment](https://github.com/Princeton-CDH/geniza/issues/633#issuecomment-1051031596)), or they are not, in which case please create a design issue

## testing notes (qa)

On the document search page, click the "Filters" button and confirm:
- [x] Alignment, spacing, and colors are correct
- [x] Visual result of clicking on each filter is correct
- [x] Visual result of tabbing through the filters is correct

gissoo:
Thank you, Ben! I think everything looks good here! 
When hovering over a selected filter there is a darker green, is this the grey hover on top of the green that appears as a darker green? That's fine, it actually works. But it's also appearing as you select a filter, when a filter is selected here the color shouldn't become the darker green, it should just be the bright green. 
![Screen Shot 2022-02-25 at 10 29 13 AM](https://user-images.githubusercontent.com/44615389/155742020-052ca7a3-9601-470f-ade4-20af9fe6b66a.png)

the other thing I noticed is that the X button on the filters window should follow the icon button focus style, so it would have a green outline on lightmode and a blue one on dark mode.

blms:
Thanks @gissoo! 
1. That is the "focus" style, since that option now has focus (i.e. if you press tab on your keyboard the focus will move to the next one). Should we differentiate that from the hover style somehow? We need some way to indicate "selected AND focused".
2. Good catch, I'll fix that!

gissoo:
@blms so the grey is also being used as the focus style? that's what I'm seeing. 
I see it now. I don't know, I'll think about it. It's fine as is for now. thanks

gissoo:
@blms for the new desktop version below are a couple of usability reqs: 
- the filter window shouldn't collapse after submitting search / it shouldn't collapse unless the user has collapsed it
- the expanded filter window shouldn't cover the sort, and results numbers
- there should be a submit/apply button on the expanded filters (or some hint of what users should do to see new results, i.e. "pressing enter") – till we have reactive search

blms:
Thank you @gissoo!

> the expanded filter window shouldn't cover the sort, and results numbers

[...]

PR #645, opened 2022-02-22
Implement scholarship filters (#630, #633, #636, #637, #646)

blms:
## visual review notes

Percy: [light mode](https://percy.io/2cf28a24/geniza/builds/16000192/changed/901901801) | [dark mode](https://percy.io/2cf28a24/geniza/builds/16000192/changed/901902054)

In scope:
- filter dialog, look/placement of the filters within it
- look/placement of "Apply" button
- placement of the "Filters" button in the [Document Search page](https://percy.io/2cf28a24/geniza/builds/15963964/changed/899951378)
- all of the above, mobile and desktop

Not in scope:
- the safari-only bug for the document type filter
- the partial search page that appears from underneath the filter dialog (it's a percy bug/quirk)
- fonts, horizontal spacing

## What this PR does

- Per #640:
  - solr boolean indices for digital editions, translations, and discussions
  - new Django form classes for boolean facet checkboxes (with some of the shared features moved into a reusable mixin)
- Per #633:
  - UI: filter button, filters modal, and styles for the new filter types
- Per #646:
  - "Apply" button UI and functionality


geniza/corpus/views.py
```diff
@@ -95,6 +95,9 @@ def get_queryset(self):
             DocumentSolrQuerySet()
             .filter(status=Document.PUBLIC_LABEL)
             .facet_field("type", exclude="type", sort="value")
+            .facet_field("has_digital_edition")
+            .facet_field("has_translation")
+            .facet_field("has_discussion")
```
rlskoeser, line 100: these facets can be added in a single call with `facet` if we don't need extra options like the exclusion and sort we have on type
```suggestion
           .facet("has_digital_edition", "has_translation", "has_discussion")
```

rlskoeser (approved):
Looks good, working well.

The only thing that was confusing to me in the UI is when the new filters have zero counts — at first this was because I failed to reindex (my fault), but there are still scenarios where you could get a result with zero documents in some or all of these categories. I think it's confusing to have no number and no checkmark; maybe this is a more of a design question — is this a case that Gissoo accounted for in the various states for this component? (I'm not finding it). Can we show the zero but disable so the filter can't be selected?

blms:
@rlskoeser Regarding disabling fields when the count is zero: it works, except that it doesn't let you remove filters when they are causing the count to be zero. For example, if you query "document" and turn on all 3 of the new filters, then you change your query to "documentary" while leaving the 3 filters applied, it's impossible to unapply them without restarting the search. A good argument for a "reset filters" option—but at the moment it causes problems. Should I leave them enabled but just show the number 0?

blms:
For future reference, since it's got a few moving parts, if we ever do implement that:

`populate_from_facets`:
```python
if count == 0:
    self.disabled = True
```

template:
```django
<label for="{{ form.has_transcription.auto_id }}" {% if form.has_transcription.field.disabled %}disabled{% endif %}>
```

css:
```css
&[disabled] {
    pointer-events: none;
    cursor: default;
    color: var(--disabled);
    .count::after {
        color: var(--disabled);
    }
}
```

blms:
Seems that it was required to put that `facet` call first before `facet_field`.

Also went ahead and added the apply button—the only tweak this required was disabling the submit debounce. Otherwise sometimes the `submit()` action wasn't called.

geniza/corpus/templates/corpus/document_list.html
```diff
@@ -43,7 +43,7 @@ <h1 class="sr-only">{{ page_title }}</h1>
         </fieldset>
 
         <fieldset id="filters" aria-expanded="false">
-            <a href="#" role="button" id="close-filters-modal">
+            <a href="#" role="button" id="close-filters-modal" data-search-target="closeModal">
```
rlskoeser, line 46: Can we name `closeModalFilters` (`closeFilterModal`? or even just `closeFilters`? )? Would like a little more context so this is more immediately obvious; and it's possible we'll have other modals.

rlskoeser (commented):
When I apply the filters, it submits the form but it reloads the page with the filter modal open.

gissoo:
@blms thank you!!! These look great. I think the only thing that does not match the designs is that when a filter is active (expanded) its drop shadow is green on light mode and pink in dark mode. If we can't do it that's fine, not a deal breaker, it just looks better.

sitemedia/js/controllers/search_controller.js
```diff
@@ -1,17 +1,20 @@
...
     connect() {
-        useDebounce(this);
+        // useDebounce(this);
     }
 
     submit() {
+        // Close modal if open before submitting form
+        window.location.href = "#";
         this.element.submit();
     }
```
blms, line 19: @rlskoeser Can you test again to confirm it's working now, and is this an OK way to handle closing it? (I couldn't figure out how else to consistently close the modal without breaking the strict coupling of the anchor in the URL and visibility…)

rlskoeser (approved):
yes, that works — let's just expand the comment about the location change in the js code a bit so we remember why this is here later on when we're working on other parts of the search form

PR #647, opened 2022-02-23
Only allow first page when sort is random

rlskoeser:
revisions to random sort:
- remove previous random seed logic to preserve current random order when paging through a particular randomly ordered set
- disable pagination on document search page if sort is random
- if a page other than 1 is requested when sort is random, redirect to the first page of results

### DIFF
--- a/geniza/corpus/templates/corpus/snippets/pagination.html
+++ b/geniza/corpus/templates/corpus/snippets/pagination.html
@@ -1,10 +1,10 @@
 {% load i18n corpus_extras %}
-{% spaceless %}
-    <nav class="pagination">
+{% spaceless %} {# disable pagination when sorted randomly #}
+    <nav class="pagination {% if form.sort.value == 'random' %}disabled{% endif %}">
         {# Translators: Label for "previous page" button in search results #}
         {% translate 'Previous' as previous_page %}
         {% if page_obj.has_previous %}
-            <a name="{{ previous_page }}" title="{{ previous_page }}" class="prev" rel="prev" href="?{% querystring_replace page=page_obj.previous_page_number %}{{ random_seed }}">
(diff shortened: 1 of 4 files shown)


blms (approved):
Looks good, but I wonder if people might be confused by the disabled pagination. We could just use the existing conditional and preserve the spacing:

`geniza/corpus/templates/corpus/document_list.html`
```diff
-   {% if is_paginated %}
+   {% if is_paginated and form.sort.value != 'random' %}
        {% include "corpus/snippets/pagination.html" %}
    {% endif %}
````

Then we can use the existing CSS to keep the spacing:
```scss
    ol {
        list-style: none;
        @include container.two-column;
        margin-top: spacing.$spacing-2xl;
        @include breakpoints.for-tablet-landscape-up {
            margin-top: spacing.$spacing-4xl;
        }
    }
    nav.pagination + ol {
        margin-top: 0;
        @include breakpoints.for-tablet-landscape-up {
            margin-top: spacing.$spacing-md;
        }
    }
```

Though then the count discrepancy between the "total count" and the amount visible might be confusing. I guess it's a tradeoff of confusion!

One other possible solution would be to have some kind of tooltip over the disabled nav to explain why it's disabled.

rlskoeser:
@blms good points. I thought it would be confusing to not show the pagination when there are so many results — as there will be by default; but of course it could also be confusing to have it there and not usable. Do we have an easy way to add a tool tip? I can't remember if there's anywhere else we're doing that in this site. Alternatively, perhaps we could use that pagination space to add language about reloading to get a different random set...

@gissoo do you have time to give a quick opinion? The context is, for random sort we're only allowing the first page of results (paging order within the same random set isn't guaranteed, and we decided 50 random documents was enough). Is it better to:
1. show pagination as disabled 
2. hide pagination entirely (but preserve spacing so the page doesn't jump)
3. hide pagination, but use that space to add language to suggest reloading to see a different random set
4. something else?

Issue #578, opened 2022-02-03
As a frontend user, when the term "See PGPID ___" appears, I would like it to link to the corresponding PGP document so that I can easily access referenced documents.


richmanrachel:
## testing notes (qa)

- [x] On the QA public site, visit a document detail page for a document with "See PGPID ___" in the description, and confirm it links to the referenced document

---

**Is your feature request related to a problem? Please describe.**
Rebecca rightfully pointed out that we use this language both for future merges as well as for related documents. This will help users more easily navigate to the page where better or related information exists. 

**Describe the solution you'd like**
Automatic link when "See PGPID ___" appears

## dev notes
- [x] create a new template filter; propose name `pgp_urlize` since we may eventually add support for shelfmarks also; use a regex to do a global search for `PGPID (\d+)` (maybe with word boundaries on both sides); use the captured digit to generate a url to the PGP document in question, and turn the text into a link
- [x] apply the template to the description text on the document detail page
- [x] unit test cases: one id, two ids, id coming directly before a period or comma

richmanrachel:
@rlskoeser - It works, but something is going wrong... there are only 2 documents being pulled by the search "see pgpid" on the front end, while there are 193 on the admin site (180 of which are public)

rlskoeser:
@richmanrachel is it possible you have some search filters applied that are limiting your results?

richmanrachel:
@rlskoeser was right. It's working!

PR #667, opened 2022-02-28
Template filter for automatic PGPID links in descriptions (#578)

blms:
## What this PR does

- Per #578:
  - Adds a template filter to convert all instances of "PGPID ###" in a passed text into links to the canonical URL to the referenced document
  - Applies that template filter to the description in the document detail view

### DIFF
--- a/geniza/corpus/templatetags/corpus_extras.py
+++ b/geniza/corpus/templatetags/corpus_extras.py
@@ -1,11 +1,13 @@
 import json
-from urllib import parse
+import re
 
 from django import template
+from django.core.exceptions import ObjectDoesNotExist
+from django.utils.safestring import mark_safe
 from natsort import natsorted
 from piffle.iiif import IIIFImageClientException
 
-from geniza.footnotes.models import Footnote
+from geniza.corpus.models import Document
 
 register = template.Library()
 
@@ -117,3 +119,21 @@ def h1_to_h3(html):
     """Convert h1 headers to h3 to match other transcription formats,
     used to avoid modeltranslation inserting elements into h1 headers"""
     return html.replace("h1", "h3")
+
+
+def replacement_link(matchobj):
+    """Given a regex match \"PGPID #\", return a link to the referenced
+    document with the original string as link text."""
+    try:
+        doc = Document.objects.get(id=int(matchobj.group(1)))
+        doc_link = doc.permalink
+        return '<a href="%s">%s</a>' % (doc_link, matchobj.group(0))
+    except ObjectDoesNotExist:
+        return matchobj.group(0)
+
+
+@register.filter
+def pgp_urlize(text):
+    """Find all instances of \"PGPID #\" in the passed text, and convert
+    each to a link to the referenced document."""
+    return mark_safe(re.sub(r"\bPGPID\b \b(\d+)\b", replacement_link, text))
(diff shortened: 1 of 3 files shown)


geniza/corpus/templatetags/corpus_extras.py
```diff
@@ -117,3 +119,21 @@ def h1_to_h3(html):
...
+
+def replacement_link(matchobj):
+    """Given a regex match \"PGPID #\", return a link to the referenced
+    document with the original string as link text."""
+    try:
+        doc = Document.objects.get(id=int(matchobj.group(1)))
+        doc_link = doc.permalink
+        return '<a href="%s">%s</a>' % (doc_link, matchobj.group(0))
+    except ObjectDoesNotExist:
+        return matchobj.group(0)
```
rlskoeser, line 132: I don't want a db hit for every pgpid in the description, it isn't worth it

geniza/corpus/templatetags/corpus_extras.py
```diff
@@ -117,3 +119,21 @@ def h1_to_h3(html):
...
+
+
+@register.filter
+def pgp_urlize(text):
+    """Find all instances of \"PGPID #\" in the passed text, and convert
+    each to a link to the referenced document."""
```
rlskoeser, line 138: suggestion: generate a placeholder link using `reverse`, replace the id with a regex replacement string, and construct the link so that all matches can be replaced in a single regex.

geniza/corpus/templatetags/corpus_extras.py
```diff
@@ -117,3 +119,21 @@ def h1_to_h3(html):
...
+    return mark_safe(re.sub(r"\bPGPID\b \b(\d+)\b", replacement_link, text))
```
rlskoeser, line 139: slight preference for named groups for readability

rlskoeser (changes requested):
Looks good, but I don't want the db hit to check if the pgpids are valid — please simplify that part.

geniza/corpus/templatetags/corpus_extras.py
```diff
@@ -121,19 +123,24 @@ def h1_to_h3(html):
...
+            lambda match: '<a href="%s">%s</a>'
+            % (
+                # replace the fake pgpid in the placeholder URL with the matched real pgpid
+                re.sub("000", match.group("pgpid"), placeholder_url),
+                # use the original entire matched string as link text
+                match.group(0),
+            ),
+            text,
+        )
+    )
```
rlskoeser, line 146: you don't need the lambda and the extra `sub` call, you should be able to do it as a single regex replacement with the matched groups

```suggestion
   # use an absolutized placeholder URL to use for each match, replacing the fake pgpid 000 with the regex
    placeholder_url = absolutize_url(reverse("corpus:document", args=["000"])).replace('000', '\g<pgpid>')
    placeholder_link = '<a href="%s">\g<text></a>' % placeholder_url
    # match all instances of PGPID ### with a link to the document
    return mark_safe(
        re.sub(
            r"\b(?P<text>PGPID\b \b(?P<pgpid>\d+))\b",
            placeholder_link,
            text,
        )
    )
```

blms: Thanks, I didn't know the `\g<group>` syntax, that's so helpful!

rlskoeser (approved):

Issue #668, opened 2022-02-28
As an admin, I want documents automatically reindexed when I add or update scholarship records, so that database edits are immediately available in the public site.


rlskoeser:
## dev notes
- [ ] turn on parasolr debug logging and test if footnote save already triggers reindex on related document (can't remember if parasolr handles this for us by default); if it doesn't, add a signal handler for footnote save
- [ ] add a signal handler for source — reindex all documents related via footnotes
- [ ] add a signal handler for creator — reindex all documents associated with all sources 

existing signal handlers are gathered in `geinza.corpus.models.DocumentSignalHandlers` and should be configured in `Document.index_depends_on`

kmcelwee:
When editing a footnote on a document change view, parasolr certainly reindexes.
```
[28/Feb/2022 15:42:54] DEBUG:parasolr.django.signals::Indexing <Document: AIU VII.A.34 (PGPID 445)>
[28/Feb/2022 15:42:54] DEBUG:parasolr.django.signals::Indexing <Document: AIU VII.A.34 (PGPID 445)>
[28/Feb/2022 15:42:54] DEBUG:parasolr.solr.base::POST http://localhost:8983/solr/geniza_winner/update/json/docs => 200: 0.014825 sec
{'commitWithin': 1000}
[28/Feb/2022 15:42:54] DEBUG:parasolr.django.signals::Indexing <Document: AIU VII.A.34 (PGPID 445)> (m2m change: post_add)
[28/Feb/2022 15:42:54] DEBUG:parasolr.django.signals::Indexing <Document: AIU VII.A.34 (PGPID 445)> (m2m change: post_add)
[28/Feb/2022 15:42:54] DEBUG:parasolr.solr.base::POST http://localhost:8983/solr/geniza_winner/update/json/docs => 200: 0.004702 sec
{'commitWithin': 1000}
[28/Feb/2022 15:42:54] INFO:django.server::"POST /admin/corpus/document/445/change/?_changelist_filters=transcription%3Dyes HTTP/1.1" 302 0
[28/Feb/2022 15:42:56] INFO:django.server::"GET /admin/corpus/document/?transcription=yes HTTP/1.1" 200 126330
```

When editing a footnote directly (which isn't normal, correct?), parasolr does not seem to reindex:
```
[28/Feb/2022 15:45:43] INFO:django.server::"POST /admin/footnotes/footnote/13/change/ HTTP/1.1" 302 0
[...]

PR #673, opened 2022-02-28
Automatically reindex documents when sources, footnotes, or creators change (#668)

kmcelwee:
@rlskoeser I got a bit caught up on the fact that parasolr can't handle foreign/generic keys. Throwing this together last minute, and tests need to be written, but just let me know if I'm missing something big!

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -335,6 +335,7 @@ class DocumentSignalHandlers:
         "tag": "tags",
         "document type": "doctype",
         "Related Fragment": "textblock",  # textblock verbose name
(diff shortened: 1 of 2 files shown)


rlskoeser (commented):
wow, my existing signal handlers worked well enough that this is all you needed to add? I wouldn't have put it off so long if I'd known it would be so simple.

there's no new code here, but probably worth some minimal testing to confirm that things are configured correctly — tests are in test_corpus_signals, hopefully you can follow the pattern used for the others

rlskoeser (approved):

Issue #663, opened 2022-02-25
Revise filtered search so that users can see and change filters while on the search page.


richmanrachel:
## testing notes (qa)

- [x] Visit the search page and confirm you can view and change filters by clicking the "Filters" button, while still seeing your search query
- [x] Confirm it is clear which filters are selected or not
- [x] Click the "Filters" button again and confirm the filters are hidden
- [ ] Submit the search and confirm you can now tell there are active filters, based on the icon by the "Filters" button

---
Like the filter search for document type before scholarship records were added, it would be advantageous to avoid the current re-direct away from the search page as filters are added/shifted. 

I would like to see clearly which filters are on or off on the same page as the search results. This intersects with #646 and #630.

richmanrachel:
@blms - this is so much better, thank you!!! The only thing that's still not clear is knowing when the filters are active - the subtle switch from black to green is perhaps too subtle. But at least it's much clearer as soon as you click to open it.

@rlskoeser and/or @gissoo - would it be possible/easy to add the names of which filters are active next to the word Filters when the box is closed?
![image](https://user-images.githubusercontent.com/68028846/156454330-24406bf3-1cc3-42d1-90d3-d5be61a9c4cc.png)

rlskoeser:
@richmanrachel including those is out of scope for the current filters, and I'm not sure that's the best place to display them, but maybe @gissoo can take it into consideration when she works on a filter reset option.

gissoo:
@rlskoeser @richmanrachel sorry just seeing this. I agree with Rebecca. We've discussed this, I already have some ideas on the placement and the UI styles, we'll need to plan the design work for it.

PR #674, opened 2022-03-01
Position search filters absolutely on desktop (#663)

blms:
## What this PR does

- Ensures search sort is not random for Percy
- Per #663:
  - Positions filters absolutely and aligns with the "Filters" text in the search form, on desktop
  - Leaves mobile functionality the same as before


rlskoeser (approved):
looks great

PR #677, opened 2022-03-02
Fix doc index prefetching error and test

kmcelwee:


geniza/corpus/tests/test_corpus_signals.py
```diff
@@ -64,3 +65,14 @@ def test_related_delete(mock_indexitems, document, join):
...
     assert document in mock_indexitems.call_args[0][0]
     assert join not in mock_indexitems.call_args[0][0]
+
+
+@pytest.mark.django_db
+def test_items_to_index(document, footnote):
+    """Test that items_to_index doesn't fail silently."""
+    # NOTE: Footnote must be included in fixtures for full prefetching to occur.
+    # NOTE: Queryset must be explicitly called to trigger prefetching error.
+    #  Other methods of accessing the queryset may not trigger anything.
```
rlskoeser, line 75: can you briefly clarify what/why other scenarios don't trigger the problem? 

is there anything further that should be tested here that isn't covered by other tests?

rlskoeser (changes requested):
some minor cleanup needed

geniza/corpus/tests/test_corpus_models.py
```diff
@@ -1121,3 +1122,16 @@ def test_thumbnail(self):
...
+class TestDocumentSignalHandlers:
+    # Other tests are found in `test_corpus_signals.py`
+
+    def test_items_to_index(document, footnote):
+        """Test that prefetching is properly configured."""
+        # Because of lazy loading, querysets must be executed to test prefetches.
+        # Footnote fixture must be included to check source/creator prefetching.
+        docs = Document.items_to_index()
+        assert docs
+        assert type(docs) is MultilingualQuerySet
```
rlskoeser, line 1137: this method isn't part of the signal handlers, it's on `Document`; but easiest to test as a standalone method

```suggestion
@pytest.mark.django_db
def test_items_to_index(document, footnote):
    """Test that prefetching is properly configured."""
    # Because of lazy loading, querysets must be executed to test prefetches.
    # Footnote fixture must be included to check source/creator prefetching.
    docs = Document.items_to_index()
    assert docs
    assert type(docs) is MultilingualQuerySet
```

kmcelwee: Wow yeah, I just had it in my head that it was in the signal handlers class, thanks

Issue #650, opened 2022-02-23
revise the "selected" state for the scholarship records filters so that it's easier to identify what's been selected.


gissoo:

gissoo:
@richmanrachel @rlskoeser This is how I was thinking to address this issue, are these scholarship records indicators better now?! I personally like the second option more – the problem is that I cannot add a fill color to this particular phosphor icon because of the way it was originally created so I ended up having to add a circle myself. But it's probably worth it because it's way more visible.

 ![Screen Shot 2022-03-01 at 5.20.51 PM.png](https://images.zenhubusercontent.com/5bf307a69d595b459ec91bb1/7269b203-977a-4018-8e3c-18b6736a9c4f)

[Screen Shot 2022-03-01 at 5.26.16 PM.png](https://images.zenhubusercontent.com/5bf307a69d595b459ec91bb1/d590b164-51df-4c43-9da3-9670a928d659)

rlskoeser:
The circle helps a lot to differentiate from the drop-down; filled in circle vs. empty circle looks much easier to visually differentiate selected/unselected

richmanrachel:
@gissoo - I agree with Rebecca. The second option is great, thanks!

blms:
Also agree! And I think there's a built in Phosphor icon with the filled in circle, if that's what you were trying to find:

<img width="530" alt="Screen Shot 2022-03-02 at 9 43 29 AM" src="https://user-images.githubusercontent.com/7234006/156384076-c8ef8b5e-8537-4e33-b297-5da1bdecc47f.png">

gissoo:
duhhhh it was in their "fill" category ...I forgot they had that!!! thanks @blms!!!  
(although there is a similar problem with the fill version :D ...we can't color the check mark white, but the check mark is pretty visible on the light mode, so it's good, I'll be the only who notices it :D ) – it won't work on certain backgrounds, but right now it works. 

just documenting here that the size of the icon square is 30x30 px on desktop and 
- For the active state: the icon style is fill, the color of the circle is #567856.
- For the default state: the icon style is "regular", and the only color on the whole icon is #567856 

- here is [the link to the desktop version on Figma](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=421%3A23976) so you can see its placement relative to the other icons below and/or above it. 

[...]

Issue #524, opened 2022-01-13
Design an "active filter" indicator for the search form


blms:
Originally posted by @gissoo in https://github.com/Princeton-CDH/geniza/issues/449#issuecomment-1011518093:
> Also, we should eventually add a visual signifier to show when there are active filters for when the filter is collapsed right now (or hidden as the original designs in Figma show)

rlskoeser:
work completed; asset added to git & specifics communicated for implementation

PR #678, opened 2022-03-02
Filter icons (#633)

blms:
## What this PR does

- Per #650:
  - Replaces filters icon with new SVG icon, which changes colors based on whether or not filters are active
- Per #524:
  - Replaces filter checkmarks with Phosphor check-circle and check-circle-fill


sitemedia/js/controllers/search_controller.js
```diff
@@ -36,6 +43,24 @@ export default class extends Controller {
...
+        if (this.filterTargets.some((target) => target.checked)) {
+            this.filterLabelTargets.forEach((label) =>
+                label.classList.add("active")
+            );
+        } else {
+            this.filterLabelTargets.forEach((label) =>
+                label.classList.remove("active")
+            );
+        }
+    }
```
rlskoeser, line 62: can't this be done server side? we just need to set an `active` class on the filter toggle if there are filters applied, right? how about a simple boolean method on the form that checks if there are filters active 

my inclination is that the we shouldn't display the active status until the filters have been applied to the search; once we implement the reactive search they will be applied automatically, but I think the server side active status still works in that case

rlskoeser (commented):
I love how this works! Just a question about whether we can't set filter status to active in the template instead of needing JS

sitemedia/js/controllers/search_controller.js
blms: Good point!

blms:
Quick question for re-review: does it make sense to return False if there are form errors? I figure it does, but can adjust otherwise

rlskoeser (approved):
It seems logical to me to return false if the form isn't valid!

Revision looks good to me.

Issue #638, opened 2022-02-21
As a front-end user, I want the document search to automatically reload when I change my search terms, filters, or other options so tha I can see the changed results more quickly. 


rlskoeser:
## testing notes (QA, round 2)

- [x] After making a search, navigate through the pages of results and click on a document. Hit the back button and confirm you remain on the page of results you were on before. Note any feedback about this overall interaction

---

## testing notes (QA)

- [x] On the document search page in the public site (QA), try entering some search terms, adding filters, and changing sort; confirm that the counts (for filters) and results (below search form) update automatically
- [x] Page through search results, click on some documents in the results, and note feedback here

---

related to #635

richmanrachel:
@rlskoeser - overall it looks great and seems to be working well!

One little (probably) unrelated hiccup I'm noticing, is that if you click on an entry and then "back," you wind up on the first page of the results even if you were on another page. Is this expected behavior?

rlskoeser:
No, it should go to the last page of results you were on — I noticed this too.

Let us know what else you find! We know there are still some things we need to address, but wanted to get some testing feedback on it.

richmanrachel:
I still am missing a "reset filters" button, because manually unclicking is odd. 

I also worry that folks won't know to expand the document type filter. Perhaps scholarship records should be grouped together in the same way?

gissoo:
@blms @rlskoeser this works well on all browsers: one Q: is it expected that you would have to click out of the search box to see updated results? results are not updated when I'm done typing till I click out of the search box – felt a bit weird. 

everything else is ok, tested on FF, GC, and Safari

@richmanrachel 
- the reset filters is on our list, just need to get to it!! 
- We've also discussed revising the way the scholarship records are positioned because stacking them won't probably be efficient – but I don't think we should "group" them as in they becoming selected by one click if that's what you mean? – (the results from earlier primary research showed that more people care about transcriptions but not necessarily translations, or they weren't looking for all three factors at the same time / in the same scenario.) 
- I think the doc type filter is clear but our revision of the position of the filters are probably going to help this as well – it's probably not efficient to spend more time on the UI for the filters till we have more filters to see what they would look like.

richmanrachel:
@rlskoeser and @blms - it works! Oddly, when I didn't have any filters on, it navigated back to the correct page but not the same spot on the page. Whereas with filters, the back button got me exactly back to where I was. 

Should I close or do you want to improve the handling for the non-filter scenario?

[...]

Issue #635, opened 2022-02-21
Implement Turbo to improve internal link speed


blms:
## dev notes

- [x] Use [Turbo](https://turbo.hotwired.dev/) convert all internal links to Turbo links.
- [x] Import Turbo in JS

blms:
Running into some issues with Turbo.

When wrapping the "main" content block with a `data-frame`:
- Links to document detail pages in search results do not receive the correct CSS and do not change the window location/URL (I guess this is all within a single "navigation context")

When just enabling Turbo in `stimulus_app.js` and using no `data-frame`s:
- Seems to result in more FOUCs and slightly worse performance than before, at least when testing locally
- Search page seems to refresh twice on submit
- Filter modal will not open/close unless it is given `data-turbo="false"` attribute
- Cannot view more than one OpenSeadragon instance in the same navigation context, even if `data-turbo="false"` is set on a containing element

rlskoeser:
@blms do you have a branch with your experiments that you could push so I can work with what you've already done?

blms:
@rlskoeser here you go! https://github.com/Princeton-CDH/geniza/tree/feature/635-turbo

rlskoeser:
@blms just pushed some changes to your branch:

- shift page-specific styles from `body` to `main` and update css 
- wrap `main` in `turbo-frame` so we get page-specific classname to trigger styling
- configure main turbo-frame with `target="_top"`; by default, links within a frame will stay within that frame, which is why loading the document page wasn't updating the url; I don't think we ever want that behavior, but I do wonder if we're losing some of the power/efficiency of turbo by doing it this way
- tweak openseadragon init to run on `turbo:load` event (proper solution will be to rewrite as a stimulus controller, but I think we have more changes we should incorporate and plan for in that change)
- placeholder styles for turbo progress bar (couldn't get our css variables to work)

take it for a spin and see what you think — any problems remaining? can we use this to turn on automatic form submission for search?

blms:
Thanks so much @rlskoeser!

I'm still seeing a couple of problems, but I think there is enough for me to take it from here.

For documentation:

- Home and content pages are not picking up their CSS
- OpenSeadragon still not working intermittently, not sure what triggers a failure or a success, but I can start rewriting as a stimulus controller so as not to break things

rlskoeser:
@blms I may have missed some of the css changes when I shifted things from body to main! I hope that's all it is.

[...]

PR #684, opened 2022-03-04
Turbo integration (#635) and automatic search (#638)

blms:
## What this PR does

- Per #635:
  - Integrates Turbo on internal links inside the `main` element
  - Creates Stimulus controller for OpenSeadragon to ensure proper loading and unloading after Turbo navigation
- Per #638:
  - Allows "automatic" search where query changes, filters, and sort are applied without manual form submission
  - Uses session storage to persist filters expanded/collapsed states


sitemedia/scss/pages/_search.scss
```diff
@@ -27,3 +27,12 @@
         padding: 0 spacing.$spacing-md spacing.$spacing-md;
     }
 }
+
+// Applying color per system theme
+.turbo-progress-bar {
+    height: 5px;
+    background-color: #436043;
+    @media (prefers-color-scheme: dark) {
+        background-color: #c37f97;
```
rlskoeser, line 36: Were you not able to get the css variables to work either? is it because of where the progress bar is added?

we should at least have comments here; maybe scss variables defined in the same place as the css variables 

I think you're missing the theme override cases here.

geniza/corpus/templates/corpus/snippets/document_result.html
```diff
@@ -107,15 +107,15 @@ <h3 class="sr-only">{% translate 'Tags' %}</h3>
...
+                        <a href="{% url 'corpus:document' document.pgpid %}" data-turbo-frame="main" target="_top">
                             <img src="{{ image.0|iiif_image:"size:width=250" }}" loading="lazy" alt="{{ image.1 }}">
                         </a>
                     </li>
                 {% endfor %}
             </ul>
         {% endif %}
         {# view link #}
-        <a class="view-link" href="{% url 'corpus:document' document.pgpid %}">
+        <a class="view-link" href="{% url 'corpus:document' document.pgpid %}" data-turbo-frame="main" target="_top">
```
rlskoeser, line 118: I don't think we need the `target="_top"` — I think turbo action set to advance and turbo frame `main` is sufficient; with `_top` I get a flash of unstyled pages; without it I think everything is working properly.

rlskoeser (changes requested):
This looks really good; tested locally and seems to be working well. 

There are a few small changes I'd like to see, noted in comments.

One more thing: right now there's no indicator when the form is being submitted. Can we do something simple keyed on the `turbo-frame[busy]` attribute? I'm imagining a very basic spinner overlay, maybe we could use the phosphor icon circle-notch.  Not sure which frame this should be specific to; probably only applies to the search page and we don't want to show the spinner elsewhere.

The other option I found in my quick search was to display the progress bar: https://github.com/hotwired/turbo/issues/61#issuecomment-752717046

rlskoeser:
@blms if you think the frame loading indicator is complicated and not needed right now, I'd be ok with waiting on that in favor of getting this out for testing sooner.

geniza/corpus/templates/corpus/snippets/document_result.html
blms: Quick note on this: the reason why I added `target="_top"` was because I kept getting scrolled most of the way down the document detail page when I navigated with this link. Turbo specifically says in the docs that links should always scroll back to the top unless a hash/anchor is in the URL, so I couldn't figure out why this was happening… the `target="_top"` was a quick fix. I should have mentioned earlier!

sitemedia/scss/pages/_search.scss
blms: Removed for now based on your comment below, we'll address it separately. I think we might want to apply themes to the `html` element instead of `body`.

geniza/corpus/templates/corpus/snippets/document_result.html
```diff
@@ -107,15 +107,15 @@ <h3 class="sr-only">{% translate 'Tags' %}</h3>
                 {# list of tuples of (IIIF image, label) #}
                 {% for image in document.iiif_images|slice:":3" %}
                     <li class="image-{{ forloop.counter }}">
-                        <a href="{% url 'corpus:document' document.pgpid %}">
```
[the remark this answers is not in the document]
blms: Yes, because of z-index! We could instead just make the z-index of the "clickable result" higher—which do you think would be better?

sitemedia/js/controllers/search_controller.js
```diff
@@ -1,38 +1,90 @@
...
 
 import { Controller } from "@hotwired/stimulus";
-// TODO: Re-enable debounce when Turbo is set up
-// import { ApplicationController, useDebounce } from "stimulus-use";
+import { ApplicationController, useDebounce } from "stimulus-use";
+import * as Turbo from "@hotwired/turbo";
 
 export default class extends Controller {
-    static targets = ["query", "sort", "sortlabel", "filterModal"];
-    // static debounces = ["submit"];
```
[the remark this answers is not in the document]
rlskoeser: ah, missed it further down due to the reformat!

geniza/corpus/templates/corpus/snippets/document_result.html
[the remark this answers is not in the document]
rlskoeser: better not to have redundant links if it's possible/easy

geniza/corpus/templates/corpus/snippets/document_result.html
blms: I started coding a fix for this where we do `window.scrollTo(0, 0)`—but when you navigate back to the search results, you end up losing your place in the search results. Probably preferable to being scrolled down when you click on the document, but not sure!

Here's the documentation I mentioned: https://turbo.hotwired.dev/handbook/drive#application-visits
> If the visit’s location includes an anchor, Turbo Drive will attempt to scroll to the anchored element. Otherwise, it will scroll to the top of the page.

rlskoeser (approved):
Wow, still some trade-offs in some of the behavior here! I think it's in good enough shape to merge, and it will be helpful to deliver for testing so that other people besides us can try it out and give us feedback.

Issue #688, opened 2022-03-07
Transcription lines appear left-aligned in admin interface


blms:
## testing notes

- [ ] In the admin interface, visit a document with a transcription, and confirm that the transcription text is right-aligned

## steps to replicate

- In the admin interface, visit a document with a transcription

rlskoeser:
confirmed the change in qa

PR #689, opened 2022-03-07
Right align transcriptions in admin (#688)

blms:
## What this PR does

- Per #688:
  - Uses `text-align: right` in admin CSS for transcriptions

## dev review notes

I think fixing this is as simple as aligning the text right, as it is in search results, but this comment gave me pause:

```css
/* left-aligned transcriptions; rtl is controlled thru html "dir=" attribute */
```

I realized the reason why it reverted to left-alignment is because of the Django fieldset and div classes I used to make transcriptions look nice in the admin interface. But I want to make sure I'm not missing some intention here behind the comment!

### DIFF
--- a/sitemedia/css/admin-local.css
+++ b/sitemedia/css/admin-local.css
@@ -100,11 +100,12 @@ a#needsreview:target {
     font-style: italic;
 }
 
-/* left-aligned transcriptions; rtl is controlled thru html "dir=" attribute */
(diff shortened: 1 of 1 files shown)


rlskoeser (approved):
I don't remember that comment! Based on the blame history, maybe it was Nick's comments; this seems fine, at least in the short term.

Issue #694, opened 2022-03-07
Implement tabs for the Hebrew site 


gissoo:
## testing notes (qa)

- [x] On a document details or scholarship records page, switch languages to Hebrew and confirm that the tabs at the top are in the right order and are styled correctly

---

Please refer to [this comment on the design issue ](https://github.com/Princeton-CDH/geniza/issues/442#issuecomment-1049248656)

gissoo:
@blms I can't confirm the correctness of the "Next" and "Previous" text in Hebrew, but the designs are correct. A couple of problems I'm having:

1. Switching to Hebrew doesn't give me anything till I change the sort from "Random" to something else since we disabled pagination – I think it should still translate even if it's disabled, because it's still visible 
2. Even when the sort is on something other than Random, the order of the pages don't change to RTL, just the words change to Hebrew till I do a hard refresh. And as of right now switching to Hebrew is not doing anything at all here (testing on firefox) It's going back to En after choosing Hebrew and pressing "go" 

note on Google Chrome: the switch is working on Google Chrome, except needs a hard refresh to change the direction of the pages again, but works on "random" sort too! 

note on Safari: it's the same as it is on Google Chrome 

ok – so looks like switching to Hebrew stops working in all the browsers after I do a hard refresh ... It's going back to En after choosing Hebrew and pressing "go"

gissoo:
@blms this works correctly for me, but only works after I close the current window and reopen it following our comments on #693 and your mention of the problem with Turbo.

Issue #693, opened 2022-03-07
Implement pagination for the hebrew site 


gissoo:
## testing notes (qa)

- [ ] After a hard refresh (shift + refresh), confirm that you can switch between languages
- [ ] On the QA site, visit the document search page, and switch your language to Hebrew. Confirm the pagination looks and works correctly

---

Please refer to [the comment on design issue #441](https://github.com/Princeton-CDH/geniza/issues/441#issuecomment-1049250603)

gissoo:
@blms I can't confirm the correctness of "previous" and "next". But the designs are correct. 

The problem is that when I choose to view the site in Hebrew:

1. nothing happens till I change the sort from "Random" to something else. 
2. even when the sort is not on "random", the pagination text changes to Hebrew but the numbers don't till I do a hard refresh ...as of right now the Hebrew switch is not working at all

blms:
Thanks @gissoo, I think this is a Turbo issue 😬  @rlskoeser but we can prioritize it for later, since we're not shipping out Hebrew yet… should I make a new issue for that?

richmanrachel:
@gissoo - Not sure if I'm supposed to be testing this or not, but everything is working for me on Chrome as expected except for the lack of space between the search result number and the document title:
<img width="689" alt="image" src="https://user-images.githubusercontent.com/68028846/160673764-49161874-7dac-483b-8aaf-820344a0b2d4.png">
In contrast, the English page looks like this:
<img width="724" alt="image" src="https://user-images.githubusercontent.com/68028846/160673859-f41ba715-9d86-4c8d-9468-ebdd2508f5b5.png">

gissoo:
@richmanrachel thanks for testing – yes I can see the issue with spacing only for the 1st doc. But I'll wait till I hear back from @blms about my earlier comments since I've already tested this, and then can test again.

blms:
@gissoo I believe the things you mentioned should be fixed now since the form does not use Turbo. Can you try those again?

@richmanrachel: I actually hadn't implemented the design for that part yet! I think this issue is just for the pagination, meaning, the "previous," "next," and page numbers links. So I think those things you mentioned would go under an issue "search results" instead.

[...]

PR #699, opened 2022-03-07
Implement right-to-left pagination (#693) and tabs (#694)

blms:
## What this PR does

- Per #693:
  - Implements right-to-left pagination functionality and styles

- Per #694:
  - Implements right-to-left tabs

## dev notes

- I noticed that the `LANGUAGE_CODE` variable was set to `get_current_language`, but it seems that was unnecessary—I guess it's set by Django to be the current language already. I noticed that the variable was already being used before it was defined for the html element's lang attribute in `base.html`. Hopefully I haven't missed something that will break deployment, but it works for me locally.

### DIFF
--- a/geniza/corpus/templates/corpus/snippets/document_tabs.html
+++ b/geniza/corpus/templates/corpus/snippets/document_tabs.html
@@ -2,7 +2,8 @@
 <!-- document detail page navigation -->
 {# Translators: accessibility text label for document detail view tabs navigation #}
 <nav aria-label="{% translate "tabs" %}">
-    <ul class="tabs">
+    {% get_current_language as LANGUAGE_CODE %}
+    <ul class="tabs" {% if LANGUAGE_CODE == "ar" or LANGUAGE_CODE == "he" %}dir="rtl"{% endif %}>
         {% url 'corpus:document' pk=document.pk as document_url %}
         <li><a href="{{ document_url }}"{% if request.path == document_url %} aria-current="page"{% endif %}>{% translate "Document Details" %}</a></li>
         {% with n_records=document.sources.count %}
--- a/geniza/corpus/templates/corpus/snippets/document_tabs.html
+++ b/geniza/corpus/templates/corpus/snippets/document_tabs.html
@@ -2,7 +2,6 @@
 <!-- document detail page navigation -->
 {# Translators: accessibility text label for document detail view tabs navigation #}
 <nav aria-label="{% translate "tabs" %}">
-    {% get_current_language as LANGUAGE_CODE %}
     <ul class="tabs" {% if LANGUAGE_CODE == "ar" or LANGUAGE_CODE == "he" %}dir="rtl"{% endif %}>
         {% url 'corpus:document' pk=document.pk as document_url %}
         <li><a href="{{ document_url }}"{% if request.path == document_url %} aria-current="page"{% endif %}>{% translate "Document Details" %}</a></li>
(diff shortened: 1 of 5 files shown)


geniza/templates/snippets/language_switcher.html
```diff
@@ -5,7 +5,6 @@
         <fieldset>
             <input name="next" type="hidden" value="{{ redirect_to }}">
             <select name="language">
-                {% get_current_language as LANGUAGE_CODE %}
```
rlskoeser, line 8: Right, this is set for us because we have `django.template.context_processors.i18n` enabled in our settings.  Good catch on removing it.

geniza/corpus/templates/corpus/snippets/document_tabs.html
```diff
@@ -2,7 +2,7 @@
 <!-- document detail page navigation -->
 {# Translators: accessibility text label for document detail view tabs navigation #}
 <nav aria-label="{% translate "tabs" %}">
-    <ul class="tabs">
+    <ul class="tabs" {% if LANGUAGE_CODE == "ar" or LANGUAGE_CODE == "he" %}dir="rtl"{% endif %}>
```
rlskoeser, line 5: do we need to put the `rtl` here? can it not be set at a higher level of context, or not yet because of the mixed content and we're only doing components?

sitemedia/scss/components/_pagination.scss
```diff
@@ -110,4 +110,11 @@ nav.pagination {
         font-family: "Phosphor";
         content: "\f044"; // phosphor arrow-right icon
     }
+    // Hebrew and Arabic reversed arrows
+    &[dir="rtl"] .prev::before {
+        content: "\f044"; // phosphor arrow-right icon
+    }
+    &[dir="rtl"] .next::after {
+        content: "\f03b"; // phosphor arrow-left icon
+    }
```
rlskoeser, line 119: will this be a problem if we're able to move the text-direction up to a higher level in the page?

rlskoeser (commented):
It looks good to me from what I can tell, just have a question about what level of the page we should be setting text direction — would like to set it at the highest level possible and then override as needed, but I understand if that's not practical (either not yet or not at all).

Is this because we're implementing components only? Will we need to revisit later?

geniza/corpus/templates/corpus/snippets/document_tabs.html
blms: Right, since we're only doing certain components they'll need to either be set RTL one at a time, or the whole page set to RTL and then individual elements set LTR until RTL is implemented. We could just do that for the few problematic elements, but it might take some time to figure out which those are. (The search results page seems to be an especially bad example!)

blms: It might also be fine to just have some things look a little broken until implemented…!

sitemedia/scss/components/_pagination.scss
blms: Yes, the `[dir="rtl"]` selector will need to match whatever level that attribute is set. Good catch!

geniza/corpus/templates/corpus/snippets/document_tabs.html
rlskoeser: IMO it's ok if they don't display correctly in the interim since we're not making this version public yet. I'd rather implement it the way we intend to have it work at the end than have to go back and refactor everything. (i.e., set rtl once at the top level based on language, and implement styles and overrides for the components as you work your way through them.)

What do you mean about the "few problematic elements" ?

blms: By "few problematic elements" I just meant: some things already are usable and work fine (like most of the content page and document detail page elements), but some look very messy without special styling applied (like most of the elements on the search results page)—so I considered the latter problematic!

rlskoeser:
Based on this https://www.w3.org/International/questions/qa-html-dir 

I think we should set the direction explicitly on the html element; I'll have to look and figure out if `modeltranslate` tells us what language it's giving us when it gives us the fallback content (if a translation isn't available).

Issue #711, opened 2022-03-09
As a front end user, I want some kind of progress/loading indicator during search, so that I know my input was accepted.


blms:
## dev notes
https://github.com/hotwired/turbo/issues/61#issuecomment-752717046
- [ ] style `.turbo-progress-bar`
- [x] use `html` element rather than `body` for applying theme overrides, so we can get the variables

richmanrachel:
@rlskoeser - is this worth keeping open? Everything loads so quickly for me that it's unnecessary on my end.

kseniaryzhova:
Search loads super quickly on the public site, so closing!

Issue #634, opened 2022-02-21
Refactor all JS to Stimulus


blms:
## dev notes

[Stimulus](https://stimulus.hotwired.dev/) was first used for the relevance sort in #550, the following should also be refactored to use Stimulus:

- [x] `search.js` (all functionality should be moved over to `controllers/search_controller.js`)
- [x] `submenu.js`
- [x] `theme_toggle.js`
- [x] `openseadragon.js` (see #684)

blms:
@rlskoeser exactly!

rlskoeser:
@blms what about openseadragon init?

blms:
@rlskoeser Ah, I didn't think to include it since it's just injecting OSD vendor code into the HTML, but it probably should be included, I'll add it to the list!

rlskoeser:
We may want to split this out into separate subtasks, but can wait and see how it goes; am wondering if the seadragon should be more of a progressive enhancement. I was looking at the site on my phone and wasn't seeing images on detail pages — which we should check on — but made me wonder if you really want deep zoom by default on mobile.

blms:
@rlskoeser Good call, and I see the mobile issue—it occurs on pages where there are images but no transcription. I'll track it as a bug!

rlskoeser:
@blms great — thanks for documenting, we should get that resolved

rlskoeser:
confirmed in qa, but noticed we're still including a main.js bundle that appears to be empty — let's remove that one before we close this

rlskoeser:
confirmed main.js removed in qa

PR #713, opened 2022-03-09
Stimulus refactor (#634)

blms:
## What this PR does

- Per #634:
  - Refactors all JS into Stimulus controllers
- Per #711:
  - Sets theme overrides on `html` rather than `body` element


rlskoeser (approved):
Code changes look good, everything working well for me locally.

PR #717, opened 2022-03-10
Turbo revisions: scroll and page number (#638)

blms:
## What this PR does

- Per #638:
  - Fixes the issue with search page number reverting to 1 on "restore" visit
  - Saves and restores scroll position as a workaround for document detail/search scrolling issues


rlskoeser (approved):
Changes look good, seems like the behavior is much better.

The lack of notification or progress indicator when you click a link and are waiting for a response is a little disconcerting, but I think we can live with it for now.

PR #761, opened 2022-04-12
Combine all webpack js/css into one bundle

blms:
## What this PR does

- Fixes an error with `webpack-dev-server` that required loading the JS output of a bundle (even if empty) alongside any CSS, by simply combining the two bundles into one


rlskoeser (approved):
Thank you for catching and resolving this! I didn't test locally, but everything you've changed looks right to me.

Issue #706, opened 2022-03-08
As a frontend user, I want to search in Arabic script and get search results from both Arabic and Judaeo-Arabic transcriptions so that I can find more content that matches my search.


richmanrachel:
## testing notes
- [x] review the [character mapping in the code](https://github.com/Princeton-CDH/geniza/blob/4d1f24f418e009d00d20d390f685400888e776a2/geniza/corpus/ja.py#L5-L35) to check that I correctly adapted the characters from the chart in Marina's document  (although the mixed ltr/rtl display is a little mixed up on some of the lines in the github view)
- [x] review [Hebrew script final character mapping](https://github.com/Princeton-CDH/geniza/blob/4d1f24f418e009d00d20d390f685400888e776a2/geniza/corpus/ja.py#L37-L44) 
- [x] try some keyword searches in arabic script on the [test site search](https://test-geniza.cdh.princeton.edu/en/documents/) — you should get matches in both arabic and judaeo-arabic when the word occurs in both forms

---- 

**Is your feature request related to a problem? Please describe.**
Millions of Arabic speakers around the world could be interested in the Cairo Geniza and would be able to understand many of documents if Hebrew script was not a blocker. 

**Describe the solution you'd like**
Use [Marina's Arabic to Judaeo-Arabic conversion chart](https://docs.google.com/document/d/1hcwWwBuPEmxc_FHpSk1JWoBZvs5Y1LOn8v6PIqaP9mA/edit) to convert Arabic queries so that they also search against Judaeo-Arabic.

rlskoeser:
@richmanrachel could you provide a few search terms I could use for testing? i.e., a few arabic terms that will match existing judaeo-arabic transcriptions

richmanrachel:
@rlskoeser - sure! 

دينار
مصحف
سنة
Is that enough? (those respectively are dinār [the gold coin/main currency], maṣḥaf [a book/usually a Bible], and sana [year])

rlskoeser:
@richmanrachel this one is ready for testing now. I tried the three terms you gave me to test with and only one of them is returning judaeo-arabic matches – I don't know if this means they don't occur, or if I've done something wrong with the mapping.

[...]

PR #767, opened 2022-04-14
Implement arabic to judaeo-arabic, apply to search

rlskoeser:
ref #706

adds some new utilities to check for arabic, convert arabic words to judaeo-arabic, and applies to search terms as an OR to match the word in either script


blms (approved):
Looks great! Lots of good python here; I didn't know about `translate` and `maketrans`.

Issue #724, opened 2022-03-11
As an admin user in document view, I'd like to be able to zoom on the fragment's IIIF image thumbnail so I can determine the language and check other metadata details as I'm writing or editing a description. 


mrustow:
## Testing notes (QA)

In the admin UI on the QA site, navigate to the list of documents. Click on some documents and confirm:
- [x] you can see images/transcriptions on documents that have them (try by filtering on documents that: only have image, only have transcription, have both, and have neither)
- [x] you can zoom and rotate IIIF images from within the document edit form
- [x] this layout and approach is OK, or comment feedback/suggestions

---

At the moment, I have to hit the edit pencil under "related fragments," then in the pop up window, scroll down, and most likely enlarge the pop-up window before I can see what's on the image. 

Also, some of the IIIF images are sideways (e.g., [T-S NS 99.55](https://cudl.lib.cam.ac.uk/view/MS-OR-01080-00001-00088/1)), so it would be great to be able to rotate the images from document and fragment view. 

## dev notes
Let's switch the admin document change form to use the same image & transcription display that we're using on the public site. That way, all improvements for display and transcription will be available in both places as we improve things and build out the functionality. 

It should be easy enough to include the template snippet. What about the javascript? Should we split out search and viewer so that we can load just what's needed?

This change should also resolve #708

richmanrachel:
@mrustow - does this earlier user story capture the same solution? https://github.com/Princeton-CDH/geniza/issues/384 

(But thank you for bringing this up again! I agree we can improve image viewing in the admin site)

mrustow:
Ah yes that’s it — sorry, I hunted for it but was searching IIIF instead of mirador so didn’t find it.

richmanrachel:
Closed #384 so that we could focus on the image aspect rather than Mirador specifically. Here's the phrasing of the previous issue for ease: 
"As a content editor, I would like to see the Mirador viewer for the document in the Admin Document Detail Page so I can properly consult the image while updating the metadata. #384"

richmanrachel:
@rlskoeser and @blms - this is great for now! Can't wait until we can edit the transcriptions right there too! (which means that there should be an empty transcription box for those without transcriptions so we can add them in the future, instead of having the image take over the whole section).

PR #770, opened 2022-04-14
Use IIIF viewer from doc detail view in admin UI (#724)

blms:
## What this PR does
- Per #724:
  - Uses the template, JS, and CSS from the document detail view IIIF viewer in the admin UI
  - Changes the `has_transcription` check to `digital_editions`, since otherwise it was showing an empty transcription container in the admin UI
  - Converted `H1` to `H3` in all transcriptions to avoid the "language dropdown" issue (I hate that they attach that to ALL `H1` elements!)
  - Splits IIIF viewer JS/CSS into a separate bundle to allow only loading that, and not the other JS/CSS, in admin

## Dev notes
- I had to include the `iiif` Webpack bundles in `base.html` rather than `document_detail.html` in order for Turbo to pick them up. Not entirely sure why that is, but also not sure if it's worth investigating since we were already including that JS/CSS before in `base.html` anyway.

### DIFF
--- a/geniza/corpus/templates/admin/corpus/document/change_form.html
+++ b/geniza/corpus/templates/admin/corpus/document/change_form.html
@@ -1,4 +1,16 @@
 {% extends "admin/change_form.html" %}
+{% load static render_bundle_csp  %}
+
+{% block extrastyle %}
+    {{ block.super }}
+    {% render_bundle_csp "iiif" "css" %}
+{% endblock %}
+{% block admin_change_form_document_ready %}
+    {{ block.super }}
+    {% render_bundle_csp "iiif" "js" attrs='defer' %}
+{% endblock %}
+
+
 {% block after_field_sets %}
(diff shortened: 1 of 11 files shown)


rlskoeser (approved):
This looks great! Changes look reasonable; pulled and tested locally, looks good to me — am eager to see what Rachel thinks of it in testing.

I wondered if we could remove the footnote transcription template snippet but checked and saw that it's still in use. I think it will probably go away eventually, but fine to leave as is for now.

I'm running into the same thing with splitting javascript or loading them piecemeal — I think turbo doesn't load new javascript in the header by default because it's just replacing the frame content. It seems like there ought to be a way to do this, so we can segment out our code a bit (and e.g. not load all the iiif stuff if someone never looks at an image). It's going to apply to the transcription editing also, so maybe we can put our heads together on a better solution in the long term — but this seems fine for now.

Issue #691, opened 2022-03-07
As a front end user, I want a filter for documents that have images, so that I can limit results to documents where I'll have ready access to visuals of the fragments.


blms:
## Testing notes (QA)

On the document search:
- [x] Enable the "Has Images" filter and confirm only documents with images show up
- [x] Combine with other filters and confirm it works as expected
- [x] Comment: should it be first in the list, or below the scholarship records filters?

richmanrachel:
@blms - it looks great! I like having the images filter first in the list. I'll double check with Marina shortly about her preferences, but I think we can close for now!

PR #771, opened 2022-04-14
Add has_image filter to document search (#691)

blms:
## What this PR does

- Allows the user to filter documents by the presence of IIIF images

### DIFF
--- a/geniza/corpus/forms.py
+++ b/geniza/corpus/forms.py
@@ -226,7 +226,7 @@ class DocumentSearchForm(forms.Form):
     }
 
     def filters_active(self):
-        # returns True if any filter fields are active in cleaned_data
+        """Check if any filters are active; returns true if form fields other than sort or q are set"""
         if self.is_valid():
             return bool(
                 {
--- a/geniza/corpus/forms.py
+++ b/geniza/corpus/forms.py
@@ -226,13 +226,15 @@ class DocumentSearchForm(forms.Form):
     }
 
     def filters_active(self):
+        # returns True if any filter fields are active in cleaned_data
         if self.is_valid():
-            return (
-                self.cleaned_data["doctype"]
-                or self.cleaned_data["has_transcription"]
-                or self.cleaned_data["has_translation"]
-                or self.cleaned_data["has_discussion"]
+            filter_fields = list(
+                filter(lambda k: k not in ["q", "sort"], self.fields.keys())
             )
+            active_fields = dict(
+                filter(lambda i: bool(i[1]), self.cleaned_data.items())
+            ).keys()
+            return bool(set(filter_fields).intersection(set(active_fields)))
         return False
 
     def set_choices_from_facets(self, facets):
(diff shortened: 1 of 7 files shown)


geniza/corpus/forms.py
```diff
@@ -200,6 +200,10 @@ class DocumentSearchForm(forms.Form):
         # Translators: label for document type search form filter
         label=_("Document Type"),
     )
+    has_image = BooleanFacetField(
+        # Translators: label for "has transcription" search form filter
```
rlskoeser, line 204: looks like you copied the comment and forgot to update it :-) 

```suggestion
        # Translators: label for "has image" search form filter
```

rlskoeser (changes requested):
Looks good — it's great that we're at a point where it's so easy to add new filters like this.

I did notice that the filter icon didn't change on the site — looks like you need to update the `filters_active` method. Should we add a comment to the part of the code where new form fields would be added as a reminder to update that method? OR, alternatively, could we consider everything except keyword search and sort to be filters and write it that way, so that we won't have to update it every time we add a new filter? (that option sounds better to me, but I could see it being a separate refactor — if you agree, you could make a new issue for that and do the simple fix now).

I'm not sure about the filter order — I don't think I would have put images first (I think scholarship records are more important). But probably best to get input from Rachel on that, and easy enough to switch order if we need to.

geniza/corpus/forms.py
```diff
@@ -226,13 +226,15 @@ class DocumentSearchForm(forms.Form):
...
-                or self.cleaned_data["has_transcription"]
-                or self.cleaned_data["has_translation"]
-                or self.cleaned_data["has_discussion"]
+            filter_fields = list(
+                filter(lambda k: k not in ["q", "sort"], self.fields.keys())
             )
+            active_fields = dict(
+                filter(lambda i: bool(i[1]), self.cleaned_data.items())
+            ).keys()
+            return bool(set(filter_fields).intersection(set(active_fields)))
```
rlskoeser, line 237: This looks more complicated than I think we need (and it's hard to read).

How about this?

```suggestion
            return bool(
                {
                    k: v
                    for k, v in self.cleaned_data.items()
                    if k not in ["q", "sort"] and bool(v)
                }
            )
```

geniza/corpus/forms.py
```diff
@@ -226,13 +226,15 @@ class DocumentSearchForm(forms.Form):
     }
 
     def filters_active(self):
+        # returns True if any filter fields are active in cleaned_data
```
rlskoeser, line 229: we should probably update the comment here (and should probably be a docstring, too)

```suggestion
        """Check if any filters are active; returns true if form fields other than sort or q are set"""
```

geniza/corpus/forms.py
blms: Looks much cleaner! Didn't realize we didn't need to actually compare it with `self.fields`, great solution.

rlskoeser (approved):

Issue #775, opened 2022-04-19
Implement the revised image permissions statement 


gissoo:
## Testing notes (qa -- round 2)

- [x] On the QA site, confirm focus styles are correct for the "Image Permissions Statement" buttons

---

## Testing notes (qa)

- [x] On the QA site, navigate to some documents with images, and confirm the image permissions statement link and list look correct.

---

[Link to mock-ups on Figma](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=1184%3A35954)

- [x] default view (collapsed) with downward phosphor icon chevron – on desktop and mobile
- [x] selected view (expanded) with upward phosphor icon chevron – on desktop and mobile
- [x] font styles and alignment for label: link style, below the transcription panel pagination, and bold shelfmark, taking up the full line when expanded – on desktop and mobile
- [x] spacing between the label and the transcription panel pagination and between each shelfmark when expanded – on desktop and mobile – marked in red on the mock-ups

rlskoeser:
@blms please implement so it supports displaying both attribution and license; we don't currently have both, but it's feasible we will in future

gissoo:
@rlskoeser I will email them this week about access and changes in our subscription, this is just unacceptable – I thought we agreed to display them this way because Marina mentioned the "shelfmark by shelfmark" and you said it made sense to you – is that not necessary anymore? – I can revise the designs and change the styling if that's the case – I'll create a chore

rlskoeser:
sorry I didn't think of this before, @gissoo!  It seemed reasonable to me to be explicit about rights and permissions when there are multiple shelfmarks, since the attribution and license may differ (probably by instutition/source).  It's also a bit more logic for us to figure out how to aggregate them, since the order may not align with the source.

I'm not sure how important it is that the logos be tied to specific shelfmarks — I think it's more important to display them once, and would prefer to display them only once.

gissoo:
@blms thank you!! The sizes and styles look correct, but some have the image/logo and some like the below don't. Are there cases where there is more than one shelfmark and one logo/right image only? I can't find one to test. If there are no cases like that or you don't want me to test those please feel free to close this PR. 

<img width="947" alt="Screen Shot 2022-04-22 at 6 31 07 PM" src="https://user-images.githubusercontent.com/44615389/164807168-abd1ca7c-b6ee-43f4-95ee-562050e11e65.png">

gissoo:
[...]

PR #778, opened 2022-04-19
Style fragment attributions (#775)

blms:
## What this PR does

- Per #775:
  - Implements new style for fragment attributions
- Adds functionality to only show attribution section when there are known fragment attributions

## dev review notes

- I'm especially looking for feedback on the `has_fragment_attribution` function—is there some way to get this in the template without an extra function on the model?


rlskoeser (commented):
I think you're right that the check for attribution can't easily be done in the template, but I think it would be simpler just to always display something. The PUL manifests (for the JTS content) don't have an attribution statement, but they do have a license uri which we should be displaying if set. How about some boiler plate text if not set in the manifest? e.g. "no attribution or license noted"

Here's the template code we used in mep for displaying rights statements: https://github.com/Princeton-CDH/mep-django/blob/7d76f7ccee2cef779c99f5e564f62da1e9574d6f/mep/people/templates/people/snippets/manifest_license.html#L13-L19

We put a local copy of the images in sitemedia for the project, could do the same here: https://github.com/Princeton-CDH/mep-django/tree/7d76f7ccee2cef779c99f5e564f62da1e9574d6f/sitemedia/img/rightsstatements_org 

The rights statement id, license label, etc. are all in the djiffy code and should be reusable. 

Per IIIF spec, we are also obligated to display logos included in the manifests. The mep logo logic is in the same template snippet https://github.com/Princeton-CDH/mep-django/blob/7d76f7ccee2cef779c99f5e564f62da1e9574d6f/mep/people/templates/people/snippets/manifest_license.html#L4-L12 — but ignore the linking logic; the rendering for these items currently links to Figgy, which we don't want to do.  

It isn't in the designs, but was thinking it would be good if we could display each unique logo once at the bottom of the attribution/license information.

Can we try these changes and see how it looks?  Gissoo may want to revise the visuals — although we should all remember this is a provisional version until we implement the new image & transcription view. Sorry for adding changes after you've worked on this.

rlskoeser:
@blms one more thing: the CUDL attribution statement ends with a sentence about the metadata, but we're not displaying their metadata. Please come up with a simple way to remove that sentence from the attribution before we display the attribution.  (I think that's allowable based on the IIIF presentation spec)

blms:
@rlskoeser Haven't written any tests yet for the new code, but does this seem like the right approach? (No logos yet either, just license/rights statement)

rlskoeser:
> @rlskoeser Haven't written any tests yet for the new code, but does this seem like the right approach? (No logos yet either, just license/rights statement)

approach seems generally sound, but would like to simplify in two places:
- we've talked about phasing out the remote manifest stuff, let's drop that; without that, then you shouldn't need the custom `license` method on the fragment and can just use the existing manifest properties 
- I think a template snippet would be clearer than a template tag for the license display (the tuple is opaque and not semantic for passing values)

Why don't we leave logos out of this PR and handle it separately? In a way, that's kind of a new feature... 

I happened to notice that the rights statement logos don't display well on the dark background by default — perhaps you've noticed as well; hopefully there's a simple css solution. (invert colors? light background color?)

blms:
@rlskoeser Made the requested changes! Only thing I noticed is that the rights statement text is now quite small in order to fit two lines of text into the single line of the attribution/license section. But I used the `title` property for the SVG as it's the recommended accessibility practice for inline SVGs—inline for recoloring purposes—so hopefully the license text label still be accessible that way.

rlskoeser (approved):
Latest updates look good to me. I agree the rights statement image is pretty small, but this is a provisional implementation anyway, so let's get it merged and move on.

PR #790, opened 2022-04-21
Setup sphinx documentation

rlskoeser:


blms (approved):
All of this looks good to me! Added a command to the workflow to exit with code 1 if there is any missing coverage. I also built the docs locally and checked them out and they look great. Exciting!

rlskoeser:
@blms thank you for adding that check! It's now failing, as I hoped — I left one item undocumented as a way of checking, and also because I wasn't sure how to document.

Could you remind me what this is for? The only place I see it used is in the view where we construct our local manifest. https://github.com/Princeton-CDH/geniza/blob/main/geniza/corpus/templatetags/corpus_extras.py#L108-L116

blms:
@rlskoeser Yes, that's all it's used for, so should probably not be a template tag. I believe it was formerly used for displaying attributions in the document detail template, but we stopped displaying that per `Document` and instead show attributions for each `Fragment`. So I think that is the only place it's used, and it's also the only reason we still have a `Document.attribution()` function.

rlskoeser:
> @rlskoeser Yes, that's all it's used for, so should probably not be a template tag. I believe it was formerly used for displaying attributions in the document detail template, but we stopped displaying that per `Document` and instead show attributions for each `Fragment`. So I think the local manifest is the only place it's used, and the local manifest is also the only reason we still have a `Document.attribution()` function.

Wow, thanks for remembering all that. That's more cleanup than I want to do at the moment, so I'm creating a cleanup chore to track the work and will add a docstring to hopefully get the sphinx coverage check passing. 😁

Issue #779, opened 2022-04-20
As a user, I would like to see historic and converted dates on the document details page so that I can easily find date information when it is known.


rlskoeser:
## testing notes (qa)

On the QA public site, check document detail pages for documents with historical dates, converted dates, and any combination/lack thereof, and confirm: 
- [x] The dates are present on those documents' detail pages (if they should be)
- [ ] The dates are formatted correctly

## dev notes
- [x] create a method on `Document` to render historical date combining display form of `doc_date_calendar` and `doc_date_original`
- [x] Update document details template to display original and converted document date if the fields are set.
 
Follow the latest version of the document details designs for placement and labeling.

In the html, let's wrap the whole thing (historic and converted date text) in a `time` element and set the converted date as the `datetime` attribute.

richmanrachel:
@rlskoeser and @blms - could we perhaps a comma before the type of calendar so it's clearer? Here's how it looks now:
<img width="248" alt="image" src="https://user-images.githubusercontent.com/68028846/165316440-6085c17f-822b-4d43-9659-c5b9c0d310ea.png">

blms:
@richmanrachel That's odd! I don't know of any data refresh, though maybe there was one—@rlskoeser?

In any case, you should be able to see newly added dates either way. Can you provide an example PGPID for one you tried to add that didn't show up?

In case it helps, here's an example of one that is showing up successfully, both on the admin and the public site:
https://test-geniza.cdh.princeton.edu/admin/corpus/document/16798/change/
https://test-geniza.cdh.princeton.edu/en/documents/16798/

rlskoeser:
@richmanrachel I haven't refreshed the test data in a while — LMK if it would be helpful.

I did notice that the site url had not been set correctly in the db, so if you were editing documents and following the 'view on site' link, it would have taken you to the production site.

here's a list of sample documents with dates: 23729, 1715, 35163, 8394, 34913, 19138, 35159, 32167, 30735, 30733

richmanrachel:
@rlskoeser - you're exactly right, the view on site link was pulling me onto the production side. Thank you so much for the list of PGPIDs

It looks pretty good! I want @mrustow's opinion before closing, though. I'll send her some screenshots.

blms:
Thanks @mrustow! That's very helpful. There are two more calendars: Kharājī and Anno Mundi. How should those be abbreviated? I know Anno Mundi should probably be "AM"…

[...]

PR #793, opened 2022-04-21
Implement historical/converted date display (#779)

blms:
## What this PR does

- Per #779
  - Implements display of historical and converted dates on the document detail view


blms:
This one should be good to go with what we have now, also opened issue #794 to address concerns.

geniza/corpus/templates/corpus/document_detail.html
```diff
@@ -70,11 +70,27 @@ <h2 class="sr-only">
                 {% endif %}
             </dl>
             {# secondary metadata #}
-            {% if document.languages.exists or document.secondary_languages.exists %}
+            {% if document.languages.exists or document.secondary_languages.exists or document.doc_date_original or document.doc_date_standard %}
```
rlskoeser, line 73: do we really need this extra check? what's the downside to including an empty secondary metadata `dl` ?

geniza/corpus/models.py
```diff
@@ -527,6 +527,14 @@ def shelfmark_display(self):
...
         return certain[0] + (" + …" if len(certain) > 1 else "")
 
+    @property
+    def historical_date_display(self):
+        if self.doc_date_calendar:
+            calendar_name = dict(self.CALENDAR_CHOICES)[self.doc_date_calendar]
+            return f"{self.doc_date_original} {calendar_name}"
+        else:
+            return self.doc_date_original
+
```
rlskoeser, line 537: Django automatically provides a display method for choice fields, please use that instead.

I think you can simplify like this:
```suggestion
        return " ".join([self.doc_date_original, self.get_doc_date_calendar_display()]).strip()
```

geniza/corpus/models.py
```diff
@@ -527,6 +527,14 @@ def shelfmark_display(self):
             return None
         return certain[0] + (" + …" if len(certain) > 1 else "")
 
+    @property
+    def historical_date_display(self):
```
rlskoeser, line 531: please add a docstring

rlskoeser: can we just rename to `original_date` or similar? display seems unnecessary since it doesn't conflict with any fields

geniza/corpus/templates/corpus/document_detail.html
```diff
@@ -70,11 +70,27 @@ <h2 class="sr-only">
...
+                                <time{% if document.doc_date_standard %} datetime="{{ document.doc_date_standard }}"{% endif %}>
+                                    {{ document.historical_date_display }}{% if document.doc_date_standard %}
+                                        ({{ document.doc_date_standard }})
+                                    {% endif %}
+                                </time>
+                            {% else %}
+                                <time datetime="{{ document.doc_date_standard }}">
+                                    {{ document.doc_date_standard }}
+                                </time>
+                            {% endif %}
```
rlskoeser, line 91: This looks more complicated than I want in a template. The two different cases are because the standardized date needs to be parenthesized if we have a historical date? 

Let's shift the logic to a document method — how about just calling it `document_date`? (I think we'll probably need this for indexing in Solr anyway.)

Should we add `CE` for the standardized date? That seems to be the convention in the dates I see in descriptions, and I think would make the date information more legible.

geniza/corpus/templates/corpus/document_detail.html
```diff
@@ -70,11 +70,27 @@ <h2 class="sr-only">
                 {% endif %}
             </dl>
             {# secondary metadata #}
-            {% if document.languages.exists or document.secondary_languages.exists %}
+            {% if document.languages.exists or document.secondary_languages.exists or document.doc_date_original or document.doc_date_standard %}
                 <dl class="metadata secondary">
-                    {# NOTE: placeholder to populate translation string #}
-                    {# Translators: label for date of this document, if known #}
-                    <!-- {% translate "Document Date" %} -->
+                    {% if document.doc_date_original or document.doc_date_standard %}
```
rlskoeser, line 75: revise to check the `document_date` method I'm suggesting you write (will be more extensible if/when we add other date options)

geniza/corpus/tests/test_corpus_models.py
```diff
@@ -437,6 +437,14 @@ def test_str(self):
         unsaved_doc = Document()
         assert str(unsaved_doc) == "?? (PGPID ??)"
 
+    def test_historical_date_display(self):
+        """Should display the historical document date with its calendar name"""
+        doc = Document.objects.create(doc_date_original="507", doc_date_calendar="h")
```
rlskoeser, line 442: please use the code defined on the class instead of a bare letter

```suggestion
        doc = Document.objects.create(doc_date_original="507", doc_date_calendar=Document.CALENDAR_HIJRI)
```

rlskoeser (changes requested):

geniza/corpus/templates/corpus/document_detail.html
blms: I was thinking about two things here: semantics and styling. If that's fine semantically, then that solves the former! Styling-wise, it's fine on desktop, but on mobile I was thinking we'd end up with this extra empty space.

Screen Shot 2022-04-22 at 1 11 01 PM

I think I came up with a solution to this though, which is just to use the `:first-child` CSS selector for that margin-top, instead of applying it to the entire `dl` 😄

rlskoeser: It seems ok to me semantically! Nothing will be announced to a screen-reader, right?

Could you use `:not(:empty)` ? (first child seems ok too, though)

rlskoeser (approved):

PR #797, opened 2022-04-21
transcription prototype

rlskoeser:
draft functionality for using annotations to create and edit transcriptions

currently supports rectangle selection only; create, read, update, delete all functional; can update text and also update image selection zones

included in this pr:
- adds & loads annotorious seadragon 
- custom annotorious editor plugin
- custom annotorious storage plugin for simple annotation server
- new transcription js endpoint in webpack 

questions/notes:
- not sure where the seams should between stimulus/turbo and this code; trying to write annotorious plugins in plain js
- no unit tests; should we add them? for which parts?
- what's the right way to restrict access to this functionality only to the users who are allowed to use it? (maybe this does need to be handled at the annotation storage server layer in some way...)


rlskoeser:
to run locally:
- download and run a local copy of simple annotation server as documented here https://github.com/glenrobson/SimpleAnnotationServer#getting-started
- configure annotation endpoint in your local settings as `ANNOTATION_SERVER_URL = "http://0.0.0.0:8888/annotation"`
- if you want to experiment with the sample I converted from escriptorium, use this url https://pgp-transcription-prototype.glitch.me/pgpid2806_altoAnnotations.json and load it as an annotation list in your local annotation server at http://0.0.0.0:8888/populate.xhtml — then look at pgpid 2806 in your local dev instance

blms (commented):
JS overall looks good, I don't see any glaring issues (and it's working well for me locally!). I might modularize a bit more (for example, the save and edit buttons)—let me know if you'd like me to do this. I imagine some of `editor.js` will be changing quite a bit when we integrate TinyMCE.

> not sure where the seams should between stimulus/turbo and this code; trying to write annotorious plugins in plain js

I wonder if it might make sense to load this inside a stimulus controller as a module, like we currently do with OSD itself. I see you flagged some places where things might be initialized multiple times, so perhaps using the stimulus controller "target connected" functionality could help.

Could even split all of this transcription code off into a separate repo and load from GitHub (or npm) as a package, though that may be more work than you want.

> no unit tests; should we add them? for which parts?

Probably should add unit tests for all functions except the unchanged vendor code. I'd be happy to write jest tests for this if you'd like me to—could be a good way for me to understand the code better.

> what's the right way to restrict access to this functionality only to the users who are allowed to use it? (maybe this does need to be handled at the annotation storage server layer in some way...)

Yes, I think there should be authentication at both the editor and storage ends. And I think simply authenticating against Django admin permissions is fine for the editor. Not sure what the best specific authentication strategy to protect the storage is, it's definitely not something I'm very strong in!

blms:
Also meant to mention: we could also have a JS linter making changes on a commit hook like we do with Python now.

rlskoeser:
@blms thank you so much for the comments.

I agree that the code does need to be more modular; I'm also not sure how much it will change once we integrate TinyMCE, although I imagine we'll still need things like save/edit/cancel buttons.

> Could even split all of this transcription code off into a separate repo and load from GitHub (or npm) as a package, though that may be more work than you want.

I've wondered about that — I think the storage plugin in particular is likely to be of interest to others and could be worth making available separately. I don't have a sense of how much extra effort this is — but it seems like it could be helpful in some ways in terms of testing and clean separation between different parts of the code. Perhaps we should find time to talk through what's involved and how to approach this? 

Already from your comments about npm packages and the stimulus integration I'm seeing it a bit more clearly: if the annotorious plugins are separate npm packages, then we just write stimulus code in the geniza app that initializes them the way we would anything else, and like we do for OSD. That approach certainly sounds appealing!

> ... I'd be happy to write jest tests for this if you'd like me to—could be a good way for me to understand the code better.

This sounds really appealing. Testing js code is definitely not my strong suit, and it would be helpful for you to understand this stuff better. And hopefully I would learn and improve from reviewing the tests you write. 

Agreed about permissions — I'll be talking to PUL about auth for the annotation server later this week. We can trust them to keep it secure! We just need a way to actually access it securely and make sure people without princeton accounts can use it.  (specifics TBD)

> Also meant to mention: we could also have a JS linter making changes on a commit hook like we do with Python now.

We have a pre-commit hook for `prettier` installed! But maybe it isn't configured correctly... Now that I think about it, I don't remember it doing anything with my js (which is I'm sure not consistent). We had to tell prettier to ignore `.html` files because it can't handle django templates. Guess I'll need to check on that.

blms:
> Perhaps we should find time to talk through what's involved and how to approach this?

Sure thing! I haven't published a package on NPM before but I know Performant has. Would love to help and learn this myself.

> Already from your comments about npm packages and the stimulus integration I'm seeing it a bit more clearly: if the annotorious plugins are separate npm packages, then we just write stimulus code in the geniza app that initializes them the way we would anything else, and like we do for OSD. That approach certainly sounds appealing!

My thoughts exactly!

> We have a pre-commit hook for `prettier` installed! But maybe it isn't configured correctly... Now that I think about it, I don't remember it doing anything with my js (which is I'm sure not consistent). We had to tell prettier to ignore `.html` files because it can't handle django templates. Guess I'll need to check on that.

I'm also thinking about some more aggressive linting with stricter rules, something like ESLint—I find it useful to help with not only code style but also JS best practices.

rlskoeser:
closing this since we're not planning to merge until transcription is at mvp; we'll use this branch as the base to merge and review transcription functionality until then

PR #798, opened 2022-04-22
Add optional shelfmark override

rlskoeser:
- new optional shelfmark override field, to allow controlling display for documents on a range of shelfmarks
- revise solr indexing and search configuration to allow searching on both display field and actual shelfmarks
- simplify shelfmark display in templates — always use the display field; new template tag to wrap individual shelfmarks in spans to preserve existing no-wrap behavior


blms (approved):
Great solution, much more elegant both on the document model and with the template tag. And thanks for fixing those instantiations of the `content` field in tests, I was wondering why those were failing locally (and only on running individual test files but not the entire suite!).

Issue #805, opened 2022-04-25
As a front-end user, I want a way to access the museum or library view of the fragment (when available), so I can see more context about the source.


rlskoeser:
## testing notes (QA)
- [x] Confirm that each fragment's shelfmark links to its view URL, if one exists for that fragment (e.g. pgpid 5472, the second shelfmark in pgpid 12409)

## dev notes
Suggest we add this to the image rights/attribution section — if a fragment has a `view_url` set, then use that to make the shelfmark into a link. Should probably set a `target="_blank"` on the link.

- [x] link shelfmark to fragment `url` if one exists
- [x] set `target="_blank"` on link
- [x] style link with underline

richmanrachel:
CUDL links work and look great! Is this what's expected for ENA right now while we determine whether it should be PUL or JTS?
<img width="321" alt="image" src="https://user-images.githubusercontent.com/68028846/165369916-83843787-8166-44fb-a87a-3b0142a89674.png">
Fyi there is no link here to the image currently.

rlskoeser:
@richmanrachel yes, I think that's what we expect for JTS content right now — Esmé added a provenance statement about them being owned by JTS, and we have a task to display that; we don't have view urls for JTS content in the db yet (should resolve to the DPUL site, not to figgy).

rlskoeser:
@richmanrachel based on this information, does this issue pass testing? If not, please specify what changes you'd like to see.

richmanrachel:
@rlskoeser - yes, thank you!

PR #809, opened 2022-04-25
Add target and styling to fragment links (#805)

blms:
## What this PR does

- Per #805:
  - Adds `target="_blank"` and link styling to fragment links

## dev notes

I already added the link and conditional in #803, but feel free to comment on that here!


rlskoeser (commented):
Document permissions doesn't seem like the right name for the template snippet, that makes me think of django permissions. `document_image_rights` maybe ?

The template seems like it has a lot of conditionals; one way we've handled the `a` or `span` tag situation before is to use the `yesno` template filter: https://github.com/Princeton-CDH/mep-django/blob/7d76f7ccee2cef779c99f5e564f62da1e9574d6f/mep/people/templates/people/snippets/manifest_license.html#L7-L10

I also wonder if the fallback text needs to be an `else`, could we just display the other contents if we have them and then use an `if not`? (seems like there are redundant checks at the moment)

rlskoeser (approved):

PR #812, opened 2022-04-27
Separate date and calendar with comma when needed (#779)

blms:
## What this PR does

- Per #779:
  - Adds a comma between original date and calendar name

## Questions

- Does this make sense? I wanted to make the function as simple as possible while allowing either original date or calendar name to be blank, but it still reads a bit weird.

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -537,9 +537,12 @@ class Document(ModelIndexable):
         """Generate formatted display for the document's original/historical date"""
         # separate with comma if both date and calendar are present, else just return whichever is present
         # TODO: remove conditional once validation is implemented, since one will never be present alone
-        return ", ".join([v for v in [self.doc_date_original, self.get_doc_date_calendar_display()] if v])
-
-            [self.doc_date_original, self.get_doc_date_calendar_display()]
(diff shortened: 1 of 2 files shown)


geniza/corpus/models.py
```diff
@@ -535,7 +535,14 @@ def shelfmark(self):
...
         """Generate formatted display for the document's original/historical date"""
-        return " ".join(
+        # separate with comma if both date and calendar are present, else just return whichever is present
+        # TODO: remove conditional once validation is implemented, since one will never be present alone
+        separator = (
+            ", "
+            if self.doc_date_original and self.get_doc_date_calendar_display()
+            else ""
+        )
+        return separator.join(
```
rlskoeser, line 545: my preferred pattern for this : 

```suggestion
        # separate with comma if both date and calendar are present, else just return whichever is present
        # TODO: remove conditional once validation is implemented, since one will never be present alone
        return ", ".join([v for v in [self.doc_date_original, self.get_doc_date_calendar_display()] if v])

```

blms: Much cleaner, thanks! I knew you'd have an idea 😄

Issue #807, opened 2022-04-25
As a front-end user, I want to sort documents by shelfmark so that I can view records organized based on owning institution and/or collection.


rlskoeser:
## testing notes (QA)

- [x] In the public site search (QA), choose sort by shelfmark and confirm the order is correct or comment feedback if not
- [x] Try this in combination with a few filters/queries as well

---

## dev notes 
- [x] add sort option to form
- [x] configure sort option to use solr field `shelfmark_s` (display version of shelfmark as string)
- [x] add a unit test to confirm behavior

Note: it would probably be preferable to use natural sorting for shelfmark, but I don't think we should invest a lot of time in that right now. I did find some examples of an alphanumeric sort solution for Solr; here's a gist we could probably adapt (we don't need all of this).  https://gist.github.com/peaeater/0b3473e7457dbd4b527e 
But maybe that could be done as a second pass...

rlskoeser:
@richmanrachel I'd noticed the documents with no shelfmark too — I agree with you, I think that is a data problem.

Thanks for flagging the relevance sort — that sounds like a bug to me. (Probably leftover from when we didn't have so many sort options!) Would you mind creating an issue?

Can you say more about what isn't working correctly when you search for t-s ?

richmanrachel:
@rlskoeser - when I sort by shelfmark and type in T-S, I expect to have a sorted list that starts with the lowest T-S shelfmark (T-S 8...) and the rest in order from there.

But my results are currently:
1. Unknown Type:
2. State Doc: AIU....
3. State Doc: AIU...+T-S 8J16.23
4. (and on an on with a lot of AIU docs taking precedence)

I understand that AIU is earlier in the alphabet than T-S, but I think that the functionality Alan actually wants is really to filter by collection so he can see the whole T-S set in order and look for numbers that are missing.

rlskoeser:
ah, I see — this is functioning correctly to me, because we have to sort all results returned in a search by the shelfmark we have for those documents.

for Alan's use case, how about searching on `shelfmark:"t-s"` and sorting by shelfmark?

richmanrachel:
@rlskoeser - your suggestion is actually a wonderful way to find joins! All of the T-S joins with AIU, Bodl, and BL fragments show up first. It's not until entry 45 that you start to see regular T-S shelfmarks. So it works, but with an asterisk?

rlskoeser:
[...]

PR #813, opened 2022-04-27
Implement sort by shelfmark (#807)

blms:
## What this PR does

- Per #807:
  - Implements sort by shelfmark (not natural sort) ~~ascending and descending~~

## Questions

- Do we need descending or just ascending?
- I guess right now we're using the dynamic field type `*_s` for shelfmark. For natural sort, am I correct that we'd need to define another `fieldType` like the [one from the gist](https://gist.github.com/peaeater/0b3473e7457dbd4b527e) and use that instead? And would we want anything other than "Left-pad numbers with zeroes" and "Left-trim zeroes to produce 6 digit numbers"? I think I can accomplish that, though not terribly confident in my solr config skills!

### DIFF
--- a/geniza/corpus/forms.py
+++ b/geniza/corpus/forms.py
@@ -159,10 +159,8 @@ class DocumentSearchForm(forms.Form):
         ("scholarship_desc", _("Scholarship Records (Most–Least)")),
         # Translators: label for ascending sort by number of scholarship records
         ("scholarship_asc", _("Scholarship Records (Least–Most)")),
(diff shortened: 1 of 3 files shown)


rlskoeser (approved):
I don't think we need both ascending and descending sort on shelfmark; ascending is enough.

I'm going to try to add natural sort on location in the database for an issue related to sorting footnotes — if that works we should probably apply to shelfmarks as well. I think we should deliver this version for testing and then circle back to refine when we're ready.

Can be merged after you remove descending shelfmark sort

PR #815, opened 2022-04-27
Shelfmark override revisions (#154)

blms:
## What this PR does

- Per https://github.com/Princeton-CDH/geniza/issues/154#issuecomment-1110228879:
  - Only display shelfmark override as a title and keep as a searchable field, but display the full list of shelfmarks on the document detail view
  - Display shelfmark override when available in list view and form title for documents in the admin interface

## Questions

- Is it redundant to index both `shelfmark` and `fragment_shelfmark` now? I know we at least need to index the individual shelfmarks for searching, and to index the override for searching + displaying in the search results page, but not sure if we need to index the joined string that will only appear on the detail view.

### DIFF
--- a/geniza/corpus/admin.py
+++ b/geniza/corpus/admin.py
@@ -185,7 +185,7 @@ class DocumentAdmin(TabbedTranslationAdmin, admin.ModelAdmin):
     list_display = (
         "id",
         "needs_review",  # disabled by default with css
-        "shelfmark",
+        "view_shelfmark",
         "description",
         "doctype",
         "all_tags",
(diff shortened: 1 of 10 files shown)


rlskoeser (commented):
the changes here don't match my understanding of what needs to change — I wasn't expecting to see any indexing changes

blms:
@rlskoeser Maybe my approach is wrong here—my thinking was that in order to get the shelfmark override to appear on the search results page, we need to index it. But was also thinking that we don't want to completely replace the shelfmark index for searching reasons.

Thinking it might make more sense to revert the changes to solr configs, and instead only change the `index_data` function so that `shelfmark_s` picks up the shelfmark override if it exists, and otherwise picks up the shelfmark. Then we just use that for display purposes, and use the fragments for searching. That seems much simpler now that I think about it. Apologies for not thinking that through enough before!

rlskoeser:
@blms I think the indexing should be what I had before — shelfmark label (override or join) as a complete string and then all individual shelfmarks for searching. Please hold off on making any further changes until I get a chance to look properly.

rlskoeser:
I adapted a couple of things from this approach in the PR I just opened; going to close this one without merging

PR #822, opened 2022-04-28
Revise logic for when shelfmark override is displayed

rlskoeser:
changes:
- implemented a `label` property that displays combined shelfmark or shelfmark override if set
- use the new label field for indexing shelfmark in solr for display
- label is displayed in search results, document title, document metadata, admin document change list
- actual shelfmark is displayed in metadata and shelfmark display field on document change form

question:
Is `label` confusing since we already have a `title` property? maybe `shelfmark_label` ? `shelfmark_display` ?

### DIFF
--- a/geniza/corpus/admin.py
+++ b/geniza/corpus/admin.py
@@ -185,7 +185,7 @@ class DocumentAdmin(TabbedTranslationAdmin, admin.ModelAdmin):
     list_display = (
         "id",
         "needs_review",  # disabled by default with css
-        "shelfmark",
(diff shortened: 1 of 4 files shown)


blms (approved):
Very clean and simple solution, especially with the use of the admin decorator inside of the Document model instead of in `admin.py`. I prefer `shelfmark_display` over `label`, but I think either is fine. Feel free to make that change or leave as is!

rlskoeser:
@blms `shelfmark_display` sounds good! `label` is too generic

Issue #802, opened 2022-04-25
 As a front-end user, I want to see logos for museums and libraries providing image content, so I have a better sense of where the content is coming from.


rlskoeser:
## testing notes (QA)

- [x] On the public site (QA), confirm that the document detail pages (for documents with images) include logos of image providers under "Image Permissions Statement" beneath the transcription.

## dev notes
Revise image attribution & rights section to display logos from manifests (as we're required to do per IIIF spec!)

Don't duplicate logos — simply display the unique set of logos. (I don't know how much it matters that the logos link to specific shelfmarks — I think it's ok if it's not one to one)

rlskoeser:
@mrustow @richmanrachel would you like us to display the JTS logo for JTS content served by PUL? The PUL logo is the one in the manifest, and based on my conversation with Esmé this isn't required per the MOU, but it isn't too much effort for us to do. Please let us know — and if so, please provide a link to the correct website or logo so we can make sure we get the correct one!

richmanrachel:
Never knew that was CUDL's logo, but it works! Closing!

PR #823, opened 2022-04-28
Implement basic solution for displaying IIIF logos (#802)

blms:
## What this PR does

- Per #802:
  - Displays unique logos in a list below the list of fragment attributions, with preliminary/basic styling
- Prevents an extra space from being added between the shelfmark and colon in the image permissions section

## Questions

- What do you make of the `alt` text? I think using the name of the organization would be preferable (and would be good for a `title` attribute as well), but I couldn't figure out a way to get that from the manifest.
- Using `ifchanged` is sufficient to guarantee uniqueness for all of my local data, but it somewhat depends on the order of fragments retrieved. Is there some way we should order the fragments first, or should the uniqueness not be handled in the template?

### DIFF
--- a/geniza/corpus/templates/corpus/snippets/document_image_rights.html
+++ b/geniza/corpus/templates/corpus/snippets/document_image_rights.html
@@ -8,9 +8,7 @@
             {% if fragment.iiif_url %}
                 <li class="fragment-permissions">
                     {% with fragment.url|yesno:"a,span" as tag %}
-                        <{{ tag }} class="shelfmark"{% if tag == "a" %} href="{{ fragment.url }}" target="_blank"{% endif %}>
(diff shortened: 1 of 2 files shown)


blms:
@rlskoeser  Yes, I wasn't sure about what size to use… unfortunately, it would have to be pretty large for the CUDL text to be legible! 64x64 px seems to be the smallest size where it's even barely legible 😆 Should we scale it up, or leave as-is for the time being?

rlskoeser:
@blms let's make the logos a little bigger so the CUDL one is at least semi-readable and then leave for Gissoo to deal with when she revises the designs for this.

rlskoeser (approved):
Regroup looks nice. Fine to merge once you tweak the images so they are slightly larger.

Issue #675, opened 2022-03-01
Percy sometimes fails to load some fonts


blms:
fonts failed: https://percy.io/2cf28a24/geniza/builds/16159573/changed/910342749?browser=safari
fonts fixed: https://percy.io/2cf28a24/geniza/builds/16159573/changed/910343117

## dev notes

- ~~Figure out why fonts are looking in `/static` instead of external URL for preloading (getting 404s as it is)~~
- [x] Try increasing timeout
- [x] is it possible to have an encrypted zip file that gets used by the build and use a github secret to decrypt? if that works, then we could have fonts local to the deploy

blms:
Seems fixed by cf4ba5c, closing unless/until it comes up again.

blms:
Reopening because this is still happening from time to time. For example: https://percy.io/2cf28a24/geniza/builds/16317605/changed/918573591

It's also odd because I'm still seeing 404s for `/static/fonts` in the [workflow runs](https://github.com/Princeton-CDH/geniza/runs/5453297249), so perhaps there is a failure with the CSS logic. Though, those 404s are consistent, while the actual behavior is inconsistent.

blms:
Confirmed working as of https://percy.io/2cf28a24/geniza/builds/17687691.

PR #826, opened 2022-04-29
Use encrypted zip to store production fonts (#675)

blms:
## What this PR does

- Per #675:
  - Uses GPG-encrypted zip (AES256) to store licensed fonts in the repo
  - Uses a passphrase stored in a GitHub Secret to decrypt the zip in the percy and lighthouse workflows
  - Unzips the fonts into the `sitemedia/fonts` folder so they should be processed by `collectstatic` and picked up by percy/lighthouse
  - Removes logic to point percy/CI at the production server for fonts


rlskoeser (approved):
All the changes look good to me — glad we finally came up with a simpler, more reliable solution.

Please add documentation to the devnotes about how this is set up and what needs to be done if/when we need to adjust the fonts included in this pacage in future.

Can be merged once it's documented!

blms:
Did a lot of unnecessary messing with RST to try and get syntax highlighting, but I guess that's not supported by GitHub! 😆  In any case, should be ready to merge now once checks complete.

Issue #855, opened 2022-05-11
As a content editor, I want Hijrī dates automatically converted to standard dates when possible, so that dates can be compared and used for filtering and sorting.


rlskoeser:
build on #854 to add support for basic date conversion support for Islamic calendar dates

richmanrachel:
@rlskoeser - year and year + month are working, but days are not working.

rlskoeser:
@richmanrachel please provide some examples of the ones that are not working and expected output

richmanrachel:
@rlskoeser - sorry, yes! For example, we were playing around with PGPID 20973 which already had a month and an uncertain year. We picked a year to play with and it added the correct range for month, year. But when we added days, it kept the same range for the full month:
<img width="491" alt="image" src="https://user-images.githubusercontent.com/68028846/168664319-2820fdca-5b5c-4add-9d81-e5b8f9cbeb36.png">

rlskoeser:
@richmanrachel did you clear out the standardized date field before saving? Because it's set to allow you to override the generated date, it won't update if a value is already set. We could tweak that logic some if helpful (e.g., if the original date has changed and the standard date has not) — but I'm sure there are scenarios where that might not be what you want.

We should probably add help text here! Should indicate supported calendars and how it behaves. Maybe something like this?
> Leave blank or clear out to automatically calculate standardized date for supported calendars.

rlskoeser:
Great, will do.

Issue #854, opened 2022-05-11
As a content editor, I want Anno Mundi dates automatically converted to standard dates when possible, so that dates can be compared and used for filtering and sorting.


rlskoeser:
## testing notes
Test editing a document in the admin and set an original date and Anno Mundi calendar, with no value in the standard date. Use the "save and continue editing" button to save the record. The standard date should be populated for you based on the original date you entered.  You should also see a read-only human readable version of the date next to the standard date field.

Test a variety of examples:
- [x] year only
- [x] month and year
- [x] month, year, and day
- [x] "first decade of ..." or similar — currently unsupported; you should see a notice
- [x] if you put content in the standardized date directly yourself, it should stay there (i.e., conversion logic shouldn't override manual entry, even if they differ)

## dev notes
this implementation should handle only the basic cases, as already prototyped in #714 

- [x] refactor `Document` date logic into a mixin to make the code more manageable
- [x] add method to calculate standard date on `Document` save if calendar is Anno Mundi
- [x] display human-readable version of the standard date for easier checking/comparison between dates
- [x] warn if date can't be parsed / calculated

richmanrachel:
@rlskoeser - it works! We're so excited

Issue #847, opened 2022-05-10
As a frontend user, I would like to see converted dates displayed in a standard, readable format so that I can easily understand the calendar information.


richmanrachel:
**Is your feature request related to a problem? Please describe.**
Improvement on current date handling.

**Describe the solution you'd like**
Date month, year. Ex. 1 January, 1232 CE.

**Describe alternatives you've considered**
MR wants this British style. RR wants to see more examples of specific date conversions from the corpus (for while MR said that having day come first makes the most sense, I'm not sure it does if that's the fuzziest part of the date -- to me "January" is more important than "1," and the year is the most important overall)

**Additional context**
There may be some additional decisions that need to be made about more complicated dates. For example:
- Thursday, 20 Iyar 4758 or 4768 AM (24 May, 998 or 1008 CE)	[Should we use 24 or 24th? Should Thursday actually be included in the converted date, and if so, should it be left untranslated for the Hebrew date "yom rivi'i"?]

mrustow:
I think we should follow CMS style, and CMS style for dates is:

1. For CE dates, either American usage (`November 11, 1968` — month first, and comma between day and year) or British usage (`11 November 1968` — day first, no comma) is fine. BUT
2. For other calendars, day first is standard, so `17 Barmūda 1465`, or `7 Jumādā I 648 AH`,  or `14 Adar II (4)768 AM`. The day must be first because some of the months have numerals following (Adar I/II, Jumāda I/II, Rabīʿ I/II).
3. When using both CE and other dates, for consistency, follow British usage.

rlskoeser:
Django by default will localize the modern dates. I've overridden the formatting for `en` locale to display as Marina requested when we have a full-precision date (year-month-day), and am using "month year CE" or "year CE" only otherwise.  We may need to adjust that and customize the formatting for other locales.  (Should "CE" be marked for translation?)  It may be good for you to see how it's working now when you switch languages on the test site, but we can review/refine non-english versions later as needed.

Right now I'm displaying full precision dates for date ranges generated from months or years in other calendars, which may not be what you want. I thought it would be easiest for you to see and respond to an initial implementation.

richmanrachel:
@rlskoeser - I like it a lot! I think @mrustow should take a look too.

richmanrachel:
We both agree it's great! Closing :)

PR #856, opened 2022-05-12
Feature/date conversion

rlskoeser:
in this PR:
- date fields and logic moved from `Document` to new `DocumentDateMixin` in `geniza.corpus.dates` 
- logic to convert/standardize Hebrew #854 and Islamic #855 dates (simple cases — year, year month, year month day)
- revised logic for displaying standard date to make it readable (but still localized) per #847 
- script to report on dates currently in the database and compare against the new calendar conversion logic (used for testing/development; expect to eventually add an update mode to correct/improve existing dates)
- regex validation for standard date field #857


sitemedia/scss/components/_shelfmark.scss
```diff
@@ -2,8 +2,9 @@
 // Styling rules for document shelfmarks.
 // -----------------------------------------------------------------------------
 
-.shelfmark span {
-    // Never break in the middle of a shelfmark
+.shelfmark span,
+time span {
```
rlskoeser, line 6: The component name isn't accurate anymore, but the styling is the exact same; suggestions?

blms (approved):
All looks good to me! I really like the organization of pulling all date-related functionality out into this module, and using a mixin there for the model-related pieces. Also, very nice regex.

Some of this work looks like it could be useful to other projects, as I think you mentioned in a meeting. Maybe just as contributions to convertdate, if nothing else.

rlskoeser:
> Some of this work looks like it could be useful to other projects, as I think you mentioned in a meeting. Maybe just as contributions to convertdate, if nothing else.

Yes, I'm interested in extracting the work that seems like it could be reusable at some point — thanks for confirming. I hadn't thought about contributing any of this to convertdate, as what I'm doing seems to be out of scope for what that library, but will think about whether it's worth opening an issue to discuss with the maintainer.

PR #869, opened 2022-05-17
filter and sort documents by date

rlskoeser:
in this pr:
- define date range field type in solr schema (requires updating configset)
- index standardized dates as date range in solr
- port range filter & widget from other CDH codebases
- display range filter on form (preliminary / unstyled)
- implement range filtering in the document search view

does not include styles or placeholder text; unit tests are incomplete

questions:
- The range input I ported over uses integer fields. Technically we could support YYYY-MM-DD, but Marina indicated year was sufficient and I'm not sure we need the additional complexity (number input is easier to validate). However, I'm not sure if we can add placeholder text to a numeric input, which might be reason enough to switch (maybe).

### DIFF
--- a/geniza/common/fields.py
+++ b/geniza/common/fields.py
@@ -1,3 +1,6 @@
+from django import forms
+from django.core.exceptions import ValidationError
+from django.core.validators import RegexValidator
 from django.db import models
 from natsort import natsort_keygen, ns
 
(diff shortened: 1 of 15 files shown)


rlskoeser:
Wow, glad to hear that we can set a placeholder in numeric inputs. Now I have to figure out a way to pass them into the widget args! 😁 

I didn't inherit the mixin because I wasn't sure about the value of min/max validation — what do you think? worth adding?

Issue #827, opened 2022-04-29
sort automatically switches relevance whenever the search term is revised


richmanrachel:
## testing notes (qa)

- [x] In the public site search on QA, start typing a search term and confirm it switches to "relevance". Then, change the sort option to something else (not "random" or "relevance"), add a new search term, and confirm the sort remains on the option you chose.

---

**Describe the bug**
When I add a new search term while trying to remain within Shelfmark A-Z sort, the site reverts the search to sort by Relevance. 

## dev notes
-  only change sort option when sort is default (i.e., user has not changed sorting)

blms:
I think this is actually one of the few times we can say "it's not a bug, it's a feature"! At least, it was planned that way. But perhaps we should revisit now that we have so many more sort options.

Here is the original issue collecting the discussions: https://github.com/Princeton-CDH/geniza/issues/550

richmanrachel:
@blms - ahh, I see! That makes sense to me. 

Right now the auto-switch is just extra confusing since all the filters stick around. But I think your solution is on the right track.

blms:
Right—maybe "filters applied" could be another condition that prevents the sort from auto-switching!

richmanrachel:
I'm not sure I fully understand how that will work, but I trust you!

rlskoeser:
@blms it works as we planned originally, but when we built it we didn't have so many sort options! I think maybe it should only change if you have the default sort selected. Also should probably only trigger if you go from no keyword text to some keyword text — not modified text, which I think it does now. I think it should not be aware of the filters, though.

richmanrachel:
Fixed! Thank you :)

PR #879, opened 2022-05-17
Only auto-sort by relevance from default sort

blms:
## In this PR

- Per #827:
  - Only auto-switch to relevance sort (by typing in the search query field) when the default sort element is checked


rlskoeser (approved):
great!

Issue #893, opened 2022-05-27
As a user I want to be able to toggle content panels so that I can view image or transcription separately or both at the same time, so I can read the content I am interested in.


rlskoeser:

gissoo:
@rlskoeser @blms [here are the revised toggles for desktop and mobile on Figma ](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=1426%3A40294)

[here you can see what it looks like when all toggles are deselected on desktop and mobile](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=1384%3A47910)

[I have uploaded the toggles on google drive here](https://drive.google.com/drive/u/0/folders/1uN0_0n-0a5utYVV-v2Z-hkE7WQVjfxp1), because I had a lot of questions on how to organize them on github – 

note: 
- the image toggles are in PNG and the transcription and translation toggles are in SVG 
- selected/active toggles on desktop are 65 x 65px, and deselected toggles on desktop are the same size as mobile toggles which is 40 x 40px. all toggles on mobile always maintain the same size.

Issue #892, opened 2022-05-27
As a user I want content panel toggles to be disabled when a record type for a document is not available, so that I know what content is available. 


rlskoeser:
## testing notes
- [x] if document has no image or transcription, image+transcription panel should not display at all (no toggle buttons)
- [x] if document has image but no transcription, transcription toggle should be disabled
- [x] if document has transcription but no image, image toggle should be disabled

## dev notes
- [x] set as disabled in template if there is no content available
- [x] implement disabled styles for toggle

gissoo:
@rlskoeser @blms you can see [disabled toggles variants here for desktop and mobile](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=1426%3A40294)

Issue #889, opened 2022-05-27
As a user I want to see all images and first available transcription for a document, so that I can see and read the content.


rlskoeser:
## testing notes
- [x] confirm all images available are displayed
- [x] confirm content from first transcription is displayed (note: transcription sync script has been updated to roughly chunk content into pages/images, but it is not completely accurate)

## dev notes
- [x] Show all images (no pagination for now)
- [x] Show image + transcription panels by default ; transcription should be roughly aligned with images; show first only, for now
- [x] implement spacing and layout per designs
- [x] adapt preliminary srcset image sizes as needed for the design
- [x] move / adapt image attribution styling from previous iiif viewer

gissoo:
@rlskoeser @blms [here are the designs for this issue ](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=1384%3A35645)

richmanrachel:
@rlskoeser - looks amazing! Closing!

PR #895, opened 2022-06-01
Style image/transcription/translation panel (#889, #892, #893)

blms:
## In this PR

- Per #889:
  - Spacing and layout of transcription and image panels, image attribution section
- Per #893:
  - Styles for the ITT panel toggle buttons
  - Transitions for the toggle buttons
  - Removed the transitions on panel container and panels
- Per #892:
  - Disabled styles for toggles

## Questions

How exactly should we handle the transitions on the panel?
- It seems we need a (vertical) transition for opening the panel container and reflowing the content, and a (horizontal) second transition for opening/closing an individual panel. Does that sound right?
- Because of the `auto` heights on the panels (which are necessary!), I don't believe it is possible to do this in pure CSS/HTML. I spent about 2 hours today trying to code this in Stimulus; kept thinking I was _so close_, but at a certain point I had to call it enough (about 110 lines of non working code…). Should I continue on this path, or should we think of a different solution?


geniza/corpus/templates/corpus/snippets/document_transcription.html
```diff
@@ -1,12 +1,18 @@
...
-    <input type="checkbox" id="images-on"  {% if document.has_image %}checked="true" {% else %} disabled="true"{% endif %}/><label for="images-on">show images</label>
+    {# translators: label for checkbox toggle to show the images panel for a document #}
+    <input type="checkbox" id="images-on" aria-label="{% translate 'show images' %}" {% if document.has_image %}checked="true" {% else %} disabled="true"{% endif %}/>
+    <label for="images-on"></label>
     {# transcription displayed by default; disable if no content is available #}
-    <input type="checkbox" id="transcription-on" {% if document.digital_editions %}checked="true" {% else %} disabled="true"{% endif %}><label for="transcription-on">show transcription</label>
+    {# translators: label for checkbox toggle to show the transcription panel for a document #}
+    <input type="checkbox" id="transcription-on" aria-label="{% translate 'show transcription' %}" {% if document.digital_editions %}checked="true" {% else %} disabled="true"{% endif %}/>
+    <label for="transcription-on"><svg><use xlink:href="{% static 'img/ui/all/all/transcription-toggle.svg' %}#transcription-toggle" /></svg></label>
```
rlskoeser, line 11: why this `svg` / `use` notation with id? is this so you can style with css?

rlskoeser (commented):
Updates look good. Thanks for adding aria labels and translation.

The toggle buttons look bigger than I expected, but I presume that's just because of whatever scale I looked at things on figma. 

The phosphor icons for mobile are not correct, is this because of the phosphor version issue you identified?

It looks like some of the images as uploaded have some artifacts (partial transparency on one or two sides), is that needed for the design or an export error? Would it be better if the borders were added in CSS instead of part of the image, and do we need a disabled image or can that be done via css? (are you even using the disabled image?)

Transcription toggle has a strange inset outline or border on desktop dark mode.

The disabled filter on the translation toggle isn't as evident in dark mode as it is in light mode (disabled image toggle is fine, though); can you decrease the opacity or brightness for translation & transcription toggles in dark mode?

geniza/corpus/templates/corpus/snippets/document_transcription.html
blms: Yes! It's for setting the "disabled" color on the SVGs. It could also be done with background images and separate files for "disabled", as I did with the PNGs.

blms:
> The toggle buttons look bigger than I expected, but I presume that's just because of whatever scale I looked at things on figma.

Should match Figma at 100% scale.

> The phosphor icons for mobile are not correct, is this because of the phosphor version issue you identified?

Yes, fixed in 2104fa6 (I'd started this branch before reverting it).

> It looks like some of the images as uploaded have some artifacts (partial transparency on one or two sides), is that needed for the design or an export error?
> ...
> Transcription toggle has a strange inset outline or border on desktop dark mode.

Good catch, I hadn't noticed these! It was an export error.

> Would it be better if the borders were added in CSS instead of part of the image, and do we need a disabled image or can that be done via css? (are you even using the disabled image?)

Borders are implemented in CSS for the SVG toggles. I've also implemented the disabled state in SVGs with CSS—but with images for the PNGs (image panel toggle), since it's easier that way to produce a "white overlay with 70% opacity".

For the PNG toggle images, this isn't possible unless I change the approach; right now they are background images in CSS so that they can be swapped out between mobile and desktop, and it isn't possible to add outline or border to background images. Maybe there's another way to swap them out for screen size though, like with a pseudo-element… would this be preferable?

> The disabled filter on the translation toggle isn't as evident in dark mode as it is in light mode (disabled image toggle is fine, though); can you decrease the opacity or brightness for translation & transcription toggles in dark mode?

Done in 4afa2e6, though I'm hesitant without Gissoo! I don't believe there are any dark mode designs at all right now, and the disabled toggles are defined as "white overlay with 70% opacity on top of all toggles," which has a slightly different effect than reducing opacity (e.g. the white background remains white).

rlskoeser (approved):
These updates all look good to me.

The way you've implemented the icons sounds good, just want to make sure I understand the approach.  They look better now that you've cleaned up the images.

Gissoo didn't do designs for a dark mode — I told her we'd do our best to infer from the light mode designs, and she and I agreed the icons would work for both. Your disabled styles for dark mode look much better to me, and we can tweak if needed after Gissoo reviews.

PR #897, opened 2022-06-02
Add labels to docdate filter inputs

rlskoeser:
Rachel wanted labels on the year inputs for more clarity about what this is. I couldn't figure out a way to pass them through as parameters, so just extended the widget so we could have a custom render template.

Please do a quick review and see if you have any concerns, and then adjust the styling so that the labels appear on the same line as the inputs.

### DIFF
--- a/geniza/corpus/forms.py
+++ b/geniza/corpus/forms.py
@@ -135,6 +135,13 @@ class BooleanFacetField(FacetFieldMixin, forms.BooleanField):
         self.widget.facet_counts = facet_dict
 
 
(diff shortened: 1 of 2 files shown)


blms (approved):
Code looks good! Will work on styling now.

blms:
All done in 02852a9—I gave it a left margin (on desktop) to match the other filter labels' left margin, I think this helps distinguish it in the hierarchy from the "Document Dates (CE)" label.

rlskoeser:
@blms the style looks great, thank you

PR #898, opened 2022-06-02
Add update & clean modes to convert dates script

rlskoeser:
in this PR:
- revise `convert_dates` manage command to add modes, make current logic the `report` mode
- added an `update` mode, which converts dates with supported calendars & formats into the new format
- added a `clean` mode, which identifies standardized dates that don't follow the new validation pattern and cleans up the ones that can be automatically fixed

still needs tests & documentation

### DIFF
--- a/geniza/corpus/dates.py
+++ b/geniza/corpus/dates.py
@@ -236,13 +236,13 @@ class DocumentDateMixin(TrackChangesModel):
         # if original date is set and conversion is supported for this calendar,
         # generate the standardized date
         if self.doc_date_original and self.doc_date_calendar in Calendar.can_convert:
-            converted_date = display_date_range(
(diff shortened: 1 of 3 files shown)


rlskoeser:
@blms I still need to write the tests for this, but would be great to get a quick review on the logic

blms (approved):
Logic looks good to me! This would only need to be run once, then the rest is handled by validation/`standardize_date` on save, right?

rlskoeser:
> Logic looks good to me! This would only need to be run once, then the rest is handled by validation/`standardize_date` on save, right?

Yes, that's right — it's likely that we would run the update/clean as a one-time thing, since all dates going forward will have validation and conversion applied. I imagine we'd want to re-run the `update` mode again as we add support for converting dates in other calendars.

PR #910, opened 2022-06-07
Related documents page

rlskoeser:
in this pr:

- initial view & template for new related documents page (replaces placeholder "external links")
- solr query to find other documents that share any of the same shelfmarks
- update document tab template to display related docs link with count
- adjust document result template to work with or without pagination object
- simplify page description for scholarship record & related docs (document title is redundant)
- update existing tests for external links tab behavior

no tests yet for new methods/views/templates; no updates to styles

### DIFF
--- a/geniza/corpus/views.py
+++ b/geniza/corpus/views.py
@@ -332,12 +332,11 @@ class DocumentScholarshipView(DocumentDetailView):
         count = doc.footnotes.count()
         # Translators: description of document scholarship page, for search engines
         return ngettext(
-            "%(count)d scholarship record for %(doc)s",
-            "%(count)d scholarship records for %(doc)s",
+            "%(count)d scholarship record",
+            "%(count)d scholarship records",
             count,
         ) % {
             "count": count,
-            "doc": doc.title,
         }
 
     def get_queryset(self, *args, **kwargs):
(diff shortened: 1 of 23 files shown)


rlskoeser:
@blms please review and then pick up to work on styles. If any tests are easy for you to add, please add them — but anything you think will be substantially easier for me to test, please make list for me to circle back to.

geniza/corpus/views.py
```diff
@@ -364,6 +363,41 @@ def get_context_data(self, **kwargs):
...
+        }
+
+    def get_context_data(self, **kwargs):
+        doc = self.get_object()
+        # if there are no related documents, don't serve out this page
+        if not doc.related_documents.count():
+            raise Http404
+
+        context_data = super().get_context_data(**kwargs)
+        context_data.update({"highlighting": {}})  # no solr highlighting
```
blms, line 397: Quick question, would this `context_data` ever be expected to have highlighting? Good as a failsafe anyway since the template will respond to highlighting, but just wanted to make sure I'm not missing somewhere highlighting would be added.

rlskoeser: No, there won't ever be highlighting because there's no keyword search available on this page.

I couldn't figure out a way to make the template not error without setting an empty highlighting here in the context. If you see an easy way to do this in the template, that would be nice.

blms: Would it be ok to do one of these two things in `dict_item`? And if so, which is preferable?

```py
    return dictionary.get(key, None) if isinstance(dictionary, dict) else None
```

```py
    try:
        return dictionary.get(key, None)
    except AttributeError:
        return None
```

rlskoeser: you're right, I think template tags & filters are supposed to fail silently

second option is preferable

Issue #906, opened 2022-06-07
server error for documents associated with Heidelberg IIIF (PGPIDs 34016, 34017, 34018 so far)


alanelbaum:
## testing notes

https://test-geniza.cdh.princeton.edu/en/documents/34016/
https://test-geniza.cdh.princeton.edu/en/documents/34017/
https://test-geniza.cdh.princeton.edu/en/documents/34018/

no need to test everything on all three, should be sufficient to confirm on one or two

- [x] public page displays properly without 500 error
- [x] admin edit page displays properly without 500 error
- [x] image thumbnails display properly in the admin "related fragments section"
- [x] image permissions in the image viewer displays "Heidelberg University Library" and a creative commons / public domain 

--- 

There's a server error when trying to access the document pages for these documents in both the admin site and the public site.

> server error for PGPIDs 34016, 34017, 34018 (but no other documents so far)

Thank you for investigating!

richmanrachel:
@rlskoeser - bug fixed! @alanelbaum - wow, these are such pretty documents!

PR #917, opened 2022-06-09
Improve iiif handling 

rlskoeser:
fixes related to a problems with heidelberg iiif manifests as documented in #906 

- fixes fallback method for getting image thumbnails from manifest dynamically for admin (was using image id instead of the service endpoint; they do not have to be the same)
- handle `attribution` value that is not plain text, with some language logic
- display creative commons license if present (depends on djiffy changes in https://github.com/Princeton-CDH/djiffy/pull/12 )
- remove license images; use copies in djiffy static files instead

expect that this will need some additional work for styling CC license images


blms (commented):
Code changes look good. I agree it makes sense to move the language handling to one of the other libraries. When you say "spec is more complicated than this", do you mean it won't always match Django language codes?

I also updated the `<use>` elements with the new ID from the djiffy PR.

For styling and testing locally, could I get access to a new DB dump with the new manifests?

Also, if you are able to update the djiffy version here I think it might fix the lighthouse issue—seems it's complaining about `href=""` on the license URI link.

Issue #801, opened 2022-04-25
As a front-end user, I want to see provenance information for images when available so that I know where images and content is coming from for various shelfmarks.


rlskoeser:
## testing notes
For JTS images that are managed/provided by PUL, the image permissions statement should now include language that makes it clear that the documents are provided by JTS, you should see the Public Domain Creative Commons license image, and you should see both the PUL and JTS logos.

I've updated _most_ (some?) of the figgy iiif manifests on the test site to pull in the revised language and license.

Here are some examples you can test with, but feel free to poke around others in the search:
- https://test-geniza.cdh.princeton.edu/en/documents/30780/
- https://test-geniza.cdh.princeton.edu/en/documents/11276/
- https://test-geniza.cdh.princeton.edu/en/documents/11199/

If you see a JTS record with "no known copyright" that means the manifest hasn't been updated yet; please ignore for the purposes of testing this change.

## dev notes
PUL manifests for JTS content have been updated with provenance metadata and a corrected license. We'll need to:
- [x] do an import + update of JTS manifests from Figgy to get the new versions (RSK may need to help with spreadsheet/command option; we should document in deploy notes)
- [x] add logic to image attribution/rights to display provenance metadata when available (maybe as a fallback to attribution? I don't think the CUDL provenance statement is particularly relevant for us to display)
- [x] add logic to handle creative commons licenses similarly to existing logic for rightsstatement.org 
- [x] add logic to display JTS logo for JTS content served by PUL
- [x] update styles as needed to handle provenance text + license image

richmanrachel:
@rlskoeser - works! Closing :)

PR #924, opened 2022-06-13
Image provenance and JTS logo (#801)

blms:
## In this PR

- Per #801:
  - Add fragment model property to get provenance metadata when available
  - Display fragment provenance—only as fallback if attribution is not available
  - Add JTS logo, and logic to display it in template when needed
  - Add note to deploy notes about running import script
- Allow the h1 to h3 template filter to fail silently on bad argument (e.g. "chunked" transcriptions)

## Questions

- I noticed in the manifests that “Provenance” seems to always be spelled with a capital P, but I’m wondering if that would always be the case? Wondering if I should handle both cases, or if that’s something that should be processed in the djiffy importer, or if it will just always be capitalized?
- I’m assuming the metadata values are always lists—that’s always true per the IIIF spec, right? Should I handle cases where this is not followed?


blms:
Converting to draft while I fix those tests 😬

geniza/corpus/templates/corpus/snippets/document_image_rights.html
```diff
@@ -10,23 +10,35 @@
...
     <ul class="logos">
         {% for logo in logos_list %}
             <li>
                 <img src="{{ logo.grouper }}" alt="" role="presentation" height="64" loading="lazy" />
             </li>
         {% endfor %}
+        {% for provenance in provenance_list %}
+            {% if "JTS" in provenance.grouper %}
+                <li>
+                    <img src="{% static 'img/logos/all/all/jts-logo.png' %}" alt="" role="presentation" height="64" loading="lazy" />
```
rlskoeser, line 38: Since we're embedding this directly, let's actually provide alt text and link to the JTS website. (Similar to what we would expect to get in `provider` information in v3 of iiif presentation api)

https://www.jtsa.edu/

geniza/corpus/templatetags/corpus_extras.py
```diff
@@ -113,7 +113,10 @@ def format_attribution(attribution):
 def h1_to_h3(html):
     """Convert h1 headers to h3 to match other transcription formats,
     used to avoid modeltranslation inserting elements into h1 headers"""
-    return html.replace("h1", "h3")
+    try:
+        return html.replace("h1", "h3")
+    except AttributeError:
+        return html
```
rlskoeser, line 119: this is to handle the difference between new tei sync html format and old? seems reasonable enough

geniza/corpus/templates/corpus/snippets/document_image_rights.html
blms: Since we need to switch colors for light and dark mode with the new logo, I used `svg`/`use` instead. I still provided the link and accessibility labels, hopefully that is sufficient.

rlskoeser (approved):

geniza/corpus/templates/corpus/snippets/document_image_rights.html
blms: Didn't think about that! Good idea for future cases like this.

PR #927, opened 2022-06-14
Feature/tei to annotation

rlskoeser:
sharing preliminary version of tei to iiif annotation as a reference for discussing transcription editor data modeling


blms:
@rlskoeser Pushed my updates!

Questions:

1. What's going on with the `IndexError`s from `self.get_edition_footnote`? (logged to stderr)
2. The selector `percent:1,1,98,98` doesn't seem to be working right with Annotorious, should we go back to `xywh`?
3. Do we need to handle "no manifest for..."?
4. Do we need to do anything with Footnote `location` when splitting them? Either way, we also probably need to update the logic in the Scholarship Records page for the split editions/digital editions.

geniza/annotations/admin.py
```diff
@@ -21,7 +22,10 @@ class AnnotationAdmin(admin.ModelAdmin):
 
     def display_content(self, obj):
         """format json content for display in admin"""
-        return mark_safe("<pre>%s</pre>" % pprint.pformat(obj.content, indent=2))
+        # use mark_safe to render <pre>, use escape() to prevent rendering content HTML
+        return mark_safe(
+            "<pre>%s</pre>" % escape(pprint.pformat(obj.content, indent=1))
+        )
```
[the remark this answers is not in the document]
blms: @rlskoeser Not strictly needed... In the pretty-printed dict of the content field, I wanted to see that the transcription lines HTML was being generated correctly. Without `escape()` it was being rendered out because of `mark_safe`—which also looked a bit strange, to have rendered HTML inside of `<pre>`. I think there's not really another place to inspect that HTML besides using browser tools, right? So this might be a useful thing even if not really directly related to this work.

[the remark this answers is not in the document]
rlskoeser: yes, I see; I wonder why I didn't notice this before? could it be browser dependent?

does seem useful to have, though

geniza/corpus/management/commands/tei_to_annotation.py
```diff
@@ -0,0 +1,347 @@
...
+                )
+                continue
+            # found the document
+            if self.verbosity >= self.v_normal:
+                self.stdout.write(doc)
+
+            # get the footnote for this file & doc
+            footnote = self.get_edition_footnote(doc, tei, xmlfile)
+            # if no footnote, skip for now
+            # TODO: generate footnote when needed! (but should we need to?)
```
blms, line 103: Is there a real-world condition that would cause this to happen?

geniza/corpus/management/commands/tei_to_annotation.py
```diff
@@ -0,0 +1,347 @@
...
+            # NOTE: pgpid 1390 folio example; each chunk should be half of the canvas
+            # (probably should be handled manually)
+            # if len(html_chunks) > len(iiif_canvases):
+            #     self.stdout.write(
+            #         "%s has more html chunks than canvases; skipping" % xmlfile
+            #     )
+            #     continue
+
+            # start attaching to first canvas; increment based on chunk label
+            canvas_index = 0
```
blms, line 158: I was wondering about this actually (and also for the generated canvases elsewhere), should this be 1-indexed? Seems that's the convention for most of the manifests, canvases, etc we're working with (and it's how we'll be ordering annotations now)

geniza/corpus/management/commands/tei_to_annotation.py
```diff
@@ -0,0 +1,347 @@
...
+
+            # remove all existing annotations associated with this
+            # document and source so we can reimport as needed
+            existing_annos = annotation_client.search(
+                source=footnote.source.uri, manifest=doc.manifest_uri
+            )
+            # FIXME: this is probably problematic for transcriptions currently
+            # split across two TEi files...
+            # maybe filter by date created also, so we only remove
+            # annotations created before the current script run?
```
blms, line 168: That seems like a reasonable solution to me!

geniza/corpus/management/commands/tei_to_annotation.py
```diff
@@ -0,0 +1,347 @@
...
+                # link to digital edition footnote via source URI
+                anno["dc:source"] = footnote.source.uri
+
+                anno.target.source.id = annotation_target
+                anno.target.source.partOf.type = "Manifest"
+                anno.target.source.partOf.id = doc.manifest_uri
+
+                # apply to the full canvas using % notation
+                # (using nearly full canvas to make it easier to edit zones)
+                anno.target.selector.value = "xywh=percent:1,1,98,98"
```
blms, line 196: I was running into trouble using this selector on annotorious, were you able to test it?

geniza/corpus/management/commands/tei_to_annotation.py
```diff
@@ -0,0 +1,347 @@
...
+        # extend to return digital edition or edition
+        # (digital edition if from previous run of this script)
+        return doc.footnotes.filter(
+            models.Q(doc_relation__contains=Footnote.EDITION)
+            | models.Q(doc_relation__contains=Footnote.DIGITAL_EDITION)
+        )
+
+    def migrate_footnote(self, footnote, document):
+        # convert existing edition footnote to digital edition
+        # OR make a new one if the existing footnote has other information
```
blms, line 228: How should we handle reconciling this with the automatic processes on `post_save` in the footnote branch? Is there something different that needs to happen here?

blms (commented):
This is all looking great so far, just added some comments/questions.

geniza/corpus/management/commands/tei_to_annotation.py
rlskoeser: I'm running into it in my testing, but I don't think my data is up to date; was going to refresh and see if it still happens.

geniza/corpus/management/commands/tei_to_annotation.py
rlskoeser: I had to do some experimenting to get a version that annotorious rendered properly; this is the one that worked.

geniza/corpus/management/commands/tei_to_annotation.py
rlskoeser: Hmm, good point... I hadn't thought about the fact that those signals would be in place when this script runs (once we merge). My inclination is to disable those signal handles for the duration of this script. That seems simpler to me — then the signals are for annotation editing going forward. What do you think?

geniza/corpus/tei_transcriptions.py
```diff
@@ -147,6 +142,85 @@ def chunk_html(self, blocks):
...
+        html_lines = []
+        list_num = 1
+        in_list = False
+        for line_number, line in lines:
+            # convertline number to integer for comparison
+            if line_number:
+                try:
+                    line_number = int(line_number)
+                except ValueError:
+                    # in at least one instance, line number is a range "16-17"
```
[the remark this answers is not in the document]
rlskoeser: yeah, `n` attribute can be pretty much anything I think

geniza/corpus/management/commands/tei_to_annotation.py
```diff
@@ -0,0 +1,347 @@
...
+                    % (len(iiif_canvases), num_canvases, len(html_blocks))
+                )
+            # if we need more canvases than we have available,
+            # generate local canvas ids
+            if num_canvases > len(iiif_canvases):
+                # document manifest url is /documents/pgpid/iiif/manifest/
+                # create canvas uris parallel to that
+                canvas_base_uri = doc.manifest_uri.replace("manifest", "canvas")
+                for i in range(num_canvases - len(iiif_canvases)):
+                    canvas_uri = "%s%d/" % (canvas_base_uri, i + 1)
```
blms, line 146: Ah, great! I see the `i + 1` here now, didn't read this section carefully enough 😄

geniza/corpus/management/commands/tei_to_annotation.py
blms: No need I think, I just missed the `i + 1` when generating the canvases.

geniza/corpus/management/commands/tei_to_annotation.py
blms: Yes, I think disabling them makes sense. We'll probably need to modify it here slightly to account for the missing `content` field though!

rlskoeser: We can't remove `content` until after this release, since we need to use it to identify and migrate the correct edition footnotes. I'm planning to adjust your changes in the other branch to account for that, and we can plan to have a micro release immediately after the transcription migration release to clean up the content field once we're done with the migration.

geniza/annotations/models.py
```diff
@@ -42,6 +43,26 @@ def uri(self):
         """URI for this annotation (absolute url)"""
         return absolutize_url(self.get_absolute_url())
 
+    @cached_property
```
blms, line 46: Love these convenience methods

geniza/annotations/signals.py
```diff
@@ -1,20 +1,34 @@
...
+    try:
+        manifest_uri = instance.target_source_manifest_id
+        source_uri = instance.content["dc:source"]
+    except KeyError:
+        # annotations created from tahqiq should have these;
+        # if not present, bail out
+        # (likely only happening in unit tests)
+        return
+
+    # TODO: can we get ids without loading source/doc from db?
```
blms, line 31: Good point, at least for the document you can do `resolve(urlparse(uri).path).kwargs["pk"]`.

blms (approved):
Looks great, ready for merge!

geniza/annotations/signals.py
rlskoeser: yes — was thinking it might be worth separating out the method to get the object from the logic to get the id

Issue #899, opened 2022-06-03
As a user I want to switch between different transcriptions when there are multiple, so that I can read all available versions.


rlskoeser:
## testing notes
Test the image+transcription viewer display with some documents that have transcriptions.
- [x] documents with a single transcription should have a label above the transcription panel indicating the editor; it should not be interactive
- [x] for documents with more than one transcription, the label should be interactive and allow you to switch between transcriptions. PGPID 9121 is a good example. (FYI transcription sync appears to be not updating some multi-edition records; will need to address)

## dev notes
- [x] investigate options for placing multiple transcriptions in the html
- [x] Add a dropdown menu for switching between transcriptions
- [x] Displaying a short version of the editor information in the drop down to identify the source of the current translation
- [x] transition should fade out one / fade in other (no overlap)

rlskoeser:
ok @blms finally had time to investigate this a bit and think about approach.

I propose that we use markup similar to a css-only image carousel like this one: https://css-tricks.com/css-only-carousel/

It will use flex + overflow to control placement; each nested div would be full-width of the panel width. Instead of making it scrollable, we make overflow hidden and then use javascript to manage the scrolling (which I think should not be too difficult — do you agree?). This was the only solution I've come up with that automatically accounts for the full height of the longest transcription. The main drawback is I think we'll need to add another layer of nested divs to contain all the transcriptions, but that doesn't seem too terrible.  I'm open to other solutions if you have ideas. 

I was testing with PGPIDs 2179 and 4721, however I had some trouble getting all the footnotes to update with the new chunked content format. If this is a problem for you, LMK. I did start adding an option to the sync tei script to take an optional list of filenames, but that didn't fully resolve the problem. My quick-fix workaround was to make the content editable in admin and fix it myself, but if it's a problem for you too I'll probably have to deal with it before we can release the new transcription panel.

I don't think I did enough to warrant checking in and starting a branch, but in case it's useful here's the preliminary template code I added while I was experimenting with how to implement:

```jinja
<div class="transcription">
      {% if forloop.first %}
          {# do we always display even if only one, as a label? #}
          {% if document.digital_editions.count > 1 %}
              <ul>
                  {% for ed in document.digital_editions %}
                      <li data-edition="ed {{ ed.pk }}">Editor: {{ ed.source.all_authors }}</li>
                  {% endfor %}
              </ul>
          {% endif %}
      <h3>Transcription</h3>
      {% endif %}

      <div class="wrap">
[...]

PR #929, opened 2022-06-15
Multiple transcriptions selector in ITT panel (#899)

blms:
## In this PR

- Per #899:
  - Dropdown menu to switch between multiple transcriptions
  - Styles for dropdown menu ported mostly from search sort dropdown
- Moved the general transcription styling from `_results.scss` to `_transcription.scss`

## Questions

- How do you feel about the overall approach to the spacing and divs of the toggles + the dropdowns? It's working well for me, except for one minor issue, which is that the transitions seem somewhat differently timed between the two areas.
- I’ve been testing primarily with PGPIDs 5384 and 9121. I think PGPID 9121 shows one drawback to the "overflow hidden + horizontal scrolling" approach, which is that it doesn’t work well when one of the editions is drastically shorter than the other. It might be a data problem, though, since it seems like the addition is a bunch of text before the Recto side starts.

### DIFF
--- a/geniza/corpus/templates/corpus/document_detail.html
+++ b/geniza/corpus/templates/corpus/document_detail.html
@@ -32,6 +32,14 @@
     {% if document.has_image or document.iiif_urls %}
         <link rel="preconnect" href="https://cdnjs.cloudflare.com/ajax/libs/openseadragon/3.0.0/images/" crossorigin />
     {% endif %}
+    {# Disable ITT Panel when no JS available #}
+    <noscript>
+        <style>
+            #itt-panel {
+                display: none;
+            }
+        </style>
+    </noscript>
 {% endblock extrascript %}
 
 {% block main %}
--- a/geniza/corpus/templates/corpus/snippets/document_transcription.html
+++ b/geniza/corpus/templates/corpus/snippets/document_transcription.html
@@ -3,11 +3,11 @@
     {# NOTE: inputs must be kept as SIBLINGS of panel-container for CSS selection/controls #}
     {# images displayed by default; disable if no images are available #}
     {# translators: label for checkbox toggle to show the images panel for a document #}
-    <input type="checkbox" id="images-on" aria-label="{% translate 'show images' %}" {% if document.has_image %}checked="true" {% else %} disabled="true"{% endif %}/>
+    <input type="checkbox" id="images-on" aria-label="{% translate 'show images' %}" {% if document.has_image %}checked="true" {% else %} disabled="true"{% endif %} data-action="transcription#togglePanel"/>
     <label for="images-on"></label>
     {# transcription displayed by default; disable if no content is available #}
     {# translators: label for checkbox toggle to show the transcription panel for a document #}
-    <input type="checkbox" id="transcription-on" aria-label="{% translate 'show transcription' %}" {% if document.digital_editions %}checked="true" {% else %} disabled="true"{% endif %}/>
+    <input type="checkbox" id="transcription-on" aria-label="{% translate 'show transcription' %}" {% if document.digital_editions %}checked="true" {% else %} disabled="true"{% endif %} data-action="transcription#togglePanel"/>
     <label for="transcription-on"><svg><use xlink:href="{% static 'img/ui/all/all/transcription-toggle.svg' %}#transcription-toggle" /></svg></label>
     {# translation disabled for now since we don't handle them in the system yet #}
     {# translators: label for checkbox toggle to show the translation panel for a document #}
--- a/geniza/corpus/templates/corpus/snippets/document_transcription.html
+++ b/geniza/corpus/templates/corpus/snippets/document_transcription.html
@@ -13,6 +13,25 @@
     {# translators: label for checkbox toggle to show the translation panel for a document #}
     <input type="checkbox" id="translation-on" aria-label="{% translate 'show translation' %}" disabled="true"/>
     <label for="translation-on"><svg><use xlink:href="{% static 'img/ui/all/all/translation-toggle.svg' %}#translation-toggle" /></svg></label>
+    <div class="dropdown-container">
+        <div></div>
+        <details class="transcription-select" aria-expanded="false" data-transcription-target="dropdownDetails">
+            <summary class="transcription-select-trigger" data-action="keydown->transcription#shiftTabCloseDropdown">
+                <span data-transcription-target="editionShortLabel">Edition: {{ document.digital_editions.0.source.all_authors }}</span>
+            </summary>
+            <ul>
+                {% for ed in document.digital_editions.all %}
+                    <li>
+                        <label for="edition-{{ forloop.counter }}">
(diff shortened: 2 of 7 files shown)


rlskoeser:
I see what you mean about PGPID 9121... You're right, though, it is a data problem — the unnumbered prose at the top is actually a description, not transcription. 

In this case the images are so long that you wouldn't see the height shifting when you switch transcriptions, but I know that won't be the case every time. I'm still happy with this approach.

It probably wouldn't be too hard for us to identify records where the length of parallel transcriptions differs substantially so that they can be flagged for research assistants to review.... Or maybe just any document with more than one transcription should be reviewed, there aren't that many of them.

geniza/corpus/templates/corpus/snippets/document_transcription.html
```diff
@@ -13,6 +13,25 @@
     {# translators: label for checkbox toggle to show the translation panel for a document #}
     <input type="checkbox" id="translation-on" aria-label="{% translate 'show translation' %}" disabled="true"/>
     <label for="translation-on"><svg><use xlink:href="{% static 'img/ui/all/all/translation-toggle.svg' %}#translation-toggle" /></svg></label>
+    <div class="dropdown-container">
```
rlskoeser, line 16: I think you're making the layout and transitions harder than necessary. Also, we're going to need something similar to this for the translations panel eventually, when we add support for that, so the more we can use the existing panel layout & grid, the better.

I suggest you move this inside the `panel-container` grid element and have a first row of control panel/label elements: an empty image div, and this drop-down/label for transcriptions.

geniza/corpus/templates/corpus/snippets/document_transcription.html
```diff
@@ -13,6 +13,25 @@
...
+        <div></div>
```
rlskoeser, line 17: are the empty divs for matching the panel sizes?

geniza/corpus/templates/corpus/snippets/document_transcription.html
```diff
@@ -55,47 +84,3 @@ <h3>{{ image_info.shelfmark }} {{ image_info.label }}</h3>
     </div>
 
 </section>
-
-{% comment %}
```
rlskoeser, line 59: thanks for finally cleaning this up 🙂

geniza/corpus/templates/corpus/snippets/document_transcription.html
```diff
@@ -32,20 +51,30 @@ <h3>{{ image_info.shelfmark }} {{ image_info.label }}</h3>
...
+                    {% endif %}
+
+                    <div class="editions">
+                        {# display transcription in chunks by index #}
+                        {% for edition in document.digital_editions.all %}
+                            <div class="transcription" id="ed-{{ edition.pk }}" data-label="{{ edition.display|safe }}">
+                                {{ edition.content.html|index:forloop.parentloop.counter0|h1_to_h3|safe }}
+                            </div>
+                        {% endfor %}
+                    </div>
```
rlskoeser, line 67: I had been planning to leave this out for the mvp version, but it looks like it wasn't too difficult to add.

I do notice that the text is shifting vertically when switching between records that have longer/shorter citations. Can we find an easy solution to make the height match somehow? Should we always allow for two lines?

rlskoeser (changes requested):
Functionality seems great. The spacing/alignment with the panels and the toggles is much better now.

When there's only one transcription, can we disable the select drop-down so it is not interactive and just acts more like a label? It's confusing right now.

I see now that the "overflow hidden + horizontal scrolling" approach isn't great for mobile; is there an easy way to make the vertical overflow hidden on mobile? 

I also noticed that switching to the second transcription and then toggling the image panel off and back on results in the transcription panel content being misaligned (shows the right side of transcription 1 and left side of transcription 2).

geniza/corpus/templates/corpus/snippets/document_transcription.html
blms: Yes!

geniza/corpus/templates/corpus/snippets/document_transcription.html
blms: One question about this: if we put the dropdown inside the `panel-container` grid, it may not match the designs. In the designs, the image or transcription panel can sometimes differ from the width/position of its own toggle, whereas the dropdown is always aligned exactly with the toggle. Maybe this is a question for Gissoo, but is that the desired effect? Or should we always have the panels _and_ the dropdown match the toggle width? 

If I move the dropdown inside the grid, it's going to match the panel width, so it will definitely be misaligned from the toggles as it is now:

Screen Shot 2022-06-21 at 2 27 25 PM

Screen Shot 2022-06-21 at 2 27 10 PM


We could adjust the panels themselves to always match the toggles, but that will differ from the designs. You can see them unaligned here in Gissoo's design:

Screen Shot 2022-06-21 at 2 33 00 PM

I also fear that this could create a problem on smaller screens where the transcription will become too narrow:

Screen Shot 2022-06-21 at 2 34 33 PM


So I guess to summarize, my question is: what should be aligned with what here? And if the alignments and spacing in my current work on the branch are correct (they match the design currently), then should we prioritize that over the grid reuse issue with the code (or is there another solution)?

geniza/corpus/templates/corpus/snippets/document_transcription.html
blms: Used some CSS to always make this at least two lines' height, let me know if this solution doesn't work!

geniza/corpus/templates/corpus/snippets/document_transcription.html
blms: @rlskoeser They are not! Only the dropdowns are aligned with the toggles now. Take a look at Gissoo's design in the third image I posted just now, and you can see how the image panel extends past its toggle, and the transcription panel is narrower than its toggle and shifted slightly to the right. So in my implementation, I tried to match the design where the dropdown is always aligned to the toggle, but the panels may not be (and usually aren't).

With the "transcription panel being too narrow" comment, I meant to say that would be a possible side-effect if we aligned the panels with the toggles exactly. Though it would likely be alleviated if we ignore the unselected translation toggle on the side—we would just need to rethink some margins and proportions. Generally it would need us to pass the designs/margins back to Gissoo, so that she can design everything with that alignment (toggles and panels aligned together) in mind.

rlskoeser (commented):
Thanks for your work on this. I started working on some adjustments Friday while you were out, thought it would be easier to just make the changes that I wanted. I added some comments here related to the changes I made, so no need to review my version.

geniza/corpus/templates/corpus/document_detail.html
```diff
@@ -32,6 +32,14 @@
...
         <link rel="preconnect" href="https://cdnjs.cloudflare.com/ajax/libs/openseadragon/3.0.0/images/" crossorigin />
     {% endif %}
+    {# Disable ITT Panel when no JS available #}
+    <noscript>
+        <style>
+            #itt-panel {
+                display: none;
+            }
+        </style>
+    </noscript>
```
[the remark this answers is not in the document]
blms: Ah, OK! I had misunderstood that this was requested at some point. It must have just been for the deep zoom and the editor portions that it should be disabled. My mistake!

Issue #891, opened 2022-05-27
As a user I want to click or tap on image controls to turn on deep zoom so I can inspect the image in more detail.


rlskoeser:
## testing notes (round 2)

Confirm the following on mobile:
- [x] When zooming, the toggle remains visible (https://github.com/Princeton-CDH/geniza/issues/891#issuecomment-1181795872)
- [x] Zoom container does not fill whole browser screen (https://github.com/Princeton-CDH/geniza/issues/891#issuecomment-1181934672)

## testing notes
Look at the image+transcription viewer for documents that have images available. 
- [x] images should be static and not zoomable by default
- [x] on desktop, interacting with the slider above the image will enable deep zoom and make the image zoomable and interactive
- [x] on desktop after zoom is enabled, sliding the slider all the way to the left will turn zoom off again
- [x] on mobile, zoom is enabled and disabled by a toggle. Once enabled, can use gestures to zoom and rotate. Once disabled, can scroll through the page normally.

## dev notes
Images are not zoomable by default; tap to zoom within existing container; x to turn off  zoom mode
On desktop, trigger via +/- or something outside image to activate
- use stimulus to load/activate OSD; so we'd probably want a data prop on the image container with the url for OSD

rlskoeser:
@blms when you revise the zoom functionality, please add the following:
- image should fade out the first time that open seadragon loads (since there's a slight delay with OSD coming in; hoping this will make it less jarring)
- when deep zoom is disabled, reset the image to original position so that turning it on again will match the static image

gissoo:
@blms here are the designs for this issue: 
1. [zoom toggle on mobile](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=1863%3A46499)
2.[ image rotation on desktop](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=1863%3A62435)

let me know if you need anything else.

richmanrachel:
Okay, mobile zoom works too! The mobile rotate is a bit too sensitive though - when trying to do normal zoom motions, it rotates the document significantly
![IMG_1468](https://user-images.githubusercontent.com/68028846/178302714-a05c9e5d-d03c-40ec-aa32-366ee4ef01e9.PNG)

richmanrachel:
@rlskoeser - please advise whether this issue should be closed or needs more attention. Thanks!

[...]

PR #933, opened 2022-06-23
Deep zoom (#891)

blms:
## In this PR

- Per #891:
  - Implements deep zoom control (OpenSeadragon) on mobile and desktop

## Questions

- Is the desktop rotation tool MVP? And should it be part of this PR or separate?
  - Decided to make it a separate branch so that I can get this merged first, rotation control is complicated!

### DIFF
--- a/geniza/corpus/tests/test_corpus_templates.py
+++ b/geniza/corpus/tests/test_corpus_templates.py
@@ -55,7 +55,7 @@ class TestDocumentDetailTemplate:
         response = client.get(document.get_absolute_url())
         assertContains(
             response,
-            '<section id="itt-panel" data-controller="iiif">',
+            '<section id="itt-panel" data-controller="transcription" data-action="click@document->transcription#clickCloseDropdown">',
         )
 
     def test_viewer_annotations(self, client, document, unpublished_editions):
--- a/geniza/corpus/templates/corpus/snippets/document_transcription.html
+++ b/geniza/corpus/templates/corpus/snippets/document_transcription.html
@@ -1,5 +1,5 @@
 {% load i18n static corpus_extras %}
-<section id="itt-panel">
+<section id="itt-panel" data-controller="transcription" data-action="click@document->transcription#clickCloseDropdown">
     {# NOTE: inputs must be kept as SIBLINGS of panel-container for CSS selection/controls #}
     {# images displayed by default; disable if no images are available #}
     {# translators: label for checkbox toggle to show the images panel for a document #}
@@ -14,6 +14,27 @@
     <input type="checkbox" class="toggle" id="translation-on" aria-label="{% translate 'show translation' %}" disabled="true"/>
     <label for="translation-on"><svg><use xlink:href="{% static 'img/ui/all/all/translation-toggle.svg' %}#translation-toggle" /></svg></label>
     <div class="panel-container">
+        {# label row #}
+        <div class="label-row"></div> {# unused (no image label, needed for column alignment) #}
+        <div class="transcription-panel label-row">
+            {# dropdown is disabled by default; enable if javascript is active #}
+            <details class="transcription-select" aria-expanded="false" data-transcription-target="dropdownDetails" data-edition-count="{{ document.digital_editions.count }}" disabled="true">
+                <summary class="transcription-select-trigger" data-action="keydown->transcription#shiftTabCloseDropdown">
+                    <span data-transcription-target="editionShortLabel">Edition: {{ document.digital_editions.0.source.all_authors }}</span>
+                </summary>
+                <ul>
+                    {% for ed in document.digital_editions.all %}
+                        <li>
+                            <label for="edition-{{ forloop.counter }}">
+                                <input type="radio" name="transcription" {% if forloop.first %} checked="true"{% endif %} value="relevance" data-action="input->transcription#changeTranscription keydown->transcription#keyboardCloseDropdown" id="edition-{{ forloop.counter }}" data-edition="ed-{{ ed.pk }}" />
+                                <span>Edition: {{ ed.source.all_authors }}</span>
+                            </label>
+                        </li>
+                    {% endfor %}
+                </ul>
+            </details>
+        </div>
+
         {% comment %}
 NOTE: if we have more transcription pages than images,
 they won't show in current implementation.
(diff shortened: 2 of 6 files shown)


geniza/corpus/tests/test_corpus_templates.py
```diff
@@ -55,7 +55,7 @@ def test_viewer(self, client, document):
         response = client.get(document.get_absolute_url())
         assertContains(
             response,
-            '<section id="itt-panel">',
+            '<section id="itt-panel" data-controller="iiif">',
```
rlskoeser, line 58: why is the controller at this level? shouldn't each deep zoom image be bound to its own instance of the controller?

sitemedia/js/controllers/iiif_controller.js
```diff
@@ -4,32 +4,194 @@ import { Controller } from "@hotwired/stimulus";
...
+        if (!OSD) {
+            this.addOpenSeaDragon(
+                container,
+                [container.dataset.iiifUrl],
+                isMobile
+            );
+            OSD = container.querySelector(".openseadragon-container");
+        }
+        // OSD styles have to be set directly on the element instead of adding a class, due to
+        // its use of inline styles
```
rlskoeser, line 40: thanks for the comment. this is an OSD limitation? we can't put our styles on the container?

sitemedia/js/controllers/iiif_controller.js
```diff
@@ -4,32 +4,194 @@ import { Controller } from "@hotwired/stimulus";
...
+                // If zooming to less than 100%, force it to 100%
+                zoom = minZoom;
+            }
+            // Set zoom slider value to the chosen percentage
+            zoomSlider.value = parseFloat(zoom);
+            this.updateZoomUI(zoom, false, zoomSlider, zoomSliderLabel);
+        });
+        if (isMobile) {
+            viewer.addHandler("canvas-release", () => {
+                // Handle issue on mobile that causes a scroll trap
```
rlskoeser, line 161: can you elaborate on the scroll trap? Is it if the image is showing mid-screen?

what happens with long images? do we / should we limit them to `100vh` on mobile ?

sitemedia/js/controllers/iiif_controller.js
```diff
@@ -4,32 +4,194 @@ import { Controller } from "@hotwired/stimulus";
...
+        if (deactivating) {
+            secondColor = "#9E9E9E";
+            slider.classList.remove("active-thumb");
+        } else if (!slider.classList.contains("active-thumb")) {
+            slider.classList.add("active-thumb");
+        }
+        slider.style.background = `linear-gradient(to right, var(--link-primary) 0%, var(--link-primary) ${percent}%, ${secondColor} ${percent}%, ${secondColor} 100%)`;
+    }
+    resetBounds(viewer) {
+        // Reset OSD viewer to the boundaries of the image, position in top left corner
```
rlskoeser, line 182: is `viewer.viewport.goHome(true);` not sufficient?

sitemedia/js/controllers/iiif_controller.js
```diff
@@ -4,32 +4,194 @@ import { Controller } from "@hotwired/stimulus";
...
+                    // reset bounds immediately and zoom to 110%
+                    this.resetBounds(viewer);
+                    viewer.viewport.zoomTo(1.1);
+                }
+            });
+        });
+        viewer.addHandler("zoom", (evt) => {
+            // Handle changes in the canvas zoom
+            let { zoom } = evt;
+            if (zoom <= minZoom) {
```
rlskoeser, line 151: how is it possible to zoom to less than `minZoom`?

rlskoeser (approved):
The code seems more complicated than I thought would be necessary, but it does seem to be working correctly on both desktop and mobile. I'm ok moving forward with this.

sitemedia/js/controllers/iiif_controller.js
blms: For some reason, on `canvas-release` the OSD container forcibly scrolls into view, which causes the mobile toggle to become hidden. I guess this is to make it easier to manipulate the image, but it makes our control inaccessible.

Long images would be a problem, yes! And the height would likely need to leave space for the toggle control, so maybe something like `100vh - height of toggle` would be a good idea.

sitemedia/js/controllers/iiif_controller.js
blms: You would think it wouldn't be, but it is! Using the scroll wheel on my mouse, I get these two `zoom` events in rapid succession:

```
zoomed to 0.8333333333333334
minZoom: 1

zoomed to 1
minZoom: 1
```

rlskoeser: ah, weird. I would have thought that the minzoom config in the OSD seadragon would prevent that.

sitemedia/js/controllers/iiif_controller.js
```diff
@@ -4,32 +4,194 @@ import { Controller } from "@hotwired/stimulus";
...
+        // Reset OSD viewer to the boundaries of the image, position in top left corner
+        const bounds = viewer.viewport.getBounds();
+        const newBounds = new OpenSeadragon.Rect(
+            0,
+            0,
+            bounds.width,
+            bounds.height
+        );
+        viewer.viewport.fitBounds(newBounds, true);
+        viewer.viewport.setRotation(0, true);
```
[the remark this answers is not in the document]
blms: I wasn't familiar with this function, but it sets the bounds incorrectly.

<details><summary>goHome</summary>

Screen Shot 2022-07-07 at 11 21 18 AM
</details>

<details><summary>My function</summary>

Screen Shot 2022-07-07 at 11 21 27 AM
</details>

sitemedia/js/controllers/iiif_controller.js
rlskoeser: ah, ok. please elaborate the comment in the code to explain this so we have it documented when we revisit this code.

The height limit you propose sounds good to me. Gissoo may already have some examples of longer images we can test with, or if not we can ask the team.

sitemedia/js/controllers/iiif_controller.js
rlskoeser: wow, neither option is great. let's leave it as you've implemented for now

sitemedia/js/controllers/iiif_controller.js
rlskoeser: no, haven't run into it yet — I'm good with tracking that on a separate issue so we can move forward

Issue #937, opened 2022-07-07
As a desktop user, I want to click to rotate the deep zoom image of a fragment so that I can view it in alternate orientations.


rlskoeser:
## testing notes

This is a provisional rotation solution for mvp of the new viewer so we can get it out sooner, so it just needs to be good enough for now.  Gissoo intended it for desktop only, but I asked Ben to make it available on mobile so you can compare the functionality with the buttons vs touch rotation and let us know what you prefer.

Please try out the rotation on the test site to confirm:
- [x] rotation buttons are disabled until deep zoom is activated
- [x] rotation buttons worked as expected once you activate deep zoom

## dev notes
Simple version of rotation for mvp version of image+transcription viewer.

- [x] add appropriate phosphor icons for rotation; place them roughly where the designed angle rotation will go (hopefully there's space).
- [x] bind rotation icons to OpenSeaDragon controls for rotate left and rotate right in OSD config
- [x] remove OSD image preloading

richmanrachel:
@rlskoeser - works perfectly! I think I still want Marina to weigh in re: mobile, but this is good to close!

PR #939, opened 2022-07-11
Basic rotation controls for ITT panel on desktop (#937)

blms:
## In this PR

- Per #937:
  - Remove OSD icons preconnect
  - Implement basic rotate controls with phosphor icons

## Questions

- Should the rotation controls always be enabled? For now, I've disabled them when deep zoom isn't enabled; I know we discussed eventually allowing users to enable/disable OSD using the rotation controls, but not sure if that makes as much sense for the MVP version. The strict 90º rotation could be quite jarring; it makes the transition much harder to handle, certainly. And not sure if a deactivation state should exist with this rotation control. But I can also easily see it being confusing to have the rotation controls simply disabled until you start zooming. Wonder if there's a good interim solution?
- With taller images, some surprising OSD stuff can happen on rotation. I noticed that the rotation axis seems to be the canvas, rather than the center of the image (or maybe I have that backwards, but something about the axis is unexpected to me). I also noticed the "bounds" of the canvas seem to change on rotation, like you can't pan as far up and down when the image is rotated. Should we try to do anything about this or just chalk it up to OSD limitations?

### DIFF
--- a/sitemedia/js/controllers/iiif_controller.js
+++ b/sitemedia/js/controllers/iiif_controller.js
@@ -4,7 +4,7 @@ import { Controller } from "@hotwired/stimulus";
 import OpenSeadragon from "openseadragon";
 
 export default class extends Controller {
-    static targets = ["imageContainer"];
+    static targets = ["imageContainer", "rotateLeft", "rotateRight"];
 
     imageContainerTargetDisconnected(container) {
         // remove OSD on target disconnect (i.e. leaving page)
@@ -48,6 +48,9 @@ export default class extends Controller {
         } else {
             OSD.style.top = "72px";
         }
+        // re-enable rotation buttons
+        this.rotateRightTarget.removeAttribute("disabled");
+        this.rotateLeftTarget.removeAttribute("disabled");
     }
     deactivateDeepZoom(container, image) {
         // Hide OSD and show image
@@ -59,6 +62,8 @@ export default class extends Controller {
             OSD.style.opacity = "0";
             image.classList.add("visible");
             image.classList.remove("hidden");
+            this.rotateRightTarget.setAttribute("disabled", "true");
+            this.rotateLeftTarget.setAttribute("disabled", "true");
         }
     }
     addOpenSeaDragon(element, tileSources, isMobile) {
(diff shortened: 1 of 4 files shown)


sitemedia/js/controllers/iiif_controller.js
```diff
@@ -144,6 +149,13 @@ export default class extends Controller {
                     viewer.viewport.zoomTo(1.1);
                 }
             });
+            // initialize desktop rotation buttons
+            this.rotateLeftTarget.addEventListener("click", () => {
+                viewer.viewport.setRotation(viewer.viewport.getRotation() - 90);
+            });
+            this.rotateRightTarget.addEventListener("click", () => {
+                viewer.viewport.setRotation(viewer.viewport.getRotation() + 90);
+            });
```
rlskoeser, line 158: These buttons could have been configured in the OSD initialization options (along with rotation increment), but since we're planning to customize further in when we implement the angle input, this seems fine.

sitemedia/scss/components/_transcription.scss
```diff
@@ -480,6 +480,33 @@
                 color: var(--on-background);
                 @include typography.body-bold;
             }
+            // rotation controls (desktop)
+            button.rotate {
+                display: none;
```
rlskoeser, line 485: How much trouble to display on mobile and fit in with the existing toggle?

rlskoeser (approved):
Looks good to me.

I agree that rotation controls should be disabled until deep zoom is enabled; I think we discussed using them to enable/disable zoom but discarded the idea. Much better to limit that to the toggle and slider.

I see what you mean about the long images, that's problematic. I'm wondering if we could limit the height of the OSD container — but if we do, then we'll probably have to customize the OSD config for the initial placement of the image to make sure that it still matches the static version. We don't have this problem in the version that's in production now because the zoom container never gets that long.

I think we need to make the OSD container wider, too, though — especially when the image panel is the only one displayed. When you zoom in, you shouldn't be restricted to seeing the portion that fits in the width of the static image. And presumably these two things are related, right?

sitemedia/scss/components/_transcription.scss
blms: Easy to do, but were we waiting on further feedback on mobile rotation to do this?

rlskoeser: Let's make them available on mobile but hold off on disabling touch rotate. Doesn't hurt to have more than one way to do it and then they can test and compare.

PR #944, opened 2022-07-12
Deep zoom revisions (#891)

blms:
## In this PR

- Per #891:
  - Fix issues with scrolling by putting OSD in its own container and constraining its height
- Use Stimulus targets rather than DOM queries to refer to elements

## Questions
- Should we increase the width of OSD when only images are checked? In that case, we'd need some way for Stimulus to pick that up, such that we could modify `resetBounds` to position OSD according to the static image. In addition, the following CSS would be required:
    ```scss
    #images-on:checked ~ .panel-container {
        [...]
        // image panel spacing
        div.img {
            margin: 0 auto;
            overflow: visible;
            .deep-zoom-container {
                overflow: visible;
            }
            .osd {
                width: 100vw; // Some other width?
            }
        }
    }
    ```
    Not sure what width makes the most sense though, 100vw might be overkill.

### DIFF
--- a/sitemedia/scss/components/_transcription.scss
+++ b/sitemedia/scss/components/_transcription.scss
@@ -569,12 +569,12 @@
                 width: 100%;
                 margin: 0 auto;
             }
-            .osd-inner-container {
+            .osd {
                 position: absolute !important;
                 top: 0;
             }
             img,
-            .osd-inner-container {
+            .osd {
                 &.hidden {
                     transition: opacity 300ms ease, visibility 0s ease 300ms;
                     opacity: 0;
--- a/sitemedia/scss/components/_transcription.scss
+++ b/sitemedia/scss/components/_transcription.scss
@@ -565,11 +565,19 @@
         }
         .deep-zoom-container {
             position: relative;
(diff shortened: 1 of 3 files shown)


sitemedia/scss/components/_transcription.scss
```diff
@@ -563,18 +563,36 @@
...
-                opacity: 0;
-                visibility: hidden;
-            }
-            &.visible {
-                transition: opacity 300ms ease, visibility 0s ease 0ms;
-                visibility: visible;
-                opacity: 1;
+            text-align: center;
+            img {
+                width: auto;
```
rlskoeser, line 571: Why did you set image width to `auto` ? I think on mobile it should still be `100%`, otherwise we're losing significant resolution on the static images (at least on the long ones). I tested locally and it looks like that still works properly (might need some kind of style tweak for non-mobile to make sure the image is centered within the panel, or maybe I just missed something in my quick check.)

rlskoeser (commented):
refactor looks good, functionality seems like works well, I just want to see the mobile image back to the previous width

Issue #916, opened 2022-06-09
As a content editor I want to use basic formatting in transcription content so that I can enter lines as numbered lists or tag when the language changes within a document.


rlskoeser:
## Testing notes (qa)

After testing https://github.com/Princeton-CDH/geniza/issues/911, confirm that when you start editing an annotation body, you can use:
- [x] line breaks
- [x] numbered lists
- [x] language tags
- [x] strikethroughs

Also confirm
- [x] you can save formatted changes 
- [x] saved content with formatting displays when you reload the page

## dev notes
integrate tinymce editor into tahqiq text editing block

revisions after testing:
- [x] style superscript/subscript
- [x] style lists
- [x] allow deleting empty content annotations

glitch prototype: 
- code https://glitch.com/edit/#!/pgp-transcription-prototype?path=editor.html%3A1%3A0
- view https://pgp-transcription-prototype.glitch.me/editor.html

richmanrachel:
This mostly works perfectly! The only feature that works in the edit field but doesn't save as such is the superscript.

The language tags stayed with the highlighted text even when I had two languages in the same transcription zone :) 

@rlskoeser and @blms - should I close because the testing features above worked, or mark as tested needs attention since there's still one available feature that doesn't save properly?

rlskoeser:
@richmanrachel I suspect we are missing styles for that tag (similar to the list), so we should kick it back and fix that before closing.

Can you elaborate what you meant about the language tags staying with the highlighted text?

richmanrachel:
@rlskoeser - great, thanks!

I meant that when I closed a transcription and re-opened it, that I could click the text and the language feature was highlighted and maintained the difference between the Hebrew and Arabic text blocks (this screen shot isn't great, but you can see that the language switcher is darker than normal because it's been activated)
<img width="512" alt="image" src="https://user-images.githubusercontent.com/68028846/180870458-0e049d11-6ece-4a0a-ad3b-3dcd1b9329f7.png">

rlskoeser:
@richmanrachel thanks for explaining — that's great to hear, and a good confirmation that it's working as hoped.

richmanrachel:
It works! Just want to bring to @gissoo's attention that the strikethrough lines through this Arabic font make the original word hard to read (sorry for the gibberish!)
<img width="510" alt="image" src="https://user-images.githubusercontent.com/68028846/182674883-08e6c420-21ac-43fe-a9a0-0ce72fbfc42b.png">

PR #945, opened 2022-07-12
Include TinyMCE from CDN in transcribe view (#916)

blms:
## In this PR

- Per #916:
  - In tandem with https://github.com/Princeton-CDH/annotorious-tahqiq/pull/5, allow editing using TinyMCE.

## dev notes

After spending some hours debugging bundling errors, I chose to use the CDN rather than Webpack. Hopefully that's OK! (For what it's worth, this is also [Tiny's recommendation](https://www.tiny.cloud/docs/tinymce/6/webcomponent-pm/))

### DIFF
--- a/geniza/corpus/templates/corpus/document_transcribe.html
+++ b/geniza/corpus/templates/corpus/document_transcribe.html
@@ -20,7 +20,7 @@
 {% endblock extrastyle %}
 {% block extrascript %}
     {% render_bundle_csp "annotation" "js" attrs='defer' %}
-    <script src="https://cdn.tiny.cloud/1/dxa2pi2u1fimtzb8r4t1mtpfckgiuj5vbpk3tdmxyxwdtu67/tinymce/5/tinymce.min.js" referrerpolicy="origin"></script>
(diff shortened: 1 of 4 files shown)


geniza/corpus/templates/corpus/document_transcribe.html
```diff
@@ -20,6 +20,7 @@
 {% endblock extrastyle %}
 {% block extrascript %}
     {% render_bundle_csp "annotation" "js" attrs='defer' %}
+    <script src="https://cdn.tiny.cloud/1/dxa2pi2u1fimtzb8r4t1mtpfckgiuj5vbpk3tdmxyxwdtu67/tinymce/5/tinymce.min.js" referrerpolicy="origin"></script>
```
rlskoeser, line 23: This is fine as long as we document it clearly in tahqiq setup instructions.  

However, given that this has a code embedded in the url, it should be configured in django settings. We should put the Tiny API key in django local settings, include it in a context processors (either template globals or maybe just the transcribe view, since we only need it on this page).

I can add it to the settings for production/qa in the ansible deploy scripts.

blms: Done, and I also added a line of CSS that will be useful for upcoming changes in block labeling, to avoid opening a new PR here just for that!

rlskoeser (approved):

PR #961, opened 2022-07-19
preliminary annotation server backend

rlskoeser:
first pass implementation of w3c annotation protocol; needs testing, permission checking, and other fine-tuning but wanted to get an initial review first

should be reviewed in tandem with https://github.com/Princeton-CDH/annotorious-tahqiq/pull/9


geniza/annotations/models.py
```diff
@@ -0,0 +1,82 @@
...
+            anno["canonical"] = self.canonical
+        if self.via:
+            anno["via"] = self.via
+        # make a copy of the base annotation data
+        base_anno = anno.copy()
+        # update with the rest of the annotation content
+        anno.update(self.content)
+        # overwrite with the base annotation data in case of any collisions
+        # between content and model fields
+        anno.update(base_anno)
```
blms, line 81: Is the idea here to catch the client somehow returning other values for one of those fields? Is there a known circumstance where that happens?

geniza/annotations/views.py
```diff
@@ -0,0 +1,90 @@
...
+class AnnotationList(View, MultipleObjectMixin):
+    model = Annotation
+    http_method_names = ["get", "post"]
+
+    paginate_by = None  # disable pagination for now
+
+    def get(self, request, *args, **kwargs):
+        # strictly speaking, annotations endpoint should return an annotation container,
+        # but that structure looks terrible to work with and we need a way to search,
+        # which is not defined in the w3c annotation protocol.
```
blms, line 29: Is that what the endpoint returns in SAS? And by searching you mean allowing the overall geniza search to search transcription content, right?

geniza/annotations/views.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # implement something similar to SAS search by uri
+        annotations = self.get_queryset()
+        # if a target uri is specified, filter annotations
+        target_uri = self.request.GET.get("target_uri")
+        if target_uri:
+            annotations = annotations.filter(content__target__source__id=target_uri)
+
+        # return json response with list of annotations
+        # TODO: eventually we'll need to limit or paginate this
```
blms, line 39: Probably need UI for that too…

geniza/annotations/views.py
```diff
@@ -0,0 +1,90 @@
...
+        json_data = json.loads(request.body)
+        anno = Annotation()
+        anno.set_content(json_data)
+        anno.save()
+        resp = AnnotationResponse(anno.compile())
+        resp.status_code = 201  # created
+        # location header must include annotation's new uri
+        resp.location = anno.uri()
+
+        # TODO: should we create log entries to document activity?
```
blms, line 56: Yes!

blms (approved):
This is looking great so far. Had a few questions, but everything looks right. I think the log entries are a good idea. Let me know if you'd like me to write tests.

geniza/annotations/views.py
rlskoeser: SAS search endpoint just returns a list of annotations. Django doesn't like returning non-dicts by default, so I thought I'd wrap it in a dict, esp. since we may need to add pagination or other properties. 

The SAS annotation endpoint returns an AnnotationList, but that doesn't seem to exist in w3c annotation; we could model it on `AnnotationPage` or `AnnotationCollection`. Strictly speaking it should be an annotation collection; I didn't like the way the results are nested (`first.items`) but I see now that it's nested that way because the `first` item is an `AnnotationPage`.  Maybe that's not terrible.

geniza/annotations/models.py
rlskoeser: Yes — initialize in the desired output order, but make sure those values take precedence. It seemed like the simplest solution to do what I wanted. I tried to account for that with the method to set content that strips out the other fields, but it would be possible to update the content without using that method (although not through the current APIs).

geniza/annotations/views.py
rlskoeser: how do you mean, UI?

geniza/annotations/views.py
rlskoeser: ok! was waffling on that, but agree it's probably a good idea. I should probably configure a minimal admin interface, then, too, so we have a way to investigate the content.

geniza/annotations/views.py
blms: I was thinking you meant limit or paginate the results for UI _reasons_, i.e. the list is too long to display, but now I realize you probably just mean to avoid the payload size from getting to large.

rlskoeser: Ah, yes. Right now the main annotations endpoint is returning everything, which I thought would be helpful as we're developing it. Once we split the tei into chunks and convert to annotations, that will be far too many to display in a single result. But was thinking we could do something simple to display counts and page links or similar.

geniza/annotations/views.py
blms: I see. Took a look at the W3C spec now. It does seem that's the preferred way to collect annotations into a list. I lean towards following the spec as closely as possible, but I also totally get the hesitation if it adds a bunch of overhead to this (as well as things like searching). I'll leave it up to you, but my vote would be for `AnnotationCollection`.

rlskoeser: I'm not opposed to supporting annotation collection at some point, but my goal at the moment is to get us a minimal working version that we can use with our existing annotation code and deliver transcription functionality for testing. We're not using the main endpoint at all, except to search for annotations by target uri so we can load them for display.

I looked again, I can't find anything about search anywhere in the w3c annotation protocol or model specs, or in the LDP doc (annotation container is an LDP). I'm biased against LDP because that's what Fedora Commons 4 was supposed to be built on, and my understanding is that it ended up being so inefficient that the Hydra/Samvera community basically abandoned it in favor of the postgres-based Valkyrie that the PUL devs created.

I did look briefly at iiif search API https://iiif.io/api/search/1.0/ — this is what SAS implements for search (separate from the base annotations endpoint), which explains why it returns an AnnotationList. I can see the value of that, actually, because an annotation list can be referenced in a iiif manifest.  This part does impact our current code, since we need to know where to search for annotations on load.  

What do you think about a separate search endpoint with a minimal implementation (for now) that returns an annotation list when you search by target uri?

blms: That makes sense to me! Good point about not using the main search endpoint at all yet—it doesn't seem needed at this point. Thank you for the detailed look into the options here.

geniza/common/middleware.py
```diff
@@ -29,3 +30,20 @@ def __call__(self, request: HttpRequest):
...
         return response
+
+
+class LocaleMiddleware(DjangoLocaleMiddleware):
+    ### customize locale middleware to exempt some urls from redirects
+    # adapted from https://code.djangoproject.com/ticket/17734
+
+    redirect_exempt_paths = ["admin", "annotations", "accounts"]
+
+    def process_response(self, request, response):
```
blms, line 41: Interesting that overriding `process_response` like this doesn't seem to be in the Django docs! Seems like it's something that could come up a lot.

geniza/annotations/views.py
```diff
@@ -125,9 +135,9 @@ def delete(self, request, *args, **kwargs):
         # should use etag / if-match
         # deleted uuid should not be reused (relying on low likelihood of uuid collision)
         anno = self.get_object()
-        anno.delete()
-
-        # create log entry to document deletion
+        # create log entry to document deletion *BEFORE* deleting
```
blms, line 163: Good call!

geniza/annotations/views.py
```diff
@@ -44,14 +44,38 @@ def get_permission_required(self):
         return ()
 
     def get(self, request, *args, **kwargs):
```
blms, line 46: This is looking really good!

blms (approved):
Tests, admin, other changes all look good! Tested and working. We can probably remove `annotorious-sas-storage` from `package.json`, then good to go I think.

geniza/annotations/views.py
rlskoeser: figured it out when I wrote the tests! it doesn't work otherwise (for obvious reasons)

geniza/common/middleware.py
rlskoeser: Yeah, I was surprised too. Glad I found the discussion & a solution that was easy to adapt, at least.

rlskoeser:
> ... We can probably remove `annotorious-sas-storage` from `package.json`, then good to go I think.

Thanks for the reminder – that crossed my mind at some point and then I forgot!  Removing and then will merge into the prototype transcription branch once all the checks pass.

Issue #956, opened 2022-07-18
As a user, when I search for a document that is only on one side of a fragment, I want to see the relevant image first so that I can preview the document more accurately.


rlskoeser:
## testing notes (qa)

After setting a "side" for a fragment on a document in #955, search for that same document in the public site search (can search like `pgpid:###`). 
- [x] Confirm that in the search results page, only the thumbnail for the selected side shows up.

## dev notes
In `Document.index_data` when indexing iiif images in Solr, only include images that are part of the current document based on recto/verso side information; if `side` is unset, don't filter images. Looks like it would be easiest to implement as an optional param to the `Document.iiif_images` method.

richmanrachel:
Beautiful, closing!

Issue #955, opened 2022-07-18
As a content editor, when I select a fragment “side” in the document edit form I want an indicator of which fragment images will be displayed so that I can confirm I’m selecting the correct side or sides.


rlskoeser:
## testing notes (qa)

In the admin edit form for a document, in the "related fragments" section, change the selected "Side" for a fragment connected to the document. Save your changes on the document, and scroll back down to the fragments section. There:
- [x]  Confirm that the selected side or sides are clearly highlighted in the image thumbnails

## dev notes
Side information is stored in `TextBlock.side` and can currently be recto, verso, or both.

- [x] revise the `Fragment.iiif_thumbnails` method used in the textblock inline display to add a class for selected image based on side. (For now assume recto / verso matches iiif image order). 
- [x] add admin css to add an indicator to thumbnails based on the selected sides (e.g. colored outline or similar)

rlskoeser:
@mrustow or @richmanrachel how do you use (or want to use) the `side` field (recto, verso, recto & verso) wen a fragment has more than two images? Do you need a way to select a subset of images as being part of the document when there are more than two images? 

Here are a few shelfmarks with more than two images in case it's helpful to look at examples (let me know if you want the whole list):
- T-S Ar.54.46
- Or.1081 J56
- T-S 10J7.6
- T-S 8J17.18

In my local development data (which is probably slightly out of sync) I found 293 IIIF manifests with more than two associated canvases out of 15,455 total cached manifests.

rlskoeser:
@richmanrachel are these cases where you'd want to make a custom iiif manifest, like Alan and I discussed on #701 ? Or is this a case where the manifest is fine but you want to select which images from it are relevant?

We'll need to discuss how to support selecting images in cases that are not simple recto/verso, I'm not sure how to approach it.

richmanrachel:
@rlskoeser - I don't think I can make a decision about this without further discussion. Is this a blocker for what you're trying to get out today? Or can we discuss it next week?

rlskoeser:
@richmanrachel Not a blocker, and agreed we should discuss. We'll work on the implementation for the simple cases for now, which will hopefully give us all more context for thinking about the other cases.

richmanrachel:
Works!

PR #966, opened 2022-07-20
Filter/select side in IIIF image/thumbnail methods (#955, #956)

blms:
## In this PR

- Per #955:
  - Add `selected_side` param to `Fragment.iiif_thumbnails()`; use in `TextBlock.thumbnail`
  - Use that to set a class `"selected"` to the thumbnail of the selected side, and add some styling to the admin CSS
- Per #956:
  - Add `filter_side` param (filter by each `TextBlock`'s side) to `Document.iiif_images()`; use in solr indexing
  
## Questions

(asked on Slack but here is probably better)

1. For side filtering, should we go by index of the image (ie assume first image always = recto) for all methods? I only saw it mentioned for `iiif_thumbnails` so wasn’t sure if I should take the same approach on `Document.iiif_images`
   - I took an approach using labels in f7b102a and 073d023, but reverted to going by order—wondering if that's the direction we're going or if we should just go by order for now.
2. For the admin view, it will only update the UI for the “selected side” when you save the change. Should we somehow recalculate the thumbnail classes when you change the dropdown? I guess the thumbnails already don’t show up at all until you save the document with the fragment attached, so maybe it’s fine as is.

### DIFF
--- a/sitemedia/css/admin-local.css
+++ b/sitemedia/css/admin-local.css
@@ -221,3 +221,8 @@ a#needsreview:target {
     clip: rect(0, 0, 0, 0);
     white-space: nowrap;
 }
+
+/* selected side thumbnail for fragment admin */
+img.selected {
+    border: 5px solid var(--accent);
+}
--- a/DEPLOYNOTES.md
+++ b/DEPLOYNOTES.md
@@ -1,5 +1,9 @@
 # Deploy Notes
 
+## 4.6.0
+
+-   This update includes Solr indexing changes to show the relevant side(s) of document images in search results. Run `python manage.py index` to reindex all content.
+
 ## 4.5.0
 
 -   Document date functionality in this release requires updating the Solr index. Run `python manage.py index` to reindex all content.
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -612,12 +612,8 @@ class Document(ModelIndexable, DocumentDateMixin):
             frag_images = b.fragment.iiif_images()
             # image indices of this TextBlock's side (0 is recto, 1 is verso)
             selected = [
-                0
-                if any(b.side == s for s in [TextBlock.RECTO_VERSO, TextBlock.RECTO])
-                else None,
-                1
-                if any(b.side == s for s in [TextBlock.RECTO_VERSO, TextBlock.VERSO])
-                else None,
+                0 if b.side in [TextBlock.RECTO_VERSO, TextBlock.RECTO] else None,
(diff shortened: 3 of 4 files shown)


rlskoeser:
In answer to your questions:

>     1. For side filtering, should we go by index of the image (ie assume first image always = recto) for all methods? I only saw it mentioned for `iiif_thumbnails` so wasn’t sure if I should take the same approach on `Document.iiif_images`

Yes, based on my last conversation with Marina about this, assume first image is always recto. We can't go by label because it's unreliable. Marina has said in the past that sometimes the scholar's opinion of recto/verso differs from the library or museum's designation, but for practical reasons it makes sense to follow what they use — my hope is that providing an indicator in the admin interface will make this easier.

>     2. For the admin view, it will only update the UI for the “selected side” when you save the change. Should we somehow recalculate the thumbnail classes when you change the dropdown? I guess the thumbnails already don’t show up at all until you save the document with the fragment attached, so maybe it’s fine as is.

That was my intention; not worth investing javascript in this, and we have other features in the admin site that function similarly, so hopefully it will be familar.

sitemedia/css/admin-local.css
```diff
@@ -221,3 +221,8 @@ a#needsreview:target {
     clip: rect(0, 0, 0, 0);
     white-space: nowrap;
 }
+
+/* selected side thumbnail for fragment admin */
+img.selected {
+    border: 5px solid var(--accent);
```
rlskoeser, line 227: I had to look to see that this is defined in the django admin base styles, interesting. I wonder if it's visible enough / distinct enough from the colors in the images, but I don't see another obvious choice in the existing color variables so let's go ahead and try it.

rlskoeser (commented):
The logic is working correctly for the two-image fragments I tried, but I find the code a little hard to read and I don't like the duplication. Is there a simple way to refactor a lookup or utility method that would do the equivalent in one place? Maybe a method on `TextBlock` that returns selected/included image indexes based on `side` variable?

There are some fragments with more than two images; I'd been thinking that choosing "recto and verso" implies selecting all images rather than first and second, as you've implemented here, but we should get input on how the team wants to handle those cases instead of assuming!

blms:
> There are some fragments with more than two images; I'd been thinking that choosing "recto and verso" implies selecting all images rather than first and second, as you've implemented here, but we should get input on how the team wants to handle those cases instead of assuming!

@rlskoeser It sounds like the team has more to discuss before making a decision on this. Should we leave it as first and second for now, or select all images?

rlskoeser:
@blms I'm fine with whichever is easier for now — esp. since no side selection is effectively all images, right?

blms:
@rlskoeser right!

Issue #958, opened 2022-07-18
As a user looking at images for a single document, I want easy access to documents on images from the same fragment that are not part of the current document.


rlskoeser:
## dev notes

This is a follow up / refinement to #957 

If a document has images that are not part of it, which have other documents, we should link to them in the space where we would normally display the transcription.   `Document.related_documents` which we use for the Related Documents tab should find the correct records; maybe call it in the view if a document includes non-document images, and then filter documents for display based on the iiif url indexed in Solr index data based on side information.

richmanrachel:
Something I noticed in the backend as I was testing #957 but is even more evident on the frontend is how the intro text still stays with the first image, regardless of whether that image is part of the document:
<img width="1051" alt="image" src="https://user-images.githubusercontent.com/68028846/185242883-579174f0-4091-4b73-be93-79a05868f24d.png">

richmanrachel:
But this feature is working as expected, so I'll close here!

rlskoeser:
> Something I noticed in the backend as I was testing #957 but is even more evident on the frontend is how the intro text still stays with the first image, regardless of whether that image is part of the document

Yes... I guess when we give you the image order override feature, then you can re-order so the non-document image isn't first?

richmanrachel:
@rlskoeser - hmmm... yes, that's probably a good idea. Now that we're replacing recto/verso with ordering it hopefully won't screw things up?

rlskoeser:
@richmanrachel I think that it will work and not cause new problems. :-) Can you make a note somewhere so we can keep this in mind as a test case to try when we test the ordering feature and the transcription editor work?

richmanrachel:
@rlskoeser - yes, will do. Thanks!

Issue #957, opened 2022-07-18
As a user viewing document details I want to see which images are not part of the document so that I understand which parts of the fragment are used for the current document.


rlskoeser:
## testing notes (QA)

For [several documents](https://github.com/Princeton-CDH/geniza/issues/957#issuecomment-1218446947) containing images that are not selected (in the admin, i.e. designated as being part of the document), check the document details page image viewer and confirm:
- [x] it is clear which images are and are not part of the document
- [x] if applicable, link or links appear under the excluded images to documents that _do_ include them

## dev notes

design work for this feature was tracked on https://github.com/princeton-cdh/geniza/issues/884 ([here is the design](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=1565%3A53416))

summary of decisions:
- images that aren't part of the current document are displayed with white box with 60% transparency on top of the image
- these images are _not_ zoomable
- we still display shelfmark and side information as for other images
- we should not display transcription content adjacent to non-document images
- if there is a document on that image, we link to it in place of transcription; Figma design says "view document details", but we will use document title (type + shelfmark). Note that there could be multiple documents. [let's make this a separate issue]

implementation:
- [x] add an included/not included indicator to information returned by `Document.iiif_images` (similar to the admin version in #955 )
- [x] add an appropriate class name or other to non-document images in `document_transcription.html`
- [x] add styles for non-document images
- [x] omit zoom behavior for non-document images
- [x] when displaying transcription content in `document_transcription.html`, skip non-document images (not sure how to do this in current forloop structure; if it's complicated let's discuss)

implementation updates after design revision and transcription format decision:
- [x] update non-document image display and links to match revised designs as documented in https://github.com/Princeton-CDH/geniza/issues/997#issuecomment-1209723393
- [x] revise sync transcript to return html in a dict of lists keyed on canvas uri; skip non-document images
- [x] in document transcription template, display transcription html by canvas uri instead of index using `dict_item`

rlskoeser:
@gissoo what should happen to the links to other documents when only the transcription panel is showing? Do they go away?

gissoo:
@rlskoeser it's fine if the links to other documents stay in the format you've mentioned "document title (type + shelfmark)"

rlskoeser:
Revised estimate from 2 pts to 3 since we had to add logic to associate html chunks with canvases skipping non-document images.

rlskoeser:
[...]

PR #969, opened 2022-07-21
Indicate excluded images in ITT viewer (#957 #958)

blms:
## In this PR

- Per #957:
  - Show a 60% white overlay on images that are not part of the document, based on `TextBlock.side`
  - Hide zoom controls for those excluded images
- Per #958:
  - Don't display transcription chunk next to image
  - Display related documents there instead

## Questions

- For related documents, I didn't fully understand the filtering you described in #958, so right now I'm just showing all related documents. Can you elaborate on "filter documents for display based on the iiif url indexed in Solr index data based on side information"? And when you said to call it in the view, did you mean the view class or the template?
  - Should there also be a label here before the link, like "See **Type: shelfmark**"?
- Also for related documents, should I create a reusable template snippet for document link with type + shelfmark? It's identical code to the document results page right now. (It's also nearly identical to the one in `document_header`, but that one uses `doctype` and `shelfmark_display` and has a fallback/translation for "Unknown type".)

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -621,6 +621,7 @@ class Document(ModelIndexable, DocumentDateMixin):
                         "image": img,
                         "label": labels[i],
                         "shelfmark": b.fragment.shelfmark,
+                        "not_in_document": b.side and i not in b.image_indices,
                     }
                     for i, img in enumerate(images)
                     # include if filter inactive, tb has no side info, or index in tb indices
--- a/geniza/corpus/templates/corpus/snippets/document_transcription.html
+++ b/geniza/corpus/templates/corpus/snippets/document_transcription.html
@@ -76,9 +76,8 @@ Also doesn't currently account for fragments without images.
                         <ul class="excluded-related">
                             {% for related_document in document.related_documents %}
                                 <li>
-                                    {% translate 'Unknown type' as unknown_type %}
                                     <a href="{% url 'corpus:document' related_document.pgpid %}">
-                                        <span class="doctype">{{ related_document.type|default:unknown_type }}</span>
+                                        <span class="doctype">{{ related_document.type }}</span>
                                         <span class="shelfmark">{{ related_document.shelfmark|shelfmark_wrap }}</span>
                                     </a>
                                 </li>
--- a/geniza/corpus/templates/corpus/snippets/document_transcription.html
+++ b/geniza/corpus/templates/corpus/snippets/document_transcription.html
@@ -47,18 +47,20 @@ Also doesn't currently account for fragments without images.
                 <div class="img" id="{{ image_info.label }}" data-controller="iiif" data-iiif-target="imageContainer">
                     <div class="img-header" data-iiif-target="imageHeader">
                         <h3>{{ image_info.shelfmark }} {{ image_info.label }}</h3>
-                        <input data-iiif-target="zoomSlider" data-action="iiif#handleDeepZoom" id="zoom-slider-{{ forloop.counter0 }}" type="range" name="zoom-slider" min="1" max="100" value="0" step="0.01" />
-                        <label data-iiif-target="zoomSliderLabel" for="zoom-slider-{{ forloop.counter0 }}">100%</label>
-                        <input data-iiif-target="zoomToggle" data-action="iiif#handleDeepZoom" id="zoom-toggle-{{ forloop.counter0 }}" type="checkbox" name="zoom-toggle" />
-                        <label for="zoom-toggle-{{ forloop.counter0 }}">{% translate 'Zoom and Rotate' %}</label>
-                        <button type="button" data-iiif-target="rotateLeft" class="rotate" aria-label="{% translate 'Rotate left' %}" disabled>
-                            <i class="ph-arrow-counter-clockwise"></i>
-                        </button>
-                        <button type="button" data-iiif-target="rotateRight" class="rotate" aria-label="{% translate 'Rotate right' %}" disabled>
-                            <i class="ph-arrow-clockwise"></i>
-                        </button>
+                        {% if not image_info.not_in_document %}
+                            <input data-iiif-target="zoomSlider" data-action="iiif#handleDeepZoom" id="zoom-slider-{{ forloop.counter0 }}" type="range" name="zoom-slider" min="1" max="100" value="0" step="0.01" />
+                            <label data-iiif-target="zoomSliderLabel" for="zoom-slider-{{ forloop.counter0 }}">100%</label>
+                            <input data-iiif-target="zoomToggle" data-action="iiif#handleDeepZoom" id="zoom-toggle-{{ forloop.counter0 }}" type="checkbox" name="zoom-toggle" />
+                            <label for="zoom-toggle-{{ forloop.counter0 }}">{% translate 'Zoom and Rotate' %}</label>
+                            <button type="button" data-iiif-target="rotateLeft" class="rotate" aria-label="{% translate 'Rotate left' %}" disabled>
+                                <i class="ph-arrow-counter-clockwise"></i>
+                            </button>
+                            <button type="button" data-iiif-target="rotateRight" class="rotate" aria-label="{% translate 'Rotate right' %}" disabled>
+                                <i class="ph-arrow-clockwise"></i>
+                            </button>
+                        {% endif %}
                     </div>
-                    <div class="deep-zoom-container">
+                    <div class="deep-zoom-container{% if image_info.not_in_document %} excluded{% endif %}">
                         <img class="iiif-image" data-iiif-target="image" src="{{ image_info.image|iiif_image:"size:width=500" }}" alt="{{ image_info.label }}" title="{{ image_info.label }}" loading="lazy"
                             sizes="(max-width: 650px) 50vw, 94vw"
                             srcset="{{ image_info.image|iiif_image:"size:width=300" }} 300w,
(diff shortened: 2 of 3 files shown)


geniza/corpus/models.py
```diff
@@ -621,6 +621,7 @@ def iiif_images(self, filter_side=False):
                         "image": img,
                         "label": labels[i],
                         "shelfmark": b.fragment.shelfmark,
+                        "not_in_document": b.side and i not in b.image_indices,
```
rlskoeser, line 624: the double negative in the template was making my head hurt a little bit, but I see now why you did it — it's easier to tell explicitly if an image is not part of a document / should be excluded, but included images could either be explicit or implicit

geniza/corpus/templates/corpus/snippets/document_transcription.html
```diff
@@ -47,18 +47,20 @@
...
-                        <label data-iiif-target="zoomSliderLabel" for="zoom-slider-{{ forloop.counter0 }}">100%</label>
-                        <input data-iiif-target="zoomToggle" data-action="iiif#handleDeepZoom" id="zoom-toggle-{{ forloop.counter0 }}" type="checkbox" name="zoom-toggle" />
-                        <label for="zoom-toggle-{{ forloop.counter0 }}">{% translate 'Zoom and Rotate' %}</label>
-                        <button type="button" data-iiif-target="rotateLeft" class="rotate" aria-label="{% translate 'Rotate left' %}" disabled>
-                            <i class="ph-arrow-counter-clockwise"></i>
-                        </button>
-                        <button type="button" data-iiif-target="rotateRight" class="rotate" aria-label="{% translate 'Rotate right' %}" disabled>
-                            <i class="ph-arrow-clockwise"></i>
-                        </button>
+                        {% if not image_info.not_in_document %}
```
rlskoeser, line 50: a one-line comment here would be helpful to indicate why this section is conditional; something like this

```suggestion
                        {# only images that are part of the current document are zoomable #}
                        {% if not image_info.not_in_document %}
```

geniza/corpus/templates/corpus/snippets/document_transcription.html
```diff
@@ -47,18 +47,20 @@
                 <div class="img" id="{{ image_info.label }}" data-controller="iiif" data-iiif-target="imageContainer">
```
rlskoeser, line 47: can we / should we omit the `data-controller` for images that are not part of the current document?

geniza/corpus/templates/corpus/snippets/document_transcription.html
```diff
@@ -69,18 +71,32 @@ <h3>{{ image_info.shelfmark }} {{ image_info.label }}</h3>
...
+                                    <a href="{% url 'corpus:document' related_document.pgpid %}">
+                                        <span class="doctype">{{ related_document.type }}</span>
+                                        <span class="shelfmark">{{ related_document.shelfmark|shelfmark_wrap }}</span>
+                                    </a>
+                                </li>
+                            {% endfor %}
+                        </ul>
+                    {% else %}
+                        {% if forloop.first %}
+                            <h3>Transcription</h3>
```
rlskoeser, line 88: it's possible a document occurs on verso only and not recto, in which case the transcription label would never be shown — I think it needs to be outside of  / separate from the in-document check

geniza/corpus/templates/corpus/snippets/document_transcription.html
```diff
@@ -69,18 +71,32 @@ <h3>{{ image_info.shelfmark }} {{ image_info.label }}</h3>
...
-                        <h3>Transcription</h3>
-                        <span data-transcription-target="editionFullLabel" class="current-transcription">{{ document.digital_editions.0.display|safe }}</span>
+                    {% if image_info.not_in_document %}
+                        {# if this image isn't actually in this document, link to related documents #}
+                        <ul class="excluded-related">
+                            {% for related_document in document.related_documents %}
+                                <li>
+                                    <a href="{% url 'corpus:document' related_document.pgpid %}">
+                                        <span class="doctype">{{ related_document.type }}</span>
+                                        <span class="shelfmark">{{ related_document.shelfmark|shelfmark_wrap }}</span>
```
rlskoeser, line 81: I see what you mean about how we're duplicating this, but I don't think a template snippet makes sense. Do we need the spans here? I guess we probably do need the shelfmark wrapping for linking to multishelfmark docs.

If we're concerned about the reuse and consistency, what about a method on `Document` like `display_title` or `formatted_title` that would be parallel to `title` but adds the spans/wrapping?

rlskoeser: Oh, but that doesn't help us with the solr results, which isn't actually a `Document` instance.  We could implement it on our solr result class, maybe, but again, not sure how much that helps.

This is still small enough I'm not sure I want the opacity of a template snippet.

rlskoeser (changes requested):
This looks pretty good, I think it's nearly there.

Please use the styles and adapt the language from Gissoo's designs:
https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=1562%3A44641

Instead of "View document details ->" let's do "View [document title] ->" (where doc title is the type + shelfmark you have already).

I'll comment separately to try to explain what I meant about filtering related documents.

rlskoeser:
> Can you elaborate on "filter documents for display based on the iiif url indexed in Solr index data based on side information"? And when you said to call it in the view, did you mean the view class or the template?

For any document on one more than one fragment/shelfmark, the related documents won't all apply to all the images that are displayed on the document detail page.  Imagine a join between A and B with three documents, where document 1 is on recto A and recto B, document 2 is on verso A, and document 3 is on verso B. When we're viewing document 1, each of the excluded images should only link to a single document. We should be able to filter these based on iiif url, since we have them in the document detail view and we're indexing them in Solr.  

My comment about view/template is that I'm not sure the easiest or most efficient way to filter related documents by iiif url. I could imagine doing some preprocessing in the view (probably in `get_context_data`) to organize the related docs by excluded image url, but it looks like it's probably easy enough to filter them in the template.

The filter is something like `if image_info.image in related_document.iiif_images` — except that I think at least some of these are piffle `ImageClient` objects and they may not compare very easily. 

It occurred to me as I wrote it up that this filter only works for documents with iiif images — but we're already working within that subset in this context anyway.

geniza/corpus/models.py
rlskoeser: maybe exclude or excluded would be clearer

geniza/corpus/templates/corpus/snippets/document_transcription.html
```diff
@@ -69,18 +71,32 @@ <h3>{{ image_info.shelfmark }} {{ image_info.label }}</h3>
                     </div>
                 </div>
                 <div class="transcription-panel">
-                    {% if forloop.first %}
-                        <h3>Transcription</h3>
-                        <span data-transcription-target="editionFullLabel" class="current-transcription">{{ document.digital_editions.0.display|safe }}</span>
+                    {% if image_info.not_in_document %}
+                        {# if this image isn't actually in this document, link to related documents #}
+                        <ul class="excluded-related">
```
rlskoeser, line 76: there will be some excluded images with no related documents; any problem with having an empty `<ul>`? should we structure this in a way that we know if we have any?

geniza/corpus/templates/corpus/snippets/document_transcription.html
blms: That actually also brings up a related problem: which chunk of the transcription is displayed on verso in that same case. Since there will only be one chunk in the transcription, it won't show up at all. One example is PGPID 4474, which you can see appears on the [wrong page in QA](https://test-geniza.cdh.princeton.edu/en/documents/4474/) and is omitted entirely on this branch.

rlskoeser: this was what I meant by my instructions in the issue:
> when displaying transcription content in document_transcription.html, skip non-document images (not sure how to do this in current forloop structure; if it's complicated let's discuss)

Not skip as in not display but skip as in don't count excluded images for matching/indexing purposes.

I'm not sure an easy way to do it, but we should try to come up with something simple since we know the way we're displaying transcriptions is going to change soon.

geniza/corpus/templates/corpus/snippets/document_transcription.html
blms: Would it be OK to just `display: none` when it's empty?

rlskoeser: that sounds like a good solution to me!

geniza/corpus/views.py
```diff
@@ -305,6 +305,15 @@ def get_context_data(self, **kwargs):
                 "page_type": "document",
                 # preload transcription font when appropriate
                 "page_includes_transcriptions": self.object.has_transcription(),
+                "related_documents": [
```
rlskoeser, line 308: probably worth a brief comment so we remember why this is here
```suggestion
                # generate list of related documents that can be filtered by image url for links on excluded images
                "related_documents": [
```

rlskoeser: Can we only generate if the document has images that are excluded? I don't know if that's something we have easy access to here, but it would prevent an extra call to Solr that is probably not needed for a lot of cases.

rlskoeser (commented):
Updates look good. Can we easily skip the solr query for related documents when it's not needed? (the majority of records probably won't)

The view related document link is green in light mode, but I don't see a color in dark mode. Can we add an equivalent color to make it more obvious that it's a link?

blms:
> The view related document link is green in light mode, but I don't see a color in dark mode. Can we add an equivalent color to make it more obvious that it's a link?

I actually was just using the exact style and colors of the "view document details ->" link from search results! But I added a pink color to make it stand out more on this page.

rlskoeser:
> I actually was just using the exact style and colors of the "view document details ->" link from search results! But I've now added a pink color to make it stand out more on this page.

Gissoo's design doesn't have an underline displaying by default, which I would prefer, but that feels like a quibble. It's certainly better now than it was; I care most that it's obviously a link, which it is now.

rlskoeser (approved):
All you revisions look good. 

But we shouldn't merge until we figure out a way to make sure transcription content displays properly with excluded images, correct?

blms:
> But we shouldn't merge until we figure out a way to make sure transcription content displays properly with excluded images, correct?

Right!

rlskoeser:
@blms just pushed changes to update the sync transcription script to structure the html as a dict keyed on canvas uri. I updated the templates a bit because it was easier to see if I was generating the new structure correctly.

Please review and fill in anything I missed, and then add tests for whatever you think is needed (maybe just a couple simple cases of the html chunking?). Might make sense to move that logic into a method, certainly would make it easier to test. (I'm not sure if we'll want to reuse that functionality for the tei to annotation script, I think I already have a provisional version of that logic — but this version has the local canvas uris, so may be better.)

blms:
@rlskoeser All of your code looks great, the templates are all working right, and I think the tests you already wrote for chunking are fine. Did you have specific additional tests in mind? Or did you mean I should write tests for `sync_transcriptions`? (Looks like we don't have any at the moment, but not sure if it's necessary.) Otherwise I think we should be good to go.

rlskoeser:
@blms so sorry I clobbered your changes, thanks for fixing. The new styles look really good.

I forgot we weren't testing the sync transcriptions script — I guess because it's temporary. I'm ok with that, if we adapt any of the html / canvas uri logic for the tei to annotation we can test it there.

rlskoeser (approved):

Issue #909, opened 2022-06-07
On the admin site, I want the tag list view to include counts for how many times its used, in order to understand the scope of tags and to help clean them later.


richmanrachel:
## testing notes
In django admin, go to the list of tags:
- [x] confirm you see a count for each tag
- [x] confirm you can sort by count

**Is your feature request related to a problem? Please describe.**
In #506 we asked for a tag merge solution. But before we merge tags, it would be helpful to understand the spread, scope, and overlap of existing tags.

**Describe the solution you'd like**
Similar to how the language+script list view is, it would be great to add a column to the tag list view with numbers of how often the tag is used.

## dev notes

Should be similar to implementation with `LanguageScriptAdmin` and queryset annotated with `Count`.  

Possibly additional complication that we're using default taggit admin view — will likely have to extend the default taggit admin model object, unregister their version, and register the customized version. It seems unlikely a custom admin model queryset will impact the autocomplete keyword lookup speed, but should be tested.

richmanrachel:
@rlskoeser - this is great! I think we can close this issue now, then can we get a CSV download of just the categories that show up on this tag list view/would it be easy to implement that as a command for this page in general?

rlskoeser:
@richmanrachel what's the use case for a CSV export of tags?

richmanrachel:
@rlskoeser - so that we can re-sort by name, for example, to determine which tags need to be merged. And to identify which of the lesser used tags should be deleted.

rlskoeser:
@richmanrachel you can sort tags by name in the admin site, which is a better place to work with them and do cleanup.

richmanrachel:
@rlskoeser - oh, fantastic! you're right. No problem then :)

PR #970, opened 2022-07-21
Extending taggit.admin.TagAdmin to include document counts

owenduffymassey:
For [issue#909](https://github.com/Princeton-CDH/geniza/issues/909).  Let me know if there are any additional issues! Thanks for the help.


owenduffymassey:
Here is how my local setup runs tags with the document counts.
Screen Shot 2022-07-21 at 4 24 19 PM

geniza/corpus/admin.py
```diff
@@ -592,3 +594,32 @@ def save_model(self, request, obj, form, change):
...
+    list_display = ("name", "slug", "document_count")
+
+    def get_queryset(self, request):
+        # The annotations we use for document count on the list view
+        # make the search too slow for autocomplete.
+        # Reset to original, unannotated queryset *only* for autocomplete
+        qs = super().get_queryset(request)
+        if request and request.path == "/admin/autocomplete/":
+            # return without annotations
+            return qs
```
rlskoeser, line 613: Did you test if taggit uses admin querysets for autocompletes? I think taggit probably custom autocomplete views, so these lines are unneeded.

geniza/corpus/admin.py
```diff
@@ -592,3 +594,32 @@ def save_model(self, request, obj, form, change):
...
+        # otherwise, annotate with counts
+        qs = qs.annotate(
+            # taggit_taggeditem_items is how to reference the taggeditems for tags
+            document_count=Count("taggit_taggeditem_items", distinct=True),
```
rlskoeser, line 617: Technically this is counting any tagged item, not just documents; we're only currently using tags for documents but I expect that to change in the future; so I suggest `item_count` for the field and `Count` for the label

rlskoeser (commented):
Implementation looks good.

Please make the following changes:
- [ ] move the custom tag admin code to `geniza.common.admin` since it isn't specific to the corpus app (and we already have a lot of code in corpus admin)
- [ ] remove unneeded autocomplete exception in `get_queryset`
- [ ] write unit tests for the new methods (get queryset & document count); can probably adapt from similar tests in the corpus app

owenduffymassey:
For the custom tag admin test, is it ok verify the item count number in the query set test? Also, what file should this test be in?

rlskoeser:
Hi @owenduffymassey 

> For the custom tag admin test, is it ok verify the item count number in the query set test? 

We run code coverage reports, so both of the methods you wrote should be tested (even though there is overlap). But that's the basic idea — create some test tags, call the queryset method, and confirm the counts are set as expected.

> Also, what file should this test be in?

Tests are organized by app, so these should be added to `geniza.common.tests`

owenduffymassey:
Ok great! Thanks for the help

rlskoeser (approved):

rlskoeser (changes requested):
Ah, my apologies! I somehow missed that you opened your PR against `main`. We use git flow, so the PR needs to be to `develop`. I edited the PR to change the target but now it's including a bunch of irrelevant changes.

Would you please see if you can easily rebase your branch against current `develop` code? Or if that doesn't work, create a new branch based on `develop` and cherry-pick the commits from this branch.

Once that is resolved, this is ready to merge in and then I can deliver for acceptance testing.

owenduffymassey:
I just rebased my code against develop, should I continue by pushing with [`git push --force`](https://riptutorial.com/git/example/4357/pushing-after-a-rebase)?

rlskoeser:
> I just rebased my code against develop, should I continue by pushing with [`git push --force`](https://riptutorial.com/git/example/4357/pushing-after-a-rebase)?

yes, go ahead!

owenduffymassey:
I just pushed the new pull request against develop! For some reason the rebase wasn't working well for me.

rlskoeser:
cherry-picked into a fresh branch from develop and merged in via PR #981

Issue #959, opened 2022-07-18
As a user, I want the full citation for a transcription in context so I know who authored it and where it came from.


rlskoeser:
## testing notes (qa)

- [x] Check the transcriptions for a few different kinds of documents—example PGPIDs could be: 606, 9121, 2691, 4508, 3254—and confirm that the citation in gray text provides the necessary information, including when switching between multiple transcriptions.

## known revisions needed

- [x] switching back to Gil edition on [pgpid 606](https://test-geniza.cdh.princeton.edu/en/documents/606/), Goitein on [pgpid 9121](https://test-geniza.cdh.princeton.edu/en/documents/9121/) loses some info (related to links?)

## dev notes

Transcription citation should be displayed in gray at the top of the transcription panel.

This was already implemented in an earlier phase even though it was not mvp, but I'm not sure it was tested directly and I noticed in passing some citations that didn't look correct. Let's check the dev implementation and be sure we're using the correct citation and then test it with all source variations.

blms:
We are indeed using the same short citation format that is used in the metadata section. A few things I noticed:

1. That citation format doesn't seem to italicize book titles—I recall vaguely that it was intended to be plaintext, but should it be?
2. Should the citation format include "[digital geniza document edition]"? Right now we're only including that on the long format, but in the designs it appears next to the transcriptions.
3. I noticed sometimes the spacing is different depending on whether or not there is a heading. It interacts strangely with the `min-height` we set on the citation. I think that's more of a problem with the transcription styling than the citation styling, though. See below:
   <details>
   <summary>Spacing differences</summary>
   
   <img width="698" alt="Screen Shot 2022-07-25 at 2 18 49 PM" src="https://user-images.githubusercontent.com/7234006/180846719-435317d1-a77b-4bda-b9c8-b729e4bb606f.png">
   <img width="677" alt="Screen Shot 2022-07-25 at 2 18 53 PM" src="https://user-images.githubusercontent.com/7234006/180846726-91d427f0-f4f1-459f-bcdf-65470802e7ee.png">
   <img width="684" alt="Screen Shot 2022-07-25 at 2 18 57 PM" src="https://user-images.githubusercontent.com/7234006/180846728-c1b2a568-dbf1-497d-a304-ee0b11037452.png">
   
   </details>
4. Unrelated, but I found during this testing that the transcription switcher doesn't work at all for documents with no images, so time to fix that!

rlskoeser:
[...]

PR #976, opened 2022-07-25
Fix transcription switcher on docs w/ no images (#899), use full citation (#959)

blms:
## In this PR

- Per https://github.com/Princeton-CDH/geniza/issues/959#issuecomment-1194461536, fixes the transcription switcher (#899) for documents with no images. (It was not working at all before due to missing Stimulus targets)

## Questions

Similar to how the chunking index in documents with images relies on the number of images, the chunking index here relies on the number of chunks in `digital_editions.0`. I wonder if we should come up with a different way to handle chunk indices, since that's also the problem we're having with https://github.com/Princeton-CDH/geniza/pull/969?

### DIFF
--- a/geniza/corpus/templates/corpus/snippets/document_transcription.html
+++ b/geniza/corpus/templates/corpus/snippets/document_transcription.html
@@ -88,9 +88,18 @@ Also doesn't currently account for fragments without images.
             {% for chunk in document.digital_editions.0.content.html %}
                 <div class="img"></div> {# empty image div needed for grid layout #}
                 <div class="transcription-panel">
-                    {% if forloop.first %}<h3>Transcription</h3>{% endif %}
-                    {# display transcription in chunks based on loop index #}
-                    <div class="transcription">{{ chunk|h1_to_h3|safe }}</div>
+                    {% if forloop.first %}
+                        <h3>Transcription</h3>
+                        <span data-transcription-target="editionFullLabel" class="current-transcription">{{ document.digital_editions.0.display|safe }}</span>
+                    {% endif %}
+                    <div class="editions">
+                        {# display transcription in chunks based on loop index #}
+                        {% for edition in document.digital_editions.all %}
+                            <div class="transcription ed-{{ edition.pk }}" data-label="{{ edition.display|safe }}">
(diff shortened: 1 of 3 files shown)


geniza/corpus/templates/corpus/snippets/document_transcription.html
```diff
@@ -88,9 +88,18 @@ <h3>Transcription</h3>
             {% for chunk in document.digital_editions.0.content.html %}
                 <div class="img"></div> {# empty image div needed for grid layout #}
                 <div class="transcription-panel">
-                    {% if forloop.first %}<h3>Transcription</h3>{% endif %}
-                    {# display transcription in chunks based on loop index #}
-                    <div class="transcription">{{ chunk|h1_to_h3|safe }}</div>
+                    {% if forloop.first %}
+                        <h3>Transcription</h3>
+                        <span data-transcription-target="editionFullLabel" class="current-transcription">{{ document.digital_editions.0.display|safe }}</span>
+                    {% endif %}
```
rlskoeser, line 94: oh, I see – we missed the stimulus target because we currently have two different ways we output the transcription content, depending on whether there are images or not

rlskoeser (approved):
revision looks good, thanks for catching and correcting

rlskoeser:
> Similar to how the chunking index in documents with images relies on the number of images, the chunking index here relies on the number of chunks in `digital_editions.0`. I wonder if we should come up with a different way to handle chunk indices, since that's also the problem we're having with #969?

I want to work backwards from what the display will be after we migrate to the new transcription/annotation data — but I don't know yet exactly how it will work. I'm thinking it will be based on canvas identifiers, which should resolve our problem with skipping over non-document images. I'm not sure exactly how it will work for docs with no images, but we're going to have canvases without images for them, so we should be able to make it consistent – and should also work to come up with a way that doesn't require us to duplicate the transcription output in the template.

blms:
> and should also work to come up with a way that doesn't require us to duplicate the transcription output in the template.

@rlskoeser In that case, should I stop creating a template snippet for that exact thing right now? 😅

blms:
@rlskoeser I switched to using `formatted_display` and using the min-height only when there are multiple transcriptions; sorry to drop those in the same PR, was working on them while you were reviewing!

Issue #978, opened 2022-07-25
As a desktop user, I want an angle control to rotate the deep zoom image of a fragment, so that I can control the rotation more finely than 90º increments.


blms:
## testing notes (qa — round two)

On documents with images on the QA site, try rotating images with the new rotation control
- [x] confirm it snaps to 0 degrees when you get near 0 (+/- 5 degrees), and this works as expected

## testing notes (qa)

- [x] On some documents with images, on desktop QA site, try rotating images with the new rotation control and confirm it works as expected / write any feedback

## known issues

- center of rotation seems off?

---

Extension of #937 that uses [andrejewski/angle-input](https://github.com/andrejewski/angle-input) and this [design](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=1863%3A62435)

richmanrachel:
@blms - overall I really like it! The only thing that might be convenient is some sort of button to snap it back to "normal" (but I was able to reload the page to get that functionality)

richmanrachel:
@blms and @rlskoeser - such an improvement, thank you!

PR #979, opened 2022-07-25
Implement angle rotation (#978)

blms:
## In this PR

- Per #978:
  - Replace +/- 90º rotation with `angle-input` library, new design

## dev notes

- The Webpack IgnorePlugin is required because of a line `require("jquery")` in the angle rotation library, which Webpack will evaluate even if you aren't using the jQuery part of the library. Fortunately, the library doesn't list `jquery` in its requirements, so we aren't forced to include it in the bundle; instead, we need to tell Webpack specifically to ignore this `require` statement.
- I added this section to support rotation events on desktop that aren't triggered by the angle input, as I recalled from our meeting that @rlskoeser found it was possible. But I don't have a Magic Trackpad or similar, so I have no way of testing if this works. Please check to make sure that the angle input updates its UI when you rotate the canvas this way—or, let me know if you can't actually rotate without using the angle input, in which case I'll remove it.
    https://github.com/Princeton-CDH/geniza/blob/f55751b1fa6c4d1d48b0709be68eec7a0fbe083f/sitemedia/js/controllers/iiif_controller.js#L152-L156


rlskoeser (commented):
Thanks for explaining and documenting the ignore plugin so well.

Code looks good; it's kind of interesting to use the new rotation, not sure what I think about it yet.

It seems like it's working properly, except the center of rotation seems to be off — any idea why that it is or how we can improve it?

I couldn't get gesture rotation to work with my magic touch pad...

rlskoeser:
@blms should we merge & deliver this version for testing and not hold it up for the rotation origin?

Issue #914, opened 2022-06-09
As an admin I want TEI transcription content migrated to IIIF annotation so that I can manage and edit it in the new transcription editor.


rlskoeser:
## testing notes
All TEI content has been migrated to the new annotation format and should be editable in the new transcription editor.  

Not sure how to test this thoroughly (and some of it will be tested under other issues), but let's start with these:
- [x] number of documents with transcriptions in public search should be similar to production
- [ ] number of documents with digital editions in admin search should be similar to production
- [x]  footnotes linking the transcription to the document should be updated as follows:
   - [x] converted from edition to digital edition
   - [x] provisional content removed
   - [x] if existing edition footnote also had a link or other document relations, it should have been preserved and a new digital edition footnote created
- [x] content linked to correct scholarship record (uses same basic logic as sync transcription, so has some similar limitations)
   - [x] check at least one document with more than one transcription
   - [x] check at least one document where transcription for a join is split across two TEI files
- [x] transcription chunked into editable sections by label and grouped with pages based on label (uses same logic as sync transcription script)
   - see https://github.com/Princeton-CDH/geniza/issues/913 for checking editable version
 - [ ] spot check transcriptions that were reported as having unicode normalization to see if there are any concerns (expectation: changes should not be visible)

## dev notes
Adapt from existing sync tei script to the degree possible (maybe early prototype iiif annotation code also helpful). Should create labeled blocks similar to what we have now in the tei -> html; attach translations to the full canvas. (Use % notation if annotorious allows. Possibly tag as full-image/unblocked or similar.) Needs to handle documents without images, with mixed images, etc.

next steps:
- [x] map source uri to dc:source on the annotation
- [x] revise tei to html to use standard `<li>` for easy editing in tinymce; try this: if an unnumbered list item occurs, map it to a paragraph instead, and if the number is not 1 set the start attribute (will need to test if this displays properly and is easily editable in tinymce)
- [x] migrate footnotes to digital editions

after first-round testing:
- [x] normalize unicode when migrating; report on affected files
- [x] empty content has ""; investigate and resolve (migration or display issue)
- [x] migration script should report on how many footnotes with content still remain
- [x] investigate why so many footnotes with content still remain

rlskoeser:
output from running the script in staging:

```
Processed 4492 TEI file(s).
Created 9411 annotation(s).
Could not find footnotes for 63 documents:
	/srv/www/geniza/data/tei_xml/4638.xml
[...]

Issue #755, opened 2022-04-04
As a user, when I’m reading transcription text, numbered lines should only wrap when necessary (based on display width), so that I can see more clearly how the lines match up with the original.


rlskoeser:
## testing notes
- [ ] review the TEI changes on bitbucket : https://bitbucket.org/benjohnston/princeton-geniza-project/pull-requests/20 
- [ ] look at a variety of documents modified and compare on production and test sites to see how the lines have been adjusted
- [ ] when lines are combined, is the order correct?

## dev notes
- [x] clean up with identity transform so semantic changes will be easier to identify
- [x] cleanup labels that are mistagged as lines to avoid consolidating with lines
- [x] cleanup line numbers that are in content instead of attributes
- [x] write an xslt to consolidate unnumbered lines following a numbered line 
- [x] run new xslt against all current TEI transcription files; review locally to check changes
- [x] open pull request for review before merging in changes
- [x] update staging site with content from pull request branch to review in context
- [x] don't wrap lines that only include dashes or dots and whitespace
- [ ] don't wrap lines with single letters and apostrophes
- [x] remove lines with only single bracket or ampersand

rlskoeser:
ok @richmanrachel,  I'm ready for a first round of review on the transcription line cleanup work.

I've modified a set of TEI documents and opened a pull request for those changes on the bitbucket transcription repository, and I've loaded that version of the transcription content into the test site for review.

What I would recommend is that you work from the bitbucket pull request I've opened so that you can look at the changes in the TEI, and then use the PGPID from the filename to look at the content on the test site; it may also be helpful to compare the same document on the production site.

There are 710 files changed — I don't expect you to look at all of them on the site! Please look at enough to see what's going on and give me feedback on my questions below; once we make some decisions I can refine the logic and run the conversion again to skip the things we don't want to wrap. 

I have some questions about specific things I noticed, wanted to get input before revising the logic:
- some transcriptions have what about lines like this `----- `;  see 1461 for an example; 1772 for another; my guess is that these should not be wrapped, but would like to know why they are in the transcription
- some lines only have spaces and periods (1685, 1690, 1816, others); should these also not be wrapped
- other documents have lines with spaces and a few characters or fractions; for example, see 1679, 1698. Should these also not be merged? What are these lines?
- 1736 has a line that looks like a label but it is in brackets; should I edit and convert to label without brackets?
[...]

PR #984, opened 2022-07-26
Fix missing styles on transcription editor (#916)

blms:
## In this PR

- Per https://github.com/Princeton-CDH/geniza/issues/916#issuecomment-1194136341:
  - Style superscript
  - Style line numbers (in transcribe view; main view uses data attribute)
- Update tahqiq to allow editing empty annotations
  - We will probably need to update the commit hash once [the tahqiq PR](https://github.com/Princeton-CDH/annotorious-tahqiq/pull/11) is merged

## Questions

- How should we handle the discrepancy between the line numbering between the editor (plain `<li>` elements) and the existing TEI transcriptions (`<li>` elements with `data-value` attributes)? Something to think about for the upcoming changes to transcription format, maybe?
- Should there be some kind of validation to ensure content in the annotation body?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,7 +4,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "geniza",
       "dependencies": {
         "@hotwired/stimulus": "^3.0.1",
         "@hotwired/stimulus-webpack-helpers": "^1.0.1",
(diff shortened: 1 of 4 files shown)


rlskoeser (approved):

rlskoeser:
> * How should we handle the discrepancy between the line numbering between the editor (plain `<li>` elements) and the existing TEI transcriptions (`<li>` elements with `data-value` attributes)? Something to think about for the upcoming changes to transcription format, maybe?

Yes, this is has been on my mind, I should add as a checklist item to the tei to annotation issue (#914). I'm using data attributes in the current conversion to handle the skipped unnumbered lines. I was hoping I could clean up those wrapped lines in the tei before migration, but that was much harder than anticipated and we iceboxed it. (#755) I may still revisit that one and see if we can scope it to a subset of content, but I think the migration still needs to account for this.

I've been thinking about how to migrate so it will display properly and be easily editable; have an idea to try when I'm able to circle back to that task.

> * Should there be some kind of validation to ensure content in the annotation body?

Yes, good idea — maybe we shouldn't allow saving unless we have at least a label or content.

PR #990, opened 2022-07-27
Snap rotation to 0 within +/- 5 degrees (#978)

blms:
## In this PR

- Remove handler on viewer to handle gesture "rotate" events triggered by interaction with the canvas itself, since per https://github.com/Princeton-CDH/geniza/pull/979#pullrequestreview-1051011969 gesture rotation doesn't seem to work on desktop anyway (and I can't test it)
- Per https://github.com/Princeton-CDH/geniza/issues/978#issuecomment-1195959854:
  - Snap rotation to 0 degrees within +/- 5

### DIFF
--- a/sitemedia/js/controllers/iiif_controller.js
+++ b/sitemedia/js/controllers/iiif_controller.js
@@ -149,15 +149,14 @@ export default class extends Controller {
             this.zoomSliderTarget.value = parseFloat(zoom);
             this.updateZoomUI(zoom, false);
         });
-        viewer.addHandler("rotate", (evt) => {
-            // Handle other changes in the canvas rotation (via Magic Trackpad, e.g.)
-            const { degrees } = evt;
-            this.updateRotationUI(-1 * parseInt(degrees));
-        });
     }
     handleRotationInput(viewer) {
         return (evt) => {
-            const angle = parseInt(evt.target.querySelector("input").value);
+            let angle = parseInt(evt.target.querySelector("input").value);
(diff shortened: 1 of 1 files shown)


rlskoeser:
@blms I think I missed something with one of your questions earlier — the last I understood, Marina wanted to have both touch gesture and button controls for rotation on mobile, since the gesture is so fiddly. Is this easy to restore?

blms:
@rlskoeser Ah, I missed that! Yes, should be fairly easy. It's the old CW/CCW buttons, right?

rlskoeser:
> @rlskoeser Ah, I missed that! Yes, should be fairly easy. It's the old CW/CCW buttons, right?

I'd rather not have two different sets of rotation controls! Can we use the angle input on mobile?

rlskoeser (approved):
logic looks good, seems to work pretty well locally

blms:
> I'd rather not have two different sets of rotation controls! Can we use the angle input on mobile?

@rlskoeser Unfortunately it doesn't really work on mobile. You can tap different areas of the angle input to choose a specific rotation, but you cant rotate the input itself—it just drags the page up and down. I'm not convinced that's less fiddly!

https://user-images.githubusercontent.com/7234006/181367398-72578fd2-78f9-4b3f-8a9b-b77c492cfec2.MP4

rlskoeser:
@blms ok, thanks for testing and documenting! Let's proceed per Gissoo's design and keep this desktop only, then.

Issue #982, opened 2022-07-26
As a content editor I want to select images in the related fragment view in order to determine which images belong with the document.


richmanrachel:
## testing notes (qa)

- [x] In the admin interface for a document on a fragment with attached images, confirm you can click the side(s) to determine which images actually belong with the document
- [x] Save and confirm the image selection is persisted
- [x] If the side is recto, verso, or recto and verso (i.e. selected image is first, second, or first and second) confirm that the "side" column in the related fragments section of the document admin correctly states this
- [x] For a fragment with more than 2 images, confirm you can select images other than just the first two

---

**Is your feature request related to a problem? Please describe.**
To improve on the work in #955, we would like to click on fragment images rather than rely on the "Side" drop down in order to handle more complicated documents.

**Describe the solution you'd like**
Click on the related fragment thumbnails to curate the correct IIIF images to view on the front end for each document.

rlskoeser:
@blms how does this approach sound:
- we add a new field to `TextBlock` for selected image indexes (either array field or comma-separated text field, whichever is easier to manipulate with js)
- add the new image index field to textblock inline edit as a hidden field
- side (recto/verso) is a shortcut to set image index to `0`, `1`, or `0,1` (on model save I think? but probably also needs to interact with js); suggest adding a mapping to the model that defines this and passes it through to the page if needed for js to use
- add javascript to make the images selectable; clicking on images toggles whether they are selected and updates the list in the hidden image index field

richmanrachel:
Love it! Great first step towards getting the full functionality we need :D

PR #991, opened 2022-07-28
Allow select/deselect images in admin (#982)

blms:
## In this PR

- Per #982:
  - Migrate `TextBlock.side` to `selected_images` array field
  - Create `side` property to guess and display side based on indices (0 = recto, 1 = verso)
  - Update admin fields, add JS to allow admins to select/deselect which images are part of a document by clicking


geniza/corpus/models.py
```diff
@@ -1091,15 +1093,15 @@ class TextBlock(models.Model):
...
-    RECTO_VERSO = "rv"
-    RECTO_VERSO_CHOICES = [
-        (RECTO, "recto"),
-        (VERSO, "verso"),
-        (RECTO_VERSO, "recto and verso"),
-    ]
-    side = models.CharField(blank=True, max_length=5, choices=RECTO_VERSO_CHOICES)
+    RECTO = _("recto")
+    VERSO = _("verso")
+    RECTO_VERSO = _("recto and verso")
```
rlskoeser, line 1098: I don't think we need to translate these — AFAIK they won't be displayed on the public site, only in the admin and in the data downloads.

rlskoeser (approved):
Sorry for the delay on reviewing! Implementation looks great.

Only change I'd like to see before merging is unflagging recto/verso labels for translation.

Issue #989, opened 2022-07-27
Implement the light/dark mode toggle so that users can use the site in the UI mode they prefer. 


gissoo:
## testing notes (qa)

Confirm the light/dark mode toggle matches the designs at:
- [x] Mobile (<900px)
- [x] Tablet (900px-1200px)
- [x] Desktop (1200px+)

---

This issue is based on the designs handed off in issue #834

## dev notes 
[desktop](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=2098%3A47371)
- on desktop it's a sticky, and that it sits below the header
- on mobile it's not a sticky and it sits in the header
- the transition is: once selected it would rotate 180 degrees such that the left half of the circle is the active mode (on desktop and mobile)
- on desktop it's 70px
- on mobile it's 40px
- for larger screens it would be great if the toggle's distance with the search form and the nav above it remains the same (doesn't get pushed to the right edge too much)
- the interaction is once selected you get the light mode, and once selected again you get the dark mode
- I've fixed the mobile version and [here is my final proposal on mobile light mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=2098%3A54059) and [mobile dark mode](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=2098%3A53568)
- note that the toggle is placed 16px to the left of the burger menu. I think the language toggles can be placed to the left of this toggle on mobile (we'll know more soon, but I expect this proposal will work in the future as well)
- note that the toggle is horizontally centered with the burger menu

blms:
@gissoo Two questions on this:

1. On desktop, for screens between 900px and about 1130px, what should happen with the light/dark mode toggle? Right now, it will either cover up or be hidden behind the search box/button, depending on z-index.
   <img width="898" alt="Screen Shot 2022-07-28 at 5 05 42 PM" src="https://user-images.githubusercontent.com/7234006/181775328-eca0a6a2-b930-4282-824a-0f39e301cc41.png">
   I know we're not specifically designing for screens of those sizes, but I think the controls should still all be clickable at that size.
2. Also on desktop, right now with `position: sticky` it's going to cover up some of the transcriptions, because the ITT viewer is wider than the normal page content. How should we handle this? Should we not show the toggle on this page?
[...]

PR #998, opened 2022-08-02
Implement revised design for light/dark toggle (#989)

blms:
## In this PR

- Per #989:
  - Implement revised design for light/dark toggle, `position: absolute` in the menu bar on mobile/tablet, and `position: fixed` in the right margin on desktop
  - Set its z-index below that of the ITT panel, and give the ITT panel a bg color
- Ensure use of `for-desktop-up` is deliberate; in all other cases, it should have been `for-tablet-landscape-up` since we use the desktop design for that, unless we need to carve out a specific exception as here


rlskoeser (approved):
Looks good. Noticed one small z-index interaction: on the homepage the toggle is under the banner and not clickable. Can be merged once you correct that.

Issue #913, opened 2022-06-09
As a content editor, I want transcription content linked to a scholarship record so that it is clearly documented who authored the transcription and where it came from.


rlskoeser:
## testing notes

- [x] on public document detail page, if logged in as a content editor and there is an existing transcription, you should see a link to "edit X's transcription"
- [x] clicking the link should take you to the transcription editor for that transcription
- [x] test a document with more than one existing transcription: the transcription content should be separated correctly by source (i.e., when you open the editor you should only be editing one transcription at a time and it should be the correct one)

## dev notes

general approach: transcriptions for different sources will be edited on different urls; we'll store the uri for the scholarship record (source) as a custom field on the annotation so that we can find them again

- differentiate editing existing transcriptions and adding new ones:
  - [x] revise transcription editor url to take a source id to)
  - [x] revise admin links from document details; we'll need separate links for  "edit x's transcription" and "add a new transcription"
- link annotations to sources:
  - [x] implement a URI method on `Source`; use the `ANNOTATION_MANIFEST_BASE_URL` configuration rather than current site absolute url; numeric db id is fine
  - [x] pass source uri as an optional config to tahqiq
  - [x] on save, set source uri as `dc:source` on the annotation (annotation top level)
  - [x] implement a search option in the annotation search view to find annotations by uri _and_ source
  - [x] on load, filter annotations by source uri to get only the annotations for the current source

see also #1002

rlskoeser:
@blms let's use `dc:source` for the scholarship record URI — I just looked at extensions, that's already one of the namespaces that's included in the annotation json context.

kseniaryzhova:
@rlskoeser Every time I click to edit a transcription from the public site, it takes me to the bottom of the page so I have to scroll all the way up to see the actual transcription. PC, Microsoft Edge, Adblocker. Everything else works though!

rlskoeser:
@kseniaryzhova oh yes! I noticed this and mention it to @blms but I don't remember if we're tracking it anywhere yet — the zoom image autoscroll functionality we wanted on the public site needs to be turned off when we're in transcription edit mode since we load all images in deep zoom mode by default for editing.

blms:
Tracked here! https://github.com/Princeton-CDH/geniza/issues/1037

rlskoeser:
did not mean to reopen this!

PR #1004, opened 2022-08-04
Link annotations to scholarship records (#913)

blms:
## In this PR

- Per #913:
  - Add `Source.uri` property
  - Allow searching annotations on `dc:source`/source URI
  - Pass source URI to Tahqiq when editing existing transcription

## Questions

- Should I raise `Http404` here if the source could not be found and `source_pk` is not None? (It will always be None when a user clicks "Add a new transcription")
   https://github.com/Princeton-CDH/geniza/blob/12515f671a1f279ddd3175ecd9c40364c07d4d85/geniza/corpus/views.py#L669-L675
  - Note: the latter behavior may change depending on our approach in https://github.com/Princeton-CDH/geniza/pull/1005 (e.g. https://github.com/Princeton-CDH/geniza/pull/1005/commits/71f5a2475ba7fb9d5e560543f0e993bc5b2a2cfc)

### DIFF
--- a/geniza/corpus/views.py
+++ b/geniza/corpus/views.py
@@ -664,7 +664,7 @@ class DocumentTranscribeView(PermissionRequiredMixin, DocumentDetailView):
 
         # get source uri if source_pk present in kwargs (i.e. editing an existing transcription),
         # and the source exists
-        source_uri = ""
+        source = None
         source_pk = self.kwargs.get("source_pk", None)
         # if source_pk is None, it's a new transcription, so no Source query
         # TODO: once add transcription view is separated, this view should not be called without
@@ -672,7 +672,6 @@ class DocumentTranscribeView(PermissionRequiredMixin, DocumentDetailView):
         if source_pk is not None:
             try:
                 source = Source.objects.get(pk=source_pk)
-                source_uri = source.uri
             except Source.DoesNotExist:
                 raise Http404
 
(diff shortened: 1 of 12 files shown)


geniza/corpus/views.py
```diff
@@ -661,18 +661,34 @@ def page_title(self):
...
+        # and the source exists
+        source_uri = ""
+        source_pk = self.kwargs.get("source_pk", None)
+        try:
+            source = Source.objects.get(pk=source_pk)
+            source_uri = source.uri
+        except Source.DoesNotExist:
+            # NOTE: Should this raise a 404 if source_pk is not None?
+            # TODO: If source_pk is None, instantiate the scholarship record choice form
+            pass
```
rlskoeser, line 675: Shouldn't try to request a source when `source_pk` is `None`; add a comment to clarify that's a new transcription

rlskoeser (approved):
This looks good, only a few small things that I think should be revised before merging.

In addition to the things I commented here, when I tested locally I noticed: we need to update the transcribe editor template to add a label indicating which source you are working on.

rlskoeser:
@blms for the add new transcription, what if we provide an interim page that prompts the user to choose (or link to create) the source scholarship record for their transcription, and only then link to the transcription edit page with the source id specified? That way the transcription edit view can require the source id, and it will always be specified when we need it.  (Maybe you've already thought about this; it wasn't exactly what I was proposing but seems like it would be simpler!)

Issue #1002, opened 2022-08-03
As a content editor, I want to create a new transcription linked to new or existing scholarship records so I can document its source and authorship.


blms:
## testing notes

From a document detail page, choose "add a new transcription".

- [x] you should be prompted to search for and select the source for the transcription you want to add
- [x] once you choose and submit the form, you should be on the transcription editor page and able to start adding transcription content
- [x] after adding the transcription, you should see "edit X's transcription" on the document details page above "add a new transcription" (may need a hard refresh)

## dev notes

Spun off from #913 

  - [x] prompt editor to choose an existing scholarship record to attach the annotations to; we should use existing django autocomplete functionality if at all possible (maybe use is a minimal admin model form?); link to source list in admin to create a new source if there is not already an appropriate one

richmanrachel:
Sorry @rlskoeser - this needs to be reopened. This works for all the documents that have images and/or existing transcriptions. But there is no prompt to create a transcription for documents with neither image nor transcription:
https://test-geniza.cdh.princeton.edu/admin/corpus/document/26619/change/?_changelist_filters=textblock__fragment__iiif_url__isempty%3D1%26transcription%3Dno

richmanrachel:
I'm not able to edit this transcription for a document without an image either (it lets me into the editor, but I can't edit): https://test-geniza.cdh.princeton.edu/en/documents/4744/transcribe/7/

richmanrachel:
I'm also having trouble drawing boxes (I think because of the auto zoom as I mentioned somewhere else) so starting a new transcription is hard :(

richmanrachel:
@rlskoeser - yes, that works for me! Do you have capacity to write the user story or should I?

rlskoeser:
I'll write it — I have some notes somewhere on what needs to happen, if I can find or remember them.

question for you: is it ok if we show a placeholder image so that this will work as much like the other cases as possible, from a code perspective?

richmanrachel:
> is it ok if we show a placeholder image so that this will work as much like the other cases as possible, from a code perspective?
@rlskoeser - yes, this works! Sorry, I thought I had responded to this already... my apologies fro keeping you waiting!

blms:
Closing, superseded by #1039 and #1044.

richmanrachel:
Yes, absolutely!
________________________________

PR #1005, opened 2022-08-08
Choose a scholarship record to associate with new transcription (#1002)

blms:
## In this PR

- Per #1002:
  - When creating a new transcription, admins will be given an intermediate screen prompting them to choose a scholarship record to associate with it using an autocomplete form, which will create a Footnote model to link the record with the Document

<details>
<summary>Questions (outdated)</summary>

- I wasn't able to figure out how to isolate the autocomplete field for use in a `ModelForm`, since the `autocomplete_fields` property only seems to exist on `ModelAdmin`, and Django doesn't seem to expose the field or widget. I ended up going the route of installing a new dependency for this, but now I'm wondering if I should have created a new `ModelAdmin` for documents or footnotes instead. (I was thinking you could only have one per model, but maybe that's only for the `@admin.register` decorator.) Anyway, does this overall approach seem OK or should I try something with `ModelAdmin` instead?
- Does the query look correct for Source autocompletion?

</details>

### DIFF
--- a/geniza/corpus/templates/corpus/add_transcription_source.html
+++ b/geniza/corpus/templates/corpus/add_transcription_source.html
@@ -1,0 +1,28 @@
+{% extends 'base.html' %}
+
+{% load admin_urls static %}
+
+{% block meta_title %}{{ page_title }}{% endblock meta_title %}
+
+{% block main %}
+    <h1>{{ page_title }}</h1>
+    <h2>Select or create source</h2>
+
+    <script type="text/javascript" src="{% static 'admin/js/vendor/jquery/jquery.min.js' %}"></script>
+
+    {{ formset.media }}
+
+    <form action="" method="post">
+        {% csrf_token %}
+        <table style="display: none">
+            {{ formset.empty_form }}
+        </table>
+
+        <fieldset class="module">
+            <table>
+                {{ formset }}
+            </table>
+        </fieldset>
+        <input type="submit" />
+    </form>
+{% endblock %}
--- a/geniza/corpus/templates/corpus/add_transcription_source.html
+++ b/geniza/corpus/templates/corpus/add_transcription_source.html
@@ -23,7 +23,7 @@
         <button class="primary" type="submit">Create transcription</button>
     </form>
 
-    <h2>Source not yet in the database?</h2>
+    <h2>Scholarship record not yet in the database?</h2>
     <p>
         <a href="{% url "admin:footnotes_source_changelist" %}" data-turbo="false">Add a new source record</a>, then return to this page.
     </p>
--- a/geniza/corpus/templates/corpus/add_transcription_source.html
+++ b/geniza/corpus/templates/corpus/add_transcription_source.html
@@ -14,11 +14,11 @@
     <h1>{{ page_title }}</h1>
     <h2>Select scholarship record</h2>
 
-    {{ formset.media }}
+    {{ form.media }}
     <form method="post" id="add-transcription-source">
         {% csrf_token %}
         <fieldset>
-            {{ formset }}
+            {{ form }}
         </fieldset>
         {# TODO: add convenience link to an unpublished source where the currently logged in user is the sole author #}
         <button class="primary" type="submit">Create transcription</button>
--- a/geniza/corpus/templates/corpus/add_transcription_source.html
+++ b/geniza/corpus/templates/corpus/add_transcription_source.html
@@ -20,11 +20,12 @@
         <fieldset>
             {{ formset }}
         </fieldset>
+        {# TODO: add convenience link to an unpublished source where the currently logged in user is the sole author #}
         <button class="primary" type="submit">Create transcription</button>
     </form>
 
     <h2>Scholarship record not yet in the database?</h2>
     <p>
-        <a href="{% url "admin:footnotes_source_changelist" %}" data-turbo="false">Add a new source record</a>, then return to this page.
+        <a href="{% url "admin:footnotes_source_changelist" %}" data-turbo="false" target="_blank">Add a new source record</a>, then return to this page.
     </p>
 {% endblock %}
--- a/geniza/corpus/templates/corpus/add_transcription_source.html
+++ b/geniza/corpus/templates/corpus/add_transcription_source.html
@@ -12,7 +12,7 @@
 
 {% block main %}
     <h1>{{ page_title }}</h1>
-    <h2>Select source</h2>
+    <h2>Select scholarship record</h2>
 
     {{ formset.media }}
     <form method="post" id="add-transcription-source">
--- a/geniza/corpus/templates/corpus/add_transcription_source.html
+++ b/geniza/corpus/templates/corpus/add_transcription_source.html
@@ -1,28 +1,30 @@
 {% extends 'base.html' %}
 
-{% load admin_urls static %}
+{% load admin_urls static render_bundle_csp %}
 
 {% block meta_title %}{{ page_title }}{% endblock meta_title %}
 
+{% block extrascript %}
+    {# jQuery required for autocomplete widget #}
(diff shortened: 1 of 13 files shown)


geniza/corpus/templates/corpus/add_transcription_source.html
```diff
@@ -0,0 +1,30 @@
...
+
+    {{ formset.media }}
+    <form method="post" id="add-transcription-source">
+        {% csrf_token %}
+        <fieldset>
+            {{ formset }}
+        </fieldset>
+        <button class="primary" type="submit">Create transcription</button>
+    </form>
+
```
rlskoeser, line 25: I don't think it should be MVP, but I'm curious: how difficult would it be to search for an unpublished source where the currently logged in user is the sole author? Could we provide a convenience link for that source in addition to this form?

rlskoeser (approved):
This looks great.

I wondered if we'd be able to use the admin autocomplete — I couldn't find any documentation of reversing the views for it, which made me suspect not. Using django-autocomplete-light is a good solution, we've used that in the past (before the admin autocomplete was easily available).

rlskoeser:
the autocomplete query looks reasonable to me

geniza/corpus/templates/corpus/snippets/document_transcription.html
```diff
@@ -3,11 +3,11 @@
     {{ annotation_config|json_script:"annotation-config" }}
 {% elif perms.corpus.change_document and document.iiif_images %}
     {# NOTE: document.iiif_images check is temporary, remove once we can use generated canvas ids #}
-    <a class="editor-navigation" href="{% url 'corpus:document-transcribe' document.id %}" data-turbo="false">
+    <a class="editor-navigation" href="{% url 'corpus:document-add-transcription' document.id %}" data-turbo="false">
         <i class="ph-plus-circle"></i>
         <span>Add a new transcription</span>
     </a>
-    {% for ed in document.digital_editions.all %}
+    {% for ed in document.editions.all %}
```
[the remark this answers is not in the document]
blms: If a footnote was added to this document without adding any digital edition content, i.e. adding a "new transcription" through this new view without actually transcribing lines, then it wouldn't show up in `digital_editions`, and thus wouldn't appear in the list. Perhaps that means we need to adjust the digital editions logic, though I thought it might also make sense to allow admins to add digital versions here for editions that are not yet digital.

This also makes me wonder if we should have a uniqueness check on Footnote for source + content object pairs; I was able to add duplicate footnotes by accident (it didn't show up in the `digital_editions` list here so I thought it wasn't created).

geniza/corpus/templates/corpus/add_transcription_source.html
blms: I think that's a great idea. I wonder if we should actually make it a one-to-one relation on their user account, to remove the guesswork; if there isn't already one, this could be a place to prompt the user to create one.

Either way, I'll add a TODO here.

rlskoeser: You know, I think ages ago we discussed having an optional 1-1 between `Creator` and `User` since some of them are the same people. We didn't have a reason to do it before, but implementing this feature would be a logical time to add it.

geniza/corpus/views.py
```diff
@@ -646,6 +651,74 @@ def form_valid(self, form):
...
+
+    def page_title(self):
+        """Title of add transcription page"""
+        return "Add a new transcription for %(doc)s" % {"doc": self.get_object().title}
+
+    def post(self, request, *args, **kwargs):
+        """Create footnote linking source to document, then redirect to edit transcription view"""
+        source_pk = int(
+            request.POST["footnotes-footnote-content_type-object_id-0-source"]
+        )
```
rlskoeser, line 695: It's confusing to me to call this an inline formset, because that's a django admin thing — and this isn't quite the same as that. Also, that post value is ugly! Do we need to use a model form and formset? Can't we just use a simple form with a single source autocomplete?

geniza/corpus/views.py
```diff
@@ -646,6 +651,74 @@ def form_valid(self, form):
...
+                Q(title__icontains=q)
+                | Q(authors__first_name__istartswith=q)
+                | Q(authors__last_name__istartswith=q)
+                | Q(year__istartswith=q)
+                | Q(journal__icontains=q)
+                | Q(notes__icontains=q)
+                | Q(other_info__icontains=q)
+                | Q(languages__name__icontains=q)
+                | Q(volume__icontains=q)
+            ).distinct()
```
rlskoeser, line 672: This is really slow for me in the autocomplete. I wonder if we need to narrow the list of fields, or add database indexing?

FYI, in dark mode the autocomplete looks like it's only returning one result; I think the non-highlighted results are white on white.

geniza/corpus/views.py
```diff
@@ -646,6 +651,74 @@ def form_valid(self, form):
...
+        return qs
+
+
+class DocumentAddTranscriptionView(PermissionRequiredMixin, CreateView):
```
rlskoeser, line 676: what benefit are we getting from `CreateView` ?

geniza/corpus/views.py
```diff
@@ -646,6 +651,74 @@ def form_valid(self, form):
...
+        source_pk = int(
+            request.POST["footnotes-footnote-content_type-object_id-0-source"]
+        )
+        source = Source.objects.get(pk=source_pk)
+        Footnote.objects.get_or_create(
+            source=source,
+            doc_relation=[Footnote.EDITION],
+            content_type=ContentType.objects.get_for_model(Document),
+            object_id=self.kwargs.get("pk"),
+        )
```
rlskoeser, line 702: We don't need to create the footnote here, or even retrieve the source — footnote will be created when/if annotations are added in the new signal handler you have in a different branch. Just redirect.

geniza/corpus/views.py
```diff
@@ -655,25 +728,21 @@ class DocumentTranscribeView(PermissionRequiredMixin, DocumentDetailView):
...
-        if source_pk is not None:
-            try:
-                source = Source.objects.get(pk=source_pk)
-            except Source.DoesNotExist:
-                raise Http404
+        # source_pk will always be an integer here; otherwise, a different (or no) route
+        # would have matched
+        source_pk = self.kwargs.get("source_pk")
+        try:
+            source = Source.objects.get(pk=source_pk)
```
rlskoeser, line 743: could consolidate if we aren't using `source_pk` anywhere else

```suggestion
        try:
            source = Source.objects.get(pk=self.kwargs["source_pk"])
```

geniza/corpus/views.py
rlskoeser: Also, what happens if a document already has footnotes?

geniza/footnotes/forms.py
```diff
@@ -0,0 +1,23 @@
...
+        }
+
+
+FootnoteInlineFormSet = generic_inlineformset_factory(
+    model=Footnote,
+    form=FootnoteInlineForm,
+    max_num=1,
+    exclude=("content", "url", "doc_relation", "location", "notes"),
+    can_delete=False,
+)
```
rlskoeser, line 23: More complicated than we need! Should just be a source choice form with the one autocomplete.

rlskoeser (changes requested):

geniza/corpus/templates/corpus/snippets/document_transcription.html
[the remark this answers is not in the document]
blms: No longer needed with the new logic (no Footnotes are created).

geniza/corpus/templates/corpus/add_transcription_source.html
blms: Split out into separate issue, thanks!

blms:
Thanks so much for your helpful review on this, @rlskoeser! Really helped me simplify how I was thinking about it.

rlskoeser (approved):
Revisions look great

Issue #1001, opened 2022-08-03
As a content editor, I want footnotes to indicate when a digital edition is available so that I can see and filter on records with and without transcription in the admin interface.


rlskoeser:
## testing notes
- [x] in admin, footnotes now have a "digital edition" document relation type
- [x] in admin, test filtering documents by "has transcription"
- [x] add a new transcription to a document that did not previously have one; confirm that a digital edition footnote was automatically created (testing may overlap with other issues, e.g. adding a new transcription falls under #1002) and that there is a log entry documenting the creation
- [x] delete the last transcription section for a document; confirm that the digital edition footnote was removed and that there is a log entry documenting the change

## dev notes
- [x] update the footnote model to remove the interim `content` field and add a new boolean  `has_transcription` (replaces the existing property)
- [x] add a signal handler that listens to annotation save events and updates the footnote appropriately when new transcriptions are added or all transcriptions are removed
- [x] update admin list views, filters, etc as needed

rlskoeser:
@kseniaryzhova ah, sorry! my testing instructions were vague, there should be a log entry for the footnote, not on the document

Issue #1000, opened 2022-08-03
As a content editor, I want new and revised transcriptions available for search immediately so that changes and new content are all available to all site users.


rlskoeser:
## testing notes

- [x] add and edit a handful of transcriptions as described in #1002 and #913
- [x] go to the public site search and search for the text of the transcriptions you added/edited, and confirm it brings up the document(s) you transcribed successfully

## dev notes

- [x] revise document index data to pull transcription content from annotations 
- [x] add a new signal handler to reindex documents when annotations associated with them are changed or added (maybe via footnote / scholarship record?)
- [x] signal handler to reindex document when annotation changes but footnote does not

richmanrachel:
@rlskoeser - we're both getting this weird pop up in the transcription editor (which we can close and work around), but just fyi:
<img width="1027" alt="image" src="https://user-images.githubusercontent.com/68028846/187286130-627ade57-0b48-41ec-a162-45984cdd8626.png">

richmanrachel:
The auto zoom is also turned on, and so we're not seeing the boxes when we hold the shift key down to try to create regions.

richmanrachel:
The search is working immediately and correctly! Only had time to test with one doc so far, but I think it's good.

rlskoeser:
thanks for alerting me to the Tiny editor warning! I thought I had it configured with the correct API key and the domain listed, will investigate

richmanrachel:
I would like to retest once it's easier for me to add new transcriptions. This is working for editing existing  transcriptions.

richmanrachel:
@rlskoeser - I had the same trouble with drawing boxes at first. After a hard refresh it worked, but sometimes it kept wanting to move the box around as I zoomed in to try to see the text better. The search is still updating automatically!

kseniaryzhova:
@rlskoeser I'm able to draw the boxes after a hard refresh and it doesn't move around when I zoom in. However, if I zoom in, the zoom bar at the top of the document disappears. I can still successfully zoom in and out with my mousepad. 
![image](https://user-images.githubusercontent.com/105385128/188002376-9a10482e-074f-4f6e-b884-7b31e322cb41.png)
https://test-geniza.cdh.princeton.edu/en/documents/8820/transcribe/933/

richmanrachel:
@rlskoeser - should I close and create a new issue for the zoom problems?

rlskoeser:
@richmanrachel yes, please do

Issue #999, opened 2022-08-03
As a content editor, I want transcription edits displayed immediately on the document details so that changes and new content are all available to all site users.


rlskoeser:
## testing notes

- [x] use the transcription editor to make a change to the transcription; load the public document details page and confirm that the change is visible

## dev notes

update public transcription display to pull content from annotation server 

If we were working with an external annotation server, we would want to query the search endpoint by manifest but I think it will be significantly faster to use the queryset.

- [x] revise digital edition footnote html content method to return a dict keyed on canvas uris (adapt sync transcription logic; must handle placeholder canvases for documents without images)
- ~revise transcription panel output to output transcriptions based on canvas uri instead of index and sectioned out by scholarship record, when there are multiple~ (already complete)

PR #1006, opened 2022-08-08
WIP change Footnote.content to has_transcription (#1000, #1001)

blms:
## Todo

When TEI-to-annotation is done:
- [ ] Merge, and make this work with it
- [ ] Update fixtures

## Questions/dev notes

- I was wondering if we need to do anything to preserve the existing content in `content` fields in the migration, and convert them to annotations, or is that going to happen separately using the transcriptions stored elsewhere?
- Do the `content_text` and `content_html` make sense for Footnote? Trying to be backwards compatible to the old behavior when we stored both within Footnotes' `content`.
- Does the signal handler make sense? Trying to handle create, update, and delete all in one place.
- In the templates, I've just commented out the sections that use `content` for now. But I think in this branch I'll need to also do #999.


rlskoeser:
Answers to your questions:

> I was wondering if we need to do anything to preserve the existing content in `content` fields in the migration, and convert them to annotations, or is that going to happen separately using the transcriptions stored elsewhere?

No need to preserve — the html and text in the content field is generated from the TEI, and we're going to migrate the TEI to our new annotation format. 

> Do the `content_text` and `content_html` make sense for Footnote? Trying to be backwards compatible to the old behavior when we stored both within Footnotes' `content`.

Yes, thanks for adding this! We'll need methods to aggregate the annotation content for display and indexing, as you say.

Let's rename to `transcription_html` and `transcription_text` since we know eventually there will be translation equivalents.

> Does the signal handler make sense? Trying to handle create, update, and delete all in one place.

I wonder if we should split out indexing from the digital edition indicator? That way we only have to handle create and delete. I don't think we should be saving the footnote just to trigger a reindex.  

> In the templates, I've just commented out the sections that use `content` for now. But I think in this branch I'll need to also do [As a content editor, I want transcription edits displayed immediately on the document details so that changes and new content are all available to all site users. #999](https://github.com/Princeton-CDH/geniza/issues/999).

Yes, that seems reasonable to me; they are all related.

geniza/annotations/signals.py
```diff
@@ -0,0 +1,34 @@
...
+    # get other annotations on this canvas and source
+    other_annotations_exist = (
+        Annotation.objects.exclude(pk=instance.pk)
+        .filter(
+            content__target__source__id=canvas_uri,
+            content__contains={"dc:source": source_uri},
+        )
+        .count()
+        > 0
+    )
```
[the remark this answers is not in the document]
rlskoeser: Actually, since you're connecting to the `post_delete` signal, the deleted annotation isn't in the database so this definitely doesn't need the exclusion and we can just check with `.exists()`

We should only do this query if the signal is a delete.

rlskoeser (commented):
This is a really good start!

geniza/annotations/signals.py
```diff
@@ -0,0 +1,34 @@
+from geniza.annotations.models import Annotation
+from geniza.corpus.models import Document
+
+
+def update_footnote(instance, **kwargs):
+    """On Annotation save or delete, update related footnote's has_transcription field
+    and trigger a reindex of related documents (via footnote save)"""
+    created = kwargs.get("created")
+    deleted = created is None
```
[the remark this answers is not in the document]
blms: Nope, if it's `post_save` for updated, then `created=False` rather than `None`! I'll make a note there.

geniza/annotations/signals.py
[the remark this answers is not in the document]
blms: Sounds like we should remove this altogether if we're not going to do anything on deleting the last annotation, but I'll save this here in case we decide we need it:

```python
    # get other annotations on this manifest and source
    other_annotations_exist = (
        Annotation.objects.filter(
            content__target__source__partOf__id=manifest,
            content__contains={"dc:source": source_uri},
        )
        .exists()
    )
```

geniza/annotations/signals.py
```diff
@@ -0,0 +1,34 @@
...
+    )
+    # update footnote(s)
+    for doc in docs:
+        footnotes = doc.footnotes.filter(source__pk=source_pk)
+        for footnote in footnotes:
+            if deleted and footnote.has_transcription and not other_annotations_exist:
+                footnote.has_transcription = False
+            elif not footnote.has_transcription and created:
+                footnote.has_transcription = True
+            footnote.save()  # ensure footnote is saved regardless for reindexing purposes
```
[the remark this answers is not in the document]
blms: That makes sense to me. In that case, it seems we don't need to attach this on `post_delete`, unless we should just do something small like remove `DIGITAL_EDITION` from `doc_relation`?

For the log entry, I guess we can't get the user from the signal handler, so ok to just use `SCRIPT_USERNAME`?

blms:
@rlskoeser Tried implementing with the new doc relation instead of `has_transcription`, as well as simplifying the `post_save` signal handler as you recommended. No tests or anything like that yet, but just wanted to make sure it's looking good before moving onto that!

geniza/annotations/signals.py
```diff
@@ -0,0 +1,44 @@
...
+            footnote.doc_relation = [Footnote.DIGITAL_EDITION]
+            footnote.save()
+            LogEntry.objects.log_action(
+                # can't get user in a signal handler, so associate with script
+                user_id=User.objects.get(username=settings.SCRIPT_USERNAME).id,
+                content_type_id=ContentType.objects.get_for_model(Footnote).pk,
+                object_id=footnote.pk,
+                object_repr=str(footnote),
+                action_flag=ADDITION,
+            )
```
rlskoeser, line 44: Sorry I wasn't clearer about the logic I want! I need to write down the big picture somewhere asap, as much as I have it figured out; maybe when you review you can help me identify any gaps I need to still fill in.

The digital edition footnotes should be separate from the edition footnotes. For the sake of this signal handler, we can assume that the appropriate edition footnotes have already been converted to digital editions by the tei to annotation script. 

For this signal handler, I think we just need to make sure that a digital edition footnote for the given source + document exists.  We can't use `get_or_create` because that will create a duplicate footnote if any of the fields differ (e.g., a note documenting emendations by a second contributor).

Did you do away with the delete signal handler? I think we still need that.

geniza/corpus/models.py
```diff
@@ -677,11 +683,12 @@ def related_documents(self):
 
     def has_transcription(self):
         """Admin display field indicating if document has a transcription."""
-        return any(note.has_transcription() for note in self.footnotes.all())
+        return self.footnotes.filter(
+            doc_relation__contains=Footnote.DIGITAL_EDITION
+        ).exists()
```
rlskoeser, line 688: This was implemented on the local versions of the footnotes to avoid an extra db call, since this property is used in the admin list view. (Although there should have been a comment to that effect!)

Please revise to implement it without an extra db call and add a comment about avoiding the extra db call for admin list view

geniza/corpus/models.py
```diff
@@ -1041,12 +1045,12 @@ def _merge_footnotes(self, doc):
...
                     # if the new footnote has content, add it
-                    if footnote.content:
+                    if Footnote.DIGITAL_EDITION in footnote.doc_relation:
                         self.footnotes.add(footnote)
                     # if the partial match has no content, remove it
                     # (if it has any content, then it is different from the new one
                     # and should be preserved)
-                    if not equiv_fn.content:
+                    if Footnote.DIGITAL_EDITION not in equiv_fn.doc_relation:
                         self.footnotes.remove(equiv_fn)
```
rlskoeser, line 1055: do we need this check anymore? or will the existing equivalence check handle this case now that we're not comparing content?

geniza/corpus/templates/corpus/document_detail.html
```diff
@@ -138,9 +138,9 @@ <h3>
         {% if document.has_transcription and user.is_authenticated %}
             <section class="transcription-link">
                 {% for ed in document.digital_editions %}
-                    {% if ed.content.text %}
+                    {% comment %} {% if ed.content.text %}
                         <p><a href="{% url "corpus:document-transcription-text" document.pk ed.pk %}">Download {% for auth in ed.source.authorship_set.all %}{% include "snippets/comma.html" %}{{ auth.creator.last_name }}{% empty %}[unknown]{% endfor %}'s edition.</a></p>
-                    {% endif %}
+                    {% endif %} {% endcomment %}
                 {% endfor %}
```
rlskoeser, line 144: commented out because it doesn't work anymore?

geniza/corpus/views.py
```diff
@@ -530,7 +530,7 @@ def get(self, request, *args, **kwargs):
         digital_editions = document.digital_editions()
         # while sync transcription is in transition, we need to check for
         # html transcription content
-        digital_editions = [de for de in digital_editions if "html" in de.content]
+        digital_editions = [de for de in digital_editions if de.content_html]
```
rlskoeser, line 533: I don't think we need this check anymore, digital edition filter should be sufficient

rlskoeser:
I tried to see how this looks in django admin and found a couple of prefetch/filter queries that still reference the old footnote contents that cause 500 errors. Also, of course, nothing shows up when I filter because the current migration removes the content field. I think we might actually need to leave it in place (but not use it...) until the tei to annotation migration is complete! I'll work on writing up the migration details so we can figure out the steps & order.

geniza/corpus/models.py
blms: I'd actually updated the original check to compare the `content_text` property:

```py
if include_content:
    compare_fields.append("content_text")
```

So maybe the new check should be something like this?
```py
# if there's a partial match (everything but content)
if equiv_fn:
    # if the new footnote has content, add it
    if footnote.content_text:
        self.footnotes.add(footnote)
    # if the partial match has no content, remove it
    # (if it has any content, then it is different from the new one
    # and should be preserved)
    else:
        self.footnotes.remove(equiv_fn)
# if neither an exact or partial match, add the new footnote
else:
    self.footnotes.add(footnote)
```

Or should we get rid of the content checking entirely?

geniza/annotations/signals.py
blms: > Did you do away with the delete signal handler? I think we still need that.

I was asking about that [here](https://github.com/Princeton-CDH/geniza/pull/1006#discussion_r942861472), what should we do on delete if it's the only annotation?

geniza/corpus/templates/corpus/document_detail.html
blms: Right, there's no `content.text` now, so needs to be updated in the upcoming #999 work, but commented out temporarily while this is a WIP.

geniza/annotations/signals.py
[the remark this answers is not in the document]
rlskoeser: @blms if the last annotation for a document+source is removed, the digital edition footnote should be removed.  Think of the digital edition footnote as the convenience marker that a transcription exists, so we can quickly check and filter without querying for annotations every time.

Yes, makes sense to use script user for the log entry. Let's add a change message to clarify that the footnote was automatically created/removed based on annotation creation/deletion.

geniza/corpus/models.py
rlskoeser: There's no point checking content — transcription is linked to document+source, so if there is more than one digital edition for that document and source they would get the same content. 

It would be nice to have validation to prevent more than one digital edition for a given source being added to a document, but I'm not sure where to put it. Also, I think we want digital edition footnotes to be separate from other footnotes from the same source, if there are any. (I have to admit, this is making me question whether it's the right solution. But I don't think we can efficiently query for annotations in bulk every place we need to know if a document has digital editions or not.)

geniza/corpus/templates/corpus/document_detail.html
rlskoeser: this was a provisional solution, let's keep it commented out and mark as something to address later (we may remove it entirely in favor of the cut + paste we still need to implement)

Issue #943, opened 2022-07-12
As a content editor, I want to be able to search for tags with or without diacritics and get the same results.


mrustow:
## testing notes

- [x] Try the below search and confirm it works as expected
- [x] try adding a new tag with diacritics — tag should be saved with the diacritics removed
- [x] confirm existing tags have had diacritics removed
- [x] confirm tag variants with and without diacritics have been merged correctly 

## deploy notes

- [ ] solr reindex

---

On the admin interface > taggit, I typed hujra (no dot under the H) into the search bar and received this result:
<img width="1436" alt="image" src="https://user-images.githubusercontent.com/73319225/178540505-ac18a681-3d3a-467b-ab31-872eaacb5e5c.png">

However, I knew that #ḥujra is associated with 4 documents. So then I searched for ḥujra and received this result:
<img width="1443" alt="image" src="https://user-images.githubusercontent.com/73319225/178540775-8976e188-818a-4fe4-b41f-c126e12019cd.png">

Lots of our tags don't use proper diacritics, so it would be better for us to be diacritic-agnostic.

Which we do seem to be already in the document view in the admin interface, and in searching the public site. 

Thanks! 🥇

rlskoeser:
@richmanrachel is it ok for tags to not allow diacritics? wondering if we can tackle this in tandem with #499 — collapse tags that are the same except for diacritics, and then add logic so that in future diacritics are ignored when creating tags.

richmanrachel:
@rlskoeser - yes! Especially since Marina already requested this, I'm happy with this solution

richmanrachel:
This didn't quite work either. The slug correctly took out diacritics, but the search doesn't include the slugs I guess? no results for mualim but returning correct result with diacritics:
<img width="451" alt="image" src="https://user-images.githubusercontent.com/68028846/184223530-c742c07c-b0eb-4ff4-92f7-f0f08dfa59ce.png">

blms:
Ah—this is a problem with the "add tag" functionality in the Tags section of the admin, rather than adding a tag to a document. Looks like our diacritic-stripping code is not run when you create a tag separately!

richmanrachel:
I didn't read through all the tags to ensure that all of the diacritics were properly dealt with, but I did look over about 400. The only potential error I saw was this one:
<img width="417" alt="image" src="https://user-images.githubusercontent.com/68028846/184896889-8c5d8ad3-febd-4363-a61e-b0460891e625.png">

[...]

Issue #499, opened 2022-01-06
As a content editor, when I search for tags to add to a document I want the search to ignore case so that I don't create variations of the same tag.


kmcelwee:
## testing notes

- [x] In the admin, find a tag that exists with capital letters, then add that tag—but all lowercase—to a document, and save, and confirm the original tag is added (rather than a new tag being created)
- [x]  confirm tags that previously existed in upper case and lower case variants (e.g., India and india) have been merged correctly (only one tag, all documents associated with that tag, log entry on the tag history documenting the merge)

---

We need to set `TAGGIT_CASE_INSENSITIVE = True` in `base.py` but because tags have already been created, we'll need to collapse existing tags  with a migration before turning this on. A quick script showed that 200 tags were duplicated when made case insensitive, so this shouldn't be done by hand. This work was started in #257.

richmanrachel:
I don't think the case sensitivity is totally working, it recommended I add the correct tag but ended up allowing me to save both versions:
<img width="198" alt="image" src="https://user-images.githubusercontent.com/68028846/184222182-877bc3e9-f4f9-4c54-a232-c3b9967a2e95.png">

rlskoeser:
@richmanrachel ah, interesting — this maybe a change since Ben wrote up the testing note. I didn't want to force all tags to lower case — for a lot of tags that would result in a loss of information (especially for proper names). 

Does it sound right to you that we should normalize the lowercase version to match the uppercase one if there is already a tag? We don't want to end up with tag variants like this again.

richmanrachel:
@rlskoeser - yes, I agree that we should normalize the lowercase to match uppercase!

blms:
@rlskoeser I thought that's actually what `TAGGIT_CASE_INSENSITIVE` was supposed to do!

https://github.com/jazzband/django-taggit/blob/eb66e6607bb2997ce63e58c59fdec0df5a474a6b/taggit/managers.py#L220-L231

rlskoeser:
@blms thanks for chiming in. I was surprised too. Any idea what's going on? I also see that there are tags that should have been merged but weren't (Formulary / formulary) and the new tag Rachel created didn't get the diacritic stripped — but I tested and `unidecode("mu'ālim")` does return `mu'alim`

blms:
> I also see that there are tags that should have been merged but weren't (Formulary / formulary)

This looks right to me, unless one was deleted since you posted this!

[...]

PR #1007, opened 2022-08-08
Make tags case- and diacritic-insensitive (#499, #943)

blms:
## In this PR

- Per #499:
  - Make tags case-insensitive, apply `lower()` to tags on creation, and rename/merge all tags to their lowercase variants in a data migration
- Per #943:
  - Rename/merge all tags to their ASCII variants in a data migration, and apply `unidecode` to tags on creation

## Notes

- I added a deploy note for reindexing, as all the documents with changed tags will need to be reindexed to make the changes searchable.


geniza/corpus/migrations/0034_merge_duplicate_tags.py
```diff
@@ -0,0 +1,88 @@
...
+    doc_contenttype = ContentType.objects.get(app_label="corpus", model="document")
+    tag_contenttype = ContentType.objects.get(app_label="taggit", model="tag")
+    (user, _) = User.objects.get_or_create(
+        username=settings.SCRIPT_USERNAME, is_staff=False, is_active=False
+    )  # generic user for script execution
+    tags_to_delete = []
+    tagged_items_to_delete = []
+
+    for tag in Tag.objects.all():
+        lower_ascii_name = unidecode(tag.name).lower()
```
rlskoeser, line 30: I would prefer not to _force_ lower case, but just to use the taggit case insensitive logic to avoid case-collisions. We should preserve the mixed-case tag where making case-insensitive results in duplicates. So logic instead should be something like:

- check if `unidecode` changes the current tag; use that as basis for new tag if changed
- do a case-insensitive search for any tags that should be merged with the current one; choose the mixed case variant if there is one
- for any tag that is being removed due to collapse, do a tagged items queryset update to replace the old tag id with the merged tag id (should not need to change anything else)
- remove the redundant tags

It might be easier to aggregate the tags first rather than iterating on`Tag.objects.all()`; then you don't have to worry about what order you see them in. But perhaps it is sufficient to sort by tag name, since you should get the upper case variants first.

geniza/corpus/migrations/0034_merge_duplicate_tags.py
```diff
@@ -0,0 +1,88 @@
...
+                    )
+                    # log tag merge on document
+                    LogEntry.objects.log_action(
+                        user_id=user.id,
+                        content_type_id=doc_contenttype.pk,
+                        object_id=document.pk,
+                        object_repr=str(document),
+                        change_message=f'replaced tag "{tag.name}" with "{lower_ascii_name}"',
+                        action_flag=CHANGE,
+                    )
```
rlskoeser, line 57: log entries should only be added to tags, not documents

rlskoeser (changes requested):
Sorry I wasn't clearer in my notes — I don't want to force lower case since that will result in a loss of information and readability (in the admin interface, at least), especially for names.

geniza/corpus/migrations/0034_merge_duplicate_tags.py
blms: @rlskoeser I made good progress on this, but I was wondering if you could help me with the aggregation query. The problem I'm running into when iterating on `Tag.objects.all()` is that, if there is more than one variant of a tag name in terms of diacritics, the ones alphabetically after the first one aren't picked up for merge. Maybe it's just a problem with my query for finding the tags to merge:

```python
for tag in Tag.objects.all().order_by("name"):
    if tag.pk not in tags_to_delete:
        ascii_name = unidecode(tag.name)

        # search for any differently-cased (incl. unidecoded) versions of this tag
        tags_to_merge = Tag.objects.filter(
            Q(name__iexact=ascii_name) | Q(name__iexact=tag.name)
        ).exclude(pk=tag.pk)
```

That's only going to pick up other tags with the exact same diacritics, or tags that are the same but in ASCII. If the first encountered tag does not have diacritics, but the second does, then it won't pick up the second one.

The problem appears when I then go to save the tag with its unidecoded name, leading to a uniqueness violation when the second variant is reached (in the data I have, "Abu Sahl Levi" is picked up first, then "Abū Sahl Levi" is the second variant):
```python
        # rename this tag to unidecoded name
        if ascii_name != tag.name:
            tag.name = ascii_name
            tag.save()
```

My idea was to skip over all the merged tags in the loop by adding them to `tags_to_delete`:

```python
        # merge all other matched tags into this one
        for matched_tag in tags_to_merge:
            # replace the matched tag in all tagged items with this tag
            matched_tag.taggit_taggeditem_items.update(tag=tag)
            # finally, mark matched tag for deletion
            tags_to_delete.append(matched_tag.pk)
```

In this case "Abū Sahl Levi" never gets added to `tags_to_delete` because it wasn't picked up in the merge queryset.

Is there a way we can write an aggregation query that will handle all diacritic variants of a tag?

rlskoeser: I think it might be easier to do the aggregation first in python! That's what I meant when I said "aggregate the tags first" but I see now that isn't clear. Let me take a quick stab at this.

PR #1013, opened 2022-08-15
Unidecode all tags on pre_save signal (#943)

blms:
## In this PR

- Per #943:
  - Adds `pre_save` signal to `taggit.Tag` to ensure `unidecode` is run on the `name` field of model instances created manually, not just on related save of a tagged instance of another model.


rlskoeser (approved):
looks good! thanks for figuring out where the problem was and documenting it clearly in the comments

Issue #1023, opened 2022-08-22
not possible to edit recto/verso for fragments without images


rlskoeser:
## testing notes

in admin site, edit a document associated with a fragment that doesn't have iiif images available
- [x] you should see placeholder recto/verso images
- [x] can select placeholder images to edit recto/verso

--- 
As reported by Alan on Slack, there's no longer a way to edit recto/verso information for documents on fragments without images, since we made the edit interface dependent on images.  (We haven't lost the existing recto/verso information, but it can't be updated.)

Not sure how much of a priority this is yet; in terms of the web application and the public site, we're primarily using this information for image display so it doesn't have a big impact on documents without images, but I think we do include it in the data export.

rlskoeser:
Alan responded that this is important to do so content editors can keep side information when editing descriptions, rather than risking things getting out of sync.

@blms let's do something simple, maybe we can just do two placeholder images in the fragment admin thumbnails when there are no images available so that the selection logic can be used as-is? (I don't remember if we're showing fragment admin thumbnails elsewhere, but I think we only want these on the fragment inline on the document change form.)

richmanrachel:
Worked! Closing!

PR #1027, opened 2022-08-23
Allow editing side info when no IIIF images present (#1023)

blms:
## In this PR

- Per #1023:
  - Use placeholder SVG images when no IIIF image available, to allow editing side info with the same UI


rlskoeser (approved):
Looks reasonable to me.

Issue #1011, opened 2022-08-10
As a content editor, I want to override the order images are displayed for a document so that I can set the images to display in logical order for joins.


rlskoeser:
## testing notes

In the QA admin site, edit a document with images.
- [x] In the field labeled "image order override", click and drag the images around to reorder them. Confirm this interaction works as expected
- [x] Save the document with a changed order and confirm the new order is persisted in the admin
- [x] View the document in the public site and confirm the order is shown correctly there as well
- [x] Try the same on a document joining multiple fragments each with images, and confirm all of the above still works

---

## dev notes
- [x] add an override image order field to `Document` (array of url fields)
- [x] add custom logic in document admin form to display all thumbnails and add javascript to drag and drop to re-order; when sorted, populate override array field with the canvas ids in order; reset option to revert back to no override
- [x] revise `Document.iiif_images` to use override order when set (should update indexing, document details, and sync transcription)
- [x] revise `DocumentManifestView` to use override order when set (maybe update to use `iiif_images`)

kseniaryzhova:
@rlskoeser it all works for me!

richmanrachel:
@rlskoeser - the re-arranging works, but the text does not adjust to correct page even when I made the cover pages not part of the document in the related fragment section: https://test-geniza.cdh.princeton.edu/en/documents/9117/

[This was the same example I showed yesterday in the meeting where JTS added the cover note about it being the autograph of a famous poet. The transcription is still sticking to the cover pages rather than the manuscript]

rlskoeser:
@richmanrachel text won't adjust until the sync transcription runs again. Do you want me to run it on a specific document or two so you can verify? (probably good to check, the tei migration will use the same logic)

richmanrachel:
@rlskoeser - I see. I just updated the page and it's working! I think we can close - maybe we can just add a note when we release this feature that you will need to wait on the hourly sync for everything to work properly?

rlskoeser:
@richmanrachel ah, thanks for checking back! Glad we set that script to run more frequently. That makes sense to me! 

It would be good if you could think of any high priority documents where you want to have research assistants review and adjust the order before we run the TEI to annotation migration — that will be a one-time migration when we run it for real in production, and we should honor any overrides that are set, but if they are not set then you'll have to override order and reassociate content after the migration.

richmanrachel:
[...]

PR #1032, opened 2022-08-24
Allow reordering document images in admin (#1011)

blms:
## In this PR

- Per #1011:
  - Add new ArrayField of URLFields, `image_order_override` (naming ok?), to Document
  - Refactor `Fragment.iiif_thumbnails` into two methods, one to generate the html itself, and one to call the html generation function correctly depending on whether iiif images are available or placeholders are needed
  - In a new method `Document.iiif_thumbnails`, use the new `Fragment.html_thumbnails` (naming ok? makes sense as a static method?) method to generate thumbnails for all of the Document's IIIF images for reordering purposes
  - Display those thumbnails in the Document admin, add new JS to allow drag and drop reordering
  - Add hidden field to Document admin form that is populated via JS reordering
  - Update `Document.iiif_images`, `DocumentManifestView` to account for order override

## Questions

- Should it be possible to reorder placeholders, or is that just for tracking side information? I assumed the latter (and thus don't show placeholders in `Document.iiif_thumbnails`) but wanted to make sure, especially since we could be dealing with joins that have some fragments without images.
- Does it make sense to deal with fragments added after reordering [like this](https://github.com/Princeton-CDH/geniza/blob/4b5ed7ce7a167a7e6e9aa1a30c98fdb6963cf294/geniza/corpus/models.py#L684-L687)? Or is that not even a case we need to consider?
- Is there a simpler way to find the first matching dict in a list in Python? I feel like this is a lot of list comprehension 😆
- How is the UX for reordering? I tried to do something similar to [this](https://codepen.io/macro6461/pen/RwWemgM), but I'm wondering if it might be confusing, e.g. it may appear that you are "swapping" two images when you are actually just placing one in a different position and shifting the other. But not sure—it's intuitive to me!


geniza/corpus/models.py
```diff
@@ -666,8 +669,38 @@ def iiif_images(self, filter_side=False):
...
+                if matches:
+                    iiif_images.append(unordered_images.pop(matches[0]))
+            # ensure images not present in the order override are still added at the end, e.g. if
+            # fragments are added to this document after order override was set
+            for remaining_images in unordered_images:
+                iiif_images.append(remaining_images)
+
         return iiif_images
 
+    def iiif_thumbnails(self):
```
rlskoeser, line 691: here too, maybe admin thumbnails would be clearer / more explicit

geniza/corpus/models.py
```diff
@@ -457,6 +457,9 @@ class Document(ModelIndexable, DocumentDateMixin):
     fragments = models.ManyToManyField(
         Fragment, through="TextBlock", related_name="documents"
     )
+    image_order_override = ArrayField(
+        models.URLField(), null=True, verbose_name="Image Order"
```
rlskoeser, line 461: Let's add some help text here. Maybe something like this?
> Drag image thumbnails to customize order when necessary (i.e. image sequence does not follow fragment/shelfmark sequence)

geniza/corpus/models.py
```diff
@@ -666,8 +669,38 @@ def iiif_images(self, filter_side=False):
...
+                    idx
+                    for idx, img in enumerate(unordered_images)
+                    if img["canvas"] == canvas
+                ]
+                if matches:
+                    iiif_images.append(unordered_images.pop(matches[0]))
+            # ensure images not present in the order override are still added at the end, e.g. if
+            # fragments are added to this document after order override was set
+            for remaining_images in unordered_images:
+                iiif_images.append(remaining_images)
```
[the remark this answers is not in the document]
rlskoeser: IDK how likely it is to happen — most likely case is I think if we add new iiif images after overriding order. This seems like a reasonable solution to me (and simpler than trying to catch and update the order elsewhere when images are added)

geniza/corpus/models.py
```diff
@@ -666,8 +669,38 @@ def iiif_images(self, filter_side=False):
...
+            unordered_images = iiif_images
+            iiif_images = []
+            for canvas in self.image_order_override:
+                matches = [
+                    idx
+                    for idx, img in enumerate(unordered_images)
+                    if img["canvas"] == canvas
+                ]
+                if matches:
+                    iiif_images.append(unordered_images.pop(matches[0]))
```
rlskoeser, line 683: What about making a dict of images keyed on canvas uri, and then adding them to the list based on uris in override order? You can `pop` values from a dict, so then you could still add the remaining canvases.

Could maybe even do this when initially constructing the list of images from fragments — create a dict with canvas uris as keys, return values if no order override (and dict order preserved as of python 3.7 so it should do what we want). Retrieving by uri would need to handle missing fragments in the case of filtering non-document images

rlskoeser (approved):
This is working really nicely, the drag and drop interaction is exactly what I would expect.

A few suggestions for renaming methods for clarity and an idea for improving the canvas sorting logic. Will answer your other questions separately.

rlskoeser:
> * Should it be possible to reorder placeholders, or is that just for tracking side information? I assumed the latter (and thus don't show placeholders in `Document.iiif_thumbnails`) but wanted to make sure, especially since we could be dealing with joins that have some fragments without images.

This actually just occurred to me in the context of the the tei to annotation migration, because we're making our own canvas uris where we don't have enough images. I don't think the recto/verso placeholders need to be included in the sort but canvases with transcription attached may need to be (eventually).

> * Does it make sense to deal with fragments added after reordering [like this](https://github.com/Princeton-CDH/geniza/blob/4b5ed7ce7a167a7e6e9aa1a30c98fdb6963cf294/geniza/corpus/models.py#L684-L687)? Or is that not even a case we need to consider?

Seems reasonable! Thanks for thinking of it.

> * Is there a simpler way to find the first matching dict in a list in Python? I feel like this is a lot of list comprehension 😆

Does my suggestion make sense and simplify things? Can sketch out code if unclear.

> * How is the UX for reordering? I tried to do something similar to [this](https://codepen.io/macro6461/pen/RwWemgM), but I'm wondering if it might be confusing, e.g. it may appear that you are "swapping" two images when you are actually just placing one in a different position and shifting the other. But not sure—it's intuitive to me!

Great! Very intuitive.

blms:
> This actually just occurred to me in the context of the the tei to annotation migration, because we're making our own canvas uris where we don't have enough images. I don't think the recto/verso placeholders need to be included in the sort but canvases with transcription attached may need to be (eventually).

I think the placeholder canvases will be easy enough—though I think it would be a problem if the placeholder gets replaced with actual IIIF images and thus the canvas URI changes. Then I guess all documents with those fragments need to get their `image_order_override` updated? 

This also makes me think we should plan for what should happen if one of the fragments gets _removed_ from the document after the order is set. Maybe remove their canvas URIs from `image_order_override` for all associated documents?

Many edge cases here with duplicating data from elsewhere in an array field like this, where it needs to be relatively synchronized!

blms:
@rlskoeser oops, forgot to tag you above

rlskoeser:
> I think the placeholder canvases will be easy enough to add at some point, especially for the case you mentioned—though I think it would be a problem if the placeholder gets replaced with actual IIIF images and thus the canvas URI changes. Then I guess all documents with those fragments need to get their `image_order_override` updated?

I agree, can probably add placeholder canvases here if/when needed. 

If new iiif images are added when we already have a transcription attached to placeholders, I think a content editor will need to go in and move the transcription blocks to the appropriate image — so I guess we'll need a way to remove placeholders? Or maybe if we do things right they just go away when there are no annotations attached to them?

> This also makes me think we should plan for what should happen if one of the fragments gets _removed_ from the document after the order is set. Maybe remove their canvas URIs from `image_order_override` for all associated documents?

I think the missing uri would be ignored if it isn't present in the actual canvases, so nothing needs to happen. The next time they edit the document will it get updated properly? Or can we tweak the logic so it does?

blms:
> I think the missing uri would be ignored if it isn't present in the actual canvases, so nothing needs to happen. The next time they edit the document will it get updated properly? Or can we tweak the logic so it does?

You're right, it's already working this way!

geniza/corpus/models.py
blms: Ended up needing to override the `ModelAdmin.get_form` method to get this to show up in the form, since I use a read-only field based on the `admin_images` property for the actual reordering, while the `image_order_override` field is a hidden input.

geniza/corpus/models.py
blms: Refactored along those lines, let me know if I missed anything!

rlskoeser (approved):
refactor looks great!

Issue #1038, opened 2022-08-25
As a transcription editor I want to edit and rearrange transcription content as numbered lists so that I can correct line wrapping introduced to match printed editions.


rlskoeser:
## testing notes
- [ ] TEI to html conversion should preserve line numbers from the TEI (unnumbered lines are being converted to paragraphs; lines are converted to ordered lists; line numbers starting with something other than 1 are converted to order list with a start attribute)
- [x] in the transcription editor, the lines should display and be editable as a regular ordered list
- [ ] test editing a wrapped unnumbered line to clean up the formatting (here's one example: https://test-geniza.cdh.princeton.edu/en/documents/646/transcribe/228/ )

## dev notes
- [x] revise tei to html conversion to generate ordered lists (with start attributes when needed) so they can be edited in TinyMCE (done as part of tei to annotation conversion)
- [x] revise list styles for transcription to remove counter and display the actual marker element from the ordered list

revisions after testing:
- update public image+transcription viewer margins to handle new list styles
   - [x] display full number two-digit numbers
   - [x] when scrolling to second transcription, numbers should not show on the left
   - [x] style unnumbered paragraphs between lines so they aren't as jarring

kseniaryzhova:
@rlskoeser @richmanrachel and I get the second number cut off in lines that have more than 1 digit - so we just see a bunch of 1s and 2s for lines in the 10s and 20s. 
![image](https://user-images.githubusercontent.com/105385128/187287664-ce3c2fe8-43ca-4272-a689-0f97b2434429.png)
https://test-geniza.cdh.princeton.edu/en/documents/646/

kseniaryzhova:
@rlskoeser we're not seeing the unnumbered line for the document you linked (the one of which I took a screenshot above). What should we do?

rlskoeser:
@kseniaryzhova I meant the text between lines 3 and 4 and between lines 19 and 20

richmanrachel:
@rlskoeser - it looks like the editor itself is handling the wrap text correctly (here's the same doc as in Ksenia's pic above):
<img width="582" alt="image" src="https://user-images.githubusercontent.com/68028846/187301570-0b0748c7-d030-43c3-9732-50da98bff9de.png">
I agree that it looks a little funny on the viewer side to not have parts of the same line indented.

richmanrachel:
Line numbers are not showing up in the search results anymore (which was a really useful feature so I knew where in the document to look):
[...]

PR #1043, opened 2022-08-29
Revise ordered list style to use actual line numbers (#1038)

blms:
## In this PR

- Per #1038:
  - Remove counter from `<ol>` lists in transcription content, and use the decimal list style, so that line numbers respect the HTML start attribute
  - Adjust margin/padding to adjust for line number marker pseudo-elements


rlskoeser (approved):
this is exactly what I wanted, thank you

Issue #1037, opened 2022-08-25
update annotation controller so deep zoom doesn't autoscroll the page when in edit mode


rlskoeser:
## testing notes (QA)

When navigating to the transcription editor for a document (QA public site document details page --> "edit X's transcription"/"add a new transcription", visible when logged into admin):
- [x] Confirm that the page does not immediately scroll itself down to the last image on the document

---
## dev notes

at minimum, we should disable the autoscroll behavior needed for the image+transcription panel viewer

if possible, it would be great if we could do some lazy-loading, maybe only initialize zoom and annotation editing when images are visible?

rlskoeser:
@blms what if instead of enabling zoom on page load we activate it based on intersection observer?
https://stimulus-use.github.io/stimulus-use/#/use-intersection

blms:
@rlskoeser That makes sense to me!

kseniaryzhova:
@rlskoeser It does not immediately autoscroll to the bottom BUT it's still not immediately showing me the existing transcription. I have to refresh again after clicking on th Edit transcription button. Are you still tiling? Let me know how you want to proceed - the slow load time seems like a separate issue so happy to close this one and write a new bug report.

blms:
@kseniaryzhova Yes, if this one is working then please close and create a new bug report for the slow load time. It might be (and this is my hunch) that they are not loading at all rather than loading slowly.

kseniaryzhova:
@rlskoeser works for us!

PR #1047, opened 2022-08-30
Use composition instead of extension for iiif+annotation controllers

rlskoeser:
Revised the iiif & annotation controllers and bindings so that we can use composition instead of extension — main pain point I want to resolve is the doubling all of the bindings in the template, but I think this will help in other ways too.

Thinking of this as a starting point / pre-step for fixing #1037


blms (approved):
Looking great! I think what needs to change for the `handleZoomSliderInput` function is that these three lines in `IIIFController`:

```py
this.resetBounds(viewer);
this.deactivateDeepZoom();
deactivating = true;
```

Should only be run when we're not using the annotation controller. Not sure what's the best approach for that; maybe another property on `this.element`?

rlskoeser:
> Looking great! I think what needs to change for the `handleZoomSliderInput` function is that these three lines in `IIIFController`:
> 
> ```python
> this.resetBounds(viewer);
> this.deactivateDeepZoom();
> deactivating = true;
> ```
> 
> Should only be run when we're not using the annotation controller. Not sure what's the best approach for that; maybe another property on `this.element`?

I think we want to make it a configurable option of the controller (and we'll probably have another one for the auto-scroll behavior).

Can you remind me, what's the logic for calling this / not calling this? Is this to deactivate zoom when the zoom slider goes to minimum?

blms:
> Can you remind me, what's the logic for calling this / not calling this? Is this to deactivate zoom when the zoom slider goes to minimum?

@rlskoeser Exactly! Which should not happen on the transcription editor (I think).

rlskoeser:
> > Can you remind me, what's the logic for calling this / not calling this? Is this to deactivate zoom when the zoom slider goes to minimum?
> 
> @rlskoeser Exactly! Which should not happen on the transcription editor (I think).

I think that's right. So, maybe a controller setting to deactivate on min-zoom that is on by default and we turn off in the annotation controller?

blms:
@rlskoeser I seem to be running into the "scroll OSD parent element on focus" bug consistently, where it covers up the controls when you start interacting with the canvas. Are you experiencing that too? I know Kseniya mentioned it too. I thought we'd squashed that a while ago…

rlskoeser:
> @rlskoeser I seem to be running into the "scroll OSD parent element on focus" bug consistently, where it covers up the controls when you start interacting with the canvas. Are you experiencing that too? I know Kseniya mentioned it too. I thought we'd squashed that a while ago…

I haven't noticed that one, no. I'm usually looking at it on my large monitor, though; and as I recall, that one is more problematic / noticeable with longer images.

Is there a way we can put the controls inside the OSD parent element so that we don't have to fight the default behavior? Or something similar, maybe extend the padding/margin on the OSD parent element so it covers our controls?

blms:
@rlskoeser Done, thanks! I think we had actually already fixed this at some point with OSD being attached to the `osdTarget` but it must have gotten clobbered in a merge, which is understandable as those merges have been nightmarish.

Speaking of merges, no clue why this one test is suddenly failing after the merge. Looks like it started failing when `develop` was merged into `prototype/transcription`. Does this order matter or should we just rewrite the test so that the order isn't taken into account?

rlskoeser:
> Speaking of merges, no clue why this one test is suddenly failing after the merge. Looks like it started failing when `develop` was merged into `prototype/transcription`. Does this order matter or should we just rewrite the test so that the order isn't taken into account?

It looks like order isn't significant there, let's rewrite so it's order-agnostic

sitemedia/js/controllers/annotation_controller.js
```diff
@@ -1,64 +1,68 @@
...
+            !this.imageContainerTarget.querySelector(".openseadragon-container")
+        ) {
+            // load configuration variables from django settings; must use DOM query due to json_script
+            const config = JSON.parse(
+                document.getElementById("annotation-config").textContent
+            );
 
-        // TODO: Better way of determining if we're on mobile?
-        const isMobile = window.innerWidth <= 900;
+            // TODO: Better way of determining if we're on mobile?
```
rlskoeser, line 27: not urgent / immediate, but the todo made me think: we could think about using a media query for this instead; that would probably be more semantic, and it looks like you can watch for changes if you want

rlskoeser:
thanks for finishing what I started on this one, let's get it merged

PR #1050, opened 2022-08-31
Index transcription as html

rlskoeser:
index transcription as html instead of plain text

still needs tests and styling revisions

### DIFF
--- a/geniza/corpus/templates/corpus/snippets/document_result.html
+++ b/geniza/corpus/templates/corpus/snippets/document_result.html
@@ -59,7 +59,7 @@
                     {% elif document.transcription %}
                         {# otherwise, display truncated transcription #}
                         {# NOTE: might be nice to display N lines instead of using truncatechars #}
-                        <div class="transcription" lang="{{ lang }}">{{ document.transcription.0|linebreaks|truncatechars_html:150 }}</div>
+                        <div class="transcription" lang="{{ lang }}">{{ document.transcription.0|safe|linebreaks|truncatechars_html:150 }}</div>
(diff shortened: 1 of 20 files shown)


geniza/corpus/templates/corpus/snippets/document_result.html
```diff
@@ -55,11 +55,11 @@ <h2 class="title">
                     {# use first language code, if any; otherwise empty lang attribute to indicate unknown language #}
                     {# NOTE: this is problematic for documents with multiple primary languages #}
                     {% if document_highlights.transcription %}
-                        <div class="transcription" lang="{{ lang }}">{% for snippet in document_highlights.transcription %}{{ snippet.strip|safe|linebreaks }}{% endfor %}</div>
+                        <div class="transcription" lang="{{ lang }}">{% for snippet in document_highlights.transcription %}{{ snippet.strip|safe }}{% endfor %}</div>
```
blms, line 58: @rlskoeser I'm running into issues here when searching on transcription text; the highlighted snippet seems to be closing `li` tags that it hasn't opened, which breaks the search results `ol`. At least, that's my best guess at what's happening; my search term was "הן" if you want to try to replicate.

rlskoeser: ooh, you know I wondered if that might come up but didn't test thoroughly enough to run into it...  let me look into options for dealing with it

blms (approved):
Changes look great, and work well. Good call on the alignment in annotation editor; I also like your decision for the paragraphs matching alignment of list items.

PR #1053, opened 2022-09-01
Count digital editions as editions when indexing

rlskoeser:
Rachel flagged this in testing, IDK if we have an issue anywhere that would make sense to test it  — in some cases after the migration we only have a digital edition footnote and no editions, and in the search results those records display as having no transcriptions, which is obviously false. This is a simple fix to make sure we count digital editions when tallying up editions.


blms (approved):
Great!

Issue #1044, opened 2022-08-30
As a content editor, I want to add and edit transcriptions for records without all IIIF images available so that transcriptions aren't limited to records with all images.


rlskoeser:
## testing notes (QA)

(these notes are similar to #1039, but refer to the transcription editor and not the public viewer)

Navigate to the QA public site document details page for one or more documents with: 
- no images, but transcription content (e.g. PGPID 6176)
- no images and no transcription content (e.g. PGPID 33472)
- both, with some but not all images (i.e. some transcription content extends past the images). (e.g. PGPID 1390)

In the transcription editor ("edit X's transcription"/"add a new transcription", visible when logged into admin):
- For a document with no images and no transcription:
  - [x] confirm that you see two gray blank rectangles as placeholder images where images would go
  - [x] shift+click on them to create new annotations on the image, should result in new transcription blocks in the right column
- For a document with transcription content but no images:
  - [x] Confirm you see placeholder images for each block of transcription
  - [x] Edit and save some transcription text
- For a document with transcription content and some but not all images:
  - [x] Confirm you see placeholder images for each block of transcription that extends past the number of images present

## dev notes
- [x] load transcription editor even if document has no images (ex. PGPID 6176)
- [x] display placeholder canvases in editor of document has no images (maybe recto/verso, canvas 1 and 2 ?)
- [x] use placeholder canvases for documents with some but not all images, ex. PGPID 1390

may have some overlap with #1039

kseniaryzhova:
@rlskoeser PGPID 3185 is the better tester for incomplete images compared to the transcription - 1390 has all images (or at least does not have any gray squares)

Issue #1039, opened 2022-08-25
As a user, I want to see all transcription content for a document even if it extends beyond the currently available iiif images.


rlskoeser:
## testing notes (QA)

(these notes are similar to #1044, but they refer to the public viewer and not the transcription editor.)

Navigate to the QA public site document details page for one or more documents with: 
- no images, but transcription content (e.g. PGPID 6176)
  - [x] Confirm you see two gray blank rectangles as placeholder images for each block of transcription
- transcription content, and some but not all images (i.e. some transcription content extends past the images). (e.g. PGPID 1390)
  - [x] Confirm you see placeholder images for each block of transcription that extends past the number of images present

## dev notes
Transcription migration is creating placeholder canvases as needed for the content (either document has no images or document has not enough images), but we don't currently display them in the image + transcription viewer and we need to.

For documents with transcription but no image, we should not display the image panel on the public site (i.e., should remain disabled).

- [x] get unique set of canvas URIs from images + transcriptions in view
- [x] some but not all image URLs: display placeholder canvases; example PGPID 1390
- [x] canvas IDs, but no image URLs: no images, no image panel; ex PGPID 6176
  - [x] should be able to remove separate html section for transcription without images

see also: #1044 load editing for transcription with no iiif images

- [x] for the public site, make the placeholder completely transparent

rlskoeser:
I had a thought about this — maybe we just gather the set of all canvas ids in the view first, based on any iiif images (with order override) and then any canvases referenced by annotations not in that set.

It occurred to me we need to make sure we handle something like reimporting iiif manifests with new canvas ids (as I think we will do with the JTS content after deskewing images and reingesting) without losing any transcription content — I think this implementation would display empty (or old/duplicate) canvases until we cleanup and reassociate, but all the content would be there.

kseniaryzhova:
@rlskoeser I'm not sure I understand the second point. I went to PGPID 1390 but I don't see placeholders, just the images for the transcription (out of order because it's an unbound notebook). Are placeholder images supposed to be other Geniza documents or is it the same gray blank squares?

blms:
@kseniaryzhova I got an email with your comment about PGPID 1390 but the comment seems to have been deleted!

My mistake; that PGPID was edited between my copy of the data and what's live on the test site. You're right that there are no placeholders for it now. Can you check the one you mentioned in the other issue, PGPID 3185, instead? (I see you checked it off in the list so maybe you already did.)

[...]

PR #1066, opened 2022-09-09
Handle missing or partial IIIF content in transcription viewer/editor (#1039, #1044)

blms:
## In this PR

- Per #1039:
  - Pass all canvas IDs from both images and transcription to document detail view context
  - Show placeholder image in viewer when there are some canvases without IIIF images
  - When there are no IIIF images, hide image panel as before
- Per #1044:
  - Allow transcription and OSD/Annotorious/tahqiq editing on placeholder images
  - Show placeholder images in the editor even if there are no IIIF images, so that new annotations can be created
- Address a couple of small display issues (these keep coming up 😅):
  - Fix zoom percentage displayed incorrectly in slider track (became obvious with max zoom of 2.0)
  - Fix truncated height of OSD container (the original solution _only_ worked when an image was taller than the viewport, but the new solution picks the minimum between our truncated height and the image's original height)

## Questions
- Do we actually want placeholder images to show up in the editor if there are no IIIF images? Or do we just want to be able to edit the text?
- Is the general approach to the placeholder ok? I basically wanted as little changed logic in the template as possible, so I just replicated the basic structure of what `iiif_images()` returns, but with placeholder content and the canvases from the transcriptions.
- What should happen when there are multiple transcriptions, and one of them needs placeholder images but the other doesn't? This is a very rare edge case; PGPID 9121 and PGPID 2691 are the only examples of this I could find.
- Is this single placeholder image OK? It's public-facing in some cases, so I wonder if Gissoo should at least weigh in. I also figured it doesn't make sense here to try and determine which are recto/verso/etc, but maybe I'm wrong.

Good test PGPIDs: 1390, 6176, 848, 4495, 9121, 2691

---

## todo

- ~rm logic to show placeholders in transcription when no iiif~
- [x] solid color with no text (change on light/dark) for placeholder


blms:
@rlskoeser That huge template diff is mostly just unindentation, I hope there's an easier way for you to view it! Maybe side by side with the original would be best.

rlskoeser:
responding to your questions first

>     * Do we actually want placeholder images to show up in the editor if there are no IIIF images? Or do we just want to be able to edit the text?

I don't feel strongly about this either way; what matters is that the text is editable, if it's easiest for them to show up that's fine with me. The transcription editor can always toggle off the display of the image column in the ITT viewer. Did you want to propose something here?

>     * Is the general approach to the placeholder ok? I basically wanted as little changed logic in the template as possible, so I just replicated the basic structure of what `iiif_images()` returns, but with placeholder content and the canvases from the transcriptions.

Yes, approach looks good. 

>     * What should happen when there are multiple transcriptions, and one of them needs placeholder images but the other doesn't? This is a very rare edge case; PGPID 9121 and PGPID 2691 are the only examples of this I could find.

Ah, right — the editions may differ but we only display one set of images. I think the logic you have now is what we want. We need to display the superset of all images & canvases referenced in associated iiif images or transcriptions for any sources. It makes sense that this should be rare, and I think as we add more iiif images 

>     * Is this single placeholder image OK? It's public-facing in some cases, so I wonder if Gissoo should at least weigh in. I also figured it doesn't make sense here to try and determine which are recto/verso/etc, but maybe I'm wrong.

Ideally Gissoo should weigh in, but she doesn't have capacity right now. A single placeholder image is fine, and I agree there's no point trying to guess or label recto/verso. However, I don't want a placeholder image with text — it's too noticeable, and it won't translate. Can we just do a solid color that it distinct enough from the background to be visible but still unobtrusive? Something like the gray you're using for the background of the placeholder image now is reasonable for light mode (use existing colors if we have something reasonable). Could use a transparent image and color it with css for light/dark mode.

blms:
@rlskoeser Oh yes, I meant to ask about that!

Here's the logic for image order override: The admin is visually populated via [`admin_thumbnails`](https://github.com/Princeton-CDH/geniza/blob/66c7cdbf4147f1c8e34ad2cc554324f10629e6aa/geniza/corpus/models.py#L251-L268). Then [the JS](https://github.com/Princeton-CDH/geniza/blob/prototype/transcription/sitemedia/js/admin.js#L30) enables interaction with the `admin_thumbnails` section of the admin, and uses the resulting values to populate the hidden `image_order_override` field.

Also, [here's how](https://github.com/Princeton-CDH/geniza/blob/66c7cdbf4147f1c8e34ad2cc554324f10629e6aa/geniza/corpus/models.py#L270-L283) the recto/verso placeholders get into the admin (for the "document included images" selection). That code gets called by `TextBlock.thumbnail`.

geniza/corpus/views.py
```diff
@@ -304,6 +305,27 @@ def get_queryset(self, *args, **kwargs):
...
+                        {
+                            "image": {
+                                "info": static("img/ui/all/all/image-unavailable.png"),
+                            },
+                            "label": "",
+                            "shelfmark": "",
+                            "canvas": canvas_uri,
+                            "excluded": False,
+                        }
+                    )
```
rlskoeser, line 328: This code seems more complicated than necessary. I think it's because of the way we're structuring our iiif image return, and I failed to communicate that I wanted to make this a `dict` instead of a list.

Would it make sense to define the placeholder image once outside the view? At least the image path should be defined somewhere more obvious, not buried in the method. (It doesn't work with the current canvas uri structure, let me comment on that separately). Empty values aren't necessary, and probably not even the `False`; django templates will ignore and not error on missing attributes.

rlskoeser (commented):
Logic is working well, and it's great to see the simplified template.

I want to revisit how we're generating `iiif_images` — we're using a dict, duplicating canvas uris, and throwing the dict structure away before returning, even though it would be more useful to have it.  I have some draft code where I was figuring out what I wanted here, helpful to push it as a reference branch?

side note: is it confusing to you that we have two different `iiif_images` methods? any thoughts on renaming to distinguish the document from the fragment ones?

rlskoeser:
> @rlskoeser Oh yes, I meant to ask about that! They are not currently included in the image order override.

Let's postpone including placeholder canvases in the image order override; when we do, hopefully we can reuse / adapt the logic for this view to get the complete set of iiif images + canvases. (They may want a way to add placeholders too... lots of potential for this to keep getting more complicated!)

It wasn't obvious from the code in `geniza.corpus.admin` where to find this logic; probably the js should be a separate file, referenced explicitly in the `media` for this form or model admin. We may need more comments, too. (But that's not part of this issue/pr; I think it should be handled as part of fixes for #1077)

blms:
New placeholder image is ready for your review @rlskoeser. I went with a semi-transparent "no" symbol (🚫) on a background color that changes according to light/dark mode. My thinking was that it's good to have some kind of indication that this means "no image" rather than that the IIIF image is still loading or has failed to load, but I can remove if it's distracting or not helpful.

> I don't feel strongly about this either way; what matters is that the text is editable, if it's easiest for them to show up that's fine with me. The transcription editor can always toggle off the display of the image column in the ITT viewer. Did you want to propose something here?

Something I thought about here is that we probably want to hang on to the placeholder images during editing, in case an editor deletes all transcription blocks on the page. In that case, they'd no longer be able to add any new ones.

But that does bring up an interesting problem that it's not really possible to add new transcriptions to documents without image using our current logic. The placeholders only show up if an existing transcription references some canvases that aren't tied to images, but if there is no image and no transcription on the document, then how would editors create a new transcription? Not sure if this is relevant for this PR.

blms:
I guess we already aren't displaying the "add a new transcription" button in that case because of this check.

https://github.com/Princeton-CDH/geniza/blob/6e9d2ced44091377113798d10312e2462aa78201/geniza/corpus/templates/corpus/document_detail.html#L155-L157

rlskoeser:
> New placeholder image is ready for your review @rlskoeser. I went with a semi-transparent "no" symbol (🚫) on a background color that changes according to light/dark mode. My thinking was that it's good to have some kind of indication that this means "no image" rather than that the IIIF image is still loading or has failed to load, but I can remove if it's distracting or not helpful.

The 🚫 symbol makes it look like something is broken to me. Please remove it.  I'll find out if/when Gissoo has capacity to give us a consult on this.

> Something I thought about here is that we probably want to hang on to the placeholder images during editing, in case an editor deletes all transcription blocks on the page. In that case, they'd no longer be able to add any new ones, unless there is a placeholder canvas available to draw on.
> 
> But that does bring up an interesting problem that it's not really possible to add new transcriptions to documents without image using our current logic. The placeholders only show up if an existing transcription references some canvases that aren't tied to images, but if there is no image and no transcription on the document, then how would editors create a new transcription? Not sure if this is relevant for this PR.

Part of #1044 is making sure we can add and edit transcriptions for records with no images at all, which includes documents that don't currently have any transcriptions. You checked off the second task (display placeholder canvases in editor if document has no images), but maybe that got lost when we revised the canvas uri logic.  I suggest we always display a minimum of two canvases (no need to label, but should use placeholder canvas uris like we do elsewhere) when the transcription editor is loaded.

You're right, we need to fix the template include logic so that we display the add transcription link. I still prefer not to show the disabled ITT panels for records with no content; I guess we move the check inside the included template, after the add transcription link.

blms:
> The 🚫 symbol makes it look like something is broken to me. Please remove it. I'll find out if/when Gissoo has capacity to give us a consult on this.

Removed!

> You checked off the second task (display placeholder canvases in editor if document has no images), but maybe that got lost when we revised the canvas uri logic. 

Apologies, I actually checked that off thinking it referred only to those with transcriptions; I clearly didn't think about the _other_ case where there were no images until later! 😅 

> You're right, we need to fix the template include logic so that we display the add transcription link. I still prefer not to show the disabled ITT panels for records with no content; I guess we move the check inside the included template, after the add transcription link.

Good idea, all set.

rlskoeser:
@blms updates look good! One more thing I didn't notice before somehow: we should not have zoom/rotate controls on the placeholder canvases. Could just be a simple css rule to hide them, maybe?

Issue #1065, opened 2022-09-09
Pass tiny API key to, and use it in, tahqiq


blms:
## testing notes (QA)

In the transcription editor for a document (QA public site document details page --> "edit X's transcription"/"add a new transcription", visible when logged into admin):
- [x] Create or edit a block of transcription text. Confirm there is no warning message about "This domain is not registered with Tiny Cloud" or about an API key.

## dev notes
- [x] Pass API key along in config via geniza
- [x] Add line during tiny instantiation to set the `api-key` attribute
- [x] Remove extraneous instantiation in HTML head

kseniaryzhova:
@rlskoeser has not appeared for me while I've been testing with transcriptions today!

PR #1071, opened 2022-09-12
Pass Tiny API key to tahqiq, remove Tiny CDN JS from header

blms:
## In this PR

- Per #1065:
  - Pass the Tiny API key along to tahqiq, instantiate Tiny via the web component rather than a direct call to the CDN JS
  - Bump tahqiq commit hash


rlskoeser (approved):
good

Issue #1055, opened 2022-09-01
As a content editor, I want to cut and paste transcription content from a Google Doc or similar and have it display properly with site styles so that I can easily add existing transcription content.


rlskoeser:
## testing notes (qa)

In the transcription editor for a document (QA public site document details page --> "edit X's transcription"/"add a new transcription", visible when logged into admin):
- [ ] Edit or create a transcription block. Paste in some content with formatting, such as bold, italic, tables, images, etc (such as the unpublished edition by Marina in the scholarship records for PGPID 646). Click "save" and confirm the extraneous formatting is removed, but list numbering remains.

## dev notes
- [x] use bleach or similar to sanitize body html content on `Annotation.save` (maybe make use of a clean method? but make sure it's called on save)

PGPID 646 is a good test case, there's an unpublished edition by Marina in a google doc linked in an existing footnote; cutting and pasting from the google doc looks like it works, but when I tried it, it introduced a bunch of in-line styles.

kseniaryzhova:
@rlskoeser extraneous formatting is removed and the list numbering remains BUT the numbers move to the ends of the lines like so: 
![image](https://user-images.githubusercontent.com/105385128/190935771-522b64f6-b094-4bca-b86d-3db3b988c74c.png)

blms:
@kseniaryzhova Odd, that suggests the list formatting got stripped out. Can you pass along the source that you copied from? I want to see how it's structured to see if there's anything we can do to preserve the list formatting.

rlskoeser:
@kseniaryzhova was it formatted as a numbered list in the version you copied from, or were the numbers just text?

kseniaryzhova:
@rlskoeser works for me - other issues discussed on Slack!

PR #1072, opened 2022-09-12
Sanitize HTML in annotations (#1055)

blms:
## In this PR

- Per #1055:
  - Add class method `sanitize_html` to Annotation
  - Use that method to clean any HTML in the body value of annotations' content set in `set_content`

## Questions

1. Is it OK that I called it `sanitize_html` instead of `clean`? Looks like `clean` is a Django hook and I didn't think we wanted to interact with/override that behavior, but maybe I'm wrong.
2. `bleach` is working well for Marina's PGPID 646 transcription, but I noticed one problem with the pasted content that causes a styling quirk: each line of content is a `p` nested inside `li`. The easy fix is just to allow this nesting and fix the styling issue with CSS, which I did with the `li > p` rule. Do we care about this semantically, or is just fixing the style sufficient?
3. In Tahqiq, we set the visible content to the editor's inner HTML after saving. But in the case of copy/paste, the inner HTML likely will no longer match what actually gets stored. Presumably the user would want to see what got stored, without having to refresh, but as of now they will see the unsanitized HTML at first, and need to refresh to see the sanitized version. Should we now always reload the content from the server for display after saving?

### DIFF
--- a/geniza/annotations/views.py
+++ b/geniza/annotations/views.py
@@ -134,7 +134,9 @@ class AnnotationSearch(View, MultipleObjectMixin):
         # implement something similar to SAS search by uri
 
         # sort by schema:position if available
-        annotations = self.get_queryset().order_by("content__schema:position")
+        annotations = self.get_queryset().order_by(
+            "content__schema:position", "created"
+        )
         # if a target uri is specified, filter annotations
         target_uri = self.request.GET.get("uri")
         if target_uri:
--- a/geniza/annotations/tests/test_annotations_views.py
+++ b/geniza/annotations/tests/test_annotations_views.py
@@ -244,7 +244,7 @@ class TestAnnotationSearch:
         results = response.json()
         assert "resources" in results and len(results["resources"]) == 4
 
-        # in absence of schema:position, should order by created date (via model Meta)
+        # in absence of schema:position, should order by created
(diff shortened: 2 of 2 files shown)


blms:
@rlskoeser This one should be ready for review now that we closed the tahqiq one!

geniza/annotations/views.py
```diff
@@ -132,7 +132,9 @@ def get(self, request, *args, **kwargs):
         # TODO: Convert this to list when > 2 options
         # implement minimal search by uri
         # implement something similar to SAS search by uri
-        annotations = self.get_queryset()
+
+        # sort by schema:position if available
+        annotations = self.get_queryset().order_by("content__schema:position")
```
rlskoeser, line 137: whoa, does this syntax work?! cool.

Please add a quick test to confirm the sorting; and can we get a secondary sort by creation time? (this is the default sort in the model, so we may get it for "free", but would be good to confirm)

blms: Based on the test I wrote it looks like we do, should I still add it to the `order_by`?

rlskoeser: hmm, doesn't hurt to be explicit I guess. defer to you

rlskoeser (approved):

Issue #1056, opened 2022-09-01
clean up footnote superscripts in TEI


rlskoeser:
## dev notes

Can be done as a one-time conversion in the TEI. 

Example:
`            <l n="12">וקד כאן עבדהא אי̇צא לחקה מר̇ץ<hi rend="superscript">3</hi> צעב ואקאם לאזם אלבית איאם ותפצל אללה</l>`

Can find the relevant files with: `grep -l rend=.superscript *.xml`

rlskoeser:
Figured out how to make the change with `sed`:

`sed -i "" 's/<hi rend="superscript">[0-9][0-9]<\/hi>//g' `grep -l rend=.superscript *.xml`

Couldn't get the one or more `\+` regex syntax to work that worked in the [online sed regex tester](https://sed.js.org/index.html), but ran the replace twice for single digit and double digit footnotes, and that seems to have done the trick.

PR #1078, opened 2022-09-13
Revise Document.iiif_images to return dict instead of list

rlskoeser:
partial refactor; updates `Document.iiif_images` to return a dict instead of a list; in reference to PR #1056

only partially tested, haven't updated all the unit tests; may well be other places that we need to update

### DIFF
--- a/geniza/corpus/views.py
+++ b/geniza/corpus/views.py
@@ -320,7 +320,7 @@ class DocumentDetailView(DocumentDetailBase, DetailView):
 
         for ed in document.digital_editions().all():
             for canvas_uri in ed.content_html.keys():
-                if canvas_uri not in images:
+                if canvas_uri and canvas_uri not in images:
                     # use placeholder image for each canvas not in iiif_images
                     images[canvas_uri] = placeholder_canvas
 
--- a/geniza/corpus/views.py
+++ b/geniza/corpus/views.py
@@ -320,7 +320,7 @@ class DocumentDetailView(DocumentDetailBase, DetailView):
 
         for ed in document.digital_editions().all():
             for canvas_uri in ed.content_html.keys():
-                if canvas_uri and canvas_uri not in images:
+                if canvas_uri not in images:
                     # use placeholder image for each canvas not in iiif_images
                     images[canvas_uri] = placeholder_canvas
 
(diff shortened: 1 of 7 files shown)


blms (approved):
I think this one should be ready for your review. This is the one remaining issue I'm struggling with (also causing the 500 error on Lighthouse):

https://github.com/Princeton-CDH/geniza/blob/2ced96efeda7f51d99b78c12956e3cf6d4aa1506/geniza/corpus/templates/corpus/document_detail.html#L23-L28

Otherwise let me know what you think of my revisions, and then we can merge this into my branch and hopefully get that PR finished up quickly as well!

geniza/corpus/views.py
```diff
@@ -308,24 +315,15 @@ def get_context_data(self, **kwargs):
...
-                            "image": {
-                                "info": static("img/ui/all/all/image-unavailable.png"),
-                            },
-                            "label": "",
-                            "shelfmark": "",
-                            "canvas": canvas_uri,
-                            "excluded": False,
-                        }
-                    )
+                if canvas_uri and canvas_uri not in images:
```
rlskoeser, line 323: can `canvas_uri` be `None` here?

blms: It should not (and maybe will not) in normal operation, but I did run into it in my testing. It might have been from an outdated annotation with the canvas set incorrectly. Should I investigate further to track down why it happened? Or maybe remove this check and then investigate if we encounter an error here in QA/production?

rlskoeser: outdated canvas sounds plausible... I like your last suggestion, remove the check but see if we have any problems with fresh data in qa

blms:
Oh yes, @rlskoeser if it wasn't clear, I meant to ask for your help with the twitter/opengraph meta issue!

rlskoeser:
OK, played with it a bit and the iiif images behavior in the template thing is very strange! Not sure what's going on. But if this is the only place we're using iiif images for templates, can we just shift the logic to the view?

rlskoeser:
your revisions to my refactor look good, thanks for fixing my weird code (IDK how I thought it worked). and yes, hooray for unit tests

blms:
@rlskoeser is this what you had in mind? If so, feel free to merge

PR #1085, opened 2022-09-14
annotation / transcription backup 

rlskoeser:
preliminary code for exporting annotations; currently all logic is in a manage command but some of it needs to be callable elsewhere and I'm not sure where it belongs

I'm exporting annotation content in annotation format as a backup that could be imported/restored and then compiled html and text versions of the transcription so it's easier to review and see textual changes. 

Would appreciate a quick review and any thoughts on where the logic could live to be callable from a script and from a method (e.g. fired on a signal when an annotation changes).


geniza/corpus/management/commands/export_annotations.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def annotation_list_name(self, canvas_uri):
+        # per annotation spec, annotation list name must uniquely distinguish
+        # it from  all other lists, and is typically the same name as the canvas.
+
+        # canvas uris vary depending on source:
+        # https://cudl.lib.cam.ac.uk/iiif/MS-TS-NS-00321-00008/canvas/1
+        # https://figgy.princeton.edu/concern/scanned_resources/f9eb5730-035c-420a-bf42-13190f97c10d/manifest/canvas/cfd65bb6-7ff5-47e8-9e92-29dd0e05baf2
+        # https://princetongenizalab.github.io/iiif-bodleian-a/manifests/canvas/1.json
+        # https://digi.ub.uni-heidelberg.de/diglit/iiif/codheidorient78/canvas/0001
```
blms, line 121: So this produces something like this? `heidelberg_codheidorient78_canvas_0001`

Maybe examples of output in the comment here would be helpful.

geniza/corpus/management/commands/export_annotations.py
```diff
@@ -0,0 +1,143 @@
...
+            os.makedirs(output_dir, exist_ok=True)
+
+            # find all annotations for this document
+            annos = Annotation.objects.filter(
+                content__target__source__partOf__id=document.manifest_uri
+            )
+            # TODO: ensure ordered by position / date created
+            # (updated from latest view logic)
+
+            # aggregate annotations by canvas id
```
blms, line 62: I wonder if this could be useful as a method on `Annotation`?

geniza/corpus/management/commands/export_annotations.py
```diff
@@ -0,0 +1,143 @@
...
+            format_path = os.path.join(
+                settings.ANNOTATION_BACKUP_PATH, "transcriptions", output_format
+            )
+            transcription_output_dir[output_format] = format_path
+            os.makedirs(format_path, exist_ok=True)
+
+        # identify content to backup based on documents with digital editions
+        docs = Document.objects.filter(
+            footnotes__doc_relation__contains=Footnote.DIGITAL_EDITION
+        ).distinct()
```
blms, line 41: Seems useful to have elsewhere, but it's such a small amount of code that it probably doesn't warrant a whole method.

geniza/corpus/management/commands/export_annotations.py
```diff
@@ -0,0 +1,143 @@
...
+            # use PGPID for annotation directory name
+            # path based on recommended uri pattern from the spec
+            # {prefix}/{identifier}/list/{name}
+            output_dir = os.path.join(annotations_output_dir, str(document.id), "list")
+            # ensure output directory exists
+            os.makedirs(output_dir, exist_ok=True)
+
+            # find all annotations for this document
+            annos = Annotation.objects.filter(
+                content__target__source__partOf__id=document.manifest_uri
```
blms, line 57: Same thought here as "documents with digital editions", but I do find I have to refer back to this query elsewhere in code every time I need it, so a convenience method (on `Document`?) could maybe be helpful.

blms (approved):
Looking good so far! 

~~You're already using a lot of methods from elsewhere, like the content text and content HTML methods; I made a few suggestions for other things we could pull out, but it doesn't seem like there's a huge amount of reusable stuff here to me.~~

~~One thing that could potentially be pulled out: do we want the ability to make AnnotationLists elsewhere in the app? I seem to recall we had discussed it briefly, but not sure of the use case.~~

EDIT: See [comment](https://github.com/Princeton-CDH/geniza/pull/1085#issuecomment-1248073063) below, I'd lost exactly what was being asked of me here 😅

blms:
> Would appreciate a quick review and any thoughts on where the logic could live to be callable from a script and from a method (e.g. fired on a signal when an annotation changes).

Ah, OK, re-reading this I understand now. Living in the `Document` model and firing on a signal when the annotation changes makes sense to me, since we want it grouped first by document and then by canvas.

geniza/corpus/management/commands/export_annotations.py
```diff
@@ -0,0 +1,143 @@
...
+                    )
+
+            # for convenience and more readable versioning, also generate
+            # text and html transcription files
+            for edition in document.digital_editions():
+                print(edition)
+                # filename should be based on pgpid and source
+                # use source id for now; slug todo
+                # should we explicitly label as transcription?
+                base_filename = "%s_source_%s" % (document.id, edition.source.id)
```
rlskoeser, line 89: I've been mulling it over — implementing slugs for sources seems like way too much overhead for right now, and still potentially brittle.  I think using slugified author last names should be sufficient, since we shouldn't have more than one transcription for a document by the same unique set of authors.  

Proposed filename format:
PGPID###_authorslug_transcription.txt (or .html)

blms: That proposal looks good to me, agree it's better to slugify by authors.

geniza/corpus/management/commands/export_annotations.py
rlskoeser: if we already had a custom document queryset it would make sense to add it there, but I'm not sure it's worth adding one for just this... IDK if we use this anywhere else

geniza/corpus/management/commands/export_annotations.py
rlskoeser: it does make sense on `Document` but that model and file are already enormous! is there a logical mixin we could create with annotation specific functionality? AnnotationTargetSourceMixin ? (ugh)

geniza/corpus/management/commands/export_annotations.py
rlskoeser: > Ah, OK, re-reading this I understand now. Living in the `Document` model and firing on a signal when the annotation changes makes sense to me, since we want it grouped first by document and then by canvas.

I was thinking we will want to run the backup two ways:
- script that updates everything or specific documents by id; could be run as a cron job on all documents to make sure everything is up to date (should do nothing if there are no changes)
- signal handler that updates annotation exports for a single document when annotations change (I'm a little worried about the signal handler delays already, which was why I was starting to look into async task queues, but I'd prefer to keep things simpler if possible). 

We'll need logic to delete outdated transcription files; maybe that only happens when the full script runs?

geniza/corpus/management/commands/export_annotations.py
blms: Maybe just AnnotationTargetMixin? It's not semantically correct in terms of IIIF, but I think it could easily mean "the annotation's target is a part of this document" to PGP code contributors. OTOH, I'm not opposed to making the Document model slightly longer to add a small convenience method like this. But I do share your concern!

geniza/corpus/annotation_export.py
```diff
@@ -0,0 +1,215 @@
...
+
+        # identify content to backup based on documents with digital editions
+
+        docs = Document.objects.filter(
+            footnotes__doc_relation__contains=Footnote.DIGITAL_EDITION
+        ).distinct()
+        # if ids are specified, limit to just those documents
+        if pgpids:
+            docs = docs.filter(pk__in=pgpids)
+        # use logging? and/or pass in stdout when running from command?
```
blms, line 51: Maybe rename param something like "logger" and then pass stdout when running from command, and use django logger as default.

geniza/corpus/annotation_export.py
```diff
@@ -0,0 +1,215 @@
...
+        if repo.is_dirty():
+            # print("Committing changes")
+            repo.index.commit("Automated data export from PGP")
+            if push_changes:
+                try:
+                    origin = repo.remote(name="origin")
+                    # pull any remote changes
+                    origin.pull()
+                    # push data updates
+                    # TODO: how to tell repo object not to output hashes?
```
blms, line 143: By output do you mean it logs to stdout?

geniza/corpus/templates/corpus/transcription_export.html
```diff
@@ -0,0 +1,24 @@
...
+    <body style="max-width: 860px; margin: auto">
+        <h1>Transcription of {{ document.title }}</h1>
+
+        <p>{{ edition.source.formatted_display|safe }}</p>
+
+        {# include language code when possible #}
+        <section dir="rtl" {% if document.languages.count == 1 and document.languages.first.iso_code %}lang="{{ document.languages.first.iso_code}}"{% endif %}>
+            {% for canvas, html_section in edition.content_html.items %}
+                <div data-canvas="{{ canvas }}">
+                    {# FIXME: why is this lines instead of a block ? #}
```
blms, line 15: Not sure without testing, but I wonder if this could be similar to the weird dict issues I was having with `iiif_images`… still don't fully understand what Django templates do to dicts

blms (commented):
This is looking really good so far, just had a few questions/comments, but nothing pressing. Is it ready for me to test? And should I take a look at the failing unit tests here?

rlskoeser:
Thanks for reviewing! I forgot I still need to write unit tests for the new code, I'll plan to handle that. 

It should be functional enough for you to try it, if you want to test running it. You can use this repo for testing, I just added the @Princeton-CDH/genizadevs group: https://github.com/Princeton-CDH/test-pgp-annotations

If you configure the repo and local path, then run the script to export one pgpid with annotations,  it should set everything up; then you can manually switch to a different branch in your local checkout if you want to export more content.

geniza/corpus/annotation_export.py
rlskoeser: yes, it prints out some content — I think it's similar to what you'd get if you ran the git command yourself

geniza/corpus/annotation_export.py
rlskoeser: that would be a nice solution, but they don't have the same interface...

geniza/corpus/management/commands/export_annotations.py
```diff
@@ -31,193 +32,196 @@ def handle(self, *args, **options):
...
+    #         Repo.clone_from(url=remote_git_url, to_path=local_path)
+    #         # then initialize the repo object
+    #         repo = Repo(local_path)
+    #     else:
+    #         # pull any changes since the last run
+    #         repo = Repo(local_path)
+    #         if remote_git_url:
+    #             repo.remotes.origin.pull()
+
+    #     return repo
```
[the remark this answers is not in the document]
rlskoeser: thanks for flagging, forgot to cleanup

geniza/corpus/templates/corpus/transcription_export.html
rlskoeser: django templates call callables (which is usually convenient and sometimes really frustrating); was that problem with a standard dict?

rlskoeser:
>  I ran into an integrity error running the digital edition migration. Do we need logic there to merge duplicates prior to creating the unique index or should I just start over with fresh data from production?

What state was your data in? I think this could happen if you had a previous migration (esp. before the new constraint was applied), but shouldn't happen if you're starting from a fresh copy.

> `AnnotationExporter.setup_repo` expects the directory to not exist until the repo is cloned; I had created the directory but not cloned the repo yet, so I ran into an error on line 212. Not sure if this needs to be handled but maybe should be made explicit somewhere in the help.

Ah, thanks for flagging. Probably simple enough to handle that case.

> `AnnotationExporter.setup_repo` also calls `self.verbosity` and `stdout`, which are not defined in its scope.

whoops, thanks for flagging (how did I miss those)

>  On line 65, `stdout.write(document)` gave me an error but `stdout.write(str(document))` worked.

👍

geniza/corpus/templates/corpus/transcription_export.html
blms: It was [this problem](https://github.com/Princeton-CDH/geniza/blob/2ced96efeda7f51d99b78c12956e3cf6d4aa1506/geniza/corpus/templates/corpus/document_detail.html#L23-L28), and IIRC when I tried to use `.values` or `.items` in the template it returned empty strings instead of IIIF image objects, and with `to_list` it started giving me back a string representation of a list…

blms:
> What state was your data in? I think this could happen if you had a previous migration (esp. before the new constraint was applied), but shouldn't happen if you're starting from a fresh copy.

Yeah, likely I need to just try with fresh data since I'm sure I've created duplicates with the old logic in my current data.

Issue #1049, opened 2022-08-31
As a content editor, I want transcription formatting preserved in search result display but ignored for search text so that I can see where in the transcription matching terms are.


rlskoeser:
## testing notes
- [x] transcription line numbers (when present) should display in the search results by default (no keyword search)
- [x] transcription line numbers (when present) should display with keywords in context if match is near the beginning of the document (won't always display if your match is from the middle of the text; is this acceptable?)
- [x] transcription keywords in context matches display the correct line number when the match is from the middle of the text

## dev notes
- [x] index html instead of plain text
- [x] update search result formatting for list instead of plain text with line breaks
- [x] calculate and add line number to html when indexing
  - [x] [parse the HTML](https://stackoverflow.com/questions/14694482/converting-html-to-text-with-python/55825140#55825140) to look for `ol > li` and use "start" attribute if present
  - [x] add either a data attribute or value attribute to each `li`, doesn't matter which
- [x] revise css to display pseudo-marker based on line number attribute for search only
- [x] update styles in admin that are suppressing line numbers

richmanrachel:
>transcription line numbers (when present) should display with keywords in context if match is near the beginning of the document (won't always display if your match is from the middle of the text; is this acceptable?)
- I'm afraid this isn't acceptable -- it's really helpful as a researcher to have all the line numbers on the search page so that I can quickly jot down where to find the term in my spreadsheet/notes without having to go into the full document.

Also, it looks a bit messy and confusing to have a mixture of numbered and un-numbered entries:
transcription line numbers (when present) should display with keywords in context if match is near the beginning of the document (won't always display if your match is from the middle of the text; is this acceptable?)

rlskoeser:
ok! I wondered if it might not be sufficient, but wanted to know for sure before doing more work. Now that we're using proper ordered lists, we don't have the sequence number on each line by default, so we'll have to do a little bit of extra work to put it in so we can display it on the search version.

richmanrachel:
Okay, thanks so much!

Also, just an oddity that's happening now with the formatting for results that are too far into the text for line numbers - there's a first line and then an indentation:
<img width="764" alt="image" src="https://user-images.githubusercontent.com/68028846/188005113-a995876c-1a0a-4b9e-ad91-5208e1100787.png">
[...]

PR #1086, opened 2022-09-14
Add line numbers to transcription lines in indexed html (#1049)

blms:
## In this PR

- Per #1049:
  - Use an `HTMLParser` subclass to parse transcription html and add line number `value` attributes
  - Restore CSS to show those attributes as line numbers in search results
  - Fix admin CSS issue with line number suppression

### DIFF
--- a/geniza/footnotes/models.py
+++ b/geniza/footnotes/models.py
@@ -1,6 +1,5 @@
 from collections import defaultdict
 from functools import cached_property
-from html.parser import HTMLParser
 from os import path
 from urllib.parse import urljoin
 
--- a/geniza/footnotes/models.py
+++ b/geniza/footnotes/models.py
@@ -1,5 +1,6 @@
 from collections import defaultdict
 from functools import cached_property
+from html.parser import HTMLParser
 from os import path
 from urllib.parse import urljoin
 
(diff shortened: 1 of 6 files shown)


geniza/footnotes/models.py
```diff
@@ -410,6 +411,43 @@ def editions(self):
         return self.filter(doc_relation__contains=Footnote.EDITION)
 
 
+class HTMLLineNumberParser(HTMLParser):
```
rlskoeser, line 414: not sure this belongs in models, could we put it in a utils file or something like that?

geniza/footnotes/models.py
```diff
@@ -410,6 +411,43 @@ def editions(self):
...
+        """Restart line numbering on <ol>, include line number with <li>, and construct
+        start tags with included attributes"""
+        if tag == "ol":
+            # restart line numbering on encountering ol
+            self.line_number = 1
+            for (attr, val) in attrs:
+                # if start present in attrs, restart to start number
+                if attr == "start":
+                    self.line_number = int(val)
+        elif tag == "li":
```
rlskoeser, line 433: I think we need a state indicator that we are within an `ol`, because that's the only time this logic should happen. (Possible we could have a `ul`)

geniza/footnotes/models.py
```diff
@@ -522,14 +560,19 @@ def content_html(self):
...
+        parser = HTMLLineNumberParser()
+        parser.feed(
+            "\n".join(
+                [
+                    section
+                    # content html is a dict; values are lists of html content
+                    for canvas_annos in self.content_html.values()
+                    for section in canvas_annos
+                ]
+            )
```
rlskoeser, line 573: I think we only want the line numbers for solr index data, not other use cases (like my github export), so I vote not applying it here by default.  What about a utility method that applies the parser and generates the output, so it's a one-line call? maybe something like `explicit_line_numbers`

geniza/footnotes/tests/test_footnote_models.py
```diff
@@ -299,6 +299,39 @@ def test_content_html(self, annotation, twoauthor_source):
...
+            }
+        )
+        digital_edition = document.footnotes.create(
+            source=twoauthor_source, doc_relation=[Footnote.DIGITAL_EDITION]
+        )
+        assert (
+            digital_edition.content_html_str
+            == '<ol start="5"><li value="5">one</li><li value="6">two</li></ol>'
+        )
+
```
rlskoeser, line 334: test case here that `ul > li`  is ignored

rlskoeser (changes requested):
Working well! Tested locally reindexing one doc and searching for a line later in the text, confirmed that I got the correct line numbers. 

Requesting some small changes in the parsing and code organization

rlskoeser (approved):

PR #1091, opened 2022-09-16
Ensure we only have one digital edition footnote per document+source combination

rlskoeser:
intially implemented as part my github backup script but cherry-picked to help with review

- new db constraint to ensure we can't create more than one digital edition footnote per document+source combination
- revise tei to annotation migration to fix the logic that was resulting in duplicate digital edition footnotes


blms (approved):
Logic looks good to me! Just pushed a fix for the tests, I had been manually creating digital editions that would be automatically created by the post-save signal.

PR #1104, opened 2022-09-19
Fix duplicate menu IDs (#1088)

blms:
## In this PR

- Per #1088:
  - Add prefix "footer-" to all menu IDs in footer

I checked all reused template snippets and that's the only way it would have resulted in a duplicated ID, as far as I can tell.


rlskoeser (approved):
yes, it was the repeated nav template that was causing the duplicate id

Issue #1106, opened 2022-09-19
Implement the Hebrew type styles so that users can easily read the website content 


gissoo:
[here are the type styles](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=2338%3A56894)

blms:
Confirmed present in QA, closing.

PR #1108, opened 2022-09-20
Implement new Hebrew type styles (#1106)

blms:
## In this PR

- Per #1106:
  - Implement all new Hebrew type styles from Figma, using new mixins and specifying new rules for `html[lang="he"]`
- Use `calc()` for calculating line height proportions, instead of manual calculation (using proportions instead of pixels for scaling)

### DIFF
--- a/geniza/corpus/templates/corpus/snippets/document_result.html
+++ b/geniza/corpus/templates/corpus/snippets/document_result.html
@@ -36,7 +36,8 @@
                 </dl>
 
                 {# description #}
-                <p class="description">
+                {# TODO: Adjust lang attribute logic for non-English descriptions #}
+                <p class="description" lang="en">
                     {# display keywords in context if any #}
                     {% with document_highlights=highlighting|dict_item:document.id %}
                         {% if document_highlights.description %}
--- a/sitemedia/scss/base/_global.scss
+++ b/sitemedia/scss/base/_global.scss
@@ -54,6 +54,9 @@ body {
     flex-direction: column;
     min-height: 100vh;
 }
+html[lang="he"] body {
+    @include typography.body-he;
+}
 
 // Center-align sections of main content using flexbox.
 main {
@@ -67,13 +70,22 @@ main {
 h1 {
     @include typography.headline-1;
 }
+html[lang="he"] h1 {
+    @include typography.headline-1-he;
+}
 
 h2 {
     @include typography.headline-2;
 }
+html[lang="he"] h2 {
+    @include typography.headline-2-he;
+}
 
 h3 {
     @include typography.headline-3;
+}
+html[lang="he"] h3 {
+    @include typography.headline-3-he;
 }
 
 // Any links without a class (i.e. in body text) get the default link style;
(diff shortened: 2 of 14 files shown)


blms:
@rlskoeser Good point, this does indeed assume all affected content in the page will be translated, so definitely will not account for some of them being in Hebrew and some in English.

So I do think we will need to manually add `lang="en"` to elements that we know will not be translated. What about things like content pages and document descriptions, which may have some translated and some untranslated content? Maybe since those are "translatable" fields there's a way to apply it programmatically when we're displaying the translated version?

For testing, I think it would actually be great if you took a look, especially since it will also give you a sense of remaining work on the Hebrew site. I think there is still some other design implementation to be done… I saw some things on the document detail page, and the ITT viewer definitely doesn't work currently in RTL.

rlskoeser (commented):
Styles look reasonable to me.

For testing the untranslated content, I think the simplest thing to do right now is to put `lang=en` on the description in search results; we will have hebrew descriptions for some records, but not all, and when we implement the logic to show them when in the hebrew site, we can adjust the lang attribute logic as needed.

I looked briefly at wagtail — from the interface I saw before, content is translated in blocks, so there ought to be a way to determine if a block is translated or not (or what language it's in). What I see in their documentation is whether a _page_ is translated, which doesn't seem helpful. We'll need an issue for that, and hopefully can get some of the wagtail translations to work with.

blms:
@rlskoeser Will any of the below content ever be in English on the Hebrew site, and if so, what element would get the `lang` attribute? Right now these are all in English in my local data.

- search form (`form` element gets lang?)
- scholarship records (`li.citation` gets lang?)
- document details metadata (i.e. shelfmark, primary language—labels and values)
- footer and site nav items (maybe each `li` or `a` gets lang? would be weird to have mixed line heights though)
- maybe others I added Hebrew variants for in 52bb18bfe0e264815cb012533d73462436e166e6 as well

Most page content will be handled by d39cdd0868b7c54b1788ebbe91a88b9eab2f0977 with `*[lang="en"]`, but I think we need finer control in the above sorts of cases since some get `meta-he` and some get `body-he`.

blms:
@rlskoeser made some edits to the above comment after running `compilemessages`, thanks for the tip!

So I'm thinking it makes sense to look _inside_ each of the sections or containers for `*[lang="en"]`, since it's really just going to be leaf nodes like `dd` that get the `lang` attribute. Still not 100% sure what to do about the search form controls and nav, but maybe can treat them the same way?

rlskoeser (commented):
Thanks for adding the `not` filters; I was hoping we could do most of this be excluding elements rather than re-applying styles; it looks like there are some cases where exclusion doesn't work, but since it's a semantic include, it looks pretty clean to me.

Do we need to document somewhere what will and won't or might not be translated? 

Do we need any font files added somewhere, or were they already added previously?

I did intend for you to add the lang attribute for the current description on this branch, I thought it might be helpful.

blms:
> Do we need to document somewhere what will and won't or might not be translated?

I think that would be very helpful, but not sure where. It would be nice to have them all in one place. Perhaps just somewhere on GitHub so it could be edited by any project contributors?

> Do we need any font files added somewhere, or were they already added previously?

If I understand correctly, and based on Figma, the Hebrew typefaces themselves aren't changing, so we can use the ones already added.

Issue #1109, opened 2022-09-20
As an admin/content editor, I want to be able to see all the images associated with a document so that I can determine whether I need to associate more images, clone the record, etc.


mrustow:
... but the greyed-out images are hard to see. So once they're no longer greyed out, the chosen images could be marked in some other way, perhaps with the yellow frames we now use for the thumbnails (or another idea you may have).

Then there's the question of do we need to zoom in on them. I'm not as worried about that, since we're just talking about eyeballing the fragment to decide what to do with it, so I say no, not at this stage. 

Thanks!

PR #1111, opened 2022-09-20
Revise admin display of non-document images (#1109)

blms:
## In this PR

- Per #1109:
  - In admin document change form, make non-document images full-width, remove overlay from them, and add yellow outline around selected images
  - Show placeholder images as gray boxes in admin view (i.e. same as editor)
- Pass new images dict to Document admin change view (we weren't doing this yet! not sure how the ITT viewer was working in admin for others, it wouldn't work at all for me due to `images` not being in the context…)

## Questions

- I've just duplicated code from `DocumentDetailView.get_context_data` here, but that's obviously not DRY and I'd prefer if it's a method called from elsewhere. Would it make sense to add a method for this on the Document model? (and move `placeholder_canvas` there too)

### DIFF
--- a/geniza/corpus/admin.py
+++ b/geniza/corpus/admin.py
@@ -27,7 +27,7 @@ from geniza.corpus.models import (
     TextBlock,
 )
 from geniza.corpus.solr_queryset import DocumentSolrQuerySet
-from geniza.corpus.views import DocumentMerge
+from geniza.corpus.views import DocumentMerge, placeholder_canvas
 from geniza.footnotes.admin import DocumentFootnoteInline
 from geniza.footnotes.models import Footnote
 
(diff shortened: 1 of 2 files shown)


rlskoeser:
> * I've just duplicated code from `DocumentDetailView.get_context_data` here, but that's obviously not DRY and I'd prefer if it's a method called from elsewhere. Would it make sense to add a method for this on the Document model? (and move `placeholder_canvas` there too)

yes, makes sense to move it to a method on `Document`

blms:
@rlskoeser Simplified by just moving into a parameter-enabled section in `iiif_images`, let me know if that seems wrong to you.

geniza/corpus/models.py
```diff
@@ -676,6 +686,13 @@ def iiif_images(self, filter_side=False):
                             "excluded": b.side and i not in b.selected_images,
                         }
 
+        if with_placeholders:
```
rlskoeser, line 689: need a comment here for context
```suggestion
        # when requested, include any placeholder canvas URIs referenced by any associated transcriptions
        if with_placeholders:
```

rlskoeser:
@blms this one should be merged into `develop` rather than our transcription prototype branch, yes? maybe that's why we're missing the fix in v4.8.1 ?

blms:
@rlskoeser Ah, great catch! For some reason, I was totally forgetting that we already have the ITT _viewer_ in develop, so was branching from the editor prototype 🤦  

But I think it's still good to have these changes regarding the new dict format (which is not in develop) present in prototype/transcription. Would it be okay to cherry-pick the commits that should be made against develop—namely, the admin style changes—and open a new PR just for those?

rlskoeser:
@blms whoa 🤯 and I forgot that the new canvas uri structure is in the transcription branch only!

If it's simpler to just keep it in the transcription branch, I think that's ok — we should be releasing it soon, and this isn't urgent. And we need to update our prototype branch from develop at some point.  If you think it's trivial to cherry-pick the styles to develop, that would be ok too.

blms:
@rlskoeser Just noticing I did the CSS in a separate commit, so if it's not urgent then let's just get it into the transcription branch as you suggest.

Issue #1089, opened 2022-09-15
remediate sort selection drop-down: interactive controls must not be nested


rlskoeser:
according to dubbot, we're violating https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html 

flagged on the sort select trigger

convert to a standard `select` element and style as able

rlskoeser:
here's the response from John Jameson:

> The Filter button does not expose state, and focus is lost on click, and the dropdown DubBot caught loses focus and is missing needed keyboard interactions. It
> 
> The “filters” button should follow a toggle button pattern – aria-expanded states: https://www.w3.org/WAI/ARIA/apg/patterns/button/
> 
> The dropdown could either just use a native browser `<select>` (hecka easiest) or implement a listbox pattern (click opens, arrow keys navigate): https://www.w3.org/WAI/ARIA/apg/patterns/listbox/
> 
> You are also reloading the page immediately on change for any of the filters – this causes focus to jump back to “Jump to main content.” A keyboard or screen reader user who wants to set a few different filters will have a frustrating road ahead – click one, tab back to form, click another, tab back to the form…basically, screen reader users prefer a separate submit button when search results aren’t AJAX loaded – see https://www.w3.org/WAI/WCAG21/quickref/#on-input

@blms I thought you had done some work along the lines of the list box pattern, is that partially implemented?  What's your estimate in hours for the two options for the sort dropbox?

blms:
@rlskoeser Thanks so much!

Yes, the list box pattern is partially implemented. I'll need to make sure the roles and interactions are 1:1 with that guideline—such as the list getting focus on click, thus allowing immediate keyboard control. One problem is that when a keyboard user interacts with the list and focuses another option, it actually gets selected and thus refreshes the search results; I suppose instead it should only do that on keypress. So I would estimate ~3 hours for that.

I've already spent 1 hour on making it a select element, I would estimate 1-2 more for putting the finishing touches on that and updating any affected unit tests.

Also, if I recall correctly the original DubBot issue was "nested interactive elements," and from a quick glance the only thing I see that matches that is the button inside the `summary` element. But maybe I'm wrong about what exactly was flagged by DubBot?

rlskoeser:
I think maybe the nested interactive elements are for the search filters.

I think we should handle the automatic form submission separately — John said that isn't as serious; it's inconvenient but not a blocker.

rlskoeser:
confirmed fixed on test site by scanning with axe DevTools

PR #1114, opened 2022-09-21
Revert sort listbox to <select> in search form (#1089)

blms:
## In this PR

- Per #1089:
  - Remove `SelectDisabledMixin` and `RadioSelectWithDisabled`, and revert form widget to default `select` element
  - Remove details/summary/button from HTML and revert to form default widget; add icon for caret-down (presentation only)
  - Remove Stimulus logic for control of the custom listbox, and adapt Stimulus logic for "relevance" option enabling/disabling
  - Adapt CSS for `select` element and Phosphor icon

### DIFF
--- a/geniza/corpus/forms.py
+++ b/geniza/corpus/forms.py
@@ -8,41 +8,6 @@ from geniza.common.utils import simplify_quotes
 from geniza.corpus.models import Document
 
 
-class SelectDisabledMixin:
-    """
-    Mixin for :class:`django.forms.RadioSelect` or :class:`django.forms.CheckboxSelect`
-    classes to set an option as disabled. To disable, the widget's choice
-    label option should be passed in as a dictionary with `disabled` set
-    to True::
-
-        {'label': 'option', 'disabled': True}.
-    """
-
-    # copied from mep-django
-
-    # Using a solution at https://djangosnippets.org/snippets/2453/
-    def create_option(
-        self, name, value, label, selected, index, subindex=None, attrs=None
-    ):
-        disabled = None
-
-        if isinstance(label, dict):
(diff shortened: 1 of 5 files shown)


rlskoeser (commented):
The styles look good. Wish we had done this from the outset, and will remember in future.

Why did you remove `SelectDisabledMixin` ? That logic hasn't changed and still pertains.  (Were we even using it, since we weren't using a `<select>` element?)

blms:
@rlskoeser We were not using it, but according to the docstring it is actually meant for radio or checkbox select widgets. 

I think it was vestigial, we should have removed it back when we removed this block in c4f237b16ad50e0d12016ef271d65a862ac3032a. (relevant discussion in https://github.com/Princeton-CDH/geniza/pull/353)

https://github.com/Princeton-CDH/geniza/blob/e567519726221b31ee6818d25fff9718ef22e55c/geniza/corpus/forms.py#L64-L75

rlskoeser:
> @rlskoeser We were not using it, but according to the docstring it can be used for radio or checkbox select widgets.
> 
> I think it was vestigial ...


It was vestigial in our previous implementation but useful in our fixed one now. I expect it to work the same way as a mixin for the `Select` widget, I think the choice field logic is the same for all of the widgets.

I'd prefer to have the logic server side and not just in the javascript, and I think th existing code should get us very close if not all the way there

blms:
I see, thanks! I will give it a try with that `__init__` code restored.

blms:
Done, it is working perfectly! Thanks for the catch. I also gave you co-author credit since all of that logic was yours to begin with 😆

rlskoeser (approved):

Issue #1120, opened 2022-09-22
display issue / artifact showing up in image + transcription in some cases


rlskoeser:
## testing notes

please review and see if you can find any records where there are weird display issues like this

![Screen Shot 2022-09-22 at 4 02 19 PM](https://user-images.githubusercontent.com/691231/191840264-bca2b305-840e-4629-89bb-12a3a93318b4.png)

kseniaryzhova:
@rlskoeser not sure how to test - I actually cannot find any transcriptions with the recto header for a text box (or any header for the first text box in a transcription, but I know they exist)

rlskoeser:
@kseniaryzhova can you edit one to add a label there and see if causes a display problem?

kseniaryzhova:
@rlskoeser Tried adding a label and it won't even save. Here's the video:

https://user-images.githubusercontent.com/105385128/192912791-97961674-a103-4d12-a679-175b7faf02a6.mp4

kseniaryzhova:
@rlskoeser edited a few documents and don't see this issue anymore - closing!

PR #1125, opened 2022-09-23
Tweak transcription list number style #1120

rlskoeser:


blms (approved):
Much better solution! I was wondering if that visual glitch issue was related to the decimal covers.

Issue #1113, opened 2022-09-21
Implement a language switch so that users can choose to view the site in English, Hebrew, or Arabic 


gissoo:
## testing notes (QA)

- [x] On the QA site, when logged in, confirm there is a language switcher with 3 languages (english, hebrew, arabic) matching the designs below
- [x] On desktop and mobile (if possible), confirm it switches the language of the current page
- [x] Navigate away and confirm the language change persists
- [x] Log out and confirm the switcher is still present, but only shows English and Hebrew
- [x] Confirm the language switcher options are properly aligned

## dev notes

@blms [here are the designs](https://www.figma.com/file/anZFStv7zBaVWnFxr95nKt/Work-in-Progress-%2F-WIP?node-id=2398%3A70147) – I'll eventually move them to the Components pages – I've provided select and hover designs for LTR and RTL pages for light and dark mode on desktop and mobile 

so on desktop and mobile the toggles will stay in the nav, for mobile this means that they are visible and usable once you've selected the burger menu 

You can see the designs on the pages if you zoom out of the Figma page – let me know if you need anything

gissoo:
@blms I've applied the Semibold Greta Sans to "E" in the designs 
I also noticed the "עִ" wasn't centered in the toggle container, I have fixed that as well, fwiw. 

"E" is 20px on desktop and 18px on mobile (semibold weight Greta Sans for default, and Bold weight Greta Sans for selected)
"עִ" is 26px on desktop and 20px on mobile (regular weight Frank Ruehl for both default and selected)
"ع" is 20px on desktop and 16px on mobile. (bold weight Hassan for both default and selected)

kseniaryzhova:
@gissoo the language symbol for Hebrew looks out of line with the English and Arabic 
![image](https://user-images.githubusercontent.com/105385128/192869659-a247fcc1-592e-411b-bff3-7bc525804b27.png)

blms:
@kseniaryzhova Ooh, good catch. That looks like a browser difference, since I'm now seeing similar on Firefox but not Safari or Chrome.

kseniaryzhova:
@gissoo couldn't test on mobile since I'm off campus, but everything else works on desktop! Should I close?

blms:
@kseniaryzhova Please leave open and I'll fix the alignment issue you raised. (Also I should mention, the Arabic font is a placeholder, still waiting for the Hassan font from @gissoo but should be ready soon.)

[...]

Issue #582, opened 2022-02-04
Language switcher does not correctly pick up the translated slug for the current page


blms:

kseniaryzhova:
@blms is this still an issue?

richmanrachel:
@mrustow - Do you care if urls include slugs in Hebrew? Because our website url is already in English script, I don't think this is important. I recommend closing if you don't have strong feelings otherwise.

blms:
@richmanrachel Taking the thumbs up as approval to close!

PR #1126, opened 2022-09-26
Implement new language switcher design (#1113)

blms:
## In this PR

- Per #1113:
  - Implement the new language switcher design
  - Use links to the translated pages, using a template tag, to populate the language switcher

## Questions

- On mobile, I ran into some z-index issues that I wasn't fully able to resolve. 
  - Any ideas about how to hide the light/dark toggle when in a menu? (Seems like needs to be between the z-index for the `nav` and the `ul` but it lives outside of both…) 
  - And any ideas about how to show the language switcher when in a sub-menu? (I was able to get it working with the main menu using `:target` but that can only work on the highest level of menu nesting.)
  - Maybe both are acceptable as-is, I tried to get them as close as I could to the designs despite these limitations.


geniza/corpus/tests/test_corpus_templatetags.py
```diff
@@ -161,3 +161,14 @@ def test_shelfmark_wrap():
...
+
+def test_translate_url(document):
+    # should translate requested URL into Hebrew
+    ctx = {"request": HttpRequest()}
+    ctx["request"].path = document.get_absolute_url()
+    assert corpus_extras.translate_url(ctx, "he").startswith("/he/")
+
+    # if a Hebrew version cannot be determined, should return the original URL
+    ctx["request"].path = "https://example.com"
+    assert corpus_extras.translate_url(ctx, "he") == "https://example.com"
```
rlskoeser, line 174: what are the cases where we can't get a translated version of the url? 

I don't suppose this would have any awareness of wagtail pages and whether a page is available in translation or not?

It seems like it would be nice to disable the switcher when a page can't actually be switched, but that might need to be out of scope for now

rlskoeser (commented):
Styles look really nice.

I see what you mean about mobile. Is there a way to make the html more semantic? Could we have a separate mobile instance of the light/dark toggle that sits in the right place in the html and is only visible on mobile? (... although wonder if that would be a more convenient place for keyboard users? 🤔 )

I think it's fine to not have the switcher display on the sub page; I forget how the menu is structured, but that doesn't seem like a problem to me.

geniza/corpus/tests/test_corpus_templatetags.py
blms: > what are the cases where we can't get a translated version of the url?

Basically I was just testing generic "bad input" to this template tag and not really thinking through a more realistic use case. In the current implementation, it'll never be the case! (At least, I think all paths within the app are translatable.) Can remove the test if we can't think of any real world examples.

> I don't suppose this would have any awareness of wagtail pages and whether a page is available in translation or not?

Right, it's just relying on Django [`translate_url`](https://docs.djangoproject.com/en/3.2/_modules/django/urls/base/), so if the slug gets translated it won't work (as in #582). Probably need to do something [like this](https://github.com/wagtail/wagtail-localize/discussions/308#discussioncomment-303188) instead. But maybe it makes sense to roll #582 in with this issue?

> It seems like it would be nice to disable the switcher when a page can't actually be switched, but that might need to be out of scope for now

I agree on both.

blms:
> I see what you mean about mobile. Is there a way to make the html more semantic? Could we have a separate mobile instance of the light/dark toggle that sits in the right place in the html and is only visible on mobile?

I'll give that a try, thanks! Was avoiding repetition but maybe that's not always the semantically accurate choice.

geniza/templates/snippets/theme_toggle.html
```diff
@@ -1,8 +1,8 @@
 {% load i18n static %}
```
rlskoeser, line 2: might be nice to have a comment block here documenting the `is_mobile` flag

rlskoeser (approved):
Looks reasonable to me.

Thanks for documenting the additional font in the deploy notes. I updated the font vault file in ansible with the new fonts, so you can check if I got it right when we deploy this.

Issue #912, opened 2022-06-09
As an admin, I want transcription content synchronized from annotation storage to a GitHub repository so that the content is backed up, versioned, and available for use in generating a text corpus.


rlskoeser:

rlskoeser:
@mrustow @kseniaryzhova I'm working on the logic to backup and version our transcription annotation content to GitHub. We'd discussed previously that we want to track who is making the edits, and we also discussed preserving the contributors to the bitbucket version when we migrate.  I have a plan for how to do this, but it does make the tei to annotation migration more complicated so I wanted to check in and make sure it's worth it.  (More complicated because as we migrate each file we need to determine the contributors for that file, then generate the corresponding annotation/transcription backup files and commit them with the appropriate co-author list. But it is feasible.)

I used `git log` to get a list of all unique usernames for everyone who's contributed to the tei bitbucket repo. I think we can probably get GitHub co-author emails for almost all of these folks, do you agree?

Alan Elbaum
Ben
Ben Johnston
Brendan Goldman
Jessica Parker
Ksenia Ryzhova
Rachel Richman
Rebecca Sutton Koeser
benj@princeton.edu
benjohnsto
mrustow
rlskoeser

Please let me know whether or not you want me to preserve this contributor information in the migration.

--- 
documenting how I got the list: `git log --pretty=format:'%an' | sort | uniq`

mrustow:
I think it’s important to preserve all contributions to the editions at the document level if at all possible — seems like the best-practices thing to do. (We can of course merge the four Ben Johnstons and two RSKs.) Not preserving them would entail too much data loss. Sorry about the complexities this entails.

rlskoeser:
@mrustow thanks for confirming! I just wanted to make sure before I implemented it. I think it's valuable but wanted to check.

rlskoeser:
upgrading to 8 points for complexity

kseniaryzhova:
@rlskoeser I keep getting this message when I open up those folders - is this an issue? I don't have problems accessing the files though, at least the ones I can see.
![image](https://user-images.githubusercontent.com/105385128/191878393-d1b1251a-51e8-4fd5-af55-a23ca478c9bb.png)

rlskoeser:
[...]

PR #1128, opened 2022-09-27
Revise transcription export directory layout

rlskoeser:
this PR revises the directory layout for the transcription export, per conversation on https://github.com/Princeton-CDH/geniza/issues/912

here's an example of the previous directory structure:
https://github.com/Princeton-CDH/test-pgp-annotations/tree/test-tei-migration

and here is an example of what the new structure will look like:
https://github.com/Princeton-CDH/test-pgp-annotations/tree/rsk-dev2

(I only tested a few records, and it looks like I tested before I added leading zeros to make the top-level directories 5 digits)

### DIFF
--- a/geniza/corpus/annotation_export.py
+++ b/geniza/corpus/annotation_export.py
@@ -1,6 +1,7 @@
 import glob
 import json
 import logging
+import math
 import os
 from collections import defaultdict
 from urllib.parse import urlencode, urlparse
@@ -69,15 +70,7 @@ class AnnotationExporter:
 
         # define paths and ensure directories exist for compiled transcription
         annotations_output_dir = os.path.join(self.base_output_dir, "annotations")
-        transcription_base_dir = os.path.join(self.base_output_dir, "transcriptions")
-        # make a dict of output dir by format / file extension
-        transcription_output_dir = {}
-        for output_format in ["txt", "html"]:
-            format_path = os.path.join(transcription_base_dir, output_format)
-            transcription_output_dir[output_format] = format_path
-            os.makedirs(format_path, exist_ok=True)
-
-        # identify content to backup
+        # identify content to export
 
         # if ids are specified, limit to just those documents
         if self.pgpids:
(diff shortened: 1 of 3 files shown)


geniza/corpus/annotation_export.py
```diff
@@ -111,18 +104,21 @@ def export(self, pgpids=None, modifying_users=None, commit_msg=None):
             # use PGPID for annotation directory name
             # path based on recommended uri pattern from the spec
             # {prefix}/{identifier}/list/{name}
-            doc_output_dir = os.path.join(
-                annotations_output_dir, str(document.id), "list"
+            doc_output_dir = self.document_path(document.pk)
```
quadrismegistus, line 107: Switch to a method for getting path is nice

geniza/corpus/annotation_export.py
```diff
@@ -199,6 +195,15 @@ def export(self, pgpids=None, modifying_users=None, commit_msg=None):
         # commit and push (if configured) all the exported files
         self.commit_changed_files(updated_filenames, remove_filenames)
 
+    def document_path(self, pgpid):
+        """Generate path based on pgpid so records are chunked by 1000s,
+        then nested by pgpid to avoid too many files in a single directory."""
+
+        # round pgpid to nearest 1000; pad with leading zeros for < 1000
```
quadrismegistus, line 202: Seems like nice straightforward solution

geniza/corpus/annotation_export.py
```diff
@@ -111,18 +104,21 @@ def export(self, pgpids=None, modifying_users=None, commit_msg=None):
...
-            os.makedirs(doc_output_dir, exist_ok=True)
+            doc_transcription_dir = os.path.join(self.base_output_dir, doc_output_dir)
+            # ensure output directories exist
+            os.makedirs(doc_annotations_dir, exist_ok=True)
+            os.makedirs(doc_transcription_dir, exist_ok=True)
 
             # get a list of existing files so we can cleanup removed content
             # - annotation lists
-            doc_existing_files = glob.glob(os.path.join(doc_output_dir, "*.json"))
+            doc_existing_files = glob.glob(os.path.join(doc_annotations_dir, "*.json"))
```
quadrismegistus, line 118: Glob doesn't sort, which always makes me nervous, but it doesn't seem like sorting is necessary here since it seems like `doc_existing_files` is just put into a set later on in `extra_doc_files = set(doc_existing_files) - set(doc_updated_files)`.

quadrismegistus (approved):

Issue #1034, opened 2022-08-24
Quotation marks not working in Arabic script searches


richmanrachel:
**Describe the bug**
As discussed in #789 and #948 , there is some problem with the frontend search and quotation marks around Arabic content. Even when a phrase is known to be in the Geniza, adding the quotation marks produces "0 results"

**To reproduce/expected behavior**
1. Go to test site.
2. Search for "ثوب دبيقي" and compare to results for Judaeo-Arabic "תוב דביקי" (a type of robe), there should be 11 results
3. See that Arabic search results are 0 when there are real entries with this phrase, despite without the quotation marks there are 145 results

rlskoeser:
🤦‍♀️ it somehow only just occurred to me, this is almost certainly due to the way I'm modifying the query to put in the Judaeo-Arabic alternates – it's mangling the phrase somehow (still, even after my last revision). I think this is why it's only a problem for Arabic. Good news is, now I know where in the code I need to work on a fix. :-)

richmanrachel:
@rlskoeser - great! Thanks so much!

rlskoeser:
A fix for exact phrase searching in Arabic is on the test site now. I tested with the example in this issue and another Arabic phrase I found by searching and it looks like it's working, but it would be good to have someone familiar with the content and scripts confirm. Please try a variety of searches and different syntax to make sure they all work as expected.

kseniaryzhova:
@rlskoeser works for me - 11 results for both Arabic search term and same search term in JA (also the same documents). Closing!

PR #1134, opened 2022-09-28
Feature/ja phrases

quadrismegistus:
Fix to https://github.com/Princeton-CDH/geniza/issues/1034. 

The solution relies on having a tokenizer for the query which separates the string into a list of words OR phrases:

```python
def tokenize_words_and_phrases(text):
    import re
    return [x for x in re.findall(r'"[^"]*"|[\w]+', text) if x]
```

Then the arabic_to_ja code can be run on each token in that list:

```python
def arabic_or_ja_allowing_phrases(text, boost=True):
    new_terms = []
    for word_or_phrase in tokenize_words_and_phrases(text):
        judeo_arabic_equivalent = arabic_to_ja(word_or_phrase)
        if contains_arabic(word_or_phrase):
            if term_is_phrase(word_or_phrase):
                new_term = f"({judeo_arabic_equivalent} OR {word_or_phrase})"
            else:
                new_term = f"({judeo_arabic_equivalent}|{word_or_phrase})"
        else:
            new_term = word_or_phrase

        new_terms.append(new_term)

    text = " ".join(new_terms)

    # fix query
    text = text.replace("((", "(")
    text = text.replace("))", ")")
    return text
```

The problem is some of my tests are failing in a bizarre way. Here is what the test is:

```python
# make sure broken quotes still work
assert (
    arabic_or_ja_allowing_phrases('"تعطل شغله', boost=False)
    == '"(شغله|שגלה) (تعطل|תעטל)'
)
```

Note that the Judeo-Arabic is *before* the Arabic.

Now see what Python thinks the test was:

* What I see in my terminal (note Judeo-Arabic is *after* the Arabic):
Screen Shot 2022-09-28 at 12 01 06

* What I see pasting that identical text (note Judeo-Arabic is before Arabic again!?):
```
        # make sure broken quotes still work
>       assert arabic_or_ja_allowing_phrases('"تعطل شغله', boost=False) == '(شغله|שגלה) (تعطل|תעטל)'
E       AssertionError: assert '(תעטל|تعطل) (שגלה|شغله)' == '(شغله|שגלה) (تعطل|תעטל)'
E         - (شغله|שגלה) (تعطل|תעטל)
E         + (תעטל|تعطل) (שגלה|شغله)
```

tldr I have a fix but going a bit crazy with word/character order issues.

### DIFF
--- a/geniza/corpus/ja.py
+++ b/geniza/corpus/ja.py
@@ -79,55 +79,32 @@ def arabic_to_ja(text):
     return re.sub(re_he_final_letters, lambda m: he_final_letters[m.group(0)], text)
 
 
-def tokenize_words_and_phrases(text):
-    # Get a list of search query tokens which are each either:
-    #    - a phrase (text inside double quotes)
-    #    - a word
-    return re.findall(r'"[^"]*"|[^ "]*[^ "]', text)
-
-
-def arabic_or_ja_allowing_phrases(text, boost=True):
-    new_terms = [
-        (
-            f"({arabic_to_ja(word_or_phrase)}|{word_or_phrase})"
-            if contains_arabic(word_or_phrase)
-            else word_or_phrase
-        )
-        for word_or_phrase in tokenize_words_and_phrases(text)
-    ]
-    new_text = " ".join(new_terms)
-    new_text = new_text.replace(") )", ")").replace("( (", "(")
-    return new_text
-
-
 def arabic_or_ja(text, boost=True):
-    """Convert text to arabic or judaeo-arabic string; boost arabic by default"""
-    # if there is no arabic text, return as is
-    if not contains_arabic(text):
-        return text
+    # any arabic phrase or word
+    sr = re.compile(r'"[ \u0600-\u06FF]+"|[\u0600-\u06FF]+')
 
-    # extract arabic words from the search query
+    # find matches
+    arabic = sr.findall(text)
 
-    arabic_words = re_AR_letters.findall(text)
-    # generate judaeo-arabic equivalents
-    ja_words = [arabic_to_ja(word) for word in arabic_words]
-    # iterate over the original and converted words together and combine
+    # find what surrounds matches
+    nonarabic = sr.split(text)
 
-    # prob something like this?
-    # ja_phrase = arabic_to_ja(text)
-    ##
+    # assert the logical
+    assert len(nonarabic) == len(arabic) + 1  # given split/findall
 
-    # add boosting so arabic matches will be more relevant,
-    # unless boosting is disabled
+    # stitch back together
     boost = "^2.0" if boost else ""
+    tokens = []
+    for i in range(len(nonarabic)):
+        nonarabic_word = nonarabic[i]
+        tokens.append(nonarabic_word)
 
-    for i, arabic_word in enumerate(arabic_words):
-        ja_word = ja_words[i]
-        # if the words differ, combine them as an OR
-        # then replace them in the original search query
-        # (preserving any existing search syntax like quotes, wildcards, etc)
-        if arabic_word != ja_word:
-            ar_or_ja_word = "(%s%s|%s)" % (arabic_word, boost, ja_word)
-            text = text.replace(arabic_word, ar_or_ja_word)
+        # arabic word?
+        if len(arabic) > i:
+            arabic_word = arabic[i]
+            ja_word = arabic_to_ja(arabic_word)
+            ar_or_ja = f"({arabic_word}{'^2.0' if boost else ''}|{ja_word})"
+            tokens.append(ar_or_ja)
 
-    return text
+    # return as list of tokens
+    return "".join(tokens)
--- a/geniza/corpus/ja.py
+++ b/geniza/corpus/ja.py
@@ -79,6 +79,72 @@ def arabic_to_ja(text):
     return re.sub(re_he_final_letters, lambda m: he_final_letters[m.group(0)], text)
 
 
+# def locate_quoted_phrases(text):
+#     import re
+#     return [m.span() for m in re.finditer(r'"[^"]*"', text)]
+
+# def extract_quoted_phrases(text):
+#     return [text[m[0]:m[1]] for m in locate_quoted_phrases(text)]
+
+# def extract_arabic_phrases(text):
+#     return [
+#         phrase for phrase in extract_quoted_phrases(text) if contains_arabic(phrase)
+#     ]
+
+
+# def extract_arabic_words_outside_phrases(text):
+#     arabic_phrases = extract_arabic_phrases(text)
+
+#     # remove those arabic phrases from original string
+#     for phrase in arabic_phrases:
+#         text = text.replace(phrase, "")
+
+#     # get remaining arabic words
+#     arabic_words = extract_arabic_words(text)
+
+#     # return words
+#     return arabic_words
+
+
+# def extract_arabic_words(text):
+#     return re_AR_letters.findall(text)
+
+
+def tokenize_words_and_phrases(text):
+    import re
+
+    return [x for x in re.findall(r'"[^"]*"|[\w]+', text) if x]
+
+
+def term_is_phrase(text):
+    text = text.strip()
+    return text and " " in text and text[0] == '"' and text[-1] == '"'
+
+
+def arabic_or_ja_allowing_phrases(text, boost=True):
+    # replace arabic phrases
+    new_terms = []
+
+    for word_or_phrase in tokenize_words_and_phrases(text):
+        judeo_arabic_equivalent = arabic_to_ja(word_or_phrase)
+        if contains_arabic(word_or_phrase):
+            if term_is_phrase(word_or_phrase):
+                new_term = f"({judeo_arabic_equivalent} OR {word_or_phrase})"
+            else:
+                new_term = f"({judeo_arabic_equivalent}|{word_or_phrase})"
+        else:
+            new_term = word_or_phrase
+
+        new_terms.append(new_term)
+
+    text = " ".join(new_terms)
+
+    # fix query
+    text = text.replace("((", "(")
+    text = text.replace("))", ")")
+    return text
+
+
 def arabic_or_ja(text, boost=True):
     """Convert text to arabic or judaeo-arabic string; boost arabic by default"""
     # if there is no arabic text, return as is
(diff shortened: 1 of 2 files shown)


geniza/corpus/ja.py
```diff
@@ -79,18 +79,89 @@ def arabic_to_ja(text):
...
+
+
+# def extract_arabic_words(text):
+#     return re_AR_letters.findall(text)
+
+
+def tokenize_words_and_phrases(text):
+    import re
+
+    return [x for x in re.findall(r'"[^"]*"|[\w]+', text) if x]
```
rlskoeser, line 116: is there a way to adapt this regex to revise the regex I was using before so that we find and extract only arabic phrases and words? I think it's fine to limit to exact phrases that only contain arabic letters and whitespace (and possibly punctuation?); seems reasonable to me to assume that if the person searching uses other characters beyond that, they know what they're doing and we shouldn't mess with their query.

geniza/corpus/ja.py
```diff
@@ -79,18 +79,89 @@ def arabic_to_ja(text):
...
+    # replace arabic phrases
+    new_terms = []
+
+    for word_or_phrase in tokenize_words_and_phrases(text):
+        judeo_arabic_equivalent = arabic_to_ja(word_or_phrase)
+        if contains_arabic(word_or_phrase):
+            if term_is_phrase(word_or_phrase):
+                new_term = f"({judeo_arabic_equivalent} OR {word_or_phrase})"
+            else:
+                new_term = f"({judeo_arabic_equivalent}|{word_or_phrase})"
```
rlskoeser, line 134: Seeing this in context, it seems obvious that we should be using the same syntax for both cases — they are equivalent, no need to complicate the code. Let's stick with the shorter format: `(term1|term2)`

geniza/corpus/ja.py
```diff
@@ -79,18 +79,89 @@ def arabic_to_ja(text):
...
         return text
 
     # extract arabic words from the search query
+
     arabic_words = re_AR_letters.findall(text)
     # generate judaeo-arabic equivalents
     ja_words = [arabic_to_ja(word) for word in arabic_words]
     # iterate over the original and converted words together and combine
 
+    # prob something like this?
```
rlskoeser, line 161: I think if you revise this regex to return phrases and words the rest of the logic may work exactly the way we want

rlskoeser (commented):
I think you have the seed of an elegant, simple solution here.  

One thing to keep in mind: this function runs on every keyword search string entered, so it needs to be quick and efficient, and only make the minimum amount of changes needed to accomplish our goals.

geniza/corpus/ja.py
quadrismegistus: One reason I turned to tokenizing the entire search query is that it makes replacing elements in the search easier. My original approach of using string replace failed because if an arabic word is present as a single search term as well as in a phrase, then (with the old code) it could have been replaced with a (judeo-arabic|arabic) equivalent already. Better to have a list of the elements in the search, I thought.

There's prob a way of modifying this so it only considers Arabic letters: ```r'"[\u0600-\u06FF]+"|[\u0600-\u06FF]+'```. But then we get a list of just those elements and it becomes difficult to return a replaced version of the original search query string.

geniza/corpus/ja.py
quadrismegistus: You're right, that's much cleaner. Here's a streamlined version of the code:

```python
def arabic_or_ja_allowing_phrases2(text, boost=True):
    new_terms = [
        (
            f"({arabic_to_ja(word_or_phrase)}|{word_or_phrase})"
            if contains_arabic(word_or_phrase)
            else word_or_phrase
        )
        for word_or_phrase in tokenize_words_and_phrases(text)
    ]
    new_text = ' '.join(new_terms)
    new_text = new_text.replace('))',')').replace('((','(')
    return new_text
```

geniza/corpus/ja.py
```diff
@@ -79,18 +79,89 @@ def arabic_to_ja(text):
...
+#     # return words
+#     return arabic_words
+
+
+# def extract_arabic_words(text):
+#     return re_AR_letters.findall(text)
+
+
+def tokenize_words_and_phrases(text):
+    import re
```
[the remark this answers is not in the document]
quadrismegistus: fixed in latest

geniza/corpus/ja.py
rlskoeser: I apologize, I didn't read your regex carefully enough! I assumed it was doing something a little bit different.

What I meant to propose was that we revise the regex to this:
```python
re_AR_tokens = re.compile(r"\"[ \u0600-\u06FF]+\"|[\u0600-\u06FF]+")
```

You can then run the conversion and substitution on the entire token and insert the replacement back into the search string. And you don't have to worry about the string replace problem you describe above because arabic words inside phrases are only matched once.

geniza/corpus/ja.py
```diff
@@ -80,28 +80,31 @@ def arabic_to_ja(text):
 
 
 def arabic_or_ja(text, boost=True):
-    """Convert text to arabic or judaeo-arabic string; boost arabic by default"""
-    # if there is no arabic text, return as is
-    if not contains_arabic(text):
-        return text
+    # any arabic phrase or word
+    sr = re.compile(r'"[ \u0600-\u06FF]+"|[\u0600-\u06FF]+')
```
rlskoeser, line 84: let's move the regex compile step outside the method (so we only compile once) and name it something meaningful

geniza/corpus/ja.py
```diff
@@ -80,28 +80,31 @@ def arabic_to_ja(text):
...
-    for i, arabic_word in enumerate(arabic_words):
-        ja_word = ja_words[i]
-        # if the words differ, combine them as an OR
-        # then replace them in the original search query
-        # (preserving any existing search syntax like quotes, wildcards, etc)
-        if arabic_word != ja_word:
-            ar_or_ja_word = "(%s%s|%s)" % (arabic_word, boost, ja_word)
-            text = text.replace(arabic_word, ar_or_ja_word)
+    # assert the logical
+    assert len(nonarabic) == len(arabic) + 1  # given split/findall
```
rlskoeser, line 93: we can drop the assertion now, right?

geniza/corpus/tests/test_ja.py
```diff
@@ -37,8 +41,42 @@ def test_arabic_or_ja__arabic():
 
 
 def test_arabic_or_ja_exact_phrase():
-    assert arabic_or_ja('"تعطل شغله"', boost=False) == '"(تعطل|תעטל) (شغله|שגלה)"'
+    # make sure basic or is working
+    assert arabic_or_ja('"تعطل شغله"', boost=False) in {
+        '("תעטל שגלה"|"تعطل شغله")',
+        '("تعطل شغله"|"תעטל שגלה")',
+    }
```
rlskoeser, line 48: do you still need the variations for the tests to pass locally? would like to clean these up to assert only what we expect, which should be arabic first based on the way you're constructing it

geniza/corpus/ja.py
```diff
@@ -80,28 +80,31 @@ def arabic_to_ja(text):
...
+    tokens = []
+    for i in range(len(nonarabic)):
+        nonarabic_word = nonarabic[i]
+        tokens.append(nonarabic_word)
+
+        # arabic word?
+        if len(arabic) > i:
+            arabic_word = arabic[i]
+            ja_word = arabic_to_ja(arabic_word)
+            ar_or_ja = f"({arabic_word}{'^2.0' if boost else ''}|{ja_word})"
```
rlskoeser, line 106: the `if boost` here is redundant given line 96

geniza/corpus/tests/test_ja.py
```diff
@@ -1,6 +1,10 @@
 from operator import contains
 
-from geniza.corpus.ja import arabic_or_ja, arabic_to_ja, contains_arabic
+from geniza.corpus.ja import (  # extract_arabic_phrases,; extract_arabic_words,; extract_arabic_words_outside_phrases,; extract_quoted_phrases,; locate_quoted_phrases,
```
rlskoeser, line 3: please clean up the commented out imports

geniza/corpus/ja.py
```diff
@@ -80,28 +80,31 @@ def arabic_to_ja(text):
...
-            text = text.replace(arabic_word, ar_or_ja_word)
+    # assert the logical
+    assert len(nonarabic) == len(arabic) + 1  # given split/findall
 
-    return text
+    # stitch back together
+    boost = "^2.0" if boost else ""
+    tokens = []
+    for i in range(len(nonarabic)):
+        nonarabic_word = nonarabic[i]
```
rlskoeser, line 99: cleaner to use enumerate
```suggestion
    for i, nonarabic_word in enumerate(nonarabic):
```

geniza/corpus/ja.py
rlskoeser, line 106: I wonder if it would be more readable to convert all the arabic words first in a list comprehension before merging the arabic and nonarabic tokens back together?

geniza/corpus/tests/test_ja.py
```diff
@@ -37,8 +41,42 @@ def test_arabic_or_ja__arabic():
...
+    assert arabic_or_ja('"تعطل شغله" etc etc شغله', boost=False) in {
+        '("תעטל שגלה"|"تعطل شغله") etc etc (شغله|שגלה)',
+        '("تعطل شغله"|"תעטל שגלה") etc etc (شغله|שגלה)',
+        '("תעטל שגלה"|"تعطل شغله") etc etc (שגלה|شغله)',
+        '("تعطل شغله"|"תעטל שגלה") etc etc (שגלה|شغله)',
+    }
+
     # proximity
-    assert arabic_or_ja('"تعطل شغله"~10', boost=False) == '"(تعطل|תעטל) (شغله|שגלה)"~10'
+    # @TODO assert arabic_or_ja('"تعطل شغله"~10', boost=False) == '"(تعطل|תעטל) (شغله|שגלה)"~10'
```
rlskoeser, line 67: is this todo still needed?

rlskoeser (commented):
It's great to see this working.

For some reason I find it a bit hard to read the logic where you're stitching the arabic and non-arabic tokens back together. I think merging the lists could be done with `zip`, but IDK if that's necessarily more readable. Adding some comments would be enough; and maybe converting the arabic tokens to the ar or ja equivalent first before merging would help.

Please cleanup the unneeded comments and the variation results in the tests. If you still need the variants locally for tests to pass, let's investigate a bit and see if we can figure it out.

We can merge it in after the cleanup is done.

geniza/corpus/tests/test_ja.py
quadrismegistus: I still need them. darned if I know why

geniza/corpus/ja.py
quadrismegistus: I meant to remove line 96 and keep this

geniza/corpus/tests/test_ja.py
rlskoeser: so weird! after you've done your other cleanup, I'll pull this branch and see if it works for me locally without them, and we can check it if works in CI too

rlskoeser (approved):

PR #1139, opened 2022-09-29
jrl iiif

rlskoeser:
in this PR:
- one-off script to crawl JRL iiif collection and harvest manifest urls & reference numbers/shelfmark
- csv resulting from this crawl
- one-off script to use that csv file to generate remixed manifests that we can publish and main ourselves
- tweaks to our manifest importing to handle the remixed manifests

partial set of the generated manifests can be seen here:
https://github.com/princetongenizalab/iiif/tree/main/jrl


blms (approved):
Looks good, I'm fine with the dark mode invert solution if it seems to work.

Do we want to keep the one-off script around? Probably best to keep track of it somewhere.

I think we're good to merge once the failing tests are addressed.

rlskoeser:
> Looks good, I'm fine with the dark mode invert solution if it seems to work.

It works ok; I'm not sure it exactly matches their light color logo, but maybe it's good enough. (Their manifests don't even have their logo, it uses the one from the iiif platform they're using! So I guess we're improving on that.)

> Do we want to keep the one-off script around? Probably best to keep track of it somewhere.

I was thinking to keep it around as a way of documenting how the manifests were generated and as a starting point for if someone has to do similar work later on (whether redoing these or a different set).  I was planning to put a  short readme in that part of the iiif repo documenting how it was created.

> I think we're good to merge once the failing tests are addressed.

yeah, whoops — missed a bunch of mock patches that have the wrong target.

Issue #1087, opened 2022-09-15
accessibility issues flagged by DubBot


rlskoeser:
- [x] view document details link on search results needs aria label: view details for [title] (multiple links with same name)
- [x] #1088
- [x] add aria label "permalink" to permalink on document details (url as text of link)
- [x] #1089
- [x] pagination nav sections: landmark must have a unique aria-label, aria-labelledby, or title 
- [x] #1090

rlskoeser:
We've closed all the tasks listed in this issue; I'll go ahead and close this one and we can create more issues as we discover problems

rlskoeser:
ah, the issues were checked off but I forgot that was for implementation and many of the changes are still in the PR that is causing a lighthouse error; re-opening this to remind myself to revisit — want to try cherry-picking fixes into a new branch one at a time so I can figure out what's breaking the lighthouse check

richmanrachel:
@rlskoeser - should this be closed now or are there issues we need to keep on our radar?

rlskoeser:
I probably never circled back to close it after the sub-tasks were completed, thanks for reminding me. Let's close it.

rlskoeser:
reopening since the changes were made but never merged due to a failing lighthouse check; changes are in #1116 and have been updated to patch against latest version of develop branch; can be merged as soon as we resolve the lighthouse check

PR #1143, opened 2022-09-30
Add aria labels to links flagged by DubBot #1087

rlskoeser:


rlskoeser:
cherry picking just one commit from the other a11y fix branch still resulted in the lighthouse failure, will need to try again; maybe I can work through the changes incrementally and test again locally to see what's breaking the build

PR #1145, opened 2022-10-03
Feature/new exporter

quadrismegistus:
Draft of PR for a new data exporting method


rlskoeser (commented):
It looks like you're on the right track. I think your next steps will be to make the new document resource match the fields in the existing admin export, and then remove the old export code (including the url and the export link, which I think we put in a local admin template override — I can point you to the right place if you have trouble finding it.)

We'll want to make sure the resource export uses the queryset prefetching we have setup in the `DocumentAdmin` so that exporting everything is at least somewhat reasonable in terms of performance.

Given how big our admin file is already, seems like it would be nice to put our resource objects in a different file (depending on how complex they get). I guess we could do `geniza.corpus.resource` since that's the naming convention that django-import-export uses, although it's kind of generic.

rlskoeser (commented):
I'm hopeful that we don't need as many field declarations and dehydrate methods as you have here — I think some of them at least should not be needed? When I implemented in PPA I was able to use a widget to customize the foreign key display and didn't have to write a dehydrate method; maybe that approach could help with some of the outputs we have here (although I don't know how that impacts import functionality).

I'm a little concerned about performance, with all these attributes and methods — I find it hard to believe that this is going to be usable to download all 30k records.  I couldn't find anything in the django-import-export docs that talked about performance or speed. What's the largest set of records you've tried it with? And can you tell if it uses the query with prefetching that we have configured in the `DocumentAdmin` object?

It might be acceptable if a full on-demand download isn't available once we have the data regularly synchronized to GitHub, and smaller downloads work... but I'd like to make it as efficient as we reasonably can.

Side note: the "resource" in `geniza.corpus.resource` is such a vague term. What do you think about `io_resource` ? (although maybe that's potentially misleading too.) We already have other export functionality that is different from this, or I might think about calling it `import_export`.

quadrismegistus:
So after discussion with @rlskoeser it seems that django-import-export is not a viable solution for our CSV export needs. It took about 26 minutes for the full 32k rows to export. I suspect this is owing to the javascript-like usage of a function `dehydrate_field(doc): ...` for each field, which necessitates recalculating complex properties of doc. We've decided to return to using the LOC tabular export module (https://github.com/LibraryOfCongress/django-tabular-export) with some improvements.

Issue #1138, opened 2022-09-29
Add Arabic type styles to the site 


gissoo:
https://github.com/Princeton-CDH/geniza/issues/1135

blms:
Confirmed present in QA

Issue #1127, opened 2022-09-27
Right-to-left site design implementation revisions


blms:
## qa checklist

- [x]  Misaligned "color bars" for metadata
- [x]  Content pages list bullets/numbers on wrong side
- [x]  Issue with short labels in nav
- [x] ITT panel transcription content hidden
- [x]  Missing space between ITT panel toggles
- [x]  Wrong spacing for collapsed ITT panel toggles
- [x]  Wrong spacing for image header label and controls
- [x]  Image header label partially offscreen on mobile
- [x]  Backwards zoom slider track coloration
- [x]  Wrong spacing for permalink

---
Remaining issues in RTL site:

- [x] [Misaligned "color bars" for metadata](https://user-images.githubusercontent.com/7234006/192540234-570c870a-1fce-4c9c-a606-ef6efedd4e38.png)
- [x] Content pages list bullets/numbers on wrong side
- [x] [Issue with short labels in nav](https://user-images.githubusercontent.com/7234006/193826008-e7069f8d-8f04-4658-99af-e8a850bcd516.png)
- ~~ITT panel [transcription content hidden](https://user-images.githubusercontent.com/7234006/192540407-4f656e0a-f079-4a58-a67d-00649f3d7b93.png)~~ this was because I was on the wrong branch…
- [x] Missing space between ITT panel toggles
- [x] [Wrong spacing for collapsed ITT panel toggles](https://user-images.githubusercontent.com/7234006/192540766-d4ed824c-a975-4bd5-b4fa-0c9441ad0ce9.png)
- [x] Wrong spacing for image header label and controls
- [x] [Image header label partially offscreen on mobile](https://user-images.githubusercontent.com/7234006/192540803-6a31c775-eb9d-4d79-b553-1a880c273f5a.png)
- [x] [Backwards zoom slider track coloration](https://user-images.githubusercontent.com/7234006/192540663-9393085e-1384-4957-a201-04813654c321.png)
- [x] Wrong spacing for permalink

blms:
Some issues remain:

- Transcriptions still not showing up in RTL even when they do in LTR. The content of the div is just completely empty so it seems it's failing to get the annotations.
- The page label somehow is still left-aligned, so it's wrongly spaced on desktop and partially offscreen on mobile. (Thought I'd fixed that!)

blms:
All confirmed fixed in QA

PR #1147, opened 2022-10-04
RTL revisions, language switcher alignment, Arabic type styles

blms:
## In this PR

- Per #1127:
  - Fix all listed RTL CSS problems in that issue
  - Wrap menu item link text nodes in `span` to prevent flex row-wrap weirdness
- Per #1113:
  - Fix alignment of items in language switcher
- Per #1138:
  - Add new Arabic fonts, type styles

## Questions

- Seems Lighthouse is failing on `total-byte-weight`, probably due to the new Arabic fonts. Should we adjust the tolerance for that, or find some way to mitigate?


blms:
@rlskoeser Still failing that lighthouse `total-byte-weight` assertion, even with the conditional 🫤

I did notice that the document detail page and search results perform the worst, which makes sense since they are both now loading the second transcription font for Arabic… but we don't actually use that font yet, since we set [all transcriptions](https://github.com/Princeton-CDH/geniza/blob/36bcbf82839dafd6cae3dfc94c2f1bdd6fcd4b70/sitemedia/scss/components/_transcription.scss#L896) to use the Hebrew font for now (we have a [TODO](https://github.com/Princeton-CDH/geniza/blob/36bcbf82839dafd6cae3dfc94c2f1bdd6fcd4b70/sitemedia/scss/base/_typography.scss#L295) about it). How should we handle that?

But, even with that, the other pages are still just barely failing their assertions, and I'm not sure why! Maybe it's sufficient for those to adjust the `total-byte-weight` tolerances slightly.

rlskoeser:
@blms thanks for investigating. I'm ok with increasing our tolerance (and documenting why!) or making it a warning instead of an error.

blms:
@rlskoeser Done—feel free to take a quick look at the code if you have time, or I can just go ahead and merge since it's mostly CSS.

rlskoeser:
@blms thanks — glanced at the diff, I think you can go ahead and merge

PR #1152, opened 2022-10-04
revise annotation export sync

rlskoeser:


geniza/annotations/models.py
```diff
@@ -42,6 +42,17 @@ def group_by_canvas(self):
                 annos_by_canvas[anno.target_source_id].append(anno)
         return annos_by_canvas
 
+    def group_by_manifest(self):
+        """Aggregate annotations by manifest uri; returns a dictionary of lists,
+        keys are manifest uri, items are lists of annotations."""
```
blms, line 47: Great to see this convenience method 🙏

geniza/corpus/management/commands/sync_annotation_export.py
```diff
@@ -1,22 +1,132 @@
...
+from geniza.annotations.models import Annotation
 from geniza.corpus.annotation_export import AnnotationExporter
+from geniza.corpus.models import Document
 
 
 class Command(BaseCommand):
     """Synchronize annotation backup data with GitHub"""
 
+    # filename for last run information (stored in current user's home)
+    lastrun_filename = os.path.join(os.path.expanduser("~"), ".pgp_export_lastrun")
```
blms, line 21: Does this need to live outside of the project directory tree? I would have put it somewhere like `data/` but maybe that won't work?

geniza/corpus/management/commands/sync_annotation_export.py
```diff
@@ -1,22 +1,132 @@
...
+            print("%d annotation log entries since %s" % (log_entries.count(), lastrun))
+
+        # generate exports based on what has been changed
+        if log_entries.exists():
+
+            # TODO: handle deletions (object no longer in db)
+
+            # make a dictionary of modified annotations and the users who modified them
+            modified_annotations = defaultdict(list)
+            for log_entry in log_entries:
```
blms, line 62: This block could maybe be a method, though I could also see wanting to reduce the amount of stuff needed to pass around, and wanting to call `self.anno_exporter.export` in this main handle method. But it could make testing easier maybe.

geniza/corpus/management/commands/sync_annotation_export.py
```diff
@@ -1,22 +1,132 @@
...
+
+        # load information about the last run of this script
+        lastrun_data = self.get_lastrun_info()
+        # if the file exists, pull out modified value for this scriptdi
+        if lastrun_data and self.script_id in lastrun_data:
+            return datetime.fromisoformat(lastrun_data[self.script_id])
+
+        # if lastrun file is not found, use last git commit on the repository
+        # (potentially unreliable if non-data repo content is updated
+        # and lastrun file does not exist, but should be ok.)
```
blms, line 122: Could potentially mitigate with protected branches if this is a big concern, though sounds like it's not. (Also from a quick glance, not entirely sure there's a protected branch rule we could use here in the first place…)

blms: Another possibility would be some kind of identifier in the commit message that indicates the commit was done programmatically by the exporter, and then finding the last git commit that matches this identifier.

blms (approved):
Looking great, I like the overall solution a lot! Will be curious to hear how you plan to handle deletions.

geniza/corpus/management/commands/sync_annotation_export.py
rlskoeser: Oh, interesting idea. I _think_ in normal operation we should have the last run file in place, so my hope is that the last git commit is just a reasonable, logical fallback so we don't re-export too much history. But you're right, we could probably filter fairly easily on commit messages if we need to.

geniza/corpus/management/commands/sync_annotation_export.py
rlskoeser: Hmm. I'm adapting this from code I wrote for PEMM, so I just did something similar to that. It doesn't need to be, but I feel like home directory might be a bit more durable than data. I think we're always going to be running it as the deploy user (or ourselves for dev), so it doesn't need to be in a shared location across accounts.

blms: That makes sense!

blms (approved):
Deletion logic and tests look great. I also didn't know about `override_settings`, that's useful!

PR #1157, opened 2022-10-05
Fix issue with FrankRuhl font vertical metrics (#1113)

blms:
## In this PR

- Per #1113:
  - Update FrankRuhl1924MF font with correct vertical metrics, and adjust margins for new version
- Update deploy notes

## Questions

- While editing the deploy notes to make note of this font change, I noticed that we hadn't consolidated the transcription prototype notes under a release version yet. Was that intended to stay separate, or is it ok to put it under 4.9.0?
- I don't seem to have delete permissions in the [Google Drive folder](https://drive.google.com/drive/u/0/folders/1efxagFKcC-Rs2clzYFBHa38yVYBdHjnk). Would it be possible for you to delete the old FrankRuhl files? They're the ones modified in December, without dashes in the filename.


rlskoeser (approved):
- added the updated fonts to ansible vault with Ben over zoom
- deploy notes were not integrated because I didn't know what version it would go in, we should clean up before we release

Issue #1150, opened 2022-10-04
add a configurable banner that can be displayed during the TEI migration and then turned off


rlskoeser:
hopefully we can adapt from the test banner we used to have and use on other sites; should make the text configurable (or maybe overridable with default test banner language), and should be able to enable/disable using a feature flags approach in django settings

rlskoeser:
Tested in qa and it works beautifully — enabled it; revised the text; disabled it.

Worth noting, it does require an nginx reload before the changes take effect.

@blms I forgot to discuss with you, is the warning banner also included in our admin base template? And if not, how much trouble to add it?

PR #1160, opened 2022-10-05
Restore test banner and make configurable (#1150)

blms:
## In this PR

- Per #1150:
  - Restore test banner JS and update CSS (replace unused font; `max-width` and padding on `p` prevent viewport overflow issue)
  - Make test banner body/header text configurable
  - Use feature flag string instead of boolean

## Questions

- Do the config options need any documentation beyond just being present with comments in `local_settings.py.sample`? (e.g. deploy notes)


rlskoeser (approved):
This looks great!

Would it be much trouble to rename everything from `TEST_WARNING_*` to something like `WARNING_BANNER` ? (now that we're using it for more than just differentiating the test site.)

I think documenting it in the local settings and deploy notes is probably enough, since it's only configurable by developers.

Issue #883, opened 2022-05-21
In .csv downloads from the admin interface, for joins, the IIIF_url field needs a space after the semicolon.  


mrustow:
## testing notes

Download a document CSV for some set of records with iiif urls and documents on multiple fragments.

Check that the urls in the iiif_urls field are separated by a space and then a semicolon, and that the urls are clickable and don't include the semicolon in the url.

----

A tiny but annoying thing, which I'm only posting because I assume it's easy to fix:

I downloaded a .csv from the admin interface. One of the rows contains a join. The iiif_url field puts the IIIF urls of all the associated fragments with a semicolon and no space between: 

https://cudl.lib.cam.ac.uk/view/MS-TS-00016-00114;https://cudl.lib.cam.ac.uk/view/MS-TS-00024-00057;;https://cudl.lib.cam.ac.uk/view/MS-TS-AS-00146-00195 
 
Google sheets reads that as a single link and can't parse it. Can the default instead be semicolon followed by space, like this?

https://cudl.lib.cam.ac.uk/view/MS-TS-00016-00114; https://cudl.lib.cam.ac.uk/view/MS-TS-00024-00057; https://cudl.lib.cam.ac.uk/view/MS-TS-AS-00146-00195 

Also for some reason, the second and third URLs are separated by _two_ semicolons. Not sure what's going on there. 

Thanks!

rlskoeser:
Downloaded a csv of records with iiif images from the test site and confirmed resolved

kseniaryzhova:
@rlskoeser in the production admin, the csv does not have a space and a semicolon between IIIF links. Instead it's a semicolon then space. In the test site admin, the csv just downloads as blank (same data set)

kseniaryzhova:
![image](https://user-images.githubusercontent.com/105385128/197599746-68c5a0b5-f3da-4502-b2e2-217b8f2fc73c.png)
confirming that the links are spearated by a space, then a semicolon and another space. But they are still not linkable, not when first opened in Excel.

kseniaryzhova:
@rlskoeser closing, because I spoke to Marina and she said this isn't a priority, we can revisit with Performant later.

PR #1161, opened 2022-10-06
Fixing semicolon problem in iiif_urls

quadrismegistus:
This should be a fix for issue #883.

Switching join string from ';' to '; ' within fields of CSV export.

Currently the fix does this for all fields (IMO this looks cleaner and is more consistent) and not only the "iiif_urls" field.

The code also switches a few list comprehensions to safer (checks against empty strings) and generator version.


rlskoeser (commented):
The reason I wasn't omitting empty strings in the lists was so that (in theory) you could match up the ones that go together across the different cells. I don't know if that was documented anywhere, but I think it's still valuable to to keep that logic.

Otherwise this looks fine. Should we use a variable for the separator to ensure we're consistent?

quadrismegistus:
Thanks for comments! Latest should resolve them?

Issue #953, opened 2022-07-15
As a content editor, I want the admin csv download to include transcription (Y/N) and translation (Y/N) so that I can filter documents to this with or without transcription or translation.


kseniaryzhova:
**Is your feature request related to a problem? Please describe.**
Building off of issues #951 and #952, we want the admin csv download to mirror the document list view options on the admin site.

**Describe the solution you'd like**
See above - add transcription (Y/N) and translation (Y/N) to the admin csv download. This EXCLUDES bibliographic information for the transcriptions and translations in order to keep the download manageable. 

**Describe alternatives you've considered**
N/A

quadrismegistus:
I'm working on this now

kseniaryzhova:
@rlskoeser confirm that I see separate columns for has translation and has transcription in the csv download, and that they are marked Y/N. Closing, thank you!

Issue #475, opened 2021-12-27
As a content editor, I want scholarship record summary information included in documents metadata so I can quickly see who has published on the document without switching context.


mrustow:
## testing notes

- [x] the documents CSV export should now include a `scholarship_records` field
- [x] should appear after the description
- [x] should include the short display version of the footnote record for all footnotes associated with that document
- [x] should handle multiple footnotes
- [x] should be empty if there are no footnotes
- [x] download should still be reasonably quick (expect it to be similar to before)

----
As a content admin, I would like footnotes included in .csv exports from the admin interface so that I can see who has published the document just by looking at the subset of docs I've exported, without referring back to the admin interface and without having to input the information manually.

Admittedly, the resulting .csv will be unwieldy, but it's already unwieldy and one can easily jettison columns, whereas it's difficult to add information that isn't there.

mrustow:
It would be hugely useful if we had info at the level that we get in the drop-down menu when adding a transcription, so, e.g., "Khan, Arabic Legal and Administrative Documents" or "Goitein, unpublished transcriptions, T-S 8." Minimal is better than maximal!

rlskoeser:
@mrustow ok, thanks for elaborating. And just to confirm, you prefer to have all this with the documents metadata, not in a secondary file?

mrustow:
right. most intuitive for me would be in the column to the right of the description, because that's where it was in the old days.

kseniaryzhova:
@rlskoeser I think we can close (or merge?), given we have a separate csv for footnotes that has all of this information listed.

rlskoeser:
@kseniaryzhova thank you for reminding me about this! I meant to take a look and see how easily we can add to the export and how much it impacts the time. I know it would be easier for Marina if we have that information included in the main documents spreadsheet.

kseniaryzhova:
@rlskoeser works for me! Thank you, closing!

PR #1163, opened 2022-10-06
New export CSV command and related changes

quadrismegistus:
This PR does:
* Create a new manage.py command called `export_csv`
* The output filename defaults to a timestamped `pgp_documents.csv`
* It writes code in a new file `geniza/corpus/csv_export.py` for:
    * Pre/fetching the documents
    * Then populating related information as the CSV file is written
* The command has a nice progress bar from `rich`.

Todo: 
- [x] Finish sphinx docs
- [x] Get tests working
- [x] Move Exporter to common
- [x] Move and adapt test from test_corpus_admin


quadrismegistus:
Are these issues we want to resolve here? These also relate to the csv export, though to the admin interface:
* #475 
* #953

rlskoeser (commented):
I thought we discussed using methods from the `tabular_export` library. I think we should use the same methods they call to generate the http response.

I think we should rename `csv_export` to `export_metadata` — that's how the project team thinks of this data (the github repo will be called pgp-metadata), and hopefully we'll be able to add JSON as well as CSV. 

I would like to use the same export logic in the admin and in the script, rather than duplicating code and risking it getting out of sync. Can you look at pulling out the common portions from admin (which is what, maybe fields and export logic?) into the new export file you've created and then update admin to reference that shared code? Maybe it's just a mixin with the relevant methods and properties

rlskoeser:
> Are these issues we want to resolve here? These also relate to the csv export, though to the admin interface:
> 
>     * [Include footnotes in .csv exports #475](https://github.com/Princeton-CDH/geniza/issues/475)
> 
>     * [As a content editor, I want the admin csv download to include transcription (Y/N) and translation (Y/N) #953](https://github.com/Princeton-CDH/geniza/issues/953)

I think it's better to do them independently so we don't have too many changes in a single PR. If you're refactoring the current document csv tabulate logic, maybe it's better to do the changes for #953 in a second pass after that change.

I'll ping Marina and Ksenia about #475 to ask what the minimum useful information is; but I think it could slow the export down significantly (have to go from document -> footnote -> source -> creator; I think we're not pulling that much in for the export currently).

My goal is that the admin csv export and the scripted csv export are basically the same, and use all the same code (with maybe a flag to exclude/include the admin-only/non-public fields).

quadrismegistus:
Keeping notes as I go. Current times for exporting all rows:
* Current admin interface: 12 min
* New manage.py logic: 12 min      # makes sense, it's roughly identical to previous, just modularized
* django-import-export: 26 min

quadrismegistus:
Okay, the latest that we discussed (along with custom query manager used across admin and CLI exports) should be in these commits.

geniza/common/utils.py
```diff
@@ -76,3 +76,18 @@ def custom_tag_string(tag_string):
...
+
+TIMENOW = time.time()
+
+
+def timeprint(*x, **y):
+    global TIMENOW
+    now = time.time()
+    print(f"[+{now-TIMENOW:.3f}s]", *x, **y)
+    TIMENOW = now
+
```
rlskoeser, line 91: If you want to keep the duration information (which I think would probably be helpful for testing/optimizing), I recommend putting that information in debug logging. Here's an example of how I've done that in the past:

https://github.com/Princeton-CDH/geniza/blob/main/geniza/annotations/signals.py#L73-L78

geniza/corpus/metadata_export.py
```diff
@@ -0,0 +1,239 @@
...
+    fn="pgp_documents.csv", docs=None, csv_fields=CSV_FIELDS, **kwargs
+):
+    timeprint("write_stream_export_data_for_docs()")
+    for row in stream_export_data_for_docs(
+        fn=fn, docs=docs, csv_fields=csv_fields, **kwargs
+    ):
+        pass
+
+
+def http_stream_export_data_for_docs(
```
rlskoeser, line 229: as we discussed : let's rename export methods so they are not document specific, and we can reuse the same approach with all our other exports; I guess they should take a queryset and a filename; what else? does it need to take a metadata generator instead of a queryset?

geniza/corpus/models.py
```diff
@@ -449,11 +450,33 @@ def unidecode_tag(sender, instance, **kwargs):
         instance.name = unidecode(instance.name)
 
 
+class MetadataManager(models.Manager):
```
rlskoeser, line 453: Instead of a secondary manager, let's use a custom queryset with a method we can call when we want to apply this prefetching.  Maybe the method on the queryset could be something like `metadata_prefetch` ; or if there are multiple groups of prefetching we need in different places, we could have two different methods (they can stack).

Then we can replace the default manager (`Document.objects`) with one generated from our queryset: https://docs.djangoproject.com/en/3.2/topics/db/managers/#creating-a-manager-with-queryset-methods

Then in the admin or metadata export, we can do `Document.objects.metadata_prefetch()` or whatever you decide to call the method(s). And in admin, this method can be applied to the existing queryset passed in (can revert to previous logic where we call `super()` and don't have to apply any ordering).

geniza/corpus/metadata_export.py
```diff
@@ -0,0 +1,239 @@
...
+    )
+    outd["languages_primary"] = doc.all_languages()
+    outd["languages_secondary"] = doc.all_secondary_languages()
+    outd["language_note"] = doc.language_note
+    outd["doc_date_original"] = doc.doc_date_original
+    outd["doc_date_calendar"] = doc.doc_date_calendar
+    outd["doc_date_standard"] = doc.doc_date_standard
+    outd["notes"] = doc.notes
+    outd["needs_review"] = doc.needs_review
+    outd[
```
rlskoeser, line 156: please preserve the comment I had about this in admin (I was going to ask why you're not using `reverse` here but the comment in the old code explains why!)

rlskoeser:
@quadrismegistus I had a thought about how some of the logic here might be grouped: what if you have something like a `DocumentMetadataExport` class that has a queryset method (default queryset, apply prefetching), an attribute or method for list of csv fields), default or basename for exported filename?

quadrismegistus:
> @quadrismegistus I had a thought about how some of the logic here might be grouped: what if you have something like a `DocumentMetadataExport` class that has a queryset method (default queryset, apply prefetching), an attribute or method for list of csv fields), default or basename for exported filename?

Interesting! What it would be a subclass of?

rlskoeser:
> Interesting! What it would be a subclass of?

Either a brand new class, or you could make a common metadata export class with common methods that we can use across the different exports we're going to implement — although I don't think we know enough yet to know how it should generalize. Was thinking of this partly because your current export methods have some code specific to `Document` so it might be easier to think of it as class methods, where the class provides the queryset, csv fields, filename, etc.

But I don't want to make this any harder than what we need to get it done quickly this week! So probably only do that if you think it will be helpful in the short term.

geniza/corpus/metadata_export.py
```diff
@@ -0,0 +1,221 @@
...
+    def csv_filename(self):
+        return (
+            f'geniza-{self.str_plural}-{timezone.now().strftime("%Y%m%dT%H%M%S")}.csv'
+        )
+
+    def get_queryset(self):
+        return self.model.objects.metadata_prefetch() if not self.qset else self.qset
+
+    def get_export_data_dict(self, obj):
+        # @NOTE: THIS NEEDS TO BE SUBCLASSED
```
rlskoeser, line 73: could raise `NotImplementedError` in the base class if this is required to be extended

geniza/corpus/metadata_export.py
```diff
@@ -0,0 +1,221 @@
...
+            fn = self.csv_filename()
+        iterr = self.iter_export_data_as_csv(pseudo_buffer=True)
+        response = StreamingHttpResponse(iterr, content_type="text/csv; charset=utf-8")
+        response["Content-Disposition"] = f"attachment; filename={fn}"
+        return response
+
+
+class DocumentExporter(Exporter):
+    model = Document
+    str_plural = "documents"
```
rlskoeser, line 125: should be able to get this from the plural verbose name on the model: `Documents._meta.verbose_name_plural`

https://docs.djangoproject.com/en/4.1/ref/models/options/#verbose-name-plural

geniza/corpus/metadata_export.py
```diff
@@ -0,0 +1,219 @@
...
+            fn = self.csv_filename()
+        iterr = self.iter_export_data_as_csv(pseudo_buffer=True)
+        response = StreamingHttpResponse(iterr, content_type="text/csv; charset=utf-8")
+        response["Content-Disposition"] = f"attachment; filename={fn}"
+        return response
+
+
+class DocumentExporter(Exporter):
+    model = Document
+    csv_fields = CSV_EXPORT_FIELDS_DOC
```
rlskoeser, line 124: let's declare the fields here instead of referencing a global variable

geniza/corpus/metadata_export.py
```diff
@@ -0,0 +1,219 @@
...
+from geniza.common.utils import Echo, timeprint
+from geniza.corpus.models import Document, TextBlock
+
+
+class Exporter(object):
+    model = None
+    csv_fields = []
+
+    def __init__(
+        self, qset=None, progress=CSV_EXPORT_PROGRESS, sep_within_cells=SEP_WITHIN_CELLS
```
rlskoeser, line 28: I would put default separator as a base class attribute instead of a global.

Not sure we need a global for progress, seems like it should be optionally on / on by default for the manage command and off by default for http?

geniza/corpus/metadata_export.py
```diff
@@ -0,0 +1,219 @@
...
+
+script_user = settings.SCRIPT_USERNAME
+site_domain = Site.objects.get_current().domain.rstrip("/")
+url_scheme = "https://"
+
+from geniza.common.utils import Echo, timeprint
+from geniza.corpus.models import Document, TextBlock
+
+
+class Exporter(object):
```
rlskoeser, line 23: I like the exporter class structure! It seems like this should work pretty well to adapt to the other exports we need to implement. 

FYI, in python 3 you can omit the `(object)` — this is exactly equivalent:

```suggestion
class Exporter:
```

geniza/corpus/admin.py
```diff
@@ -540,19 +428,10 @@ def merge_documents(self, request, queryset=None):
...
-        )
 
-        return export_to_csv_response(
-            self.csv_filename(),
-            self.csv_fields,
-            self.tabulate_queryset(queryset),
-        )
+        queryset = queryset or self.get_queryset(request)
+        exporter = DocumentExporter(qset=queryset, progress=False)
+        return exporter.http_export_data_csv()
```
rlskoeser, line 434: this is pretty slick

rlskoeser (commented):
this seems like it's really coming along and will generalize well for the other exports we need to add/revise

quadrismegistus:
Thanks for all these very helpful comments! I should have incorporated them in the last few commits.

geniza/corpus/management/commands/export_metadata.py
```diff
@@ -0,0 +1,28 @@
+import time
+from asyncore import write
```
rlskoeser, line 2: what is this? (do you by any chance have the vs code setting turned on that tries to autoimport for you? that was always adding weird imports to my code until I found it and turned it off)

geniza/corpus/metadata_export.py
```diff
@@ -0,0 +1,223 @@
...
+from django.conf import settings
+from django.contrib.postgres.aggregates import ArrayAgg
+from django.contrib.sites.models import Site
+from django.db.models.query import Prefetch
+from django.http import StreamingHttpResponse
+from django.utils import timezone
+from rich.progress import track
+
+script_user = settings.SCRIPT_USERNAME
+site_domain = Site.objects.get_current().domain.rstrip("/")
```
rlskoeser, line 13: This requires hitting the database, so you shouldn't set it as a global. Probably best to get it in exporter init

geniza/corpus/models.py
```diff
@@ -449,11 +450,35 @@ def unidecode_tag(sender, instance, **kwargs):
...
+                    "textblock_set",
+                    queryset=TextBlock.objects.select_related(
+                        "fragment", "fragment__collection"
+                    ),
+                ),
+            )
+            .annotate(shelfmk_all=ArrayAgg("textblock__fragment__shelfmark"))
+            .order_by("shelfmk_all")
+            # .order_by("id")   # @TODO: Where does this go?
+            # .prefetch_related("secondary_languages", "log_entries")
```
rlskoeser, line 472: this one is needed for csv export but not for admin; secondary prefetch method we can call in the export code but skip in admin?

geniza/corpus/models.py
```diff
@@ -449,11 +450,35 @@ def unidecode_tag(sender, instance, **kwargs):
...
+                Prefetch(
+                    "textblock_set",
+                    queryset=TextBlock.objects.select_related(
+                        "fragment", "fragment__collection"
+                    ),
+                ),
+            )
+            .annotate(shelfmk_all=ArrayAgg("textblock__fragment__shelfmark"))
+            .order_by("shelfmk_all")
+            # .order_by("id")   # @TODO: Where does this go?
```
rlskoeser, line 471: I think both of these `order` calls should be removed

geniza/corpus/metadata_export.py
rlskoeser: In fact, this is probably why your unit tests are failing!

rlskoeser (commented):
I think this is getting close, would love to get it merged so we can deliver for testing and work on the follow up data exports. I think if we can get the test & doc checks passing, we should go ahead and do that.

I updated from develop and dealt with the merge conflict because it was closer to my current dev environment. (FYI you'll need to switch your local env to python 3.9)

I added placeholders to the sphinx docs for the new export code; you can build locally (should be instructions in the dev notes) to see what methods its flagging, but should also get an alert from the docs coverage check on this PR.

rlskoeser:
FYI: tweaked the queryset code in a couple of places (it was more obvious when I was testing locally than on the PR review for some reason). I restored the additional prefetches we had for the CSV before that we don't need for admin list view, and it sped up my export substantially.

quadrismegistus:
Ok, finished 3 of 4 todos at top. The 4th todo is getting all the unit tests to pass. There are 3 still failing:

```
FAILED geniza/corpus/tests/test_corpus_admin.py::TestDocumentAdmin::test_tabulate_queryset
FAILED geniza/corpus/tests/test_corpus_admin.py::TestDocumentAdmin::test_export_to_csv
FAILED geniza/corpus/tests/test_corpus_models.py::test_items_to_index - Asser...
```

These look like old export test functions. I can remove or clean them tomorrow morning. After that we should be pretty much good to go?

geniza/common/tests.py
```diff
@@ -329,3 +331,25 @@ def test_item_count(self):
...
+@pytest.mark.django_db
+def test_base_exporter():
+    exporter = Exporter()
+
+    # raises correct error?
+    try:
+        exporter.get_export_data_dict(obj=None)
+        assert False  #
+    except NotImplementedError:
+        pass
```
rlskoeser, line 355: oh, actually — you can use pytest raises for this

https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

should look something like this:

```suggestion
def test_base_exporter():
    exporter = Exporter()

    # raises correct error?
    with pytest.raises(NotImplementedError):
        exporter.get_export_data_dict(obj=None)
```

Pretty sure this doesn't hit the db, so you don't need the django_db mark

rlskoeser (approved):
one suggestion for the expected exception test, but otherwise I think this is ready to merge!

Issue #1164, opened 2022-10-06
Exempt IIIF URLs from i18n patterns


blms:
> maybe we need to define separately and include in our main urls.py differently? might mean the urls will have a different namespace, but once we get the url config right I think everything else should work

Originally posted by @rlskoeser on Slack

rlskoeser:
@blms I checked an annotation created by the tei migration after this fix was pushed to the test site, and the document and source uris in the annotation both look good to me. Is there any other checking you want to do on this?

blms:
@rlskoeser Great, thanks! I also just checked the `/iiif/annotations/` endpoints for a few documents with transcriptions and they all look fine. I think we're good to close.

PR #1165, opened 2022-10-06
Exempt IIIF URIs from i18n patterns (#1164)

blms:
Spent some time trying to get the regex pattern matching right until I finally realized I just needed to move this url config above the `documents/` match from wagtail 😂 

## In this PR

- Per #1164:
  - Move the IIIF URIs from `corpus/urls.py` to `corpus/uris.py`
  - Add additional url pattern in global `urls.py`, moving them to namespace `corpus-uris`
  - Update all references to the old namespace

### DIFF
--- a/geniza/annotations/tests/test_annotations_signals.py
+++ b/geniza/annotations/tests/test_annotations_signals.py
@@ -26,7 +26,8 @@ class TestCreateOrDeleteFootnote:
                     "source": {
                         "partOf": {
                             "id": reverse(
-                                "corpus:document-manifest", kwargs={"pk": document.pk}
+                                "corpus-uris:document-manifest",
+                                kwargs={"pk": document.pk},
                             )
                         }
                     }
--- a/geniza/annotations/tests/test_annotations_views.py
+++ b/geniza/annotations/tests/test_annotations_views.py
@@ -172,7 +172,9 @@ class TestAnnotationSearch:
 
     def test_search_uri(self, client, document, annotation):
         # content__target__source__id=target_uri)
-        manifest_uri = reverse("corpus:document-manifest", kwargs={"pk": document.pk})
+        manifest_uri = reverse(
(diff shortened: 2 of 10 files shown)


blms:
> Changes look great. Is it helpful if I pull and test anything locally? (and if so, anything specific to test?)

I don't think so!

> Do we have the same problem for our source uris? or does it not happen there because we're not currently using a real url?

Correct, I just checked all the annotations and the source URIs are built straight from the base URI with no i18n prefix.

@rlskoeser I noticed that in tests where we check for 404 on bad manifests and annotation list views, we're now getting a 302 to the i18n version, and then 404 there. Is that OK, or should we be intercepting those 302s and making sure they're just 404s? (If we should, I think we may need to do one of [these solutions](https://code.djangoproject.com/ticket/17734).)

rlskoeser:
> @rlskoeser I noticed that in tests where we check for 404 on bad manifests and annotation list views, we're now getting a 302 to the i18n version, and then 404 after following the redirect. Is that OK, or should we be intercepting those 302s and making sure they're just 404s? (If we should, I think we may need to do one of [these solutions](https://code.djangoproject.com/ticket/17734).)

The redirected 404 is confusing, it's better if we can fix. Doesn't look too hard...

And we already have a `LocaleMiddleware` ! We obviously ran into this before. Right now it's only checking the first part of the url path, but it shouldn't be too hard to add another check for the new urls we want to exempt.

blms:
@rlskoeser Is this solution ok, or too brittle? I considered using regex to match any "/iiif/" but I think we'll only ever encounter it in the place I've specified.

rlskoeser:
> @rlskoeser Is this solution ok, or too brittle? I considered using regex to match any "/iiif/" but I think we'll only ever encounter it in the place I've specified.

I'd rather not use a regex if we don't need one! I think this is fine — but could we add a quick example uri in the comment, just as a reminder of what this is for / what it's doing? I think once you add that it's good to merge

Issue #1174, opened 2022-10-07
As a frontend user, I want search results to include partial matches of words in transcriptions, so that I can search by substrings of words.


blms:
## testing notes

Try searching for transcription content without using full words; you should get matches where the word contains your partial word, and phrases should be most relevant. This should work with or without quotes. Two details to be aware of:
- I have the ngram filter configured with two letters as a minimum, so two letters will be enough to match a word but one letter will not. LMK if this is not sufficient.
- Right now Solr is highlighting the entire word instead of just the part of the word that matched, which is frustrating. I did a little investigating into this and found examples of other people wrestling with this problem. None of their solutions worked immediately for me, and I thought it was more important to get the partial search functionality out.  (We should probably make a separate issue for that.)

Here's one of my test searches to get you started:
- [search documents for partial transcription text (test site)](https://test-geniza.cdh.princeton.edu/en/documents/?q=%D7%A2%D7%A9%D7%94+%D7%91%D7%99%D7%AA+%D7%99%D7%A2%D7%96%D7%A8+%D7%9C%D7%94&docdate_0=&docdate_1=&has_transcription=on&sort=relevance)
- [same partial search, but as an exact phrase](https://test-geniza.cdh.princeton.edu/en/documents/?q=%22%D7%A2%D7%A9%D7%94+%D7%91%D7%99%D7%AA+%D7%99%D7%A2%D7%96%D7%A8+%D7%9C%D7%94%22&docdate_0=&docdate_1=&has_transcription=on&sort=relevance)

richmanrachel:
The search for דלאל (and other terms in Ju-Ar) is much better than before! I'm getting more variations on the word without having to use the *.

kseniaryzhova:
@rlskoeser all good for transcription search!

Issue #1162, opened 2022-10-06
shelfmark sort in document search should use natural sorting


rlskoeser:
What I recommend is that we implement natural sorting for shelfmarks in Solr. I know how it needs to be done but didn't want to do too many changes at once, and I wasn't sure how essential this is. If you want to create an issue, I can add notes to the documentation / example config that I found (or maybe even add it quickly in next week!)

_Originally posted by @rlskoeser in https://github.com/Princeton-CDH/geniza/issues/863#issuecomment-1270055023_

kseniaryzhova:
@rlskoeser sort works logically, thank you! Only concern is the sorting by shelfmark for joins (also logical, but can be confusing if the first shelfmark on the join is not what you searched for). Kicking over to Performant as it will probably need help text/more time to improve!

richmanrachel:
Good call to close for now. Was writing an idea for help-text, but then realized that only Geniza researchers would likely search this way (so Alan and others can muddle through for now). 

I think with Performant, it would be ideal if the shelfmark sort privileged whatever was in the search bar (so assuming someone wants to page through T-S AS, they would put T-S AS in the search and then sort by shelfmark). Then joins between T-S AS ___ + AIU would appear in the search results with the T-S AS join in the place that the T-S AS fragment would appear if it were not a join. 

@kseniaryzhova - does that make sense?

Issue #1151, opened 2022-10-04
As a frontend user, I want keyword search for Seleucid dates to give me complete matches first so that I can browse by decreasing relevancy in the date field. 


kseniaryzhova:
**Is your feature request related to a problem? Please describe.**
Syntax search for Seleucid dates (document_date: 1450 seleucid) brings back 5 exact matches. [https://test-geniza.cdh.princeton.edu/en/documents/?q=document_date%3A+%221450+seleucid%22&docdate_0=&docdate_1=&sort=relevance](https://test-geniza.cdh.princeton.edu/en/documents/?q=document_date%3A+%221450+seleucid%22&docdate_0=&docdate_1=&sort=relevance) Keyword search for 1450 Seleucid brings back 1,364 results (and the first two are actually shelfmark matches - the 1450 is part of the shelfmark and not in the date field). [https://test-geniza.cdh.princeton.edu/en/documents/?q=1450+seleucid&docdate_0=&docdate_1=&sort=relevance](https://test-geniza.cdh.princeton.edu/en/documents/?q=1450+seleucid&docdate_0=&docdate_1=&sort=relevance) 

**Describe the solution you'd like**
I want keyword searches for Seleucid dates (when I use Seleucid dates as a phrase "1450 seleucid") to give me exact date matches first, followed by other partial matches. 

**Describe alternatives you've considered**
@rlskoeser suggested we tweak phrase boosting to make terms together more important in the search results.

richmanrachel:
@rlskoeser - I'm still getting the same results as Ksenia did :(

kseniaryzhova:
@rlskoeser keyword search is giving me exact matches first, followed by the shelfmark matches and then partial matches everywhere else. Works for me!

PR #1192, opened 2022-10-12
document search improvements

rlskoeser:
Solr config changes in this PR:
- new custom field for transcription text, with bigram filter #1174 
- revise boosting for date text #1151 and phrase matches across the board
- natural sorting for shelfmarks #1162


blms (approved):
Looks great, thanks for the clear test cases so I understand the implications of the solr changes.

Issue #1176, opened 2022-10-10
Transcriptions are failing to save in the editor


blms:
## Testing notes (QA)

In the transcription editor for a document:
- [x] Create and save a new annotation and confirm it persists in the editor
- [x] Edit and save an existing annotation and confirm the change persists in the editor
- [x] Delete an existing annotation and confirm it is deleted successfully

## dev notes

- [x] Fix save failure condition
- [x] raise events on save/delete failure
- [x] raise events on save/delete success
- [x] display errors/success messages in frontend

blms:
About to jump into a meeting, so before I forget: I think the save failure is another Annotorious/async race condition.

First we `await` the [Annotorious promise](https://github.com/Princeton-CDH/annotorious-tahqiq/blob/e208fd859002052c7bb2e260fba6b0ca547b7f96/src/index.ts#L349-L350), which raises the following event:

https://github.com/Princeton-CDH/annotorious-tahqiq/blob/e208fd859002052c7bb2e260fba6b0ca547b7f96/src/storage.ts#L36-L39

That event handler `await`s [saving the annotation](https://github.com/Princeton-CDH/annotorious-tahqiq/blob/e208fd859002052c7bb2e260fba6b0ca547b7f96/src/storage.ts#L133) to storage.

Back in the original function, we [fetch all annotations](https://github.com/Princeton-CDH/annotorious-tahqiq/blob/e208fd859002052c7bb2e260fba6b0ca547b7f96/src/index.ts#L354-L355) from storage after the Annotorious promise resolves.

But there's no guarantee that our event handler has itself resolved. It's entirely possible (and indeed, it appears to be the case) that we are calling `this.storage.loadAnnotations()` before it has been saved to the storage). Should be as simple as calling that method in `storage.ts` `handleUpdateAnnotation` instead.

kseniaryzhova:
@rlskoeser works for me, thank you!

PR #1196, opened 2022-10-12
Show alert toasts on transcription editor (#1176)

blms:
## In this PR

see also https://github.com/Princeton-CDH/annotorious-tahqiq/pull/26

- Per #1176:
  - Add a new div for alerts bound to a new stimulus controller
  - In the new stimulus controller, listen for `tahqiq-alert` events, in order to populate div with alerts
  - Add basic styles for different alert statuses
  - Bump tahqiq commit hash

## Questions

- Please test locally, is the overall look and feel OK?
- Should we add publishing the new tahqiq + bumping the version number in PGP (if applicable) to the release checklist?

### DIFF
--- a/sitemedia/js/controllers/alert_controller.js
+++ b/sitemedia/js/controllers/alert_controller.js
@@ -1,0 +1,46 @@
+import { Controller } from "@hotwired/stimulus";
+
+export default class extends Controller {
+    initialize() {
+        // bind "this" so we can append elements
+        this.boundAlertHandler = this.handleAlert.bind(this);
+    }
+    connect() {
+        // avoid adding a duplicate event listener, it happens sometimes(?)
+        // regardless of the number of images on the page; bound to alerts
+        // div in document_transcribe.html... TODO: figure out why
+        if (!this.element.alertListenerInitialized) {
+            document.addEventListener("tahqiq-alert", this.boundAlertHandler);
+            this.element.alertListenerInitialized = true;
+        }
+    }
+    disconnect() {
+        document.removeEventListener("tahqiq-alert", this.boundAlertHandler);
+    }
+    handleAlert(evt) {
+        // create a new div with the alert message and appropriate classes
+        const { message, status } = evt.detail;
+        const alert = document.createElement("div");
+        alert.textContent = message;
+        alert.className = "alert alert-visible";
+        if (status) {
+            alert.classList.add(`alert-${status}`);
+        }
+
+        // append the alert as child of alerts div
+        this.element.appendChild(alert);
+
+        // longer timeout for error messages for readability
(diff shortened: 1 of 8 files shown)


blms:
One other note: the event listener needs to be independent of Tahqiq instances, rather than binding one per instance; not sure if a new stimulus controller is best approach, but maybe better than binding it to theme or menu…

sitemedia/js/controllers/alert_controller.js
```diff
@@ -0,0 +1,46 @@
+import { Controller } from "@hotwired/stimulus";
+
```
rlskoeser, line 2: Please add comment with an overview/summary of what this controller is for.

Do you think it could be useful for other notifications at some point?

sitemedia/scss/components/_alert.scss
```diff
@@ -0,0 +1,53 @@
...
+        &.alert-success {
+            background-color: var(--alert-success);
+        }
+        // error msg, keeping this the same cross-theme to ensure it stands out
+        &.alert-error {
+            // allow copy-paste
+            pointer-events: all;
+            opacity: 1;
+            color: white;
+            background-color: #ce2029;
```
rlskoeser, line 50: should this be defined as a variable so it's with the other colors, even if it doesn't need to change for dark/light mode?

rlskoeser (approved):
I pulled and built tahqiq branch and this branch and tested locally, and it works great. I like the visual display, seems like a nice simple solution that will be visible anywhere you're working.

I forced some errors to see what would happen — when I shut down my django runserver, I got notifications about network errors; when I forced a 500 error in the annotation view, I got an error notification that said there was a 500 error saving the record. This seems like a huge improvement and will make it a lot easier for editors to see what's happening and tell us what errors they're getting when things go wrong.

rlskoeser:
> Should we add publishing the new tahqiq + bumping the version number in PGP (if applicable) to the release checklist?

Ah, good point. I was thinking of it as being included under this point in the checklist, but I see that it isn't quite there:

> Check python requirements for any internal dependencies that should be released (or at least pinned to a specific git commit)

But the software release checklist is org-wide. Maybe we can add a similar line about releasing any internal JS depenpencies?

blms:
> Maybe we can add a similar line about releasing any internal JS depenpencies?

That makes sense to me, releasing any internal JS dependencies and bumping the version number in `package.json` for the project.

sitemedia/js/controllers/alert_controller.js
blms: Yes, I think so. Right now it's only connected to a div in the transcribe template, but it could easily be moved anywhere and extended to respond to other kinds of events.

PR #1197, opened 2022-10-12
transcription lang attribute & primary script

rlskoeser:
I added document properties for primary language code and primary script, so that we can use them in document details template and index in solr for use in search results.

I put the script as a data attribute on the transcription container because I thought it would be a useful fallback for styling transcription content when we don't have language codes (or maybe is what we should rely on anyway, since it's more specific); I wasn't sure what attribute to put it in. Should it be a class instead?


blms (approved):
This looks great! Clear code, thorough test cases. Exactly what I need for the next steps.

As far as the attribute, I think the data attribute is just fine. We don't need to do anything class-specific (like JS `classList` manipulation), so they're functionally and semantically identical AFAICT.

PR #1203, opened 2022-10-13
Exports include transcription (Y/N) and translation (Y/N)

quadrismegistus:
This is a PR response to issue #953.

### DIFF
--- a/geniza/common/metadata_export.py
+++ b/geniza/common/metadata_export.py
@@ -26,6 +26,7 @@ class Exporter:
     model = None
     csv_fields = []
     sep_within_cells = " ; "
+    true_false = {True: "y", False: "n"}
 
     def __init__(self, queryset=None, progress=False):
         self.queryset = queryset
@@ -34,7 +35,7 @@ class Exporter:
         self.site_domain = Site.objects.get_current().domain.rstrip("/")
         self.url_scheme = "https://"
 
-    def csv_filename(self) -> str:
+    def csv_filename(self):
         """Generate the appropriate CSV filename for model and time
 
         :return: Filename string
@@ -44,7 +45,7 @@ class Exporter:
         str_time = timezone.now().strftime("%Y%m%dT%H%M%S")
         return f"geniza-{str_plural}-{str_time}.csv"
 
-    def get_queryset(self) -> QuerySet:
+    def get_queryset(self):
         """Get the queryset in use. If not set at init, this will be all objects from the given model.
 
         :return: QuerySet of documents to export
@@ -61,7 +62,7 @@ class Exporter:
         """
         raise NotImplementedError
 
-    def iter_export_data_as_dicts(self) -> Generator[dict]:
+    def iter_export_data_as_dicts(self):
         """Iterate over the exportable data, one dictionary per row
 
         :yield: Dictionary of information for each object
@@ -80,7 +81,7 @@ class Exporter:
         # save
         yield from (self.get_export_data_dict(obj) for obj in iterr)
 
-    def serialize_value(self, value: object) -> str:
+    def serialize_value(self, value: object):
         """A quick serialize method to transform a value into a CSV-friendly string.
 
         :param value: Any value
@@ -90,7 +91,7 @@ class Exporter:
         :rtype: str
         """
         if type(value) is bool:
-            return "y" if value else "n"
+            return self.true_false[value]
         elif value is None:
             return ""
         elif type(value) in {list, tuple, set}:
@@ -100,18 +101,7 @@ class Exporter:
         else:
             return str(value)
 
-    def serialize_key(self, key: object) -> str:
-        """Serialize the keys in the serialized dictionary. For now, this is identical to serializing values. Keys ought to be strings but this method enforces it.
-
-        :param key: Any object but it ought to be a string
-        :type key: object
-
-        :return: That object as a safe string
-        :rtype: str
-        """
-        return self.serialize_value(key)
-
-    def serialize_dict(self, data: dict) -> dict:
+    def serialize_dict(self, data):
         """Return a new dictionary whose keys and values are safe, serialized string versions of the keys and values in input dictionary `data`.
 
         :param data: Dictionary of keys and values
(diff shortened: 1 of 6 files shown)


geniza/corpus/models.py
```diff
@@ -486,12 +486,15 @@ def metadata_prefetch(self):
...
-                ),
+                # @TODO: This is causing error?
+                #       ValueError: 'textblock_set' lookup was already seen with a different queryset. You may need to adjust the ordering of your lookups.
+                #
+                # Prefetch(
+                #     "textblock_set",
+                #     queryset=TextBlock.objects.select_related(
+                #         "fragment", "fragment__collection"
+                #     ),
+                # ),
```
rlskoeser, line 497: ah, now that I see it in context maybe I know what is happening here — is it possible that there's some context in which you're calling the prefetch twice?

geniza/common/metadata_export.py
```diff
@@ -78,6 +78,22 @@ def iter_export_data_as_dicts(self):
...
+
+        :param value: Any value
+        :type value: object
+        :return: Stringified value
+        :rtype: str
+        """
+        if type(value) is bool:
+            return "y" if value else "n"
+        else:
+            return str(value)
```
rlskoeser, line 92: this is nice, could we do the same thing for lists?

geniza/corpus/models.py
```diff
@@ -851,6 +854,15 @@ def has_transcription(self):
...
 
+    def has_translation(self):
+        """Does document have a translation?"""
+        return any(
+            [Footnote.TRANSLATION in note.doc_relation for note in self.footnotes.all()]
+        )
+
+    has_translation.short_description = "Translation"
+    has_translation.boolean = True
+
```
rlskoeser, line 865: looks fine; the short description and boolean are for django admin, so if we're not using it there we don't need them

needs a unit test — parallel to the has transcription one I imagine

(now I'm wondering how much effort to generalize `has_x` for all our types; maybe something for the next time we add one)

geniza/corpus/models.py
rlskoeser: maybe admin queryset + export — are we calling the prefetch even if the queryset is passed in?

quadrismegistus: Perhaps it's this line in admin.py:

```python
def get_queryset(self, request):
        return super().get_queryset(request).metadata_prefetch()
```

quadrismegistus: I believe it is. This solves the tests and works on admin and CLI:
```python
def get_queryset(self, request):
        return super().get_queryset(request) #.metadata_prefetch()
```

geniza/common/metadata_export.py
quadrismegistus: Good idea. How about this:
```python
elif type(value) in {list,tuple,set}:
            return self.sep_within_cells.join(
                self.serialize_value(subval)
                for subval in sorted(list(value))
            )
```

geniza/corpus/models.py
quadrismegistus: > looks fine; the short description and boolean are for django admin, so if we're not using it there we don't need them

I see, gotcha

> needs a unit test — parallel to the has transcription one I imagine

How about this:
```python
def test_has_translation(self, document, source):
        # doc with no footnotes doesn't have translation
        assert not document.has_translation()

        # doc with empty footnote doesn't have translation
        fn = Footnote.objects.create(content_object=document, source=source)
        assert not document.has_translation()

        # doc with translation footnote does have a translation
        fn.doc_relation = [Footnote.TRANSLATION]
        fn.save()
        assert document.has_translation()
```

> (now I'm wondering how much effort to generalize `has_x` for all our types; maybe something for the next time we add one)

I agree that this is a good formulation and yeah let's keep it in mind if we end up expanding this/related fields.

quadrismegistus:
Looks like all tests have passed, including Sphinx/etc. Code coverage for me locally is 97%+ for the relevant files.

geniza/common/metadata_export.py
```diff
@@ -78,23 +80,51 @@ def iter_export_data_as_dicts(self):
...
+        """Return a new dictionary whose keys and values are safe, serialized string versions of the keys and values in input dictionary `data`.
+
+        :param data: Dictionary of keys and values
+        :type data: dict
 
-    def iter_export_data_as_csv(self, fn=None, pseudo_buffer=False):
+        :return: Dictionary with keys and values safely serialized as strings
+        :rtype: dict
+        """
+        return {self.serialize_key(k): self.serialize_value(v) for k, v in data.items()}
```
rlskoeser, line 123: Why is `serialize_key` needed? would really rather not have the extra method call just to enforce it's a string, when we control the keys in this code. At 30k row that will add up.

geniza/corpus/admin.py
```diff
@@ -312,7 +312,7 @@ def get_deleted_objects(self, objs, request):
         return deletable_objects, model_count, perms_needed, protected
 
     def get_queryset(self, request):
-        return super().get_queryset(request).metadata_prefetch()
```
rlskoeser, line 315: This might fix the unit tests, but it will make the admin list view unusable. I think the right way to fix it is for the export code to only apply the `metadata_prefetch` when an existing queryset is not passed in.

rlskoeser: In the `get_queryset` method of the Document exporter, we are currently applying all prefetching. I think it should be revised so that if `self.queryset` is set we only apply the things after the `metadata_prefetch` (and document that this is the case — if a large queryset is passed in, metadata prefetching should be applied before exporting).

geniza/common/metadata_export.py
```diff
@@ -78,7 +80,51 @@ def iter_export_data_as_dicts(self):
...
+        """A quick serialize method to transform a value into a CSV-friendly string.
+
+        :param value: Any value
+        :type value: object
+
+        :return: Stringified value
+        :rtype: str
+        """
+        if type(value) is bool:
+            return "y" if value else "n"
```
rlskoeser, line 93: could these defined as a lookup on the class? for potential override, similar to separator.  e.g. something like
```suggestion
            return self.true_false[value]
```

rlskoeser (commented):
Thanks for working on this.  Seems like it's close.

I prefer not to have type annotations when we aren't using them consistently across the codebase, please remove.

geniza/corpus/admin.py
quadrismegistus: The admin list view still works for me, if that's what I think it means? But I see what you mean about prefetching going on even when .queryset is set. 

This is current get_queryset in document exporter:
```python
def get_queryset(self):
    return (
        super()
        .get_queryset()
        .metadata_prefetch()
        .prefetch_related("secondary_languages", "log_entries")
        .order_by("id")
    )
```
How about this -- is this what you mean?

```python
def get_queryset(self):
    qset = self.queryset or self.model.objects.all().metadata_prefetch()
    qset = qset.prefetch_related("secondary_languages", "log_entries").order_by("id")
    return qset
```

quadrismegistus: I'm assuming that that secondary prefetching is necessary no matter whether an initial queryset was set or not.

rlskoeser (approved):
Looks good to me.

Issue #1202, opened 2022-10-13
Digital edition switcher not present in admin


blms:
## Testing notes (QA)

- [x] In the QA site admin for a document with multiple digitized transcriptions, such as PGPID 646, confirm you can click on "Edition: [Name]" (directly above the transcription text) and see a dropdown.
- [x] Confirm you can choose a different editor, and that it changes the visible transcription text to show the different editions.

## dev notes

In the admin ITT panel, for documents with multiple digital editions, the dropdown to switch between different digital editions does not appear.

- [x] Include in `iiif` bundle

kseniaryzhova:
@rlskoeser @blms works for me! Closing!

Issue #1194, opened 2022-10-12
Apply Arabic and inline type styles in transcriptions


blms:
- [ ] Apply type styles based on primary language (already done for Hebrew but not Arabic)
- [ ] Apply type styles for inline language tags

kseniaryzhova:
@blms @rlskoeser I do see these decimals on the transcription display in both public and admin
![image](https://user-images.githubusercontent.com/105385128/196517132-8a6febec-4a6b-495b-93a3-771a298104fe.png)

rlskoeser:
Closing this based on the primary functionality passing tests; should not have been kept open for the small display issue. If that remains an issue we can open a new ticket to investigate.

Issue #670, opened 2022-02-28
"Combining dot above" rendered incorrectly in transcriptions


blms:
## testing notes

Review the second transcription (by Goitein and Friedman) on PGPID 9083 linked below and check if the combining dot renders correctly. Recommend testing in multiple browsers, since before we saw different behavior on this with different browsers.

----

Unicode Character “◌̇” (U+0307) Combining Dot Above is missing from Frank Ruhl 1924 MF, so it does not render correctly in transcriptions.

Example: https://geniza.princeton.edu/en/documents/9083/

<img width="209" alt="Screen Shot 2022-02-28 at 1 52 26 PM" src="https://user-images.githubusercontent.com/7234006/156051957-8f96f623-1a43-4d54-9d99-3aa894454dda.png">

blms:
Additional missing glyphs in Arabic:
https://geniza.princeton.edu/en/documents/9140/

blms:
Seems to behave differently on different browsers, and is not resolved by Unicode normalization. Another example: https://geniza.princeton.edu/en/documents/4767/

May be related to fallback font behavior. Noting some different options here:

Default (Frank Ruhl):
![Screen Shot 2022-09-29 at 10 38 51 AM](https://user-images.githubusercontent.com/7234006/193061665-63d5c3c7-da6b-45b5-94ca-bf685f5ba01d.png)

`font-family: serif` or `font-family: Georgia` (what I thought our fallback was already!)
![Screen Shot 2022-09-29 at 10 38 18 AM](https://user-images.githubusercontent.com/7234006/193061720-a04a8f0e-5598-49df-b00b-ab65b43642ac.png)

`font-family: 'Greta Sans Hebrew Regular'`
![Screen Shot 2022-09-29 at 10 38 41 AM](https://user-images.githubusercontent.com/7234006/193061879-3b0bf15f-ee48-424e-b611-a28c1bc036f7.png)

`font-family: Times New Roman`
![Screen Shot 2022-09-29 at 10 38 29 AM](https://user-images.githubusercontent.com/7234006/193061943-8eac749a-003b-457e-b18d-9dff2851e80c.png)

All tested on Chrome.

kseniaryzhova:
@rlskoeser @blms It works partially (same transcription, test site)
![image](https://user-images.githubusercontent.com/105385128/196521958-a12e9223-7b74-4033-af70-951df4b78747.png)

rlskoeser:
[...]

PR #1204, opened 2022-10-13
Transcription styles (#1194) and admin fixes (#1193, #1202)

blms:
## In this PR

- Per #1194:
  - Apply Arabic type styles, and conditional type styles using inline lang attributes, to transcriptions
  - Adjust decimal cover
- Per #1193:
  - Apply proper fonts, sizes, line heights, decimal cover to transcription viewer in admin
- Per #1202:
  - Add digital edition switcher to admin (move it to `iiif` bundle from `main`)
- Ensure `alert_controller` excluded from `main` bundle (it should only be in `annotation`; this is why I was getting double `connect`s in Stimulus)

## Questions

- Should this include what we discussed with [`dir` attributes in the HTML](https://www.w3.org/International/questions/qa-html-dir)? If so, should I just apply RTL to all transcriptions for now, or do we want logic to determine it based on language direction?

### DIFF
--- a/sitemedia/css/admin-local.css
+++ b/sitemedia/css/admin-local.css
@@ -105,7 +105,11 @@ a#needsreview:target {
     white-space: pre-line;
 }
 
-/* Transcription styles in admin */
+/*
+ *
+ * Transcription styles in admin
+ *
+ */
 @media (min-width: 900px) {
     .form-row #iiif-viewer {
         padding: 10px;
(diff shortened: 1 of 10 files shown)


sitemedia/css/admin-local.css
```diff
@@ -114,16 +118,45 @@ a#needsreview:target {
...
+    font-family: "FrankRuhl 1924 MF Medium", serif;
+    line-height: 32px;
+}
+/* arabic script */
+.form-row .transcription[lang="ar"] p,
+.form-row .transcription[lang="ar"] li,
+.form-row .transcription[data-lang-script="arabic"] p,
+.form-row .transcription[data-lang-script="arabic"] li {
+    font-family: "Amiri", serif;
+}
```
rlskoeser, line 153: I wish we didn't have to duplicate these, but I guess it isn't that much.

If you think about a good way to consolidate common styles so we can get rid of the duplication at some point, LMK.

sitemedia/js/main.js
```diff
@@ -7,7 +7,7 @@ const application = Application.start();
 const context = require.context(
     "./controllers",
     true,
-    /^.*\/(?!iiif_controller)(?!annotation_controller).*\.js$/
+    /^.*\/(?!iiif_controller)(?!annotation_controller)(?!alert_controller)(?!transcription_controller).*\.js$/
```
rlskoeser, line 10: would you add a comment here about what this is doing? this is what you mean about the double error?

rlskoeser (approved):
Looks great. I downloaded all the new fonts and tested locally. I checked in search results with records that were tagged as primary language arabic, and I also tested with a transcription that had mixed content, and tagged the arabic section in the transcription editor — it's really obvious when it switches to the arabic font, and it looks great.

One question that I wish had occurred to me sooner: are the Arabic glyphs even in the Hebrew font? Could we just always have the Arabic transcription font applied as a fallback?

To your question about RTL: I think we should put it in the html where appropriate, and I'm pretty sure that all our transcriptions are RTL, so no need to check language.

rlskoeser:
I forgot to test styles in admin. Fonts are being loaded; I think there's a discrepancy in the marker spacing between admin and public site — possibly related to Arabic font size? Two digit line numbers are getting cut off slightly on the right in admin. (I'm testing with PGPID 1242)

blms:
> One question that I wish had occurred to me sooner: are the Arabic glyphs even in the Hebrew font? Could we just always have the Arabic transcription font applied as a fallback?

@rlskoeser That's a really good question. In fact, looking into it, it seems we can just restrict fonts to certain Unicode ranges. Should I just apply both font faces to all transcriptions restricting them by Unicode range? https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range

sitemedia/js/main.js
blms: Yes, the double error was that the stimulus controller for alerts was being loaded twice (once in the `main` bundle and once in the `annotation` bundle).

rlskoeser:
Do the changes in this PR also address https://github.com/Princeton-CDH/geniza/issues/670 ?

(Asking in part because we need to make sure the unicode range includes all of the characters we expect/allow)

I tested that one locally in both Firefox and Vivaldi and I don't see the boxes, although I'm not sure I'm seeing the accents either.

blms:
> Do the changes in this PR also address https://github.com/Princeton-CDH/geniza/issues/670 ?

@rlskoeser Not yet, it's the Goitein and Friedman transcription that has the problems though, I'm still seeing them on Chrome.

I could actually just set that broken character range to a different font like Times New Roman. I'll play around and see what I can do!

blms:
@rlskoeser That solution seems to work really well, including fixing the missing glyphs and the marker spacing. It has the side effect of reducing the repetition as well.

Would you mind taking another look locally to see if it works/if I missed anything? Your checks in the review were really helpful! 

FWIW, I've been testing with PGPIDs 646, 1390, 5468, 6176, 19011, and 36266.

rlskoeser (approved):
This is great! So glad you found this unicode range font feature, amazing. 

I found a good test case: a letter in JA with Arabic address (PGPID 3271). The Arabic is rendering beautifully without having to tag the language in the editor.

The only thing I can think of we might not be handling is coptic numerals, but I'm not sure we have any transcriptions or characters for those (and not really in scope here, either!). But you have a fallback font configured (and one of the ones Gissoo said is most complete in terms of glyphs), so that seems fine to me.

Issue #1195, opened 2022-10-12
Add newly assigned ISSN to footer


rlskoeser:
ISSN: 2834-4146

There isn't a placeholder for it in our existing PGP footer designs, but please do something similar to what we did in the S&co footer, which was implemented based on Gissoo's advice (check both mobile and desktop versions). https://shakespeareandco.princeton.edu/

rlskoeser:
confirmed the new ISSN is present on the footer on the test site and looks good in both desktop and mobile

PR #1205, opened 2022-10-13
Add ISSN to footer (#1195)

blms:
## In this PR

- Per #1195:
  - Add ISSN to footer, matching S&co styles
  - Fix resulting lighthouse issue

## Questions

- Does ISSN need a translation? (I assumed yes!)
- To match the S&Co site I needed to make a few other changes to the layout. Can you [take a quick look at Percy](https://percy.io/2cf28a24/geniza/builds/22188898/), and see if the change is acceptable?

### DIFF
--- a/geniza/locale/ar/LC_MESSAGES/django.po
+++ b/geniza/locale/ar/LC_MESSAGES/django.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: princeton-geniza-project\n"
 "Report-Msgid-Bugs-To: cdhdevteam@princeton.edu\n"
-"POT-Creation-Date: 2022-10-14 11:19-0400\n"
+"POT-Creation-Date: 2022-10-14 12:15-0400\n"
 "PO-Revision-Date: 2022-09-22 19:18\n"
 "Last-Translator: \n"
 "Language-Team: Arabic\n"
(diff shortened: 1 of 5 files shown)


rlskoeser (approved):
Looks great to me. Thanks for figuring out a way to adjust the design to accommodate the change.

I'm not sure about translating ISSN, but seems reasonable to me to flag it! Maybe crowdin will have suggested translations we can use or adapt.

rlskoeser:
@blms would you regenerate the translation files before merging?

blms:
@rlskoeser Sure, thanks for the reminder! I definitely haven't been doing that lately 😓 

Any idea why / if it matters that it's now prepending every path with `geniza/` in the translation files? I'm running it from the same place as usual.

rlskoeser:
> @rlskoeser Sure, thanks for the reminder! I definitely haven't been doing that lately 😓

I haven't run it in a while lately either, hard to remember...

> Any idea why / if it matters that it's now prepending every path with `geniza/` in the translation files? I'm running it from the same place as usual.

Our devnotes say you need to run it from inside the geniza directory — when I run it that way locally I don't get the `geniza/` path that you did

blms:
> Our devnotes say you need to run it from inside the geniza directory — when I run it that way locally I don't get the geniza/ path that you did

Wow, great catch! I should read the dev notes more. I guess I must have done that in the past, so I _didn't_ actually run it int he same directory as usual 😅

Issue #1191, opened 2022-10-11
OpenSeadragon navigator should not be visible on placeholder images


blms:
## testing notes (QA)

- [x] In the transcription editor in the QA site, for a document (such as PGPID 1390) with more transcription blocks than images, confirm there is no "navigator" (the mini zoom position indicator with the red outline) in the top left of the gray placeholder rectangles.
- [x] Confirm the navigator is still present on the actual images.

kseniaryzhova:
@rlskoeser @blms works for me! Closing!

PR #1206, opened 2022-10-13
Disable navigator for placeholder images (#1191)

blms:
## In this PR

- Per #1191 
  - Disable OSD navigator for placeholder images

### DIFF
--- a/sitemedia/js/controllers/iiif_controller.js
+++ b/sitemedia/js/controllers/iiif_controller.js
@@ -97,7 +97,8 @@ export default class extends Controller {
         const url = this.osdTarget.dataset.iiifUrl;
 
         // allow placeholder image (url ending in .png instead of .json)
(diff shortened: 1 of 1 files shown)


sitemedia/js/controllers/iiif_controller.js
```diff
@@ -97,7 +97,8 @@ export default class extends Controller {
         const url = this.osdTarget.dataset.iiifUrl;
 
         // allow placeholder image (url ending in .png instead of .json)
-        const tileSource = url.endsWith(".png") ? { type: "image", url } : url;
+        const isPlaceholder = url.endsWith(".png");
```
rlskoeser, line 100: this assumes current implementation of iiif only for real images; that's fine with me, just might be worth noting somewhere (although not sure where)

rlskoeser (approved):
looks reasonable to me! any reason to test locally?

blms:
No need to test locally @rlskoeser!

PR #1213, opened 2022-10-20
Improve document metadata export speed

rlskoeser:
@quadrismegistus check it out! I was able to improve the prefetching and the export is much faster now.

It was easier to investigate now that we have a script export — I turned on django logging so I could see db calls,  added a 1s sleep so I could see what was going on, and used that to make sure all the db calls were made at the beginning and none per individual document.  Now the script to export all documents completes in around a minute.

### DIFF
--- a/geniza/common/metadata_export.py
+++ b/geniza/common/metadata_export.py
@@ -24,7 +24,7 @@ class Exporter:
     model = None
     csv_fields = []
     sep_within_cells = " ; "
-    true_false = {True: "y", False: "n"}
+    true_false = {True: "Y", False: "N"}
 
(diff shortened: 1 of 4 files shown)


quadrismegistus:
Excellent! So it looks like adding `log_entries__user` to the prefetching was the crucial step?

rlskoeser:
> Excellent! So it looks like adding `log_entries__user` to the prefetching was the crucial step?

Yeah, that was one piece — we were querying for log entry users on every record! We also needed prefetching on `footnotes` (which I'm surprised we didn't have before), and there was one place we were calling `last()` on a footnote queryset we'd already fetched — which makes another unnecessary db call.

Issue #1097, opened 2022-09-19
As a content admin, I want document data exports synchronized to github so that there is a publicly accessible, versioned copy of project data available for researchers.


rlskoeser:
## dev notes
- [ ] reimplement existing admin document data export with django-import-export
- [ ] adjust fields as needed by existing issues (to be added here) 
- [ ] make a public version of the export that exempts admin fields (notes, ...)
- [ ] write a manage command that can run the public export and sync to a configured github repo

quadrismegistus:
Fields in adminexporter not to be included in public: notes, needs_review, url_admin, status

Make sure that status == public as well

rlskoeser:
some initial testing notes: 
- we're committing even when there are no changes, so we're getting a lot of empty commit messages 
- let's make the commit message more like the annotation backup; no need to include the date since that's elsewhere in the commit metadata

PR #1217, opened 2022-11-07
PublicDocumentExporter

quadrismegistus:
This is to fix https://github.com/Princeton-CDH/geniza/issues/1097.

### DIFF
--- a/geniza/common/metadata_export.py
+++ b/geniza/common/metadata_export.py
@@ -34,7 +34,7 @@ class Exporter:
         self.url_scheme = "https://"
 
     def __iter__(self):
-        yield from self.iter_export_data_as_dicts()
+        yield from self.iter_dicts()
 
     def csv_filename(self):
         """Generate the appropriate CSV filename for model and time
@@ -63,7 +63,7 @@ class Exporter:
         """
         raise NotImplementedError
 
-    def iter_export_data_as_dicts(self):
+    def iter_dicts(self):
         """Iterate over the exportable data, one dictionary per row
 
         :yield: Dictionary of information for each object
@@ -113,7 +113,7 @@ class Exporter:
         """
         return {k: self.serialize_value(v) for k, v in data.items()}
 
-    def iter_export_data_as_csv(self, fn=None, pseudo_buffer=False):
+    def iter_csv(self, fn=None, pseudo_buffer=False):
         """Iterate over the string lines of a CSV file as it's being written, either to file or a string buffer.
 
         :param fn: Filename to save CSV to (if pseudo_buffer is False), defaults to None
@@ -135,8 +135,7 @@ class Exporter:
             )
             yield writer.writeheader()
             yield from (
-                writer.writerow(self.serialize_dict(docd))
-                for docd in self.iter_export_data_as_dicts()
+                writer.writerow(self.serialize_dict(docd)) for docd in self.iter_dicts()
             )
 
     def write_export_data_csv(self, fn=None):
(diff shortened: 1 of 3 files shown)


geniza/corpus/metadata_export.py
```diff
@@ -135,11 +136,16 @@ def get_export_data_dict(self, doc):
...
+
+        if not "notes" in self.exclude_csv_fields:
+            outd["notes"] = doc.notes
+        if not "needs_review" in self.exclude_csv_fields:
+            outd["needs_review"] = doc.needs_review
+
+        if not "url_admin" in self.exclude_csv_fields:
+            outd[
+                "url_admin"
+            ] = f"{self.url_scheme}{self.site_domain}/admin/corpus/document/{doc.id}/change/"
```
rlskoeser, line 148: I think the the admin doc exporter should extend `get_export_data_dict` and add these fields, rather than adding conditional behavior in the base class

geniza/corpus/metadata_export.py
```diff
@@ -165,3 +172,14 @@ def get_export_data_dict(self, doc):
...
+
+class AdminDocumentExporter(DocumentExporter):
+    exclude_csv_fields = set()
+
+
+class PublicDocumentExporter(DocumentExporter):
+    exclude_csv_fields = {"notes", "needs_review", "url_admin", "status"}
+
+    def get_queryset(self):
+        return super().get_queryset().filter(status="P")
```
rlskoeser, line 185: nice. but use the constant instead of hard-coding  (`Document.PUBLIC`)

geniza/common/metadata_export.py
```diff
@@ -33,6 +33,9 @@ def __init__(self, queryset=None, progress=False):
         self.site_domain = Site.objects.get_current().domain.rstrip("/")
         self.url_scheme = "https://"
 
+    def __iter__(self):
+        yield from self.iter_export_data_as_dicts()
```
rlskoeser, line 37: while we're here, can we shorten this method name?

geniza/corpus/metadata_export.py
```diff
@@ -165,3 +172,14 @@ def get_export_data_dict(self, doc):
...
 
         return outd
+
+
+class AdminDocumentExporter(DocumentExporter):
+    exclude_csv_fields = set()
+
+
+class PublicDocumentExporter(DocumentExporter):
+    exclude_csv_fields = {"notes", "needs_review", "url_admin", "status"}
```
rlskoeser, line 182: can we just set `csv_fields` to the base list of fields minus these? I'm not sure what the exclude property gets us

geniza/corpus/metadata_export.py
```diff
@@ -165,3 +159,21 @@ def get_export_data_dict(self, doc):
...
+        outd["status"] = doc.get_status_display()
+        outd[
+            "url_admin"
+        ] = f"{self.url_scheme}{self.site_domain}/admin/corpus/document/{doc.id}/change/"
+
+        return outd
+
+
+class PublicDocumentExporter(DocumentExporter):
+    def get_queryset(self):
```
rlskoeser, line 178: Do we need to subset the list of csv fields here, or does it happen automagically/implicitly because we're using csv `DictWriter`? 

If so, that seems slick but probably worth a comment noting that fact

rlskoeser (commented):
Revised approach looks good to me. Thanks for shortening the method names, seems more readable to me now

geniza/corpus/metadata_export.py
quadrismegistus: This is so automagic since the keys aren't populated in the dictionary csv DictWriter receives. Will add a note

rlskoeser:
is the new public document exporter used anywhere yet? or would that be the next step after we merge this, adding it to the manage command?

quadrismegistus:
No new public one isn't used yet besides in a test function but it will be used by the new manage.py command. Should we merge this first?

rlskoeser:
> No new public one isn't used yet besides in a test function but it will be used by the new manage.py command. Should we merge this first?

Yes, let's merge this in first. When we get to the manage command, let's use the existing one you already started on.

Issue #1219, opened 2022-11-07
documents in admin should be sorted by shelfmark by default


rlskoeser:
reported by Alan on Slack — I think this is a regression in v4.10 related to the changes we made for the document export

the `ArrayAgg` annotation is still present  (included in `metadata_prefetch` and probably shouldn't be, I think we only need for admin) but we lost the default sort for the admin queryset

## dev notes
- [x] move shelfmark query aggregation from `metadata_prefetch` to the document admin queryset method
- [x] sort by shelfmark by default in document admin queryset
- [x] revise `shelfmark_display` property on `Document` class to use the aggregation property `shelfmk_all` for admin sorting — should be able to add it to the existing `admin.display` decorator (this should make the column clickable in admin for combining / reversing sort)

quadrismegistus:
Should we close this?

rlskoeser:
Tested in admin on qa site — there's a long join showing up first and I'm not sure exactly why (my guess: aggregated shelfmark doesn't follow document / textblock order; this one doesn't show in my local dev instance, but my data is out of date). Otherwise it looks like records are sorted by shelfmark, and shelfmark is clickable to adjust sort order. This should address the immediate regression.

PR #1220, opened 2022-11-07
Ordering by shelfmark again

quadrismegistus:
Fix for https://github.com/Princeton-CDH/geniza/issues/1219

### DIFF
--- a/geniza/corpus/admin.py
+++ b/geniza/corpus/admin.py
@@ -317,7 +317,6 @@ class DocumentAdmin(TabbedTranslationAdmin, SortableAdminBase, admin.ModelAdmin)
             .get_queryset(request)
             .metadata_prefetch()
             .annotate(shelfmk_all=ArrayAgg("textblock__fragment__shelfmark"))
-            # .order_by('shelfmark_override')
             .order_by("shelfmk_all")
         )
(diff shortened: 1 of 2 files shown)


geniza/corpus/admin.py
```diff
@@ -388,7 +395,9 @@ def merge_documents(self, request, queryset=None):
             status=303,
         )  # status code 303 means "See Other"
 
-    @admin.display(description="Export selected documents to CSV")
+    @admin.display(
+        description="Export selected documents to CSV", ordering="-shelfmark"
+    )  # @TODO: This ordering doesn't affect the order of the export because in the exporter we're ordering by ID
```
rlskoeser, line 400: ah, sorry — this is not the field I meant you to put the shelfmark ordering on; it should go on the `shelfmark_display` property on `Document`

rlskoeser (commented):
These changes look right — just need to cleanup the commented out bits.

It would be nice if we could figure out a way to test this in unit tests so that we would catch it if we caused a regression in future.... Maybe we could test the admin queryset with a couple of fixtures and confirm it sorts them by shelfmark? (would need to make sure the default sorting doesn't give the same result)

Issue #1100, opened 2022-09-19
As a content admin, I want scholarship records exported to github so that there is a publicly accessible, versioned copy of project data available for researchers.


rlskoeser:
## testing notes
test the following scenarios for exporting to CSV:
- [x] test exporting a subset of sources (scholarship records)  from django admin
- [x] test exporting all sources from django admin
- [x] test exporting a subset of footnotes from django admin
- [x] test exporting all footnotes from django admin (likely slow because includes transcription content; please report if it finishes and if it's usable or not)

### retesting notes
- [x] confirm that footnotes with no transcription content don't display "None" in the content field
- [x] confirm that opening the downloaded CSV in Excel automatically reads transcription content properly

## dev notes
- [x] re-implement existing tabular export for footnotes with new export class
- [x] re-implement existing tabular export for sources with new export class
- [x] check if revised source export resolves standing issue https://github.com/princeton-cdh/geniza/issues/278 
- [x] make public version available for including in data export manage command

kseniaryzhova:
@rlskoeser so the footnote download did not have transcription content aside from the fact that transcriptions exist or not and the download was super quick (for all footnotes). Is the csv for footnotes supposed to have actual transcription content?

rlskoeser:
@kseniaryzhova I should have probably given you links, just to make sure we were talking about the same things.

Were you testing the footnotes export from this page? https://test-geniza.cdh.princeton.edu/admin/footnotes/footnote/

It looks to me like it does include the text content of the transcription — happy to remove it if it's not useful / important to have here or to have in this format. 

And the sources export should be tested from this page: https://test-geniza.cdh.princeton.edu/admin/footnotes/source/

rlskoeser:
@kseniaryzhova so weird!

Would you try downloading the subset from this link and report back? 
https://test-geniza.cdh.princeton.edu/admin/footnotes/footnote/?q=5454
select them and then use the action -> export selected to csv -> go

I'm getting `None` for content for the first one and transcription content for the second one.  (We need to fix the `None` at least, but I'm still not sure if we should have the content here or not – if it continues to not work for you maybe we should drop it.)

I tried exporting all footnotes and I think it failed; it was definitely slow. So that slowness might be enough reason to drop the content from this export, since we have the content in the annotation backup repo already.

rlskoeser:
discussed with @kseniaryzhova and finally figured out what's going on here:
[...]

Issue #278, opened 2021-08-25
As a content editor, I want to view source URLs when I download the sources CSV in order to more easily find/update external sources.


richmanrachel:
Context: there are a few Asana cleaning tasks pertaining to the cleaning/checking of external links (particularly Marina's google doc translations). When creating a work flow for the RAs, I discovered that the current sources CSV download does not include the URLs, and therefore it is difficult to develop a proper workflow.

In conversation with @rlskoeser on 8/17/21, she said it should be easy to add the existing URL field to the CSV. Thanks!

rlskoeser:
@richmanrachel I believe this was addressed as part of Ben's work on #313 — would you confirm in qa and comment either way?

richmanrachel:
@rlskoeser, no it didn't work. I do see a URL tab on the CSV now, but it's not filled in for any of the unpublished sources.

richmanrachel:
![image](https://user-images.githubusercontent.com/68028846/145244849-3df6ed6a-dce4-47b2-a72c-16d9657f6b31.png)

rlskoeser:
@richmanrachel I'm not sure we have any URLs on sources in the database yet! Would you test adding one and downloading again?

rlskoeser:
noting from my brief testing: selecting items and exporting partial data seems to include the urls, but exporting all records as csv does _not_ ; unclear what should cause this, since it is running basically the same code (the querysets may be slightly different?)

kseniaryzhova:
@rlskoeser works for me - I see both public and admin site URLs in the csv. Not clickable until you put it in Google Sheets but that's fine. Thank you!

PR #1222, opened 2022-11-08
Rewrite footnotes source tabular export to use new exporter class

rlskoeser:
ref #1100 #278

rewrites the existing source tabular export to use the new metadata export code

### DIFF
--- a/geniza/footnotes/admin.py
+++ b/geniza/footnotes/admin.py
@@ -5,10 +5,8 @@ from django.contrib.admin import SimpleListFilter
 from django.contrib.contenttypes.admin import GenericTabularInline
 from django.contrib.sites.models import Site
 from django.db import models
-from django.db.models import Count
 from django.db.models.fields import CharField, TextField
 from django.db.models.functions import Concat
-from django.db.models.query import Prefetch
 from django.forms.widgets import Textarea, TextInput
 from django.urls import path, reverse
 from django.utils import timezone
@@ -18,6 +16,7 @@ from modeltranslation.admin import TabbedTranslationAdmin
 from tabular_export.admin import export_to_csv_response
 
 from geniza.common.admin import custom_empty_field_list_filter
+from geniza.footnotes.metadata_export import SourceExporter
 from geniza.footnotes.models import (
     Authorship,
     Creator,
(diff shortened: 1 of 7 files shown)


geniza/footnotes/metadata_export.py
```diff
@@ -0,0 +1,58 @@
+from geniza.common.metadata_export import Exporter
+from geniza.footnotes.models import Footnote, Source
+
+
+class SourceExporter(Exporter):
```
quadrismegistus, line 5: Cool!

geniza/footnotes/models.py
```diff
@@ -96,6 +98,24 @@ def __str__(self) -> str:
...
+            Prefetch(
+                "authorship_set",
+                queryset=Authorship.objects.select_related("creator"),
+            )
+        )
+
+    def footnote_count(self):
+        "annotate with footnote count"
+        return self.annotate(Count("footnote", distinct=True))
+
```
quadrismegistus, line 117: slick!

geniza/footnotes/admin.py
```diff
@@ -149,22 +148,16 @@ def get_queryset(self, request):
         return (
             super()
             .get_queryset(request)
+            .metadata_prefetch()
+            .footnote_count()
```
quadrismegistus, line 152: very nice. Is `footnote_count` not inside `metadata_prefetch` because in the event we have a queryset (from admin) we don't need to prefetch but we do need to footnote_count?

quadrismegistus (approved):
Looks good! This approach seems to be working really well.

geniza/footnotes/admin.py
rlskoeser: yes, that was my thought — they aren't necessarily related and we could easily want one and not the other

geniza/footnotes/tests/test_footnote_admin.py
```diff
@@ -134,6 +134,22 @@ def test_footnotes(self):
...
+        content = b"".join([val for val in response.streaming_content]).decode()
+
+        # spot-check that we get expected data
+        # - header row
+        assert "source_type,authors,title" in content
+        # - some content
+        assert str(source.source_type) in content
+        assert source.title in content
+        assert str(twoauthor_source.source_type) in content
+        assert twoauthor_source.title in content
```
quadrismegistus, line 151: Nice testing approach for the streaming

quadrismegistus (approved):
LGTM!

PR #1226, opened 2022-11-10
Refactor footnote export to use new exporter code

rlskoeser:
same thing as what I just did for sources, I refactored the old tabular_export to use our new metadata export code

I also removed the sorting from the common metadata export serialization code — there are cases where the order may be meaningful and order shouldn't be changed.

FYI, I think both of these will need public variants — we don't want to include the admin url in our public exports, but it's useful for the admin downloads.

### DIFF
--- a/geniza/footnotes/metadata_export.py
+++ b/geniza/footnotes/metadata_export.py
@@ -56,3 +56,39 @@ class SourceExporter(Exporter):
             # construct directly to avoid extra db calls
             "admin_url": f"{self.url_scheme}{self.site_domain}/admin/footnotes/source/{source.id}/change/",
         }
+
+
+class FootnoteExporter(Exporter):
+    """
+    A subclass of :class:`geniza.common.metadata_export.Exporter` that
+    exports information relating to :class:`~geniza.footnotes.models.Footnote`.
+    """
+
+    model = Footnote
+    csv_fields = [
+        "document",  # ~ content object
+        "document_id",
+        "source",
+        "location",
+        "doc_relation",
+        "notes",
+        "url",
+        "content",
+        "admin_url",
+    ]
+
+    def get_queryset(self):
+        return self.queryset or self.model.objects.all().metadata_prefetch()
+
+    def get_export_data_dict(self, footnote):
+        return {
+            "document": footnote.content_object,
+            "document_id": footnote.content_object.pk,
+            "source": footnote.source,
+            "location": footnote.location,
+            "doc_relation": footnote.get_doc_relation_list(),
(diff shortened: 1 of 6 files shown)


rlskoeser:
Ah, I should probably note — I haven't tested the full download, and I kind of suspect it will be slow because I'm including plain text version of transcription content (and because the way annotations are currently linked is inefficient).  Do you think it's worth including the text content here if we can?

geniza/footnotes/metadata_export.py
```diff
@@ -56,3 +56,39 @@ def get_export_data_dict(self, source):
...
+        "location",
+        "doc_relation",
+        "notes",
+        "url",
+        "content",
+        "admin_url",
+    ]
+
+    def get_queryset(self):
+        return self.queryset or self.model.objects.all().metadata_prefetch()
```
quadrismegistus, line 81: So we're only running .metadata_prefetch in the event we don't have an incoming queryset (i.e. in the event of an admin interface select X to download operation). But in that event, are we running FootnoteAdmin.get_queryset which ends up running .metadata_prefetch? Still a little confused when we are and aren't using it. SourceExporter was using a custom queryset as manager?

geniza/footnotes/metadata_export.py
```diff
@@ -56,3 +56,39 @@ def get_export_data_dict(self, source):
...
+            "document": footnote.content_object,
+            "document_id": footnote.content_object.pk,
+            "source": footnote.source,
+            "location": footnote.location,
+            "doc_relation": footnote.get_doc_relation_list(),
+            "notes": footnote.notes,
+            "url": footnote.url,
+            "content": footnote.content_text or "",
+            "admin_url": f"{self.url_scheme}{self.site_domain}/admin/footnotes/footnote/{footnote.id}/change/",
+        }
```
quadrismegistus, line 94: I like how extensible this has become

quadrismegistus (approved):

geniza/footnotes/metadata_export.py
rlskoeser: yeah, me too! it was really easy to refactor tot e new format

geniza/footnotes/metadata_export.py
rlskoeser: yeah, my assumption is that `metadata_prefetch` has already been called if we're given a queryset — because the only place right now where I know we call it with a subset of records is from the admin, and we're using the same prefetching there, so it should be applied already.

We already had a custom queryset for footnotes with other functions, so I just moved the existing admin fetching into a reusable method.

PR #1227, opened 2022-11-10
Getting data exports also to sync to repository

quadrismegistus:


geniza/corpus/management/commands/export_metadata.py
```diff
@@ -1,9 +1,106 @@
...
+        if not self._remote_url:
+            if hasattr(settings, self.remote_url_settings_key):
+                self._remote_url = getattr(settings, self.remote_url_settings_key)
+            else:
+                self._remote_url = os.path.join(
+                    self.default_remote_url_org, self.repo_name
+                )
+        return self._remote_url
+
+    @property
```
rlskoeser: could use `cached_property` from functools instead of the `self._repo` logic here

geniza/corpus/management/commands/export_metadata.py
```diff
@@ -1,9 +1,106 @@
...
+    def local_path(self):
+        if not self._local_path:
+            if hasattr(settings, self.local_path_settings_key):
+                self._local_path = getattr(settings, self.local_path_settings_key)
+            else:
+                self._local_path = os.path.abspath(
+                    os.path.join(settings.BASE_DIR, "..", self.repo_name)
+                )
+        return self._local_path
```
rlskoeser, line 43: my preference would be to require the configuration in local settings and raise a `CommandError` with an appropriate message if it isn't present (at least, for any setting we want to require and make sure is configured properly)

rlskoeser (commented):
This looks like a great start to me.  Anything you particularly want feedback on?

Might want to think about how to generalize the way you are setting and calling paths and exporters since we may have a few, but we won't have so many that it will be a big deal.

geniza/corpus/management/commands/export_metadata.py
quadrismegistus: 🤯

quadrismegistus: Good example of a Python innovation I just haven't picked up yet

rlskoeser: it's so hard to keep up with all the new things! for a while I was using a third-party cached property package, so I was delighted when I figured out there was a built in version added at some point (I forget what version of python)

quadrismegistus:
I've been going crazy because everything seems to be working but every time I re-export the documents.csv, git thinks the file changed even though it's correctly sorting by id. The git diff terminal command wasn't cutting it for showing me what was changed nor was github (file too large). Github Desktop showed me the issue:

Screen Shot 2022-11-14 at 13 32 32

It's the lists of values we're joining by semicolons. Because they're not sorted, and were originally sets, their order is non-deterministic and we're left with non-meaningful git differences.

I think we should reintroduce the sorting code used for serializing set values in metadata_export.

rlskoeser:
> I've been going crazy because everything seems to be working but every time I re-export the documents.csv, git thinks the file changed even though it's correctly sorting by id. The git diff terminal command wasn't cutting it for showing me what was changed nor was github (file too large). Github Desktop showed me the issue:
>
> It's the lists of values we're joining by semicolons. Because they're not sorted, and were originally sets, their order is non-deterministic and we're left with non-meaningful git differences.
> 
> I think we should reintroduce the sorting code used for serializing set values in metadata_export.

Whoa! good catch. Can you think of a way to force this to be implemented in the subclassses? It seems better for us to think about the ordering logic there, but as a fallback it does seem reasonable to sort sets in the base exporter.

I'm surprised the contributors is a set and not a list... I guess we must have done that to make sure they were unique. I wonder if there could be a meaningful order in that case, or if order doesn't matter...

quadrismegistus:
Solution was to return lists, sets, tuples, anything you want from the export_dict function, and leave it to the csv writer function to join them, sorted, by sep_within_cells. In this latest push there is only one invocation of sep_within_cells instead of a bunch of them jammed into the export_dict function. This has added advantage that export_dict could also feed into a json file format and preserve the list, etc data structures.

quadrismegistus:
Not sure, — I can re-export a few times and see if I get differences.

quadrismegistus:
Looks like it. iiif_urls for instead is a list but is sorted non-deterministically:
Screen Shot 2022-11-14 at 14 38 01

rlskoeser:
> Looks like it. iiif_urls for instead is a list but is sorted non-deterministically

If that's true, then that's a bug — iiif url order should correspond to fragment shelfmark order (in theory you should be able to match them back up — a different structure would be better in the  json)

quadrismegistus:
Where is the bug here? In textblock_set.all() ?

Screen Shot 2022-11-14 at 14 47 10

quadrismegistus:
Let me make sure it's not a problem further down the pipe from there

quadrismegistus:
Actually nevermind I think I was comparing different sorting behaviors. It seems to be working now

geniza/corpus/management/commands/export_metadata.py
```diff
@@ -1,30 +1,189 @@
...
+        parser.add_argument("-p", "--path", type=str, default="", required=False)
+        parser.add_argument("-u", "--url", type=str, default="", required=False)
+
+        parser.add_argument("-q", "--quiet", action="count", default=0)
 
     def handle(self, *args, **options):
-        began = time.time()
-        ofn = options["output_filename"]
-        self.print(f"Exporting data as CSV to: {ofn}")
+        from pprint import pprint
```
rlskoeser, line 166: looks like this is unused

geniza/corpus/management/commands/export_metadata.py
```diff
@@ -1,30 +1,189 @@
...
-        parser.add_argument(
-            "-o", "--output_filename", type=str, default=ofn, required=False
-        )
+        parser.add_argument("-w", "--write", action="store_true")
+        parser.add_argument("-s", "--sync", action="store_true")
+
+        parser.add_argument("-p", "--path", type=str, default="", required=False)
+        parser.add_argument("-u", "--url", type=str, default="", required=False)
+
+        parser.add_argument("-q", "--quiet", action="count", default=0)
```
rlskoeser, line 163: django manage commands already have a verbosity argument that we inherit, would be better to use that one for consistency

https://docs.djangoproject.com/en/3.2/ref/django-admin/#cmdoption-verbosity

geniza/corpus/management/commands/export_metadata.py
```diff
@@ -1,30 +1,189 @@
...
+
+    def add_and_commit(self):
+        for fn in self.paths:
+            if os.path.exists(fn):
+                self.repo.index.add(fn)
+        self.repo.index.commit(
+            "Auto-syncing @ " + timezone.now().strftime("%Y-%m-%d %H:%M:%S")
+        )
+
+    def write_local(self):
```
rlskoeser, line 104: This name seems kind of opaque to me. generate_csvs ? export_csvs ? generate_exports ? (maybe not csv-specific if we still hope to add json at some point)

geniza/common/utils.py
```diff
@@ -89,3 +91,35 @@ def __enter__(self, *x, **y):
 
     def __exit__(self, *x, **y):
         return
+
+
+class Timer:
```
rlskoeser, line 96: I see the potential value of this utility code, but if you write code you need to test & document it too

rlskoeser: was looking to see if there was an existing tool for this we could use (maybe as a dev dependency but not prod), and saw that there is a [context manager decorator]( https://docs.python.org/3.9/library/contextlib.html#contextlib.contextmanager). Maybe you already were aware and it doesn't work in this case, since you want to optionally customize print behavior, but thought I would share in case note. Looks like they added several variants in v3.7.

geniza/corpus/management/commands/export_metadata.py
```diff
@@ -1,30 +1,189 @@
...
+            # commit?
+            with self.timer("Adding any changes"):
+                self.add_and_commit()
+
+            # push data updates
+            with self.timer("Pushing any changes"):
+                result = origin.push()
+                if result:
+                    info = result[0]
+                    self.print(f"Result = {info.flags}")
```
rlskoeser, line 143: what is this number?

geniza/corpus/management/commands/export_metadata.py
```diff
@@ -1,30 +1,189 @@
...
+
+    def sync_remote(self):
+        """Sync local repository content with origin repository. Assumes
+        :meth:`setup_repo` has already been run, and any new or modified
+        files have been committed."""
+        try:
+            with self.timer("Pulling repository"):
+                origin = self.repo.remote(name="origin")
+                # pull any remote changes since our last commit
+                origin.pull()
```
rlskoeser, line 132: I think you need to pull remote changes _before_ generating new versions of the data exports

DEPLOYNOTES.md
```diff
@@ -1,5 +1,9 @@
 # Deploy Notes
 
+## 4.11
+
+-   Must configure **METADATA_REPO_REMOTE_URL** and **METADATA_REPO_LOCAL_PATH** in local settings. These will determine locations for (respectively) the remote metadata repository (e.g. "https://github.com/Princeton-CDH/test-geniza-metadata") and the local files for the metadata exposts and syncing (e.g. "~/github/test-geniza-metadata").
```
rlskoeser, line 5: For the annotation exports we have **ANNOTATION_BACKUP_GITREPO** and **ANNOTATION_BACKUP_PATH** — could we make the names more consistent, since they are parallel?

It looks like I didn't do this for the annotation backup script, but would you add a draft note here of the cron job that should be set up to export/sync metadata, and what options it should be run with? (We'll configure the cron job via ansible, but it's nice to have documentation about when it was added and how it should be run.)

quadrismegistus:
Got it, thanks for comments. I'll work through these tomorrow

geniza/common/utils.py
quadrismegistus: Didn't know about that decorator actually but yeah it doesn't seem like it would work in this context since we want something can store the variable (it's not a try/catch/finally pattern).

quadrismegistus: Let me see if I can find some good timer code replacement otherwise I can write a few tests, a simple timer seems useful for things like this

geniza/corpus/management/commands/export_metadata.py
quadrismegistus: This is a numeric flag indicating the push result. I was printing for my own purposes, we can hide

geniza/corpus/management/commands/export_metadata.py
quadrismegistus: Good catch, ok, reshuffling code a bit in response

geniza/common/utils.py
quadrismegistus: We could try this?
https://pypi.org/project/contexttimer/
Not sure if any of this is worth it -- happy to just get rid of all timery stuff for now

rlskoeser: Yeah, I did a quick look and found that one too — looks like it hasn't been updated in a while, but if it works we could use it. 

Otherwise, maybe you can just write one test that covers all of this and add some comments or docstrings.

Or another option would be to customize the rich progressbar so it will report total time when completed.  I think right now it is only counting down, would be nice to have count down + total time. Is there anyway to start the progressbar sooner so that it will include the queries and gathering data in the total? (Or does it already?)

I like the impulse, though — I have timer + logging code that I just refer to and copy instead of refactoring into something reusable. Would be nice

geniza/corpus/metadata_export.py
```diff
@@ -66,39 +66,35 @@ def get_export_data_dict(self, doc):
...
-        collections = set(
-            [
-                fragment.collection.abbrev or fragment.collection.name
-                if fragment.collection
-                else ""
-                for fragment in all_fragments
-            ]
-        ) - {
+        collections = {
+            fragment.collection.abbrev or fragment.collection.name
```
rlskoeser, line 93: should use `str` method of `Collection`, which will do the right thing (combines lib abbrev and abbrev, I think)

geniza/corpus/metadata_export.py
```diff
@@ -66,39 +66,35 @@ def get_export_data_dict(self, doc):
...
+            if fragment.collection
+            else ""
+            for fragment in all_fragments
```
rlskoeser, line 96: this is one place I would retain order in this export: if a document has multiple fragments from different collections, if fragment order matches collection order you can figure out which ones are from where

(although granted it's not a great way to structure the data, and most people probably won't bother with splitting those out and reassociating)

rlskoeser (commented):
Your updates all look reasonable to me. If you get it where you're happy with it and get the code coverage check passing, I think it's ready to merge.

geniza/corpus/metadata_export.py
quadrismegistus: Why not just this instead:

```python
libraries = [
            fragment.collection.lib_abbrev or fragment.collection.library
            for fragment in all_fragments
            if fragment.collection
        ]
        collections = [
            str(fragment.collection) 
            for fragment in all_fragments 
            if fragment.collection
        ]
```

quadrismegistus:
Ok, checks passing, including codecov with new tests. We might be able to merge soon

rlskoeser:
@quadrismegistus updated as follows:
- merged in develop and reconciled conflicts
- removed the `git pull` call in the sync repo method (you can't pull when you have uncommitted changes)
- added the public fragments exporter
- added docstrings to your repo class to appease the docs coverage test

PR #1235, opened 2022-11-15
Separate admin and public exports for sources and footnotes

rlskoeser:
Splits out admin and public versions of source and footnote exporter
- makes the base exporter the public versions (for use in metadata export script)
- create admin variants that add admin url
- use admin exporter for admin csv export functionality
- unit tests for the changes

#1100 - updates for footnote export based on problems found in testing:
- `content_text` was returning text `None` when there was no content
- added UTF-8 byte order mark at beginning of CSV so that Excel will read content correctly

### DIFF
--- a/geniza/common/metadata_export.py
+++ b/geniza/common/metadata_export.py
@@ -1,3 +1,4 @@
+import codecs
 import csv
 
 from django.conf import settings
@@ -132,8 +133,12 @@ class Exporter:
             else Echo()
         ) as of:
             writer = csv.DictWriter(
-                of, fieldnames=self.csv_fields, extrasaction="ignore"
+                of,
+                fieldnames=self.csv_fields,
+                extrasaction="ignore",
             )
+            # start with byte-order mark so Excel will read unicode properly
+            yield codecs.BOM_UTF8
             yield writer.writeheader()
             yield from (
                 writer.writerow(self.serialize_dict(docd)) for docd in self.iter_dicts()
--- a/geniza/common/metadata_export.py
+++ b/geniza/common/metadata_export.py
@@ -127,18 +127,19 @@ class Exporter:
         :yield: String of current line in CSV
         :rtype: Generator[str]
         """
+        csv_filename = fn or self.csv_filename()
         with (
-            open(self.csv_filename() if not fn else fn, "w")
+            codecs.open(csv_filename, "w", encoding="utf-8-sig")
             if not pseudo_buffer
             else Echo()
         ) as of:
+            # start with byte-order mark so Excel will read unicode properly
+            yield codecs.BOM_UTF8
             writer = csv.DictWriter(
                 of,
                 fieldnames=self.csv_fields,
                 extrasaction="ignore",
             )
-            # start with byte-order mark so Excel will read unicode properly
-            yield codecs.BOM_UTF8
             yield writer.writeheader()
             yield from (
                 writer.writerow(self.serialize_dict(docd)) for docd in self.iter_dicts()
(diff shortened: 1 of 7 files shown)


geniza/common/metadata_export.py
```diff
@@ -132,8 +133,12 @@ def iter_csv(self, fn=None, pseudo_buffer=False):
             else Echo()
         ) as of:
             writer = csv.DictWriter(
-                of, fieldnames=self.csv_fields, extrasaction="ignore"
+                of,
+                fieldnames=self.csv_fields,
+                extrasaction="ignore",
             )
+            # start with byte-order mark so Excel will read unicode properly
+            yield codecs.BOM_UTF8
```
quadrismegistus, line 141: Nice! huh, never seen this problem or solution before

geniza/corpus/tests/test_metadata_export.py
```diff
@@ -156,18 +157,19 @@ def test_http_export_data_csv(document):
...
     assert headers_d.get("Content-Type") == "text/csv; charset=utf-8"
     assert headers_d.get("Content-Disposition") == f"attachment; filename={ofn}"
-
     assert response.status_code == 200
 
-    def iter_http_response_lines_str(response):
-        for x in response:
-            yield x.decode()
+    content_lines = [x.decode() for x in response]
+    # first line should be byte order mark)
```
quadrismegistus, line 163: Does the fact that second line is now header instead of first mess up any other CSV readers?

geniza/common/metadata_export.py
rlskoeser: I've run into it before, which is why I recognized it immediately when I finally figured out why Ksenia and I were seeing different results.

geniza/corpus/tests/test_metadata_export.py
rlskoeser: Actually, I think technically it should only be the first character and not the first line... I'm glad you commented on that, because I hadn't really thought about that detail. Is the way I'm yielding it here putting it on it's own line? 

I saw some discussion on stack overflow about this — I think some csv readers will handle this fine, others may not. There's an `encoding='utf-8-sig'` parameter you can use in python when you open files like this to read that signature character and not interpret it as part of the first header. We should think about how/where to document, but I think it's important that the files open properly in Excel by default — otherwise the only way to get the unicode to display properly in Excel is to _import_ (instead of opening) and manually selecting UTF-8.

quadrismegistus: If it's passing the tests? And the csv files look good? Then I think what's happening is only that content_lines maybe ought to be changed as a variable name. What's being yielded from StreamingResponse is I believe ultimately what's being yielded in iter_csv -- the first yielded obj, this special character; the second yielded object, a line string (with line ending character attached I believe); and the rest, line strings (with endings attached).

However, I just looked back at the extra yield statement and if we're only yielding `codecs.BOM_UTF8` and not writing it to `of` as well, then it will probably appear in the http-streamed csv files but not the csv files from the exports (those created by the non-http export function). This is because the non-http csv function is really just using dictwriter to write to the file object directly; the http variant is exploiting the fact that dictwriter also yields its output, in order to use pipe output into the streaming handler.

geniza/common/metadata_export.py
quadrismegistus: In short (see my comment below) I think all we need to do to make this UTF-8 character yield statement work for both non-http and http-streaming exports, we just need to add a `of.write(codecs.BOM_UTF8)` above or below this line, so that it's both written to the file buffer and yielded out for anyone iterating over the written-out strings (i.e. the http streamer).

rlskoeser: Thank you for commenting and explaining all this — I thought that I wasn't getting it to work but I was looking at the wrong export file 🤦‍♀️ 

I've confirmed that we're getting one (and only one) BOM for both csv file on disk and http stream, so I think we're good on this one now

PR #1236, opened 2022-11-15
Implement fragment export

rlskoeser:
new fragment metadata export, with admin variant; wired into fragment admin, unit tested

### DIFF
--- a/geniza/corpus/metadata_export.py
+++ b/geniza/corpus/metadata_export.py
@@ -1,11 +1,11 @@
 from geniza.common.metadata_export import Exporter
-from geniza.corpus.models import Document
+from geniza.corpus.models import Document, Fragment
 
 
 class DocumentExporter(Exporter):
     """
     A subclass of :class:`geniza.common.metadata_export.Exporter` that
-    exports information relating to :class:`~geniza.corpus.models.Documents`.
+    exports information relating to :class:`~geniza.corpus.models.Document`.
     Extends :meth:`get_queryset` and :meth:`get_export_data_dict`.
     """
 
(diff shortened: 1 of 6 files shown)


geniza/corpus/metadata_export.py
```diff
@@ -1,11 +1,11 @@
 from geniza.common.metadata_export import Exporter
-from geniza.corpus.models import Document
+from geniza.corpus.models import Document, Fragment
 
 
 class DocumentExporter(Exporter):
     """
     A subclass of :class:`geniza.common.metadata_export.Exporter` that
-    exports information relating to :class:`~geniza.corpus.models.Documents`.
+    exports information relating to :class:`~geniza.corpus.models.Document`.
```
quadrismegistus, line 8: woops nice catch

geniza/corpus/metadata_export.py
```diff
@@ -189,3 +189,85 @@ class PublicDocumentExporter(DocumentExporter):
...
+            data.update(
+                {
+                    "collection": fragment.collection,
+                    "library": fragment.collection.library,
+                    "library_abbrev": fragment.collection.lib_abbrev,
+                    "collection_name": fragment.collection.name,
+                    "collection_abbrev": fragment.collection.abbrev,
+                }
+            )
+        return data
```
quadrismegistus, line 255: Hm. Will these fragment data be included in the metadata exports? At least any redundant values there will be guaranteed to match. I suppose the redundancy in this code snippet here is also guaranteed? Since it's just a matter of pulling additional attributes off the `fragment.collection` object.

geniza/corpus/metadata_export.py
```diff
@@ -189,3 +189,85 @@ class PublicDocumentExporter(DocumentExporter):
...
+
+
+# NOTE: may want a public version of fragment exporter
+# that would limit to fragments associated with public / non-suppressed documents
```
quadrismegistus, line 259: Should we make this a separate PR or issue? Sounds easy enough, could do it via modifying get_queryset ?

quadrismegistus (approved):
Awesome, this looks great!

geniza/corpus/metadata_export.py
rlskoeser: My plan was to include this alongside the documents export — there's overlapping information, but for documents that run across multiple fragments I think it may be easier to work with this vs. trying to split out multiple parallel fields in the document csv.

After seeing this, I'm not sure how valuable it is to have all the collection details on each row. I wonder if the short name is enough if there is a lookup somewhere that you can reference if you're not familiar with the collections — but also not sure I want to go to the effort of making a collections.csv to sit next to this as a reference.

geniza/corpus/metadata_export.py
rlskoeser: Yeah, if we go that route it would be similar to documents, where we'd have a public variant and limit it in the `get_queryset`.  I'm ambivalent about whether it's part of this PR or a separate one, what do you think?

quadrismegistus: Maybe we ought to just tuck it into this one because it's a fairly small change and the Public/Admin is a distinction we've introduced already into the Exporter family and probably all future Exporters should have them.

geniza/corpus/metadata_export.py
quadrismegistus: Yeah, agreed, not sure it's worth an extra csv file

rlskoeser: I'll ask if they want the extra collection fields when we have the team to acceptance testing

geniza/corpus/metadata_export.py
rlskoeser: added a public version if you want to take another quick look before I merge

quadrismegistus: Looks great!

Issue #918, opened 2022-06-09
As a content admin I would like to see counts and/or be able to export user log entries so that I can quantify how much work a content editor has contributed to the database. 


richmanrachel:
## testing notes
test log entry csv export — should be available from the [log entry admin list](https://test-geniza.cdh.princeton.edu/admin/admin/logentry/) on the test site, and should work like the other csv exports:
- [x] download all
- [x] select and download for a subset
- [x] CSV should include dates, usernames, information about what kind of record was changed (content type), information about what was changed, and indication of what action was taken (addition, change, deletion)

----

**Is your feature request related to a problem? Please describe.**

@rlskoeser - this is not a priority item, but I have been thinking about it for a while, so it would be helpful to know how difficult/easy this is to execute.

Just as we asked for a copy of all of Alan's log entries, it would be helpful as a Project Manager to do this for all the RAs. Every quarter when I prepare for a Geniza Lab meeting, I try to quantify what work has been done since our last meeting in order to praise and highlight the work the team has done.

Right now I have to go manually by each username and month and add all the numbers together. It would be nice if I could choose a full time frame for all users and export the data to sort by user.

**Describe the solution you'd like**
Have a "export selected user logs to CSV" action

**Describe alternatives you've considered**
I'm open to other options!

kseniaryzhova:
@rlskoeser works for me, closing!

PR #1237, opened 2022-11-15
Admin LogEntry csv export #918

rlskoeser:
Quick admin-only csv export for log entries (not intended for inclusion in metadata exports)

I remembered I had a gist from when I've done this in the past as a one-off: https://gist.github.com/rlskoeser/f9312d27eef352e16ccfa25a31132ca5

Adapted that to our new exporter code and tweaked the logentry admin we're using to add it.

### DIFF
--- a/geniza/common/metadata_export.py
+++ b/geniza/common/metadata_export.py
@@ -180,10 +180,11 @@ class LogEntryExporter(Exporter):
         "action",
     ]
 
+    #: map log entry action flags to text labels
+    action_label = {ADDITION: "addition", CHANGE: "change", DELETION: "deletion"}
+
     def get_queryset(self):
         return super().get_queryset().select_related("content_type", "user")
-
-    action_label = {ADDITION: "addition", CHANGE: "change", DELETION: "deletion"}
 
     def get_export_data_dict(self, log):
         return {
--- a/geniza/common/metadata_export.py
+++ b/geniza/common/metadata_export.py
@@ -1,6 +1,7 @@
 import csv
 
 from django.conf import settings
+from django.contrib.admin.models import ADDITION, CHANGE, DELETION, LogEntry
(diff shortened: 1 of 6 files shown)


geniza/common/metadata_export.py
```diff
@@ -165,3 +166,32 @@ def http_export_data_csv(self, fn=None):
...
+        "content_type_app",
+        "object_id",
+        "change_message",
+        "action",
+    ]
+
+    def get_queryset(self):
+        return super().get_queryset().select_related("content_type", "user")
+
+    action_label = {ADDITION: "addition", CHANGE: "change", DELETION: "deletion"}
```
quadrismegistus, line 186: Suppose we could move this up a few lines to keep it consistent with other attributes

quadrismegistus (approved):
Nice!

geniza/common/metadata_export.py
rlskoeser: good point, will do

geniza/corpus/management/commands/convert_dates.py
```diff
@@ -37,9 +37,11 @@ def handle(self, *args, **options):
 
         # find all documents with original dates set;
         # limit to calendars that we support converting
-        dated_docs = Document.objects.exclude(
-            doc_date_original="", doc_date_calendar=""
-        ).filter(doc_date_calendar__in=Calendar.can_convert)
+        dated_docs = (
+            Document.objects.exclude(doc_date_original="", doc_date_calendar="")
+            .filter(doc_date_calendar__in=Calendar.can_convert)
+            .order_by("pk")  # order by pk to ensure order is determinate
```
quadrismegistus, line 43: Nice

quadrismegistus (approved):

Issue #1102, opened 2022-09-19
As a content admin, I want data exports to include information about who made edits when possible, so that I see who contributed to changes in project data.


rlskoeser:
## testing notes
for testing you should be looking at commits on the test metadata repository here: https://github.com/Princeton-CDH/test-geniza-metadata/commits/main
- [x] metadata exports should now only generate commits when there are changes
- [x] commits should include co-author information 

## dev notes
- [x] add functionality to manage command that syncs data exports to github to track when it was last run
- [x] query django admin log entries to get the list of users who have contributed to the changes since last run (limit to appropriate content types)
- [x] generate a co-author commit message including all contributors since the last time the script was run

kseniaryzhova:
@rlskoeser works for me, closing!

PR #1241, opened 2022-11-17
Add co-author commits to metadata exports #1102

rlskoeser:
this pr revises metadata export script to check log entries to determine if relevant changes have been made, and who made them, and includes co-author message in the commit (now one commit per file)

includes the following changes:
- move existing annotation sync co-author commit logic into a method we can use in both scripts
- generalize last-run logic from annotation export script for reuse in this one
- add content type filters to each exporter — used to identify relevant log entries for that export
- revise metadata export to use log entries to check if the export is necessary and get co-authors
- commit each exported file individually with co-author info

notes:
- had a little bit of trouble figuring out how to adapt my annotation export code with this; feels like there should be more overlap between the two, but the approach was different enough I couldn't see an easy way to generalize it
- I think the metadata export script is complicated because we're trying to do too much — do we need the non-repo export functionality? can we simplify the options and paths and just always do what we want the cron job behavior to be? what are the use cases for the other version of the script?


rlskoeser:
@quadrismegistus using the script more to test document export, and confirmed that in practice we can't really separate write from sync behaviors — it puts the github repo in a bad state and the script fails. (You can't pull if you have local changes).

If we care about the non-git export functionality (which we might, it's kind of useful at least for testing) then I propose we split out a sync metadata export script that is for the git sync version of the export only, and does what the cron job needs to do by default.

geniza/corpus/management/commands/export_metadata.py
```diff
@@ -22,6 +26,16 @@ class MetadataExportRepo(Timerable):
...
 
+    #: default commit message
+    default_commit_msg = "Automated metadata export from PGP"
+
+    exports = {
+        "documents": PublicDocumentExporter,
+        "fragments": PublicFragmentExporter,
+        "sources": SourceExporter,
+        "footnotes": FootnoteExporter,
+    }
```
quadrismegistus, line 37: Nice! We have enough now it's good we're wrangling them together programmatically.

geniza/corpus/management/commands/export_metadata.py
```diff
@@ -101,35 +115,49 @@ def paths(self):
...
+        # pull any remote changes before making any local modifications
+        self.repo_pull()
+
+        # get all log entries since the last run
+        log_entries = LogEntry.objects.filter(action_time__gte=lastrun)
+
+        # if there are NO changes, bail out
+        if log_entries.count() == 0:
+            self.print("No changes since last run")
+            return
```
quadrismegistus, line 130: Good, it's nice we don't have to actually run the export when nothing changes (and the cron job will exit swiftly)

geniza/corpus/management/commands/export_metadata.py
```diff
@@ -101,35 +115,49 @@ def paths(self):
...
+
+        # run all configured exports
+        for export_name, exporter in self.exports.items():
+            with self.timer(f"Exporting { export_name } data"):
```
quadrismegistus, line 134: Cool. I hope this is still useful.

geniza/corpus/management/commands/export_metadata.py
```diff
@@ -101,35 +115,49 @@ def paths(self):
...
+                subset_logentries = log_entries.filter(**exporter.content_type_filter)
+                # if there are no changes, move to next export
+                if not subset_logentries.count():
+                    self.print("No changes for %s" % export_name)
+                    continue
```
quadrismegistus, line 139: Ooh v nice! So we only export precisely the dataset that changed.

geniza/corpus/management/commands/export_metadata.py
```diff
@@ -101,35 +115,49 @@ def paths(self):
...
+
+                export_path = self.get_path_csv(export_name)
+                exporter(progress=self.progress).write_export_data_csv(export_path)
```
quadrismegistus, line 142: Cool line

geniza/corpus/management/commands/export_metadata.py
```diff
@@ -232,10 +280,21 @@ def handle(self, *args, **options):
...
         if options["write"]:
-            mrepo.export_data()
+            # determine script last run
+            lastrun = self.script_lastrun(mrepo.repo)
+            # store the datetime immediately after this query for the next run
+            new_lastrun = timezone.now()
+            mrepo.export_data(lastrun, sync=options["sync"])
+            # when we write and sync, update last run info
+            if options["sync"]:
+                self.update_lastrun_info(new_lastrun)
```
quadrismegistus, line 292: Cool. Thanks for catching this.

geniza/corpus/metadata_export.py
```diff
@@ -193,6 +205,12 @@ class FragmentExporter(Exporter):
         "last_modified",
     ]
 
+    # queryset filter for content types included in this import
+    content_type_filter = {
+        "content_type__app_label__in": ["corpus"],
+        "content_type__model__in": ["document", "fragment", "collection"],
+    }
+
```
quadrismegistus, line 213: This is a nice solution to make sure we're exporting the right docs and objects.

quadrismegistus (approved):
Looks good! Thanks for digging deep on this exporter and figuring out how to get the github syncing to work properly.

geniza/corpus/management/commands/export_metadata.py
rlskoeser: Yes, that's the idea — although it's at a fairly rough level of granularity, and it could be that parts of the model were changed that don't actually impact the export.

geniza/corpus/metadata_export.py
rlskoeser: yeah, but it may be hard to keep it updated if we modify the exports and include data from other models

PR #1242, opened 2022-11-17
Feature/1238 footnote content

rlskoeser:
Update to use new annotation->footnote foreign key to generate footnote content html more efficiently

### DIFF
--- a/geniza/annotations/conftest.py
+++ b/geniza/annotations/conftest.py
@@ -3,7 +3,7 @@ from django.urls import reverse
 
 from geniza.annotations.models import Annotation
 from geniza.corpus.models import Document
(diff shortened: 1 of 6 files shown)


rlskoeser:
@blms this diff is showing metadata export code that wasn't part of the changes I made after branching — I think it must be in `develop` that you don't have; maybe merged after you branched? (this is my fault because I forgot and used git flow to make a feature branch and then rebased on yours, didn't think about potential discrepancies.)   LMK how you prefer to handle — if you can merge develop into your branch, or if I should make a new version of this branch without that commit.

blms (approved):
Model method updates look great, and thanks for the test fixes!

PR #1243, opened 2022-11-17
Update annotation tests failing due to missing required footnote

rlskoeser:
update failing annotation tests:
- provide required footnote when instantiating (where possible / where makes sense)
- mark as skipped tests that depend on in-progress changes like signal handling (should be re-enabled and updated as needed as that work is completed)


blms (approved):
Thank you for handling the tedious part of this! Lifesaver.

PR #1244, opened 2022-11-17
Include scholarship records in documents metadata export #475

rlskoeser:
Per Marina's request in #475 , adding scholarship record summary to documents export.

I figured out the prefetching necessary to make this perform reasonably well. (And also improved a couple of other places.)


geniza/corpus/metadata_export.py
```diff
@@ -49,10 +53,30 @@ def get_queryset(self):
...
+                    queryset=Footnote.objects.select_related(
+                        "source",
+                        "source__source_type",
+                    ).prefetch_related(
+                        "source__authorship_set__creator",
+                        "source__languages",
+                        "source__authors",
+                    ),
+                ),
+            )
```
quadrismegistus, line 77: Oh wow, tricky. Just curious, do we lose any performance here, or not really since only with the admin interface does this method (of clearing prefetches) do something significantly different?

geniza/corpus/metadata_export.py
```diff
@@ -112,6 +136,10 @@ def get_export_data_dict(self, doc):
         outd["type"] = doc.doctype
         outd["tags"] = doc.all_tags()
         outd["description"] = doc.description
+        # include short display version of scholarship records;
+        # need to use set since some sources have duplicate footnotes
+        # in order to keep track of multiple links / PDFs
+        outd["scholarship_records"] = set([fn.display() for fn in doc.footnotes.all()])
```
quadrismegistus, line 142: I like using the set comprehension syntax `{fn.display() for fn in doc.footnotes.all()}` but it's just style

quadrismegistus (approved):

geniza/corpus/metadata_export.py
rlskoeser: I'm clearing and then reapplying all of the ones we need for this export — there's a discrepancy in how we're prefetching footnotes for admin and footnotes for the data export (now that we need more details for the data export), and this seemed like the simplest solution.

geniza/corpus/metadata_export.py
rlskoeser: Ah! I don't think I knew you could do that. That does look nicer. (Although I confess that I do get a little bit of visual confusion between that kind of set notation and dicts sometimes)

PR #1258, opened 2022-12-05
Remove wagtail pages using deleted CreditsPage model

blms:
## In this PR

- Add a migration (and unit tests) to delete all wagtail `Page` instances still using the deleted `CreditsPage` content type.

I found this bug while experimenting with Wagtail to help troubleshoot other issues. There were 500 errors on some of the translated "container pages" due to child pages using this stale deleted model. 

Those pages will be outdated anyway—they were autogenerated credits pages using the old Contributor model, and only still exist among the Hebrew and Arabic content—so safe to delete.


rlskoeser (approved):
wow, wild that the pages can exist after the custom model is deleted. thanks for catching & cleaning up

Issue #1216, opened 2022-10-26
transcription labels in search results are RTL


rlskoeser:
transcription labels in search results are inheriting the RTL text direction we have set on the transcription block; we should exempt them (which we already do on the document detail view)

![Screen Shot 2022-10-26 at 4 20 23 PM](https://user-images.githubusercontent.com/691231/198128929-b95a0e9e-c027-4bff-bb38-eebf529114ff.png)

kseniaryzhova:
@rlskoeser transcription labels seem LTR aligned to me - punctuation is where it should be, etc. Closing!

PR #1265, opened 2022-12-07
address transcription label text direction

rlskoeser:
ref #1216 

- explicitly set text direction for transcription labels
- set text direction in html for transcription block in itt panel, for consistency with search results


blms (approved):
This looks right to me. The only additional implication I can think of is potential interference with `text-align` properties on search results and the ITT panel. If you've checked both of those in LTR and RTL layouts (and in the admin) and they're aligned correctly, then I think it's good to go!

rlskoeser:
> ... The only additional implication I can think of is potential interference with `text-align` properties on search results and the ITT panel. If you've checked both of those in LTR and RTL layouts (and in the admin) and they're aligned correctly, then I think it's good to go!

Thanks for suggesting I check all of those. I just checked them now, and this change doesn't seem to impact anything else.

PR #1267, opened 2022-12-07
revise anotation search view for annotation model changes

rlskoeser:
part of #1238

Issue #1238, opened 2022-11-16
Revise annotation model to link footnotes using foreign keys instead of URIs


blms:
## Testing notes

On the QA admin site:

- [x] Check a few random transcriptions and confirm they are still present/accessible.
- [x] Check Log Entries for new footnotes added by the "script" user on Dec. 16, 2022, 2:16 pm. Click on some of the footnote objects: they should be "digital edition" footnotes with associated transcription content.
  - [x] Go to the document they belong to (PGPID/title should be in the "Content object" field) and confirm the digital edition for that author/source has transcription content (i.e. using Edit X's Edition, or the dropdown menu above the transcription text)
  - [x] On that document page, scroll down to the Footnotes section. Locate the digital edition footnote for the transcription you were just looking at, and check the "delete" box next to it at the far right end of the table. Click "save" at the bottom of the document page and confirm you get an error message, telling you that you can't delete a footnote with associated annotations here.

On the QA public site:

- [x] Check a few random transcriptions and confirm they are still visible on the public site; try editing them and confirm they show up in the editor too. (note: recent beta-tested editor fixes are not present on the site yet)

Please also test:

- #1207 
- #1280 

## dev notes

Epic to collect related or affected bugs/chores.

Initial steps: 

- [x] revise annotation model to add FK to footnote
- [x] data migration to associate existing footnotes + annotations explicitly
  - [x] if digital edition footnote cannot be found but source and document can, then create new digital edition (with LogEntry)
  - [x] generate report of any remaining problem annotations
- [x] if you delete a footnote from the document edit page, warn or prevent deletion
- [x] revise from json code to set footnote based on source + manifest uri when saving
- [x] revise to json code to serialize source & manifest uri when displaying
- [x] update annotation search methods
- [x] move signal handler footnote code into annotation view methods
- [x] everywhere we are looking up documents/sources by uri, we now get them by model accessors
- [ ] #1279
- Decided not to do this to avoid altering Tahqiq: ~remove `Document.from_manifest_uri`, `Document.id_from_manifest_uri`, `Source.from_uri`, and `Source.id_from_uri` and replace all usage with lookups on `Footnote`~

blms:
Here is a list of known "digtial editions" without :

<details>
<summary>Digital editions without annotations</summary>

<ul>
[...]

PR #1268, opened 2022-12-07
revise annotation model serialization to JSON for annotation model changes

rlskoeser:
part of #1238

should be reviewed after #1267


blms (approved):
Looks great, thanks!

Issue #1260, opened 2022-12-05
As a content editor, I want the log entry to record and differentiate between users who input someone else's transcription versus users who created a new transcription so I can give the appropriate credit where it's due.


kseniaryzhova:
## Testing notes

On the [QA site admin](https://test-geniza.cdh.princeton.edu/admin/), go to a document and add a footnote (not a digital edition). Click "save and continue editing." 
- [x] Then, add a transcription using the transcription editor. You can choose any source for this.
- [x] Save the transcription, then make an edit to it, and save it again.

Return to the document in the admin, and click the "History" button. 
- [x] Under "Change history" you should see an entry for the footnote you added manually. 
- [x] Under "Related footnote(s) change history", you should see an entry for the digital edition you just made by transcribing. Clicking on that digital edition should show you the source (including author information) and allow you to add notes.
- [x] Under "Related annotation(s) change history", you should also see history entries for the transcription you made, and the edits you made to it.
- [x] Your user name should be attached to each entry, with a date and timestamp.
- [x] Confirm this makes it clear who added or edited transcriptions, and who is the original author of each (via source on the footnote).

---

**Is your feature request related to a problem? Please describe.**
When a user inputs a transcription, it does not differentiate between users who input their own transcriptions versus users who input others' transcriptions (like from published secondary sources) and those who edited existing transcriptions. This will help us give credit where it's due and also give us a trail to follow if there are issues, questions, etc. 

**Describe the solution you'd like**
I want the log entry in the admin interface to include information about who input a transcription but was not the original author (input), who created the transcription and input it (created by) and who edited an existing transcription (edited). 

**Describe alternatives you've considered**
Original authors of transcriptions who did not input the transcription get credit via their sources so that's solved in a different area of the document detail page in admin.

**Additional context**
By log in the admin interface we mean this. Marina made changes to the transcription yet they were not recorded as such: ![image](https://user-images.githubusercontent.com/105385128/205720488-e1ab3120-1089-4fe2-84c6-805589296d94.png)

rlskoeser:
There are log entries on the annotation objects that track who is creating and updating them and when; for example, here's the annotation for PGPID 25821 https://geniza.princeton.edu/admin/annotations/annotation/996bdaa9-8709-4791-b1c0-2cadb81caec8/history/?_changelist_filters=q%3D25821 

[...]

PR #1287, opened 2023-01-06
Transcription logging (#1260)

blms:
## In this PR

- Per  #1260:
  - Add log entries for related footnotes and annotations to document history view/template


rlskoeser (approved):
I'm not pulling any code to run & check locally, but the implementation looks reasonable to me. Glad this approach will work instead of adding redundant log entries on document.

Issue #1288, opened 2023-01-09
Some JRL images not displaying correctly


kseniaryzhova:
## Testing notes

On the [QA site](https://test-geniza.cdh.princeton.edu/), visit a document where both sides used to be marked "recto", such as PGPID 28296. It should now have the second image marked "verso".

---
**Describe the bug**
Some JRL images are displaying with two distinct issues:

- ~~[Only one side of the fragment is displaying](https://geniza.princeton.edu/en/documents/32579/)~~
  - Probably a manchester issue, contact manchester and revisit
- [Both sides of the fragment are marked "Recto"](https://geniza.princeton.edu/en/documents/28296/)

Many JRL images are displaying correctly, so this does not affect the whole collection. 

**To reproduce**
1. See above examples, but also try searching by the [JRL collection with images in admin](https://geniza.princeton.edu/admin/corpus/document/?textblock__fragment__collection__id__exact=23&textblock__fragment__iiif_url__isempty=0) and looking through some of the shelfmarks and how the images display in the editor. 

**Expected behavior**
I want each JRL fragment to display accurately-labelled images of the recto and verso. 

**Additional context**
RSK suggests revising the jrl_iiif.py script to override labels for verso

kseniaryzhova:
Looked into the shelfmarks that are displaying only one image (so far it's JRL Gaster ar. 1, 11, 111, and 124). Their versos are visible on FGP, but are not visible on Manchester's site (via LUNA), probably because the versos on all of these documents are completely blank. This is a Manchester issue, since we get our IIIF manifests from their site and cannot display images that aren't there.

kseniaryzhova:
@blms Yes assume the second fragment is a verso and ignore the label from Manchester.

kseniaryzhova:
@blms works well for this one example I found! (I looked at some other JRL images too and they are displaying recto/verso as they should). Thanks, closing!

PR #1294, opened 2023-01-12
Ensure 2nd image in JRL IIIF is always "verso"

blms:
## In this PR

Per #1288 
- Use `row.sequence` to find the second image in each set from JRL IIIF, and ensure that its label reads "verso" and not "recto"

## Questions

- Is there a way to test this locally, and what settings should I use to do so?
- When this is ready to go, can you help me walk through the steps to run this script on the server and deploy? I imagine it's similar to what we did with the IIIF CSV import, but since it involves the extra step of publishing to GitHub (and it involves existing records) I want to make sure I don't mess it up.

### DIFF
--- a/requirements/scripts.txt
+++ b/requirements/scripts.txt
@@ -1,0 +1,6 @@
+# requirements for the scripts in the scripts directory
+
+iiif
+pandas
+python-slugify
+ratelimit
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -3,6 +3,8 @@
 This directory contains stand-alone scripts associated with
 Princeton Geniza Project that are not part of the web application proper.
 
+Requirements for these scripts can be found in `requirements/scripts.txt`.
+
 ## IIIF
 
 Scripts for generating and managing static iiif content (manifests and
@@ -11,6 +13,7 @@ iiif image api level 0 images) for import and display in PGP.
 -   bodleian_iiif.py : generate iiif maniests from Bodleian TEI XML
 -   tile_images.py : generate static image tiles; includes extra image sizes needed for PGP application
 -   manifests_to_csv.py: generate a CSV file for importing IIIF urls into PGP
+-   jrl_iiif.py: generate remixed iiif maniests from Manchester JRL manifests
 
 ### Bulk editing
 
--- a/scripts/jrl_iiif.py
+++ b/scripts/jrl_iiif.py
@@ -116,12 +116,19 @@ def combine_manifests(csvfilepath, output_dir):
                 v.label: v.value for v in manifest.sequences[0].canvases[0].metadata
             }
             reference_number = row.reference_number
+
+            # NOTE: not every record has a folio; use reference number as fallback
+            folio = canvas_metadata.get("Folio", reference_number)
+            # override folio to use "verso" for second image, in case of double recto
+            if row.sequence == "2" and "recto" in folio:
+                folio = folio.replace("recto", "verso")
+
             if current_shelfmark == shelfmark:
                 # add canvas + image from current manifest to the new one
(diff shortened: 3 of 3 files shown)


rlskoeser (approved):
logic looks sound

rlskoeser:
> * Is there a way to test this locally, and what settings should I use to do so?

Yes, you should be able to run the script locally. The csv needed for input is checked into the scripts directory, and you specify an output directory. It's a python script and I set it up with argparse, so it has minimal help output for usage.

> * When this is ready to go, can you help me walk through the steps to run this script on the server and deploy? I imagine it's similar to what we did with the IIIF CSV import, but since it involves the extra step of publishing to GitHub (and it involves existing records) I want to make sure I don't mess it up.

There's some info in the scripts readme, but may not be enough: https://github.com/Princeton-CDH/geniza/tree/main/scripts

Feel free to expand on that as you figure things out (I can review) and please add new sections as you add new scripts.

I think the basic steps you need to take are roughly this:
- clone the genizalab iiif repo that I invited you to https://github.com/Princeton-CDH/geniza/tree/main/scripts
- make a new branch for your changes
- run the jrl script on the csv in the scripts directory and specify the jrl folder in the iiif repo as your output dir
- review changes locally 
- check into git, push, open pr so we can both review changes
- when the pr is merged, changes will automatically be published to the github pages site

I think that this set of changes should not change any of the manifest urls, so running the import iiif script won't help here...I think we'll need to do is use the djiffy import manifest command to update the cached manifest/canvas info in the db. I can help with more specifics when you get to that point if needed. (You can probably test this locally with one or two urls, but I don't remember if there is a good way to do it in bulk except via manifest collection.)

blms:
@rlskoeser Thanks for the detailed steps! I just opened a PR on that repo: https://github.com/princetongenizalab/iiif/pull/1

As I mentioned there, one manifest went missing. It actually caused an `AttributeError` on `manifest.sequences`, meaning somehow it didn't fail on `IIIFPresentation.from_url`, but made it to the next step and ran into an error there instead. Maybe it's because it's [returning JSON](https://luna.manchester.ac.uk/luna/servlet/iiif/m/ManchesterDev~95~2~3397~108777/manifest).

Should we handle this in the script and report on any failures, or just remove the problematic entry from the CSV entirely? I have changes locally to do the former if desired.

I also wrote up a quick `requirements.txt` and `requirements.lock` since I was missing some deps, if we want to include those in the folder. Or we could include the requirements in `requirements/dev.txt`. Let me know if you have thoughts on that!

rlskoeser:
> @rlskoeser Thanks for the detailed steps! I just opened a PR on that repo: [princetongenizalab/iiif#1](https://github.com/princetongenizalab/iiif/pull/1)
> 
> As I mentioned there, one manifest went missing. It actually caused an `AttributeError` on `manifest.sequences`, meaning somehow it didn't fail on `IIIFPresentation.from_url`, but made it to the next step and ran into an error there instead. Maybe it's because it's [returning JSON](https://luna.manchester.ac.uk/luna/servlet/iiif/m/ManchesterDev~95~2~3397~108777/manifest).

Ah, could be.

> Should we handle this in the script and report on any failures, or just remove the problematic entry from the CSV entirely? I have changes locally to do the former if desired.

Seems simpler to remove from the CSV

> I also wrote up a quick `requirements.txt` and `requirements.lock` since I was missing some deps, if we want to include those in the folder. Or we could include the requirements in `requirements/dev.txt`. Let me know if you have thoughts on that!

I was thinking of these as separate from the main codebase — I don't expect to run them on the application server, so don't want to install any reqs for these that we don't need there. I'm fine with a separate scripts requirements file or a section in the dev reqs with a comment.

Issue #1252, opened 2022-11-23
As a content editor, I want a warning or validation to prevent adding more than one digital edition footnote for the same document source to avoid creating duplicates.


kseniaryzhova:
## Testing notes

In the [QA site admin](https://test-geniza.cdh.princeton.edu/admin/), on a document, try to create two footnotes by attaching the _same_ source twice and choose "digital edition" as the relation for both. It should give you an error when you try to save the document.

---

**Is your feature request related to a problem? Please describe.**
Outlined in #1200 - we do not have a way to prevent duplicate digital editions appearing in the same footnote. 

**Describe the solution you'd like**
I want an error message, warning or some kind of validation when I add more than one digital edition footnote for the same source

kseniaryzhova:
@blms saving a second digital edition (of the same source) gives me a server error message like this: 
![image](https://user-images.githubusercontent.com/105385128/212405347-b988b1b9-6454-4844-b371-928d568c580b.png)
I don't think this makes it clear to users what went wrong/what they should fix.

kseniaryzhova:
@blms Works! I get the error message on the document detail page explaining the issue! Thanks, closing!

PR #1296, opened 2023-01-16
Display proper validation errors on duplicate digital editions (#1252)

blms:
## In this PR

Per https://github.com/Princeton-CDH/geniza/issues/1252#issuecomment-1382314066:
- Raise validation errors in the `clean` methods of footnote admin inlines and the single footnote admin form, if there are multiple digital editions on the same source/document. Fixes issue where it would just throw a DB integrity error due to the uniqueness constraint violation, and present the user with an opaque 500 error.

## Notes

It's pretty annoying that DB integrity errors don't automatically generate validation errors! 

Also realized I did a lot of extra work for something that wasn't necessary: validating on deletion of footnotes from the Source admin. You can't even do that in the first place! 🤦 

I was originally planning on making this a hotfix (thought showing an error would be simple!), but I think the code changes here are substantial enough to warrant review. Wondering about both the formsets and the testing approach.

### DIFF
--- a/geniza/footnotes/admin.py
+++ b/geniza/footnotes/admin.py
@@ -4,10 +4,12 @@ from django.contrib import admin
 from django.contrib.admin import SimpleListFilter
 from django.contrib.contenttypes.admin import GenericTabularInline
 from django.contrib.contenttypes.forms import BaseGenericInlineFormSet
+from django.contrib.contenttypes.models import ContentType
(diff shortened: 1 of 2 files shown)


rlskoeser (approved):
wow, complicated formset stuff. everything looks reasonable, appreciate the clear comments and detailed test cases

Issue #1295, opened 2023-01-12
De-selected images display yellow border in viewer and editor when no image on TextBlock is selected


kseniaryzhova:
## Testing notes

In the [QA site admin](https://test-geniza.cdh.princeton.edu/admin), edit a document that joins multiple fragments. On one of the joined fragments, deselect _all_ images and save. The document should no longer show the yellow border for the deselected images. The same should be true if you only deselect _some_ of the images on the fragment.

---

**Describe the bug**
Deselected images in the related fragments section still have a yellow border around them in both the viewer and the transcription editor (indicating they are part of the PGPID). 

**To reproduce**
1. Go to [Halper 389 ](https://geniza.princeton.edu/admin/corpus/document/2587/change/?_changelist_filters=q%3DHalper%2B389%26textblock__fragment__iiif_url__isempty%3D0%26transcription%3Dyes)or [Halper 414](https://geniza.princeton.edu/admin/corpus/document/35364/change/?_changelist_filters=q%3DHalper%2B389).
2. In the related fragments section, select the correct fragment and de-select the other fragment.
3. See error.

**Expected behavior**
When I purposefully deselect fragment images, I want the yellow border to be removed in both the viewer and the editor, to serve as a visual clue that those images do not belong with this PGPID.

blms:
Closing as `wontfix` per meeting on 2/1/2023. Current behavior is acceptable.

PR #1298, opened 2023-01-19
Ignore TextBlock side info when excluding images (#1295)

blms:
## In this PR

Per #1295:
- Fixes issue where sometimes deselected fragment images would still appear as part of the document in the admin

## Notes

After some debugging, I figured out that the exclusions were only failing when there were _no_ selected images in a given `TextBlock`. Seems like the code I removed is a vestigial check from back when `TextBlock.side` was an actual property and not something computed based on `selected_images`—[this is the oldest code I could find](https://github.com/Princeton-CDH/geniza/blame/ee6d8d1e0fb377fe7168e92ef9983da82b6ace28/geniza/corpus/models.py#L624) with this check.

Unit tests are all still passing, and the admin and public images seem to be working as expected, but maybe I'm missing something. Wish I had documented why I was not counting exclusion unless `b.side` was defined. Let me know if you think of any reason why this check would need to be here!

Not directly connected to this check, but a baked-in assumption: I guess the feature was designed thinking that at least _one_ image would be selected in a given TextBlock, or else why attach it as part of the join? Perhaps that's a mistaken assumption and I need to think through the other implications with the rest of the team. That said, such a decision/conversation is not a blocker for moving forward with this fix, as far as I can tell.

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -754,7 +754,7 @@ class Document(ModelIndexable, DocumentDateMixin):
                             "label": labels[i],
                             "canvas": canvases[i],
                             "shelfmark": b.fragment.shelfmark,
-                            "excluded": b.side and i not in b.selected_images,
+                            "excluded": i not in b.selected_images,
                         }
 
         # when requested, include any placeholder canvas URIs referenced by any associated transcriptions


rlskoeser (approved):
thanks for the detailed explanatory notes in the pr — I think you are right, this must be leftover from older code when side was not a computed property

PR #1302, opened 2023-01-26
Make new TextBlocks select all images by default, migrate empty selected_images (#1295)

blms:
## In this PR

Per #1295, fixes the following issues:

- New TextBlocks were always created with 0 images selected, now they will have _all_ images selected on creation
- Migrates existing TextBocks with 0 images selected to have _all_ images selected in the database, as this is by far the more common case than intentionally having 0 images selected

## Notes

Since we'd previously treated 0 images selected as equivalent to all images selected, and we also selected 0 images on TextBlock creation, after #1298 suddenly the majority of TextBlocks' images appeared as non-document images. You can see the result on the QA site now 😬 

I've let Ksenia know that this will require some manual cleanup for those that were intended to have 0 images selected, but I suspect there are few of those.


rlskoeser (commented):
the code looks reasonable, but I think there is at least one case not handled — what happens when new iiif images are loaded? will all the images be deselected by default?

(do we really have a use case for deselecting all images?)

blms:
> the code looks reasonable, but I think there is at least one case not handled — what happens when new iiif images are loaded? will all the images be deselected by default?

@rlskoeser Great catch, they would be. I guess that should go in the `Fragment.save` method, when adding a `iiif_url` or `manifest` for the first time, right? To update all associated `TextBlocks` with no selected images? I think that should cover both IIIF scripts and manual entry.

> (do we really have a use case for deselecting all images?)

I'm not sure, and I intend to bring it up in my meeting with MR/RR/KR next week. But even if not, my thought is that it's better that such an edge case behaves as expected rather than treating all images as selected when technically none are.

rlskoeser:
> @rlskoeser Great catch, they would be. I guess that should go in the `Fragment.save` method, when adding a `iiif_url` or `manifest` for the first time, right? To update all associated `TextBlocks` with no selected images? I think that should cover both import scripts and manual entry.
> 
> EDIT: Or maybe a signal handler makes more sense since TextBlock is a different model…

Updating fragment save seems reasonable — you can access related textblocks and catch if iiif url has changed. This won't account for the associated manifest being updated, but should get you closer.

> > (do we really have a use case for deselecting all images?)
> 
> I'm not sure, and I intend to bring it up in my meeting with MR/RR/KR next week. But even if not, my thought is that it's better that such an edge case behaves as expected rather than treating all images as selected when technically none are.

I was thinking of manual selection really as a way to indicate deselection, so zero is the same as all (no action). IDK if that's technically any simpler, but seems lie it would avoid problems like these when iiif changes (add/remove/update).

blms:
>I was thinking of manual selection really as a way to indicate deselection, so zero is the same as all (no action). IDK if that's technically any simpler, but seems lie it would avoid problems like these when iiif changes (add/remove/update).

Definitely technically simpler—enough so that I want to see if MR/RR/KR agree that there's no real use case for 0 selected before moving forward with any further work on this. Thank you for the quick review in any case!

blms:
Closing with issue #1295 marked `wontfix`! Thanks again Rebecca for the points made here.

Issue #506, opened 2022-01-06
As a content editor, I want to merge similar tags so I can consolidate redundant tags and decrease clutter in the database.


richmanrachel:
## Testing notes

In the [QA site admin](https://test-geniza.cdh.princeton.edu/admin), go to the Taggit > Tags section.

- [x] Select several tags that should be merged into one canonical tag. Make note of the documents tagged with the non-canonical variations.
- [x] There should be a dropdown menu item at the top of the Taggit > Tags list that says "Merge selected tags". Choose that dropdown option and click "ok". You should be taken to a form.
- [x] There should be a list of all the tags you chose, and each tag should show the number of documents tagged with it.
- [x] Choose the canonical tag in this list and click "submit". You should be taken to the page for that tag, showing all of its tagged documents.
- [x] All documents that used to be tagged with the other tags should be tagged with the canonical tag. The other tags also should not appear in the list under Taggit > Tags.

---

**Is your feature request related to a problem? Please describe.**
The tagging feature is currently cluttered due to similar tags (slaves vs. slavery, etc). This will help us clean the data and maintain a more useful set of tags in perpetuity. 

**Describe the solution you'd like**
In the Tag List, select the tags you want to merge in the action bar and click go.

## dev notes
New admin action to merge tags, similar to #225 — needs a review screen so editor can choose which tag should remain and confirm the change. Should create a log entry documenting the merge.

mrustow:
Can I make a plug for this issue? I'd like to do some tag cleanup. 

Also, in the admin interface > Taggit > Tags, how hard would it be to add a column with the number of times the tag appears in the database? If there are zeros that aren't also duplicates, it would be nice to be able to delete them without having to click into the tag to see whether it's being used for a document. If you agree, I'd be happy to write a user story.

thx :)

mrustow:
@kseniaryzhova flagging this: can we put this in the queue?

kseniaryzhova:
@blms works like a dream, thanks! Closing!

PR #1305, opened 2023-01-30
Add tag merge feature (#506)

blms:
## In this PR

Per #506:
- Add feature to tag admin allowing admins to merge several tags into one primary tag

## Questions

- This may not be entirely DRY, considering how much was cribbed from the document merge feature, but it seems like there is just enough variance in each method that it might not be straightforward to generalize. Is this repetitive enough that it's worth the time to generalize? (I guess one reason to do so would be if there are other kinds of merges they will need to do in the future, but I'm not sure there are.)
- Do you think it's ok to not require a rationale here? I think tag merges are pretty straightforward and less impactful so I didn't think it was necessary.

### DIFF
--- a/geniza/corpus/forms.py
+++ b/geniza/corpus/forms.py
@@ -390,8 +390,8 @@ class TagMergeForm(forms.Form):
         label="Select primary tag",
         queryset=None,
         help_text=(
-            "Select the primary tag, which the other selected tags will be merged into. "
-            "All tagged items will be combined on the merged tag."
+            "Select the primary tag, which will replace the names of the other selected tags. "
+            "All items tagged with the other selected tags will then only be tagged with the primary tag."
         ),
         empty_label=None,
         widget=forms.RadioSelect,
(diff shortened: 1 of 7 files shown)


geniza/corpus/forms.py
```diff
@@ -374,3 +376,40 @@ def clean(self):
...
+    def label_from_instance(self, tag):
+        return "%s (%s tagged)" % (tag.name, tag.item_count)
+
+
+class TagMergeForm(forms.Form):
+    primary_tag = TagChoiceField(
+        label="Select primary tag",
+        queryset=None,
+        help_text=(
+            "Select the primary tag, which the other selected tags will be merged into. "
```
rlskoeser, line 393: not sure this wording makes sense here like it does for documents; it's really more about renaming, right? (but this is a quibble, you can get input from the team on what makes sense to them)

rlskoeser (approved):
Looks reasonable to me.

To answer your questions:
1. Too early to try to generalize (and probably not your job 🙂 ), since we only have two merges in this app.
2. I agree, rationale seems unnecessary here and not important like it is for documents.

Issue #1230, opened 2022-11-10
Allow placeholder images to have unique labels corresponding to the shelfmark


kseniaryzhova:
## Testing notes

On the [QA site](https://test-geniza.cdh.princeton.edu/), find a few documents that include fragments with missing images: documents with no images whatsoever, and joins where only some of the fragments have images. For each of these types:

- [x] Open the transcription editor (create or edit a transcription). Placeholder canvases (that don't currently contain transcription text) should be labeled with the shelfmark and recto/verso.
- [x] If there is transcription text already for those fragments, the existing placeholders should still be unlabeled. However, you should now be able to move them to the correct, labeled placeholders.

(see also #1229)

---

**Is your feature request related to a problem? Please describe.**
Related to #1229 - if there are more than two fragments without IIIF images in a document with a transcription, we need to be able to distinguish between the fragments so we can accurately attach transcriptions. 

**Describe the solution you'd like**
I want a display for each placeholder that gives the shelfmark labels similar to what IIIF image document displays have on the site now.
![image](https://user-images.githubusercontent.com/105385128/201203717-d06ddafb-5e5d-435e-bdf8-af9f7a7b4783.png)

**Describe alternatives you've considered**

**Additional context**
This will also be useful for rearranging the placeholder images and might be a way to attach/update images to documents with existing transcriptions.

kseniaryzhova:
@blms works, thank you!

Issue #1229, opened 2022-11-10
Include two placeholder images for each fragment associated with a document


kseniaryzhova:
## Testing notes

On the [QA site](https://test-geniza.cdh.princeton.edu/), find a few documents that include fragments with missing images: documents with no images whatsoever, and joins where only some of the fragments have images. For each of these types:

- [x] There should now be two placeholder images for each fragment that does not have any images.

(see also #1230)

---

**Is your feature request related to a problem? Please describe.**
When a join has fragments that both do and do not have IIIF images, there are no placeholder images for the fragments without IIIF images. This makes it impossible to attach transcriptions to the correct fragment (we have had to attach to the fragments with IIIF images). Ex: [https://geniza.princeton.edu/en/documents/35279/](https://geniza.princeton.edu/en/documents/35279/)

**Describe the solution you'd like**
I would like for two placeholder images to be generated for every fragment associated with a document (so for a fragment that has transcription but no image)

**Additional context**
We have placeholder images for documents without images but with existing transcriptions, the issue is when the document contains some but not all IIIF images (like joins).

richmanrachel:
One additional detail I had in mind to prevent clutter: Is it possible for the two placeholder images to appear on the backend, but not the frontend unless actively selected?

I imagine the logic for this might be challenging, so feel free to disregard if it's too much of a barrier.

kseniaryzhova:
@blms and they have the shelfmark name with the recto and verso indicated - woo! Closing, thank you!

PR #1306, opened 2023-01-30
Add placeholders with labels for imageless fragments (#1229, #1230)

blms:
## In this PR

Per #1229:
- Revise `DocumentTranscribeView` placeholder image logic to include exactly two placeholders for each fragment without images in the transcription editor, so that joins with multiple fragments will show enough placeholders to add all transcription content

Per #1230:
- On placeholder images, show labels with each associated fragment's shelfmark (and recto/verso for 1st and 2nd image)
- Include TextBlock id in placeholder canvas URI
- Revise `Document.iiif_images` to look for that TextBlock id and associate the placeholders with the correct shelfmark labels

## Notes

I think the method of using the TextBlock id in the placeholder URI is a little bit hacky, but it seems to work. My thinking was that this way, if images get added after transcriptions, the placeholder images will still show up with the correct labels by the transcription, so it will be easy to identify which fragment and side they belong to. 

A potential drawback for that type of document is that for the time between new images being added and transcription content being re-assigned from the placeholders, you end up with double entries on the viewer for each of those fragments—one for the real image and one for the placeholder. But I think that's better than no label for the placeholder (which is the problem in #1230).

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -240,10 +240,15 @@ class Fragment(TrackChangesModel):
         canvases = []
         # use images from locally cached manifest if possible
         if self.manifest:
-            for canvas in self.manifest.canvases.all():
-                images.append(canvas.image)
-                labels.append(canvas.label)
-                canvases.append(canvas.uri)
+            manifest_canvases = self.manifest.canvases.all()
+            # handle no canvases on manifest; cached QS will not incur extra DB hit
+            if not manifest_canvases.count():
+                return None
+            else:
+                for canvas in manifest_canvases:
+                    images.append(canvas.image)
+                    labels.append(canvas.label)
+                    canvases.append(canvas.uri)
 
         # if not cached, load from remote url
         else:
@@ -257,6 +262,7 @@ class Fragment(TrackChangesModel):
                     canvases.append(canvas.uri)
             except (IIIFException, ConnectionError, HTTPError):
                 logger.warning("Error loading IIIF manifest: %s" % self.iiif_url)
+                return None
 
         return images, labels, canvases
 
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -1,5 +1,7 @@
 import logging
+import re
 from collections import defaultdict
+from copy import deepcopy
 from functools import cache, cached_property
 from itertools import chain
 
@@ -764,7 +766,24 @@ class Document(ModelIndexable, DocumentDateMixin):
                 for canvas_uri in ed.content_html.keys():
                     if canvas_uri not in iiif_images:
                         # use placeholder image for each canvas not in iiif_images
-                        iiif_images[canvas_uri] = Document.PLACEHOLDER_CANVAS
+                        iiif_images[canvas_uri] = deepcopy(Document.PLACEHOLDER_CANVAS)
+                        uri_match = re.search(
+                            r"textblock\/(\d+)\/canvas\/(\d)\/", canvas_uri
+                        )
+                        if uri_match:
+                            # if this was created using placeholders in the transcription editor,
+                            # try to ascertain and display the right fragment shelfmark and label
+                            try:
+                                iiif_images[canvas_uri][
+                                    "shelfmark"
+                                ] = TextBlock.objects.get(
+                                    pk=int(uri_match.group(1))
+                                ).fragment.shelfmark
+                            except TextBlock.DoesNotExist:
+                                pass
+                            iiif_images[canvas_uri]["label"] = (
+                                "recto" if int(uri_match.group(2)) == 1 else "verso"
+                            )
 
         # if image_order_override not present, return list, in original order
         if not self.image_order_override:
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -739,8 +739,9 @@ class Document(ModelIndexable, DocumentDateMixin):
         :param with_placeholders: if there are digital editions with canvases missing images,
             include placeholder images for each additional canvas (default: False)"""
         iiif_images = {}
+        textblocks = self.textblock_set.all()
 
-        for b in self.textblock_set.all():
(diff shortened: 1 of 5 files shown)


geniza/corpus/views.py
```diff
@@ -752,13 +753,24 @@ def get_context_data(self, **kwargs):
...
-        # pass two placeholder canvases for use in editor
-        if not context_data["images"]:
-            canvas_base_uri = "%siiif/canvas/" % self.get_object().permalink
-            for i in [1, 2]:
-                canvas_uri = "%s%d/" % (canvas_base_uri, i)
-                context_data["images"][canvas_uri] = Document.PLACEHOLDER_CANVAS
+        # per each fragment without images, pass two placeholder canvases for use in editor
+        for b in self.object.textblock_set.all():
+            frag_images = b.fragment.iiif_images()
+            if frag_images is None or not frag_images[0]:
```
rlskoeser, line 759: why do you need to check `frag_images[0]` ?

geniza/corpus/models.py
```diff
@@ -764,7 +766,24 @@ def iiif_images(self, filter_side=False, with_placeholders=False):
...
+                        )
+                        if uri_match:
+                            # if this was created using placeholders in the transcription editor,
+                            # try to ascertain and display the right fragment shelfmark and label
+                            try:
+                                iiif_images[canvas_uri][
+                                    "shelfmark"
+                                ] = TextBlock.objects.get(
+                                    pk=int(uri_match.group(1))
+                                ).fragment.shelfmark
```
rlskoeser, line 781: this looks like an extra db call per placeholder canvas

geniza/corpus/models.py
```diff
@@ -764,7 +766,24 @@ def iiif_images(self, filter_side=False, with_placeholders=False):
                 for canvas_uri in ed.content_html.keys():
                     if canvas_uri not in iiif_images:
                         # use placeholder image for each canvas not in iiif_images
-                        iiif_images[canvas_uri] = Document.PLACEHOLDER_CANVAS
+                        iiif_images[canvas_uri] = deepcopy(Document.PLACEHOLDER_CANVAS)
+                        uri_match = re.search(
+                            r"textblock\/(\d+)\/canvas\/(\d)\/", canvas_uri
+                        )
```
rlskoeser, line 772: regex seems like overkill, but I guess I can see the convenience. recommend using named matches so the group reference is more semantic when you pull textblock and canvas ids

geniza/corpus/models.py
rlskoeser: you should be able to find the textblock from the current document object, which should avoid extra db calls. if you use `document.textblock_set.filter` I think it still hits the db hit, so maybe a list comprehension to get the right textblock by id.  (it does feel a bit weird but I get the logic)

rlskoeser (commented):
Looks ok. A few comments / small concerns.

What about all the existing placeholders — is any action needed to migrate all of the transcriptions attached to placeholder canvases to use the new format? or will it need to be done manually? and if so, will things look weird until that cleanup is done?

geniza/corpus/views.py
blms: It's a very rare case, but it is possible that `fragment.iiif_images()` could return the list `[[], [], []]`. It would happen when `fragment.iiif_url` is defined but `fragment.manifest` is not, and then there is an error retrieving the remote manifest. It happens in testing, which is why I added it!

rlskoeser: Ah. is it an error in the other code that we're returning that? 

Please add a brief comment here to document this rare case, just so we know why/when it could happen.

Would it help if we got rid of the remote manifest loading? I remember we planned to phase that out but didn't finish all the refactoring. It makes sense to me it could happen in test; I hope it won't happen in production, or at least only very rarely.

blms: Good point, it probably makes more sense to just `return None` in that case from the other code.

> Would it help if we got rid of the remote manifest loading?
Yes, but we'll need to finish up the refactor as you say! I'm not sure where it falls in priority.

rlskoeser: > Yes, but we'll need to finish up the refactor as you say! I'm not sure where it falls in priority.

low priority — not your job unless it starts impacting the site either for feature development or is related to a bug (seems unlikely)

blms:
> What about all the existing placeholders — is any action needed to migrate all of the transcriptions attached to placeholder canvases to use the new format? or will it need to be done manually? and if so, will things look weird until that cleanup is done?

I'm looking at how I might do this automatically. I could just assume that an annotation on the first placeholder canvas = the first fragment, recto; and the second placeholder canvas = the first fragment, verso. It would be problematic for joins, though, if the intended annotated canvas is not actually the first fragment. 

Some of the annotation labels contain shelfmarks, so I could try to use that to match them.

Another option is just leaving the existing ones on label-less canvases and making it a manual data cleanup task to migrate them to the appropriate fragment. In which case, yes, things will look weird until then! (But only in the transcription editor; they will look the same in the public view.)

editor for such cases: 

Screen Shot 2023-02-06 at 1 35 40 PM

public:

Screen Shot 2023-02-06 at 1 35 55 PM

rlskoeser:
@blms if things only look weird in the transcription editor but look fine on the public site then I think you should not spend time attempting to automate anything (which I think would be pretty hard to get right).

Do we not display labels for placeholder canvases? (I've forgotten!)


Will the new placeholder canvases make it easier to add new iiif images and move transcriptions to the right images?

blms:
> Do we not display labels for placeholder canvases? (I've forgotten!)

We do not! In fact, we were also hiding the entire "image header"—which made sense to hide the controls, but it also hid the label section.

> Will the new placeholder canvases make it easier to add new iiif images and move transcriptions to the right images?

Yes, that's the idea here, especially with the new canvas matching with textblocks, so it will appear under an accurate label even after the new images are added.

Issue #1308, opened 2023-02-02
Updated transcriptions possibly failing to populate in search index


blms:
## Testing notes

- [x] On the [QA site](https://test-geniza.cdh.princeton.edu/), create a new transcription with some easily searchable text (like a nonsense word). Perform a keyword search for that text, and the document should show up in the results.
- [x] Edit that transcription with some different searchable text and save your edits. You should now be able to perform a search for the updated text and it should show up in the search results.
- [x] Delete the transcription block(s) you created. You should no longer be able to find the document using either of the previous keyword searches.

---

Issue encountered by @mrustow, who updated a transcription and days later the search index still does not reflect the changes. See the following document/search, where the portion above the first numbered line was removed:

- [PGPID 9131](https://geniza.princeton.edu/en/documents/9131/)
- [Search results](https://geniza.princeton.edu/en/documents/?q=%D7%97%D7%A6%D7%A8%D7%A0%D7%90+%D7%A0%D7%97%D7%9F+%D7%90%D7%9C%D7%A9%D7%94%D7%95%D7%93+%D7%90%D7%9C%D7%9B%D7%90%D7%AA%D7%9E%D7%99%D7%9F+%D7%A9%D7%94%D7%90%D7%93%D7%AA%D7%A0%D7%90+%D7%90%D7%A1%D7%A4%D7%9C&docdate_0=&docdate_1=&sort=relevance)
- [Transcription revision history](https://github.com/princetongenizalab/pgp-text/commits/main/09000/9131)

![image](https://user-images.githubusercontent.com/7234006/216383717-7b10aafd-38ca-440b-aa59-5825895b462a.png)
![image](https://user-images.githubusercontent.com/7234006/216383735-4e0c2b1e-8bd8-463d-acb3-d81107115874.png)

kseniaryzhova:
@blms works for me! Checked both Arabic and JA!

PR #1309, opened 2023-02-02
Reindex annotated documents after annotation save, on update and create (#1308)

blms:
## In this PR

Per #1308:
- On annotation update, wait until after annotation's `save()` method is called to reindex associated document, to ensure latest revision is in index
- On annotation creation, reindex associated document

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -423,6 +423,7 @@ class DocumentSignalHandlers:
         "footnote": "footnotes",
         "source": "footnotes__source",
         "creator": "footnotes__source__authorship__creator",
+        "annotation": "footnotes__annotation",
     }
 
     @staticmethod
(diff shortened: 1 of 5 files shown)


rlskoeser (changes requested):
Please implement using signal handlers like we do when other objects are updated; see the `index_depends_on` dict on `Document` and `DocumentSignalHandlers` in `geniza.corpus.models`

Now that annotation has foreign key access to documents through footnotes, it should work similarly to the other models — hopefully no code logic changes needed, just some dict entries and a queryset filter config. 

If the comments on that code isn't enough for you to figure it out efficiently, LMK

blms:
Good catch, thanks!

geniza/corpus/models.py
```diff
@@ -1103,6 +1104,10 @@ def index_data(self):
             "post_save": DocumentSignalHandlers.related_save,
             "pre_delete": DocumentSignalHandlers.related_delete,
         },
+        "annotations.annotation": {
+            "post_save": DocumentSignalHandlers.related_save,
+            "pre_delete": DocumentSignalHandlers.related_delete,
+        },
```
rlskoeser, line 1110: awesome that this is all it takes, glad the existing signal handler is working so well

rlskoeser (approved):

Issue #1261, opened 2022-12-07
As content editor using the transcription editor, I want the image to be sticky so that I can always have the image beside the text as I scroll down.


kseniaryzhova:
## Testing notes

On the [QA site](https://test-geniza.cdh.princeton.edu/), find a document with a transcription text that is much longer than the image.

- [x] Open the transcription editor for that transcription. Do a hard refresh, just in case. There should now be a "pop out" button beneath each image. 
- [x] If you click the "pop out" button, you should get an image that sticks on the left half of your screen. (Yes, it's not beautiful, but it's just for editors!)
- [x] You should still be able to zoom, rotate, and draw annotation boxes on the image.
- [x] You should still be able to edit the transcription text content on the right side of the image.
- [x] There should be an "X" button in the top right of the image that lets you close the "pop out" viewer, and it should bring you back to the normal view.
- [x] Confirm this is an acceptable solution for editing images with long transcription content

---

**Is your feature request related to a problem? Please describe.**
When the transcription text is much longer than the image, when you scroll down past a certain point the image disappears and it's hard to compare text to image (basically you have to scroll up and down, forget your place in the transcription, etc).

**Describe the solution you'd like**
I want the images to be sticky so they move with me as I scroll down a text.

kseniaryzhova:
@blms works as you've described; I'm going to run it by Marina and Rachel to make sure this is acceptable but I think this works even better than the sticky image.

kseniaryzhova:
@blms went over it together with Marina and Rachel - works for us, opening separate issue for improvements. Thanks!

PR #1310, opened 2023-02-02
Add "pop out" button to images in editor (#1261)

blms:
## In this PR

Per #1261:
- Allows users to "stick" an image to the left side of the screen in order to continue transcribing when the transcription extends past the visible portion of the image

## Notes

I spent a lot of time trying to get `position: sticky` to work, or mimic that behavior with JS, but I wasn't able to get it to work with the current HTML/CSS structure. In order to avoid making significant changes, I came up with a compromise: a "pop out" button that only appears in the editor, which takes up half of your screen with the image, in a fixed position.

It's not the best looking thing in the world, but it's just for the editor, and seems to work OK.

Would appreciate a checkout and test locally if you have the time!

(Also, the html changes aren't that significant; recommend hiding whitespace changes in the diff)

### DIFF
--- a/geniza/corpus/templates/corpus/snippets/document_transcription.html
+++ b/geniza/corpus/templates/corpus/snippets/document_transcription.html
@@ -67,25 +67,31 @@
             {% for canvas_uri, image_info in images.items %}
                 {# only images that are part of the current document are zoomable #}
                 <div class="img{% if image_info.excluded %} excluded{% endif %}{% if image_info.placeholder %} placeholder{% endif %}" id="{{ image_info.label }}"{% if not image_info.excluded %} data-controller="iiif {% if edit_mode %}annotation" data-iiif-edit-mode-value="true{% endif %}" data-canvas-url="{{ canvas_uri }}" data-iiif-target="imageContainer" data-annotation-target="imageContainer"{% endif %}>
-                    <div class="img-header" data-iiif-target="imageHeader">
-                        <h2>{{ image_info.shelfmark }} {{ image_info.label }}</h2>
-                        {% if not image_info.excluded %}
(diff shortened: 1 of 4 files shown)


rlskoeser (commented):
Do you know why the sticky positioning doesn't work with the viewer layout? I did some brief looking, and I see that grid makes things more complicated, but I see examples of simple cases where it works like we wanted. (I looked at https://melanie-richards.com/blog/css-grid-sticky/ and https://ishadeed.com/article/position-sticky-css-grid/ )

The pop-out does seem kind of hacky, but might be good enough for transcription editing. (It doesn't address readers looking at transcriptions, though.)  I wonder if the team should start thinking about a full-screen view...

A few notes: 
- the pop-out container interacted a little strangely with the header — the header goes over it, and can obscure the x to close it. the footer goes over it, but that seems ok. (might only happen with longer images)
- suggest you disable the pop-out for mobile (maybe just hide the button)

blms:
> https://melanie-richards.com/blog/css-grid-sticky/ and https://ishadeed.com/article/position-sticky-css-grid/ 

I wasn't able to get this method working—neither `start` or `baseline`—and I'm not exactly sure why, but I suspect it has something to do with our special positioning to get the image height down to the viewport size. I just played with it for another hour and wasn't able to make any progress (messed with all the properties on the `panel-container` including `align-items`, `overflow`, `justify-items`, messed with all of the grid items `position` and `align-self`).

FWIW, both of these methods also fail when there are more than one row in the grid, so that would need to be addressed if we went that route anyway.

Screen Shot 2023-02-06 at 2 30 37 PM

rlskoeser (approved):

Issue #1255, opened 2022-11-29
Searching descriptions for words/phrases in quotations works unpredictably


kseniaryzhova:
## testing notes

In the [QA site search](https://test-geniza.cdh.princeton.edu/en/documents/):
- [ ] Try the two searches mentioned in the bug report below, as well as some other exact phrases, and confirm results are as expected
  - Note that while only exact matches will appear, the actual text "Abū l-Munā" won't be highlighted in results. This is due to the "l-". Let me know if it's worth investing the additional ~3-5 hours to address this.
- [ ] Try some searches with both quoted and not quoted phrases, such as `"he decided" buy` and `"he decided" AND buy`

---

**Describe the bug**
Searching for phrases such as "he divorced" in quotations yields 12 results, none are precise matches but are very close. Searching for full names such as "Abu l-Muna" or even "Abū l-Munā" (both attested in the descriptions) yields 0 results

**To reproduce**
1. Go to [public site search](https://geniza.princeton.edu/en/documents/)
2. Search for "Abū l-Munā" (with quotations)
3. [See results ](https://geniza.princeton.edu/en/documents/?q=%22Ab%C5%AB+l-Mun%C4%81%22&docdate_0=&docdate_1=&sort=relevance)
4. Repeat with a non-name phrase such as "he divorced". [See results.](https://geniza.princeton.edu/en/documents/?q=%22he+divorced%22&docdate_0=&docdate_1=&sort=relevance)

**Expected behavior**
I want search terms in quotations to give precise results (while still ignoring diacritics) no matter what Latin script word or phrase I'm searching for. Searches for "Abu l-Muna" or "Abū l-Munā" should pull up at least a dozen (if not more) exact matches.

kseniaryzhova:
@blms Searching "Abū l-Munā" brings back 117 precise results with diacritics. Searching "Abu l-Muna" brings back 117 of the same exact results.

blms:
Marking as closed based on meeting discussion 3/1/2023. We agreed that:
- The comments above (searching without diacritics bringing up diacritic results, and vice versa) is the desired behavior
- Further work will be tracked in #1335, and further issues as they arise

PR #1319, opened 2023-02-16
Allow exact search with doublequotes (#1255)

blms:
## In this PR

Per #1255:
- Create a new field type called `text_nostem` that does whitespace tokenization, unicode folding, hypen preservation, and end-punctuation removal
- Create a copyfield from `description_en_bigram` of that type called `description_nostem`
- When doublequoted phrases are present in a keyword query, excluding those scoped to a certain field (such as `shelfmark:"Bodl ms heb"`), scope that part of the query to the `description_nostem` field (while including it again, bare)
- Boost the `description_nostem` field so that such queries show exact hits first (but shelfmark still takes precedence over description)

## Notes

- Playing around with this, and with boosts, I managed to get two different versions working:
   1. Only scope the quoted query to the `description_nostem` field, which excludes all non-exact matches. No highlighting whatsoever.
   2. Include the query twice, once bare and once scoped to the `description_nostem` field. With boosting, the exact results show up first, with highlighting, but non-exact matches sometimes still show up beneath it.
   - My sense is that the second one is preferable, so I went with that.
- One quirk of this is that the `description_nostem` field preserving the hyphenated tokens seems to prevent the highlighting of exact matches for  "Abū l-Munā", but that seems OK considering it's an edge case and it does still pull up all the relevant entries first.
- Can you also test this locally and try to stress test it and break it? I tried a few different things and adjusted, but I'm sure I'm missing things.


geniza/corpus/solr_queryset.py
```diff
@@ -91,6 +91,16 @@ def _search_term_cleanup(self, search_term):
         # to avoid it being interpreted as a boolean
         search_term = self.re_shelfmark_nonbool.sub("BL or", search_term)
 
+        # regex to match terms in doublequotes, but not following a colon, at the
+        # beginning of the string or after a space. (non-greedy to prevent matching
+        # the entire string if there are multiple sets of doublequotes)
+        search_term = re.sub(
+            r'\B(?<!:)(".+?")',
```
rlskoeser, line 98: I think you should also exclude fuzzy and proximity searches, which could be used with quotes. I think adding a word boundary after the end quote should be sufficient, but would be good to add test cases.

solr_conf/conf/managed-schema
```diff
@@ -258,6 +258,18 @@
...
+    <!-- Simple text field with no stemming, used to find exact matches. -->
+    <fieldType name="text_nostem" class="solr.TextField">
+      <analyzer>
+        <tokenizer class="solr.WhitespaceTokenizerFactory"/>
+        <filter class="solr.LowerCaseFilterFactory"/>
+        <filter class="solr.ICUFoldingFilterFactory"/>
+        <!-- ensure hyphenated words are joined together, even if they were separated by whitespace -->
+        <filter class="solr.HyphenatedWordsFilterFactory"/>
+        <!-- remove all punctuation from the end of tokens -->
+        <filter class="solr.PatternReplaceFilterFactory" pattern="[,\.!?:;]$" replacement=""/>
```
rlskoeser, line 270: Is there a punctuation pattern match (similar to `\d` for numeric) we can use here? There are probably other unicode punctuation that you might need to include.

solr_conf/conf/managed-schema
```diff
@@ -287,6 +299,9 @@
   <copyField source="tags_ss_lower" dest="tags_t" maxChars="30000" />
   <copyField source="type_s" dest="type_t" maxChars="300" />
 
+  <!-- copy description to a nostem field for exact matches -->
+  <copyField source="description_en_bigram" dest="description_nostem" maxChars="30000" />
```
rlskoeser, line 303: what is the default value for`stored` ? maybe good to explicitly set stored to false

rlskoeser (commented):
Some initial comments. I don't have time to pull and try locally right now, but can try to find time for that next week.

I think the same non-stemming search is needed for transcription as well as description. Since we're not using the new field for highlighting, I suggest you revise this to create one generic non-stemmed text copy field that both of them get copied into. (I don't think the boosting matters as much to differentiate between exact matches with these two.)

The highlighting doesn't work in your first case scenario because you don't have any search terms hitting the fields being used for highlighting. You should be able to resolve this by specifying a highlight query (`hl.q`,  `q=` in parasolr) with the version of the query that is not scoped to the non-stem content field. You should probably ask the team which they prefer (my guess is exact matches only based on past conversations).

I couldn't get phrases to highlight properly either, they skip over the small words or word-portions like `l-`; I'm not sure if there's a good solution, but seems more important to get this functionality out without that.

solr_conf/conf/managed-schema
blms: Looks like we might be able to do `\p{Punct}` or `\p{IsPunctuation}`. There's [some interesting complexity](https://stackoverflow.com/a/49289766) with those but the two in combination should cover everything we need. I had avoided doing something like this due to the inclusion of the hyphen, but now I realize that should actually be completely resolved by the previous filter and hyphens that are truly at the end of the token _should_ be discarded.

solr_conf/conf/managed-schema
blms: Does this need to be set separately for the `copyField` call and the `dynamicField` that it matches?

rlskoeser: I don't think so, I think it's defined on the dynamic field type

blms: I thought so as well! In that case, already done 😄

geniza/corpus/solr_queryset.py
blms: I wasn't able to get it working with just a word boundary, but either way they both use `~`, so I'll just match that explicitly.

blms:
> You should be able to resolve this by specifying a highlight query (hl.q, q= in parasolr) with the version of the query that is not scoped to the non-stem content field.

@rlskoeser I'm struggling with this—how/where do I specify a query only for highlighting and not as a search query? If I include both `keyword_query=` (scoped to field) and `q=` (scoped to highlight fields) as arguments to `raw_query_parameters`, it still seems to treat the latter as a search query as well. Is it supposed to go somewhere else?

rlskoeser:
> how/where do I specify a query only for highlighting and not as a search query?

Sorry I wasn't clear. Specify as an arg when you configure the highlighting, parasolr _should_ turn the `highlight(..., q=)` into `hl.q=`

Here-ish: https://github.com/Princeton-CDH/geniza/blob/main/geniza/corpus/views.py#L168-L180

blms:
@rlskoeser Thanks for the pointer! Still can't get it working yet though 😞 

This is what the solr debug logging looks like when I pass the quoted query as the `q=` argument to `highlight()`.

I also tried setting the `q=` argument to `description_en_bigram:"he divorced"` and `description:"he divorced"`, but still no dice.

```py
{
    "start": 0,
    "fq": ["item_type_s:document", "status_s:Public"],
    "fl": "id:id,type:type_s,status:status_s,shelfmark:shelfmark_s,document_date:document_date_t,original_date_t:original_date,collection:collection_ss,tags:tags_ss_lower,description:description_en_bigram,notes:notes_t,needs_review:needs_review_t,pgpid:pgpid_i,old_pgpids:old_pgpids_is,input_year:input_year_i,input_date:input_date_dt,num_editions:num_editions_i,num_translations:num_translations_i,num_discussions:num_discussions_i,scholarship_count:scholarship_count_i,scholarship:scholarship_t,transcription:text_transcription,language_code:language_code_s,language_script:language_script_s,iiif_images:iiif_images_ss,iiif_labels:iiif_labels_ss,has_image:has_image_b,has_digital_edition:has_digital_edition_b,has_translation:has_translation_b,has_discussion:has_discussion_b,score:score",
    "q": "{!type=edismax qf=$keyword_qf pf=$keyword_pf v=$keyword_query}",
    "sort": "score desc",
    "rows": 2,
    "hl": True,
    "hl.fl": "description_en_bigram,text_transcription",
    "f.description_en_bigram.hl.snippets": 3,
    "f.description_en_bigram.hl.method": "unified",
    "f.description_en_bigram.hl.requireFieldMatch": True,
    "f.description_en_bigram.hl.q": "he divorced",
    "f.text_transcription.hl.method": "unified",
    "f.text_transcription.hl.fragsize": 150,
    "f.text_transcription.hl.requireFieldMatch": True,
    "f.text_transcription.hl.q": "he divorced",
    "facet": True,
    "facet.field": [
        "has_image_b",
        "has_digital_edition_b",
        "has_translation_b",
        "has_discussion_b",
        "{!ex=type}type_s",
    ],
    "f.type_s.facet.sort": "value",
    "keyword_query": 'content_nostem:"he divorced"',
}
```

rlskoeser:
The per-field option ` "f.text_transcription.hl.q": "he divorced"` should be equivalent — see https://solr.apache.org/guide/solr/latest/query-guide/highlighting.html#common-highlighter-parameters

parasolr sets it that way since you're highlighting a specific field and the highlight query could vary.

It could be that you need to specify the query parser also... ? 

If the highlight query is set (as it seems to be) but you aren't getting highlighted terms in your results, then I would try running an equivalent query in the solr web ui — add enough of the query options listed in the debug output here to get a basic search working and see if you can tinker with the highlight query options to see if you can get the highlight query to work.

rlskoeser:
@blms if you're not able to resolve then perhaps I can take a look next week when I pull and review your code locally

blms:
@rlskoeser Thanks very much for the help. I may indeed have to ask you to take a stab at it; here's what I've been able to get so far.

The only param I have been able to get highlighted results with is this:
```
"hl.q":"{!type=edismax qf=$keyword_qf pf=$keyword_pf v='he divorced'}",
```

I can't get that with parasolr, except by adding `raw_query_parameters`. An identical value for the param `f.description_en_bigram.hl.q` doesn't seem to work:
```
"f.description_en_bigram.hl.q":"{!type=edismax qf=$keyword_qf pf=$keyword_pf v='he divorced'}",
```

Nor do any of the following:
```
"f.description_en_bigram.hl.q":"{!type=edismax qf=$keyword_qf pf=$keyword_pf}he divorced",
"f.description_en_bigram.hl.q":"{!type=edismax v='he divorced'}",
"f.description_en_bigram.hl.q":"{!type=edismax}he divorced",
"f.description_en_bigram.hl.q":"he divorced",
```

Same lack of results regardless of whether or not I add any of these:
```
"hl.qparser":"lucene",
"f.description_en_bigram.hl.qparser":"lucene",
"f.description_en_bigram.hl.qparser":"edismax",
"f.description_en_bigram.hl.qparser":"{!type=edismax}",
"f.description_en_bigram.hl.qparser":"{!type=edismax qf=$keyword_qf pf=$keyword_pf}",
```
etc.

Sorry I wasn't able to make any further progress without help. I think this is a bit beyond my solr capabilities!

rlskoeser:
@blms interesting. Any reason not to just use the working `hl.q` via raw query parameters? You do want the same query for both highlight fields, after all.   You could just make a note in the code why you're doing it that way.

I could think about having parasolr support highlight options that are not field-specific in the future, but since you have a working option it seems ok to me to just go with that.

I wonder if this version would work: `{!type=edismax}"he divorced"` ... but I'm not sure it really matters since you have a working option.

blms:
@rlskoeser Had to make a slight adjustment; turns out I'd neglected to add `multiValued="true"` to the dynamic field. Oddly, the only difference is that I now have to specify `hl.qparser`, but still none of the other types of highlight queries—i.e., the ones with `f.description_en_bigram.hl.q`—worked. (Curiously, I also was able to just do `{!type=edismax}"he divorced"` until I made this change, but now the bits with `qf` and `pf` are required.)

In any case, I think we have a working version now. Thanks so much for your help!

geniza/corpus/solr_queryset.py
```diff
@@ -91,13 +97,9 @@ def _search_term_cleanup(self, search_term):
...
 
-        # regex to match terms in doublequotes, but not following a colon, at the
-        # beginning of the string or after a space. (non-greedy to prevent matching
-        # the entire string if there are multiple sets of doublequotes)
-        search_term = re.sub(
-            r'\B(?<!:)(".+?")',
-            # include the term twice: once bare, and once scoped to description_nostem
-            lambda m: f"{m.group(0)} description_nostem:{m.group(0)}",
+        search_term = self.re_exact_match.sub(
+            # include the term twice: once for highlighting, and once scoped to content_nostem
```
rlskoeser, line 101: this comment is out of date now, right?

rlskoeser (commented):
Ok, I've pulled the changes, updated solr config, and reindexed so I could test locally.

comments based on my testing:
- if you search for an exact phrase and a keyword, the keyword doesn't get highlighted (e.g. `"he decided" AND buy` or `"he decided" buy` [implicit OR]); I think my suggestions on the code will address this
- I had some trouble with exact phrase searching on transcription text, but hopefully this was just because I was testing while I was indexing; I tested again with a different record and it seems ok
- searching on a single term with quotes gets unstemmed matches _only_ (`"decided"`), which is great

There's no way to do partial matches — which may be out of scope for this feature, but I think will limits the use of this for searching unstemmed transcription content — you can do exact, unstemmed matches, but only on full words. 

I'm wondering if there's something we could use besides quotes to signal a search should be against the unstemmed content, so that it could contain wildcards, but I'm reluctant to come up with custom search syntax that no one will know about! (Maybe there could be two variants that both work...  trying to think of a common character not already in use, maybe could do something like `<he decid*>` or perhaps better `exact:(he decid*)`).

This is probably something for you to discuss with the team and decide what (if anything) else is still needed once they start using this feature.

geniza/corpus/views.py
```diff
@@ -185,6 +185,20 @@ def get_queryset(self):
                     )
                     .also("score")
                 )  # include relevance score in results
+                # if an exact search was performed using double quotes, add to highlighting
+                exact_queries = DocumentSolrQuerySet.re_exact_match.findall(
```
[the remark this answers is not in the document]
blms: Made an attempt at this in eba3be9b68adb6fe9c86785e3b4825d094c85436 but I don't think it's quite right. Thank you for helping out with this!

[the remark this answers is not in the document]
rlskoeser: @blms I have some small suggestions, but what is it that's not working with this?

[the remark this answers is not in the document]
blms: @rlskoeser How odd! Yesterday when I was trying this, searching for `"he decided" buy` or `"he decided" AND buy` was leading to no results at all. Now it seems to work exactly as expected. 🤯  Maybe I'd done something wrong yesterday?

Issue #1320, opened 2023-02-21
Styles missing for JTS logo


blms:
## Testing notes

- [x] On the test site, on documents containing JTS fragments, under "Image Permissions Statement", the JTS logo should be a reasonable size, and its color should change between light and dark mode

---

I know these existed at some point, probably got clobbered somewhere

![image](https://user-images.githubusercontent.com/7234006/220377443-b81dbe37-34ef-4c46-b586-dd40b6ac57c0.png)

blms:
Confirmed fixed in QA 👍

PR #1321, opened 2023-02-21
Fix JTS logo issue (#1320)

blms:
## In this PR

Looks like there was still some leftover code from the old IIIF viewer, pre-"ITT panel" work. `_iiif.scss` was referencing a nonexistent id `#iiif-viewer`, so none of its styles were being applied. Instead, its only practical use was its `@use` import of `_transcription.scss` (with which it shared a ton of duplicated, but unused, CSS!).

Per #1320:
- Remove outdated/unused CSS, move the JTS CSS block to `_transcription.scss`, and update Webpack to use that directly instead of indirectly via `_iiif.scss`.


rlskoeser (approved):
👍

PR #1326, opened 2023-02-23
Adjust arabic phrase regex to ignore whitespace between two quoted terms

rlskoeser:


blms (approved):
Ah, excellent! So I didn't have to work on it today after all 😄 

Thanks, looks great!

PR #1329, opened 2023-02-24
Improve speed for indexing all content in Solr

rlskoeser:
This PR is an attempt to improve speed when indexing all content in Solr using `manage.py index`

I discovered recently (for PPA) that prefetching is incompatible with iterating over a queryset in chunks (which the parasolr indexing script does). Parasolr 0.9 adds a method that can be extended to apply prefetching to a chunk of iterated records.  

In this update:
- require newly released parasolr 0.9
- add prefetching logic for chunks of records when indexing
- improve methods used to generate `index_data` for `Document` records to use prefetched data and avoid additional db calls

There are still some db calls being made for each record, and I haven't been able to figure out why — I've narrowed it down to the `fn.display` call that generates the footnote citation, but I'm not sure what in the `Source` or `Footnote` method is hitting the database. To see the calls that are being made, adjust your django logging to `DEBUG` when you call the index; to make it easier to inspect, you can add a `sleep(1)` call to the `index_data` method for debugging only.

Even without this, it's still an improvement. I haven't done a full index run, but develop was estimating completion around 38 minutes and my branch is closer to 7. I think if we could figure out the remaining db call it could finish in 1-2 minutes.

This branch is still WIP as you'll see; some cleanup and testing updates would be needed before we could proceed, but review & feedback at this stage would be helpful.


geniza/corpus/models.py
```diff
@@ -1080,18 +1141,26 @@ def index_data(self):
                 "num_translations_i": counts[Footnote.TRANSLATION],
                 "num_discussions_i": counts[Footnote.DISCUSSION],
                 # count each unique source as one scholarship record
-                "scholarship_count_i": self.sources().count(),
+                "scholarship_count_i": len(source_relations.keys()),
                 # preliminary scholarship record indexing
                 # (may need splitting out and weighting based on type of scholarship)
-                "scholarship_t": [fn.display() for fn in self.footnotes.all()],
+                "scholarship_t": [fn.display() for fn in footnotes.all()],
```
blms, line 1147:
```suggestion
                "scholarship_t": [fn.display() for fn in footnotes],
```

blms (approved):
Looking great so far, super speedy! I think I found the problem also; a redundant `.all()` on the footnotes queryset. With that change it's now taking only 2 minutes to complete the entire index. 🤯

geniza/corpus/models.py
rlskoeser: Wow, thank you for finding that! I thought that using `.all()` would just return the prefetched content, but there must be cases where it doesn't

blms:
From a glance, looks like it may be related to the `scholarship_count_i` change. I'll take a closer look!

rlskoeser:
@blms oh wow, good catch — it's no longer a queryset so `.count()` doesn't work — good thing we have tests for this!

blms (approved):
Looks great! Tested locally as well. This is going to be a huge time saver. 👍

Issue #1331, opened 2023-02-27
On tag change, document indexing is one revision behind


blms:
### Testing notes (QA)

In the [QA site admin](https://test-geniza.cdh.princeton.edu/admin):

- [x] Add a tag to a document—can be a new or existing tag (try both!)—and save the document. Then, on the public site search, search for the document by pgpid (i.e. `pgpid:1234`)
- [x] In the results, the tags list should reflect the new tag(s) added
- [x] Try the same with removing a tag from a document, and modifying and saving a tag's name

---

### dev notes
Similar to https://github.com/Princeton-CDH/geniza/issues/1308, document index lags one revision behind after modifying tags.

To reproduce:
1. Add a tag to a document
2. Search for it on the search page by pgpid; tags will not have changed
3. Add another tag to the document
4. On search, tags will now reflect the revision made in step 1, but not the revision made in step 3

kseniaryzhova:
@blms works as it should, thank you! Closing!

PR #1332, opened 2023-02-27
Fix issue with outdated tag set in document index (#1331)

blms:
## In this PR

Per #1331:
- Index document, refreshed from DB, after change in m2m tag relationships

## Questions

- Is it ok that this means documents cannot have tags programmatically added or removed during a transaction without discarding the prior changes in that transaction? How can we mitigate?

---

## previous notes

After investigating #1331, it seems that when `index_data` is called after a document is saved, the updated tag set is not yet present in `self.tags.all()`.

My attempt at a workaround is to check specifically for the related change via a change in `TaggedItem`. This does trigger a reindex with the correct tag set when you change tags on a document. However, it seems that `index_data` is called again for the document save _after_ it's called for the `TaggedItem` save, which means that it overwrites the index data to be the same as before (outdated tag set).

Is there some way to ensure the `TaggedItem` related save indexing trigger always happens after the document save indexing trigger? Or is the approach wrong?

(Not exactly the same issue, but also wondering if it's related to this: https://github.com/jazzband/django-taggit/issues/818)


rlskoeser (commented):
Indexing on taggit relationship post save does seem logical... I'm puzzled about the order you're reporting; could it be that the tags are cached somehow when we save the document? 

I just checked, parasolr is listening to the `post_save` signal on the object itself. I don't know what order that happens in, or why there would be a discrepancy between them.  

Possibly this stack overflow is related? Seems to be discussing _updating_ m2m on post save, but the clear behavior may also be relevant for what we're trying todo. https://stackoverflow.com/questions/1925383/issue-with-manytomany-relationships-not-updating-immediately-after-save

blms:
Yeah, I'm at a bit of a loss here! It does indeed seem to be referring to an old/cached copy of the relationship on `post_save` of the Document. Thanks for the link; I'll look into that, and I also see reference there to a signal `m2m_changed` which I wasn't aware of.

blms:
@rlskoeser I have a working version of this now, but with one pitfall: because of the `refresh_from_db` call during the `m2m_changed` signal handler, programmatically changing tags (using `tags.add` or `tags.remove` in code) is potentially dangerous in terms of data loss. Those calls in code will trigger the `m2m_changed` signal, and thus `refresh_from_db`, wiping any unsaved changes made to the document before the tags were added/removed.

You can see an example of this in the [document merge code changes](https://github.com/Princeton-CDH/geniza/pull/1332/files#diff-49fb6b6e83aa40f8957e7d4240bfa28684a3f510cbf920e2a47b670e4985c35dR1268-R1304).

Is this acceptable? I couldn't figure out any other way for the tag set to be up-to-date when the indexing code is called.

geniza/corpus/models.py
```diff
@@ -1219,10 +1229,6 @@ def index_data(self):
             "post_save": DocumentSignalHandlers.related_save,
             "pre_delete": DocumentSignalHandlers.related_delete,
         },
-        "tags": {
-            "post_save": DocumentSignalHandlers.related_save,
-            "pre_delete": DocumentSignalHandlers.related_delete,
-        },
```
rlskoeser, line 1225: We still need these handlers, though, right? If someone renames a tag without changing any m2m relationships, we should reindex all the documents with that renamed tag so they reflect the update

geniza/corpus/models.py
```diff
@@ -485,6 +486,15 @@ def unidecode_tag(sender, instance, **kwargs):
         """Convert saved tags to ascii, stripping diacritics."""
         instance.name = unidecode(instance.name)
 
+    @staticmethod
+    def tagged_item_change(sender, instance, action, **kwargs):
```
rlskoeser, line 490: the instance here is the document with the m2m relation?

geniza/corpus/models.py
```diff
@@ -1287,10 +1290,15 @@ def merge_with(self, merge_docs, rationale, user=None):
         needs_review = [self.needs_review] if self.needs_review else []
 
         for doc in merge_docs:
-            # add merge id to old pgpid list
-            self.old_pgpids.append(doc.id)
             # add any tags from merge document tags to primary doc
+            # NOTE: This must be done before any other changes to self because it will fire
+            # m2m_changed signal which calls self.refresh_from_db()
```
rlskoeser, line 1295: is this refresh from db likely to be problematic elsewhere? it seems like it could lead to unexpected behavior that might be hard to troubleshoot

in the signal handler, instead of refreshing, could we get a new instance from the db without modifying the local copy?

rlskoeser (commented):
Thanks for catching this, m2m relationships and signals are tricky to think about and handle properly in terms of indexing. (Wondering if there's anything here we could adapt for common parasolr indexing behavior or at least document in the examples...)

one question for you about the `refresh_from_db` call, otherwise I think this is good

geniza/corpus/models.py
blms: Great idea, much better than refresh from db, thanks!

rlskoeser (approved):

Issue #1333, opened 2023-02-27
Tags may be saved with identical names, case-insensitive


blms:
## testing notes

In the [QA admin site](https://test-geniza.cdh.princeton.edu/admin/taggit/tag/), find an existing tag.

- [x] Using the "Add tag" button, try to add a new tag with the same name. You should receive an error on save.
- [x] Try the same name, but use different capitalization from the original tag. You should still receive an error on save.
- [x] Create a tag with a completely different name and save it. It should save successfully. Now, on the edit page for your new tag, rename it to be the same name as a different existing tag, but with different capitalization. You should again receive an error on save.

## dev notes

It is currently possible to save two tags with the same name, with the only difference being that different letters are capitalized. Since Taggit uses the query `tag = manager.get(name__iexact=name)` when `TAGGIT_CASE_INSENSITIVE=True`, saving a document with one of these tags results in a `MultipleObjectsReturned` error. Simplest solution is to use case-insensitive uniqueness validation.

example: https://geniza.princeton.edu/admin/taggit/tag/?q=judaeo-arabic+poetry

kseniaryzhova:
@blms works as expected! Closing, thank you!

PR #1334, opened 2023-02-27
Add case-insensitive unique validation to tag name field (#1333)

blms:
## In this PR

Per #1333:
- Validate tag names are unique, case-insensitive, during create or update of a tag within the tags admin

### DIFF
--- a/geniza/common/admin.py
+++ b/geniza/common/admin.py
@@ -71,8 +71,15 @@ admin.site.unregister(Tag)
 
 
 class TagForm(ModelForm):
-    """Extends the default tag admin form to validate uniqueness, case-insensitive,
-    on tag names."""
+    """
+    Extends the default tag admin form to validate uniqueness, case-insensitive,
+    on tag names.
+
+    NOTE: This is needed because the Tag model does not have a DB-level case-insensitive uniqueness
+    constraint applied out of the box, and adding such a constraint is not trivial at the moment.
+    TODO: Once Django is updated past 4.0, a "functional unique constraint" may be added to a custom
+    Tag model to solve this problem; then, this form override will no longer be needed.
+    """
 
     class Meta:
         model = Tag
--- a/geniza/common/admin.py
+++ b/geniza/common/admin.py
@@ -4,6 +4,7 @@ from django.contrib.admin.models import LogEntry
 from django.contrib.auth.admin import UserAdmin
 from django.contrib.auth.models import User
 from django.db.models import Count
+from django.forms import ModelForm, ValidationError
 from django.http import HttpResponseRedirect
 from django.urls import path, reverse
 from taggit.admin import TagAdmin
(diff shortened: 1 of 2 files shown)


rlskoeser (commented):
Could have sworn we did something related to tag case sensitivity before (although when I looked quickly I couldn't find any issues on github). Did we lose the `TAGGIT_CASE_INSENSITIVE` configuration, or is it just not sufficient?  It seems like there should be something built-in to taggit — surprised you need to override the tag form. (And this is only for editing tags directly, right? does it help at all when you add/edit tags via document edit form?)

I checked our config, we do have `TAGGIT_CASE_INSENSITIVE = True` and we also have a custom tag string method configured - `TAGGIT_TAGS_FROM_STRING = "geniza.common.utils.custom_tag_string"` — I remember now that was to solve a different problem, so maybe it isn't relevant here.

blms:
@rlskoeser That's right—`TAGGIT_CASE_INSENSITIVE` is involved when adding tags to a document, and that's what causes the case insensitive lookup (`tag = manager.get(name__iexact=name)`). I think the problem is that this doesn't create any kind of DB-level uniqueness validation. It seems that according to Taggit, you are only ever meant to be creating tags using the tags field on a taggable model, in which it makes sense that all that the setting would affect is the lookup. But since we created an admin to create and edit tags, we need to make sure that case-insensitivity is enforced there as well.

rlskoeser:
> @rlskoeser That's right—`TAGGIT_CASE_INSENSITIVE` is involved when adding tags to a document, and that's what causes the case insensitive lookup (`tag = manager.get(name__iexact=name)`). I think the problem is that this doesn't create any kind of DB-level uniqueness validation. It seems that according to Taggit, you are only ever meant to be creating tags using the tags field on a taggable model, in which it makes sense that all that the setting would affect is the lookup. But since we created an admin to create and edit tags, we need to make sure that case-insensitivity is enforced there as well.

Ah, thanks for clarifying / reminding me.  

Would it be simpler just to enforce case-insensitive uniqueness at the db level?

Could extend the tag model, and override the defaults so that the database table name and fields are unchanged, the only migration would be to add the db constraint. But I'm not sure how much work to update our tag managers to use our local tag class... 
https://django-taggit.readthedocs.io/en/latest/custom_tagging.html#custom-tag 

If you decide the form is simpler solution, then please expand the comment to document why this is necessary / how it differs from the other taggit case-insensitive code.

blms:
> Would it be simpler just to enforce case-insensitive uniqueness at the db level?

It seems we have a two options for doing this:

- `CICharField`, which requires adding the `citext` postgres extension (and I've seen reports of this not working on StackOverflow)
- Upgrade to Django 4.0+ to use [functional unique constraints](https://docs.djangoproject.com/en/4.0/releases/4.0/#functional-unique-constraints)

My sense is that it's simpler for now to do the admin validation, and leave it as a chore to update to the more proper DB-level constraint once we upgrade to Django 4+. Will add the requested comments!

rlskoeser:
@blms thanks for investigating! Good to know there's a nice solution once we get to django 4.0. Documenting and going with this for now seems good.

rlskoeser (approved):

Issue #1335, opened 2023-02-28
Some partial search results in description not boosted by relevancy


kseniaryzhova:
### testing notes (QA - round two)
In the [QA site document search](https://test-geniza.cdh.princeton.edu/en/documents/):
- [ ] Searching `אלמרכב AND אלצ` should return the results, in correct order for relevance, that Ksenia was trying to achieve with `*אלמרכב אלצ`

---

### testing notes (QA)
In the [QA site document search](https://test-geniza.cdh.princeton.edu/en/documents/), try the following kinds of searches:

- [x] For partial names where you only have one letter of a longer word, use an asterisk. For example, the partial search `Harun b. Y` should be written `Harun b. Y*`.
  - [x] This search should bring up the correct results, above "Yevr" shelfmarks.
- [x] Searches with single character words in them, such as `"Ḥārūn b. Yaʿīsh"`, should work as expected (that search should bring up PGPID 32249)
  - [x] The "b." should be highlighted in results, including in searches like `"Moshe b. Levi"`

Note: `"Naṣr b. Sālim"` still will not work due to #1475. There might be other records where the space is a unicode character. If something fails and you're not sure, I can check.

Note: `shelfmark:"T-S 12.30"` is also still not putting the most relevant result first, but not sure why; seems like a relevance score issue, and we're having issues with the shelfmark scoped search anyway. Created a new issue for that as well: #1476

---

**Describe the bug**
Searching for Harun b. Y (partial name) should yield at least 5 relevant results with the 5 relevant results appearing first. Currently I don't event get matches for Harun in the first page of the results in the public site search: [](https://geniza.princeton.edu/en/documents/?q=Harun+b.+Y&docdate_0=&docdate_1=&sort=relevance)

**To reproduce**
1. Go to public site search.
2. Search for Harun b. Y and sort by Relevance.
3. See results: [https://geniza.princeton.edu/en/documents/?q=Harun+b.+Y&docdate_0=&docdate_1=&sort=relevance](https://geniza.princeton.edu/en/documents/?q=Harun+b.+Y&docdate_0=&docdate_1=&sort=relevance)

**Expected behavior**
I want partial word search in descriptions sorted by relevancy to bring up the most relevant matches first. For this particular case the five Harun b. Y names should appear first, followed by all the Harun's and then everything else.

**Additional context**
The Abu l-Mun partial search works fine still (which was the previous test case).

## dev notes

[...]

PR #1339, opened 2023-03-02
Set min gram size to 2, preserveOriginal=true for solr edge ngram filter (#1335)

blms:
This does seem to fix the issue with highlighting on exact matches including single-character words (such as `"abu l-muna"`), and it seems to generally improve the results for #1335, but it's still producing some strange results. For example, `Harun b. Y` now matches "Ḥārūn b. Yaʿīsh", but it also now matches words like "buenos", "bien", "buen" in Ladino, and those results appear before "Ḥārūn b. Yaʿīsh". Later results include matches like "basmala" and "Bibas".


rlskoeser (commented):
Wow, right — single character matches makes search pretty much useless. I think even two characters sometimes gets a lot of irrelevant results.

I did a little bit of looking; I found this issue on elastic search https://github.com/elastic/elasticsearch/issues/13710 — problem is likely similar since both use lucene under the hood, but that doesn't solve our problem.

I did find this, in Solr 8.11 docs https://solr.apache.org/guide/8_11/filter-descriptions.html#edge-n-gram-filter - there's a `preserveOriginal` option. I think it was added in 8.6, but PUL is running 8.4 (recently figured that out, actually thought it was 8.6).  If you can confirm this option does what we need, I can request that PUL upgrade Solr — although IDK what kind of time frame it will be.

blms:
@rlskoeser Was able to test this just now, with the query `Harun b. Y`:

<details>
<summary>Results with minGramSize="1" preserveOriginal="true" (19,417 total results)</summary>

Screen Shot 2023-03-08 at 12 19 43 PM
Screen Shot 2023-03-08 at 12 19 54 PM
Screen Shot 2023-03-08 at 12 19 59 PM
Screen Shot 2023-03-08 at 12 20 03 PM

</details>


<details>
<summary>Results with minGramSize="2" preserveOriginal="true" (11,495 total results )</summary>

Screen Shot 2023-03-08 at 12 16 25 PM
Screen Shot 2023-03-08 at 12 16 30 PM
Screen Shot 2023-03-08 at 12 16 33 PM
Screen Shot 2023-03-08 at 12 16 36 PM
 
</details>


And for comparison's sake:



<details>
<summary>Results with minGramSize="1" preserveOriginal="false" (19,417 total results)</summary>

Screen Shot 2023-03-08 at 12 22 44 PM
Screen Shot 2023-03-08 at 12 22 52 PM
Screen Shot 2023-03-08 at 12 22 58 PM
Screen Shot 2023-03-08 at 12 23 03 PM

</details>

<details>
<summary>Results with minGramSize="2" preserveOriginal="false" (11,421 total results)</summary>

Screen Shot 2023-03-08 at 12 27 42 PM
Screen Shot 2023-03-08 at 12 27 46 PM
Screen Shot 2023-03-08 at 12 27 49 PM
Screen Shot 2023-03-08 at 12 27 52 PM

</details>

(apologies for the repost, forgot to reindex 🤦 )

I'm not seeing a huge difference in the results based on `preserveOriginal`; what do you think?

rlskoeser:
@blms I think `minGramSize="1"` is too small, will always result in too much noise in the search results (and it also makes the preserve original flag redundant, since there is no word smaller than 1 character). I think your test case with a single `Y` is maybe a little misleading, since there are individual `y` terms showing up in the results (suspect that because they are so many they overwhelm the relevancy score). I'm curious how that search would do if you search on `Y*` instead, but not sure if that helps us.

What I do see is that the `b.` is highlighted in the ` minGramSize="2" preserveOriginal="true` case — which I think is the behavior we want, since previously it was being thrown out. The results from before you indexed seemed slightly better for this case, which made me wonder if we might want different rules for indexing and search terms here — but I also wonder if that's because you're using `y` as a test case, which does match actual one-letter words.

blms:
@rlskoeser Searching `Harun b. Y*` with `minGramSize="2" preserveOriginal="true"` seems to be the definite best result set. I think it's reasonable to ask users searching on partial words by a single character to include an asterisk.

<details>
<summary>result set</summary>

Screen Shot 2023-03-09 at 4 13 32 PM
Screen Shot 2023-03-09 at 4 13 37 PM
Screen Shot 2023-03-09 at 4 13 42 PM
Screen Shot 2023-03-09 at 4 13 47 PM

</details>

So it looks like we do want the solr upgrade on PUL's side, and once that's available we can merge this and deploy to dev with `preserveOriginal="true"` in the managed schema.

rlskoeser:
@blms those results look so much better! I agree about wildcards for partial word matches, that seems reasonable.

rlskoeser (approved):
Great to see the expected behavior tested!

Issue #1338, opened 2023-03-01
Shift the date help text in the admin interface closer to the actual fields


kseniaryzhova:
## testing notes (round 2)

In the [QA site admin](https://test-geniza.cdh.princeton.edu/admin), the help text should now be revised as requested in https://github.com/Princeton-CDH/geniza/issues/1338#issuecomment-1470361412. (May require a hard refresh)

---

## testing notes

In the [QA site admin](https://test-geniza.cdh.princeton.edu/admin), the help text should now be easy to visually associate with the correct date-related fields. (May require a hard refresh)

---

**Is your feature request related to a problem? Please describe.**
The help text for the date on document and document date fields in the admin interface are currently too far down from the actual input field and do not look like they belong with these fields.

**Describe the solution you'd like**
Please shift the current help text on both fields upwards so it is just below the input field, like the help text underneath the calendar dropdown menu (sam field).

**Additional context**
![image](https://user-images.githubusercontent.com/105385128/222242462-6e605449-736c-4c9f-884c-ec7c09120cc5.png)

kseniaryzhova:
@blms still not shifted up - the placement is as it was before. I hit refresh several times.
![image](https://user-images.githubusercontent.com/105385128/223832000-d0b359bb-e666-4f29-8187-aa5b735bdc33.png)

kseniaryzhova:
@blms I think the best solution is to rename the actual date fields to make them one line. For Document Date (Standardized) please change to "CE Date" and take out the CE Date in the help text, along with the parenthesis around the Gregorian and Julian calendar explantion. For "Date on document (original):" just put "Date on document:" and capitalize the E in explicit in the help text.

kseniaryzhova:
@blms exactly what we wanted, looks great! Closing and thank you!

PR #1340, opened 2023-03-02
Improve help text display for document admin date (#1338)

blms:
## In this PR

Per #1338:
- A bit of admin CSS to manually override the document date section to display as a grid, with a gap wide enough to accommodate the help text issue
  - Help text issue was happening because the field label is more than one line long and pushing down the help text, which we can't otherwise address without changing the HTML structure (or doing hacky negative margin stuff)


rlskoeser (approved):
CSS override seems like a reasonable solution for customizing admin UI!

Issue #870, opened 2022-05-17
As a content editor, I want a way to track inferred dates for documents in a structured way so that it can be used for filtering, sorting, and display.


rlskoeser:
## testing notes (round 2)

On the [QA site admin](https://test-geniza.cdh.princeton.edu/admin/corpus/document), the "inferred datings" section should now appear immediately after the existing document date fields.

---

## testing notes

On the [QA site admin](https://test-geniza.cdh.princeton.edu/admin/corpus/document):

- [x] When adding/editing a document, there should now be a section for "Inferred datings" at the bottom
- [x] You should be able to add as many of these as you want
- [x] For each, notes and standardized date should be required
- [x] Standardized date must match the format in the help text, and should give an error if it does not

---

There may be more than one inferred date per document. Each will have a display date that indicates how it should appear on the frontend. Each will also have a notes field where content editors can add rationale, source, etc.

## dev notes
- [x] add new `Dating` model 
- [x] should have a FK relation to `Document`, so that we can have more than one `dating` entry per document (scholarly disagreement, for example)
- [x] add `datings` inline to `Document`
- [x] should have `display` field (similar to current "Date on document (original)")
- [x] should have standard date field, and use the same validation for standard date field used for document date
- [x] should have `notes` freetext field for tracking how it was inferred
---

### Is your feature request related to a problem? Please describe.
Many Geniza documents are dated inferrentially - there is no year on the document itself but other clues allow researchers to pinpoint a range of years (end of the 11th century, 1120s, mid-14th century) and we want this information to be reflected in the database/document detail page (both the date range and WHY we think it's this range).

### Describe the solution you'd like
I want the document detail page to have a section where a content admin can enter inferred dates with less precision (date range, century, etc). I want it to be clear that this is an inferred date and I want the content admin to have to choose precise dates for the range (end of the 11th century could be registered as 1080-1099)

kseniaryzhova:
@blms works as it should, keeping open to hear back from MR/RR about general layout/how this dates info should be presented.

kseniaryzhova:
@blms We would like to move the fuzzy date field closer to the original date field and the description, since we will be pulling fuzzy date information from the description.

kseniaryzhova:
@blms works beautifully and it's straightforward and makes sense. Yay thank you!

PR #1341, opened 2023-03-03
Add Dating model and formset to Document (#870)

blms:
## In this PR

Per #870:
- Add `Dating` model, which has a many-to-one association with `Document` (many datings per document, one document per dating), to allow adding fuzzy inferred dates at any level of specificity
- Use a `display_date` field for the fuzzy date and a `standard_date` field (using the regex from `doc_date_standard`) for the date/range used for filter/sort
- Add formset to `Document` admin

## Questions

- Thoughts on overall approach?
  - I wanted to make it a separate model since it came up in conversation that a document may need more than one of these, for example, if scholars disagree. I decided on M2O because with M2M we'd need a `through` model for the `notes` field—and because content editors may disagree about the meaning of "late xth century", and we actually decided we want to support those differences rather than enforcing one definition!
- The team decided they would prefer a simple free text `notes` field over attaching footnotes to a Dating, as content editors would otherwise likely end up often needing to attach the same source to both the Document and the Dating. But I could also see footnotes as being better for properly linking data and referring back to full bibliographic citations. Any thoughts or preferences from your perspective?


rlskoeser (commented):
Interesting! Thanks for explaining the rationale for the approach, very helpful.

You make a good case for not using a m2m relationship; through model is certainly cumbersome, and finding the "correct" copy of the "same" date would probably be annoying at best and difficult and confusing at worst (and not particularly helpful). Text notes field seems reasonable; if the dating is based on a scholarly record that discusses the document, shouldn't it already be present as a footnote on the document? No need to double up on footnotes. 

some questions:
- Will the date on document ever be migrated to use this model, or are they intentionally distinct? (I remember that date on document and inferred date are conceptually distinct)
- Does document dating not need support for alternate calendars? (is conversion typically handled by the scholars making the inferences?)
- Will this provide a basis for allowing fuzzier dates / larger ranges?
- Are the next steps planned out and accounted for? i.e., will these dates be indexed for search and filtering, displayed in search results, displayed in document details, included in dataset exports?

rlskoeser:
one more thought — since a document can have multiple inferred dates, it will make indexing, sorting, and dataset exports trickier

blms:
> Will the date on document ever be migrated to use this model, or are they intentionally distinct? (I remember that date on document and inferred date are conceptually distinct)

Indeed, intentionally distinct. Though, I didn't ask the team if they might want to support multiple dates _on_ the document as well.
 
> Does document dating not need support for alternate calendars? (is conversion typically handled by the scholars making the inferences?)

Exactly, this is intended to be a pretty free-form field that scholars can use however they see fit. That could include a range of dates within an alternate calendar, for example, which would then require additional logic to convert. Rather than try to cover all bases automatically, we decided it would be best to leave it up to each scholar.

> Will this provide a basis for allowing fuzzier dates / larger ranges?

That's the general idea here; one could say "12th century" and the range would span whatever the user enters into the standardized field as a range. Unless you mean a range that is too large to be represented in that field?

> Are the next steps planned out and accounted for? i.e., will these dates be indexed for search and filtering, displayed in search results, displayed in document details, included in dataset exports?

Display in search results and document details are under way with Gissoo, I believe. Indexing for search and filtering (and maybe dataset exports) are planned, but the idea for now is just to get the data entry unblocked.

blms:
> > Will the date on document ever be migrated to use this model, or are they intentionally distinct? (I remember that date on document and inferred date are conceptually distinct)
> 
> Indeed, intentionally distinct. Though, I didn't ask the team if they might want to support multiple dates on the document as well.

@rlskoeser Responses via Slack:

> **Rachel**: there are often copies of legal documents that contain both the date of original issuance as well as the date of the copy production. I haven’t felt squeezed by the limitation too frequently because the second date would go in the description, but it would be cool to capture both more formally!

> **Alan**: It’s also pretty common for the same legal document to be written on one day and signed a couple months later - could be a good argument for ability to enter two dates. I usually just pick one.

Sounds like it would be nice to have, but maybe out of scope for this issue. Thoughts? And anything else I should tackle on this PR?

rlskoeser:
@blms out of scope for this issue makes sense to me! And once there is support for more dates with notes, the team could decide on whether they wanted to use this dating field to track secondary document dates — right now you have it labeled for inferred dates, but it doesn't seem like there's anything structural about that (although logic based on it would matter... e.g., always use date on document if set as primary date for display, sorting, filtering).

One follow up question about date conversion from historical calendars (question for the team): they should think about whether they want to make the date conversion we built for the existing document date field available somewhere on the site as a utility. That way they could use the same conversion logic for these inferred dates and have the same degree of precision. (I'm not sure if there's an easy way to integrate it, but at least the conversion logic could be made available.)

rlskoeser (approved):

PR #1349, opened 2023-03-15
Customize admin templates for document to allow inlines between other fields (#870)

blms:
## In this PR

- Customize admin document `change_form` template and `fieldset` snippet to allow moving inlines between other fields
- Move datings inline after document date fields, per #870

### DIFF
--- a/geniza/corpus/admin.py
+++ b/geniza/corpus/admin.py
@@ -203,6 +203,7 @@ class DocumentDatingInline(admin.TabularInline):
     )
     min_num = 0
     extra = 1
+    insert_after = "standard_date"
 
 
 @admin.register(Document)
--- a/geniza/corpus/templates/admin/corpus/document/change_form.html
+++ b/geniza/corpus/templates/admin/corpus/document/change_form.html
@@ -14,6 +14,21 @@
 {% endblock %}
 
 
+{% block field_sets %}
+    {% for fieldset in adminform %}
+        {% include "admin/corpus/document/snippets/fieldset.html" with inline_admin_formsets=inline_admin_formsets %}
+    {% endfor %}
+{% endblock %}
+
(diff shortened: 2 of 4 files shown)


rlskoeser (commented):
This seems ok; my main concern is the potential for it to break when we update to newer versions of Django. Could you add a unit test that confirms that the expected inline formsets are included on the page as expected, so that if it breaks with an update it will be flagged?

blms:
@rlskoeser Very good point! I was thinking, I wished the `fieldset.html` template was structured in blocks so that we didn't have to copy basically the entire thing from the Django source code. Your idea is a great solution to this problem.

geniza/corpus/tests/test_corpus_templates.py
```diff
@@ -656,3 +658,42 @@ def test_related_list(self, client, document, join, empty_solr):
...
+
+        # Dating inline should be immediately after standard_date field (which is the value of
+        # DocumentDatingInline.insert_after)
+        soup = BeautifulSoup(response.content)
+        standard_date_field = soup.find(
+            "div", class_=f"fieldBox field-{DocumentDatingInline.insert_after}"
+        )
+        assert standard_date_field.find_next_sibling("div")["id"] == "dating_set-group"
+        dating_inline = soup.find("div", id="dating_set-group")
+        assert dating_inline.find_parent("fieldset") is not None
```
rlskoeser, line 691: whoa, hadn't thought about checking sequence like this; I suspect if it breaks it will not be present at all, but this seems like a useful way to bring this to our attention in future upgrades if something changes!

rlskoeser (approved):
Thanks for adding the test — looks to me like it should be sufficient to flag a problem if something changes that breaks the template

PR #1361, opened 2023-04-21
Bash script to count number of transcriptions in pgp-text repo

rlskoeser:
Adds the bash script I adapted for counting number of transcriptions over time in the pgp-text repo

### DIFF
--- a/scripts/count_transcriptions.bash
+++ b/scripts/count_transcriptions.bash
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+# script for counting the number of transcriptions and uinque pgpids
+# over time in the pgp-text git repository. 
+# Generates a csv file of counts per date.
+# To run, check out a local copy of the pgp-text repo,
+# and run this bash command in the base directory of that repo
+
+# NOTE: regexes will need revising when translations are added
+
+
+# adapted from https://blog.benoitblanchon.fr/git-file-count-vs-time/
+
+OUTPUT=transcription_stats.csv
+
+# create output file with a CSV header
+echo "date;transcription files;documents" > $OUTPUT
+
+# function that counts files matching the specified regex
+count() {
+    git ls-tree -r --name-only $COMMIT | grep $1 | wc -l | sed 's/ //g'
+}
+
+count_pgpids() {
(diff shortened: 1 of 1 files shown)


blms (approved):
This is awesome! Thanks for adding it to the repo.

Only note: When I try to run the script as-is with `zsh` it fails on `[ "$PREV_DATE" == "$DATE" ]`, but works if I change it to a single equals. I think it's just a zsh problem, going by [this stackoverflow answer](https://unix.stackexchange.com/a/564085) anyway, so not a big deal. But looks like using a single equals works just as well so maybe easy enough to fix.

rlskoeser:
@blms have you worked with any github actions that act as cron jobs to add data to a repo?

scripts/count_transcriptions.bash
```diff
@@ -0,0 +1,47 @@
...
+
+count_pgpids() {
+	git ls-tree -r --name-only $COMMIT | grep $1 | sed -E 's/(^.*PGPID)// ' | sed -E 's/_.*$//' | uniq | wc -l | sed 's/ //g'
+}
+   
+# for each commit in log
+git log --pretty="%H %cd" --date=short | while read COMMIT DATE
+do
+    # skip commits made on the same day
+    [ "$PREV_DATE" == "$DATE" ] && continue
```
blms, line 32:
```suggestion
    [ "$PREV_DATE" = "$DATE" ] && continue
```

blms:
> @blms have you worked with any github actions that act as cron jobs to add data to a repo?

I have not, the closest thing would be github actions that build an output and push it up to a certain branch (for github pages) triggered by a change on a different branch.

rlskoeser:
Thanks for the fix, tested locally and it successfully ran the way I'm running it (`bash ~/workarea/github/geniza/scripts/count_transcriptions.bash` )

rlskoeser:
@blms updated to make the regexes specific to transcription text files (so it won't count translations once you add them, until we add logic to account for those). I also added a q option to suppress output for when we set this up to run in a github action (I'm cribbing from stack overflow, IDK if the bash options will also work on zsh)

blms (approved):
Can't get the bash options to work on zsh but I think that's fine? They work perfectly well running the command with bash (as you've now noted in the comments!). Good to go!

Issue #1360, opened 2023-04-19
Dropdown header menu partially hidden behind search filters (z-index)


blms:
## Testing notes (QA)

On the QA [public site search](https://test-geniza.cdh.princeton.edu/en/documents/), with filters expanded, open the About dropdown. The below issue should not occur.

---

![image](https://user-images.githubusercontent.com/7234006/233093372-2c63fb63-f8f7-49cc-913c-b519dfc2b991.png)

richmanrachel:
Yes, @blms - I think you fixed it! First screenshot from production site, second screenshot from test site: <img width="1170" alt="image" src="https://github.com/Princeton-CDH/geniza/assets/68028846/297b6549-97e2-4671-9df8-79a8ec7e1ec4">

<img width="969" alt="image" src="https://github.com/Princeton-CDH/geniza/assets/68028846/7fc7b1f0-2644-45e3-bfe9-dedb000e2719">

Issue #1356, opened 2023-04-17
As a front-end user, I want to be able to search on the content of translations, so that I can find documents relating to terms that only appear within translations.


blms:
## testing notes (QA)

On the [QA public site search page](https://test-geniza.cdh.princeton.edu/en/documents/):
- [x] After adding some translations, you should be able to search on words and phrases in translations.
- [x] Keywords should be highlighted in the translation snippets in results.
- [x] There should be a "has translation" toggle that limits results to only digital translations.

## dev notes

- [x] add translation annotations to solr indexing code
- [x] ensure they are included in search query config
- [x] add them to highlighting and display them on the search results page

blms:
Looks like this didn't get closed, closing now per above comment

PR #1362, opened 2023-04-24
Translation search (#1356)

blms:
## In this PR

Per #1356:
- Support for faceting, keyword search, and highlighting on translations
- Styling for translations in search results

Per #1360:
- Fix issue with partially hidden "About" menu on search page

## Notes

- Not sure why I committed that about menu commit on this branch last week, but maybe it makes sense since it was happening on the search page. Can cherry pick if you prefer a separate PR.
- Would it be worth migrating `SourceLanguage` records over to `LanguageScript` records? Granted, I'm not sure how we would do this without the script information already present for sources (infer the script based on language?), but it otherwise seems like duplicate models to me right now.


geniza/corpus/models.py
```diff
@@ -1126,6 +1136,16 @@ def index_data(self):
                 content = fn.content_html_str
                 if content:
                     transcription_texts.append(Footnote.explicit_line_numbers(content))
+            elif Footnote.DIGITAL_TRANSLATION in fn.doc_relation:
+                content = fn.content_html_str
+                if content:
+                    translation_texts.append(Footnote.explicit_line_numbers(content))
+                    # TODO: How to handle multiple translations in different languages?
```
rlskoeser, line 1143: Since translations are into modern languages (I think? wondering about Goitein's translations into Hebrew), you may want to index them separately by language. Not sure how many languages we're including or how complicated it will get. You probably don't need to do it on the first pass! (Maybe we don't do it for all languages but a subset where we have enough that it matters / prioritized by the team).

I'm imagining something like `translation_text_en`, `translation_text_de`, `translation_text_he` using dynamic field types, and then should be possible to search across them using wildcard field names (I think, haven't done it myself in practice).

geniza/corpus/templates/corpus/snippets/document_result.html
```diff
@@ -68,6 +68,22 @@ <h2 class="title">
                     {% endif %}
                 {% endwith %}
 
+                {# translation: keywords in context if any, or excerpt #}
+                {% with document_highlights=highlighting|dict_item:document.id lang=document.translation_language_code %}
+                    <!-- indicate language when possible, or at least change of language (if unknown) for screen readers -->
+                    {% if document_highlights.translation or document.translation %}
+                        <div class="translation" lang="{{ lang }}" dir="ltr">
```
rlskoeser, line 75: Where are you getting language code? this may be an argument for indexing as `transcription_en` etc so you always know what the language is....

Not sure you should assume translation is LTR, I thought there were Hebrew translations — although I'm sure the project team could advise better on this. (Also would be nice to support translations in other LTR languages in future.)

geniza/corpus/views.py
```diff
@@ -175,14 +178,20 @@ def get_queryset(self):
...
                         method="unified",
                         fragsize=150,  # try including more context
                         requireFieldMatch=True,
                     )
+                    .highlight(
+                        "translation",
+                        method="unified",
+                        fragsize=150,
+                        requireFieldMatch=True,
+                    )
```
rlskoeser, line 194: I guess you have to call highlight twice the way it's currently implemented. We might want to update parasolr with support for multiple highlight fields and the same common options (esp. if you're going to need multiple translation fields). You could define highlight params once and pass in as keyword args if they are the same, but for only two it's probably easier to keep them readable.

rlskoeser: Solr will take a list of fields or even a wildcard (which you'd probably want if you have multiple translation fields by language), but it will probably take a little bit of effort for parasolr to handle, since right now we set highlight parameters per field only.

rlskoeser (approved):
Looks good to me. Added some comments with some thoughts on how to handle transcription language. 

Feels like there's a fair bit of redundancy between translation and transcription logic, but I don't see an obvious way to generalize/refactor, and it's only two right now, so probably not worth it.

geniza/corpus/models.py
blms: Interesting idea! The reason I made the comment was that I was wondering how to handle indexing `translation_language_code_s` when there are multiple translations so that they can get the `lang` and `dir` attributes appropriately in HTML. That would definitely solve that problem—are there other advantages of indexing them separately by language? Leveraging stemmers and the like for each?

rlskoeser: yes, exactly - will allow language-specific searching and indexing (hence my comment about modern languages: I think there will be reasonable Solr support for most of the translation languages)

geniza/corpus/templates/corpus/snippets/document_result.html
blms: I'm indexing the language code as `translation_language_code_s`, but it's actually pretty unstable because it just picks the language code of the last translation it happens to encounter in the loop through Footnotes. Maybe I shouldn't do that!

I had actually intended this to respect RTL/LTR of that language code, similarly to how I did it with the main translation viewer, but must have forgotten. That in itself is quite hacky:

https://github.com/Princeton-CDH/geniza/blob/882b91b084885aac531787dc8966c2fb44964efa/geniza/corpus/templates/corpus/snippets/document_transcription.html#L189

It's just hard-coded checking for a couple of specific RTL languages that are likely to have translations. So at least one thing that I should probably do is add a `direction` field to the `SourceLanguage` model and use that when possible. (Or perhaps even merge that with `LanguageScript` and add `direction` to that model! Any reason why those two are separate?)

blms:
@rlskoeser Ok, revised SourceLanguage to add the direction field. This approach should work well enough for search results to respect RTL/LTR direction until we start indexing translations separately by language; it also will continue to work in the ITT panel, where indexing doesn't come into play.

geniza/footnotes/migrations/0028_sourcelanguage_direction.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 3.2.16 on 2023-05-01 15:06
+
+from django.db import migrations, models
+
+
+def set_rtl_langs(apps, schema_editor):
+    # set known existing RTL languages' direction to RTL
+    SourceLanguage = apps.get_model("footnotes", "SourceLanguage")
+    # Arabic, Hebrew, Judaeo-Arabic, and Ottoman Turkish are all RTL
+    rtl_langs = SourceLanguage.objects.filter(code__in=["ar", "he", "ISO 639-2", "ota"])
```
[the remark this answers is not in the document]
blms: Yup, noticed and fixed that right after requesting review! 😆 It was a problem in production but I've corrected in both places.

And, huh—there are some being tagged as being in Judaeo-Arabic but I am not sure they _should_ be. https://geniza.princeton.edu/admin/footnotes/source/?languages__id__exact=6

Maybe we need clearer instructions for source language?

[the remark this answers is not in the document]
rlskoeser: Oh, interesting; they're including original language if the source includes transcriptions? I see the logic in that from the researcher side but not sure that's the best place to track that information (or if it's even needed)...

Probably something for you to discuss and decide with the team in terms of what is meaningful/important to them and practical for you, particularly with regards to translation. I had imagined the source language as being the language of the scholarship / primary language. If they need to track other languages maybe you'll need to split out primary and secondary languages, but do push them to think about whether that is really necessary to track as structured data before complicating the data model too much.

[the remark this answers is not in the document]
blms: Thanks Rebecca, great advice! Will do.

Issue #1353, opened 2023-04-12
As a content editor, I want a way to automatically import Google Doc translations into the translation module, so that I can easily prepare existing translations from multiple sources for ingest.


kseniaryzhova:
## Testing notes

On the [QA site admin](https://test-geniza.cdh.princeton.edu/admin):

- [x] Check the Log Entries, there should be many entries of new Digital Translation footnotes and annotations added by the "script" user
- [x] Check 10 or so documents with new translations added from the Lieberman Google docs and confirm the translation is there, with the correct footnote location (if applicable) and source
- [x] Check them on the public site as well and confirm they're showing up, both in the translation viewer and on the "scholarship records" page
- [x] On the public site search, "has translation" should now show at least 85 documents with translations
- [x] Search for terms in some of the translations and confirm they show up in the search results

---

**Is your feature request related to a problem? Please describe.**
We are going to have a lot of translations on Google Docs to upload to the PGP (once the published material is OCRed). It would be amazing to have an automated way to upload them.

**Additional context**
Working with @blms on a template for Google Docs so we can put them in a spreadsheet and then upload that content onto the PGP.

kseniaryzhova:
@blms works as it should! Unrelated to this, but when/why did we decide to move the footnotes up above the editor in document detail?

PR #1377, opened 2023-05-18
Add ingest script for Google Docs translations

blms:
## In this PR

Per #1353, add a script to ingest translations from the Google Doc template.

## Notes

I left this as a draft because I had a question about handling authentication.

The way it works is:
- you enable the Google Drive API for a "project" in Google Cloud, with the permission to read, export, and list files
- you grant the application an API credential for that project in the form of an OAuth Client ID, and it gives you a `credentials.json` secrets file
- when running the script, `get_credentials` will read that JSON file and open a Google login OAuth consent screen in your browser
- upon granting access through the OAuth consent screen, it generates a second `token.json` file, which it then uses to actually make the queries
- if the Google account you logged in with on the OAuth screen has appropriate permissions on the requested drive/folder, it will be able to perform those read, export, and list actions

For now, I'm just storing them in my local `data/` directory and configuring their paths through Django local settings. Would it work to just store those files somewhere on the QA and prod servers? Will they need to be re-populated on every deploy? (Also, I imagine a PGL account should be the one to create the Google Cloud project and credentials file, but that's not really a code issue)

One other question: does any of this strike you as needing unit testing?


blms:
Realizing now I didn't provide these!

Example doc: https://docs.google.com/document/d/1cd89Y30J9SdjST738ER1otaN6CEOsX_cKY4PJ17W29A/edit
Branch for my test exports: https://github.com/Princeton-CDH/test-pgp-annotations/tree/blms-dev

geniza/corpus/management/commands/googledoc_to_annotation.py
```diff
@@ -0,0 +1,389 @@
...
+
+        # disconnect solr indexing signals
+        IndexableSignalHandler.disconnect()
+
+        # initialize annotation exporter; don't push changes until the end
+        self.anno_exporter = AnnotationExporter(
+            stdout=self.stdout,
+            verbosity=options["verbosity"],
+            push_changes=False,
+            # this will be overwritten
```
rlskoeser, line 79: The comment here is ambiguous. I guess the base commit message is required?

geniza/corpus/management/commands/googledoc_to_annotation.py
```diff
@@ -0,0 +1,389 @@
+"""
+Script to convert translation content from Google Docs template
```
rlskoeser, line 2: I think I'm missing context about the bigger picture of the setup for this script and how you imagine it will be run, what the workflow is. Do you expect the script to be run manually? Cron job? One-time or ongoing? etc.

I was initially going to ask for an explanation via github comments, but I think it would make sense to expand the docs and add that larger context here, and make sure it gets included in the sphinx docs.

geniza/corpus/management/commands/googledoc_to_annotation.py
```diff
@@ -0,0 +1,389 @@
...
+        self.stats = defaultdict(int)
+        self.stats["files"] = 0
+        self.stats["created"] = 0
+        self.script_run_start = timezone.now()
+
+        # get script user for log entries
+        self.script_user = User.objects.get(username=settings.SCRIPT_USERNAME)
+
+        # get drive id from settings
+        self.drive_id = settings.GOOGLE_DRIVE_ID
```
rlskoeser, line 66: should raise a command error if not configured with instructions to add

geniza/corpus/management/commands/googledoc_to_annotation.py
```diff
@@ -0,0 +1,389 @@
...
+            # this will be overwritten
+            commit_msg="PGP translation export from Google Docs migration",
+        )
+        self.anno_exporter.setup_repo()
+
+        # use rich progressbar without context manager
+        progress = Progress(
+            MofNCompleteColumn(), *Progress.get_default_columns(), expand=True
+        )
+        progress.start()
```
rlskoeser, line 88: progbar should be optional if we need to run this as a cron job

geniza/corpus/management/commands/googledoc_to_annotation.py
```diff
@@ -0,0 +1,389 @@
...
+        fetch_task = progress.add_task("Fetching list of files...", total=None)
```
rlskoeser, line 89: I would move the logic to get the list of files into a separate method

geniza/corpus/management/commands/googledoc_to_annotation.py
```diff
@@ -0,0 +1,389 @@
...
+
+        # start the download, then loop through all chunks until file is downloaded
+        downloader = MediaIoBaseDownload(file, request)
+        done = False
+        while done is False:
+            _, done = downloader.next_chunk()
+
+        return file.getvalue()
+
+    def new_translation_annotation(self):
```
rlskoeser, line 212: would it make sense to move this to the annotation model object?

geniza/corpus/management/commands/googledoc_to_annotation.py
```diff
@@ -0,0 +1,389 @@
...
+            return
+
+        # get the first canvas only; researchers want all translation content on first img
+        canvas_base_uri = doc.manifest_uri.replace("manifest", "canvas")
+        iiif_canvas = next(
+            iter(doc.iiif_images(filter_side=True)),
+            f"{canvas_base_uri}1/",  # default in case there are no images
+        )
+
+        # get or create a digital translation footnote
```
rlskoeser, line 280: in what cases would the digital translation already exist?

geniza/corpus/management/commands/googledoc_to_annotation.py
```diff
@@ -0,0 +1,389 @@
...
+        footnote.save()
+        # log creation
+        if created:
+            self.log_addition(
+                footnote,
+                "Created new footnote for migrated Google Docs digital translation",
+            )
+        else:
+            # remove all existing annotations associated with this
+            # document and footnote so we can reimport as needed
```
rlskoeser, line 298: oh, reimport sounds like context I am missing in the bigger picture

geniza/corpus/management/commands/googledoc_to_annotation.py
```diff
@@ -0,0 +1,389 @@
...
+                    % (
+                        len(existing_annos),
+                        pluralize(existing_annos),
+                        footnote.source,
+                        doc.pk,
+                    )
+                )
+                # not creating log entries for deletion, but
+                # this should probably only come up in dev runs
+                existing_annos.delete()
```
rlskoeser, line 315: If this is a dev only scenario lets figure out how to make sure it doesn't happen on a non-dev system. Optional flag to clear out previous imports? rely on debug config to allow this case without an error?

geniza/corpus/management/commands/googledoc_to_annotation.py
```diff
@@ -0,0 +1,389 @@
...
+            db_anno.save()
+            # log entry to document annotation creation
+            self.log_addition(db_anno, "Migrated from Google Doc translation")
+
+            self.stats["created"] += 1
+
+        # export html/txt translation files to github backup
+        self.anno_exporter.export(
+            pgpids=[doc.pk],
+            commit_msg="Translation migrated from Google Doc - PGPID %d" % doc.pk,
```
rlskoeser, line 388: I would call this an import, not a migration

rlskoeser: wondering if you could set the base commit message on the class and then use it as the starting point for this one (reduce redundancy and also the unused commit message)

geniza/corpus/management/commands/googledoc_to_annotation.py
```diff
@@ -0,0 +1,389 @@
...
+
+        return anno
+
+    def process_file(self, name, html_file):
+        soup = BeautifulSoup(html_file, "html.parser")
+        # first table is metadata, second table is translation
+        tables = soup.find_all("table")
+        # extract the footnote metadata
+        metadata = tables[0].find_all("td")
+        # handle earlier and later template revisions
```
rlskoeser, line 237: what does this mean?

geniza/corpus/management/commands/googledoc_to_annotation.py
```diff
@@ -0,0 +1,389 @@
...
+
+        anno.target.source.type = "Canvas"
+        anno.target.selector.type = "FragmentSelector"
+        anno.target.selector.conformsTo = "http://www.w3.org/TR/media-frags/"
+
+        return anno
+
+    def process_file(self, name, html_file):
+        soup = BeautifulSoup(html_file, "html.parser")
+        # first table is metadata, second table is translation
```
rlskoeser, line 233: can you add some validation / check that the file has the supported structure? what happens if someone accidentally drops a file into this folder that doesn't belong there?

rlskoeser (commented):
Added some initial comments, but I'm missing context and bigger picture workflow about how you intend this to work.

Some of my concerns relate to whether this is a one-time import or is meant to be an ongoing workflow. If ongoing, I'd like to know how stable the APIs are going to be and how we handle it when they change.

Regarding credentials: 
- yes, API credentials should be created from a PU account. Is anyone on the PGL technical enough to do this and do they have a systems account to use for it? I remember drive credentials being kind of a pain when I did it before. The project I did before I used my cdhdevteam systems account for the credentials; we could do that again.  
- the ansible deploy will need to be updated to handle the credentials; they can encrypted with ansible vault and added to the deploy; we'll need to add deploy logic to install them in the correct place, but I think we're already using a project-specific role so there should be a place to put it. Copying files to a location isn't too difficult.
- will QA and Production need different credentials?

> One other question: does any of this strike you as needing unit testing?

If it's a one-time import then I'd be ok with rigorous acceptance testing. If it's meant to be used on an ongoing basis then it absolutely needs testing.

rlskoeser (approved):
Your revisions look good.

Since this is a one-time thing rather than ongoing, seems fine to me for you to create the developer credentials and add them to the ansible geniza vault and local settings, and then remove them once they are no longer needed.  (I can help with the ansible bits if you need.)

blms:
@rlskoeser Returning to this—I realized I could simplify, since it's a one-time run, and just supply the client ID and secret when running the command instead of requiring anything to go into the ansible vault.

rlskoeser:
> @rlskoeser Returning to this—I realized I could simplify, since it's a one-time run, and just supply the client ID and secret when running the command instead of requiring anything to go into the ansible vault.

Great, that sounds much simpler!

rlskoeser (approved):

Issue #1394, opened 2023-06-15
Merging two documents with digital content footnotes for the same source results in unique constraint violation


blms:
### testing notes (QA)

On the [QA site admin](https://test-geniza.cdh.princeton.edu/admin):

- [x] Check PGPID 29344 and PGPID 4474. Note their digital edition footnotes' differences ("location" and "notes" fields) and review their transcriptions.
  - [x] After reviewing their transcriptions, try merging PGPID 29344 into PGPID 4474. There should not be an error.
  - [x] The two transcriptions should now be combined on the merged document's Khan transcription, with separate annotations (blocks of transcription) for each.
  - [x] In the Footnotes section, the "location" and "notes" fields on the single merged digital edition footnote should capture the data from both of the footnotes pre-merge.
- [x] If possible, try to merge two other documents that both have transcription content on the same source—but this time, with _no_ data in the "location" or "notes" fields in the digital edition footnotes on either document. You may need to create such transcriptions if you can't find them. (This is the situation with Ksenia's merge on PGPID 5200, but that's already been completed.)
  - [x] All transcription content should be present, none deleted or missing.

---

### dev notes

There is an interesting edge case where one document, PGPID 4474, has a transcription from Geoffrey Khan. Another PGPID, 29344, was created and refers to the same fragment/images. This second PGPID also has the transcription sourced from Khan, but with emendations by Yusuf.

An attempt to merge these two documents into 4474 results in a unique constraint violation:

```py
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "one_digital_edition_per_document_and_source"
DETAIL:  Key (source_id, object_id, content_type_id)=(135, 4474, 8) already exists.
```

The solution here is either to discard Khan's original transcription and replace it with the emended one, or to take Yusuf's emendations and consider them a separate source, reassign the source, then perform the merge.

But it is possible this error could happen again, so it would be helpful to show a more explanatory error message and give the user those two options to consider.

---

- also check that "digital content" footnotes don't get merged with non-digital ones during document merge. always create a new footnote!

kseniaryzhova:
@blms I merged PGPIDs 5200 and 5201 - they were duplicates, with the same transcription (from the same source) split verso and recto (verso on one PGPID, recto on the other). So the text was different on each one, but the source was the exact same. I was able to merge but one of the transcriptions disappeared.

blms:
[...]

PR #1405, opened 2023-06-26
Fix document merge function footnote bugs (#1394)

blms:
## In this PR

Per #1394:
- Fix bug where merging a document with a "digital edition" or "digital translation" footnote on both the merge source and target documents, on the same Source, could result in lost data or a unique constraint violation
  - Migrate all annotations to the footnote on the merge target, and delete the merge source's footnote
  - If there are no annotations, remove the digital content doc relation from the merge source's footnote and just add the footnote to the merge target document


geniza/corpus/models.py
```diff
@@ -1362,9 +1362,38 @@ def merge_with(self, merge_docs, rationale, user=None):
...
+                    defaults={
+                        "notes": footnote.notes,
+                        "location": footnote.location,
+                        "url": footnote.url,
+                    },
+                )
+                # reassign each annotation's footnote to the footnote on the merge result doc
+                for annotation in footnote.annotation_set.all():
+                    annotation.footnote = self_fn
+                    annotation.save()
```
rlskoeser, line 1383: Is the problem that `includes_footnote` comparison doesn't account for annotations?  Should it include that when checking, or should it be optionally included? (... trying to think when you would not want to include annotations when comparing footnotes, but it could be an option you specify when you call the includes method if we think we don't always need it).

Alternately, would it make sense to move annotations to the equivalent footnote if we don't want to include annotations in the footnote equality check?

I'm not totally following this logic here but I think it might be more complex than needed / not in the right place. Seems like the simplest solution would just to include annotations in the footnote equivalence check, but I may be misunderstanding the problem.

geniza/corpus/models.py
```diff
@@ -1362,9 +1362,38 @@ def merge_with(self, merge_docs, rationale, user=None):
...
+                # if there is otherwise not a match, add the footnote to the merge result document
+
+                # first remove any digital doc relations to avoid unique constraint violation;
+                # footnote should not have such a relation anyway if there are 0 annotations
+                if Footnote.DIGITAL_EDITION in footnote.doc_relation:
+                    footnote.doc_relation.remove(Footnote.DIGITAL_EDITION)
+                    footnote.save()
+                if Footnote.DIGITAL_TRANSLATION in footnote.doc_relation:
+                    footnote.doc_relation.remove(Footnote.DIGITAL_TRANSLATION)
+                    footnote.save()
```
rlskoeser, line 1394: do these get added back later? or is it a data error if they're showing up here, since we already know they don't have annotations?

geniza/footnotes/models.py
```diff
@@ -439,7 +439,12 @@ def includes_footnote(self, other):
 
         for fn in self.all():
             if (
-                all(getattr(fn, val) == getattr(other, val) for val in compare_fields)
+                # if either of the footnotes have annotations, do not treat as a match
+                not (fn.annotation_set.exists() or other.annotation_set.exists())
+                # compare all ohter fields
```
rlskoeser, line 444:
```suggestion
                # compare all other fields
```

geniza/corpus/models.py
rlskoeser: ok, made the mistake of commenting before reading through everything - I see you've made some updates to the includes footnote method and are explicitly declaring footnotes with annotations to not be equivalent (which seems reasonable). 

I think you should add a comment and make this more explicit: these are two different cases where there's no equivalent footnote. If I'm understanding correctly, the options are to find an appropriate footnote and re-associate the annotations or create a new footnote. IMO notes and locations should be preserved, because they may include important information about where the content is located or if the edition has been revised by someone.

I still think parameterizing the equiv fn check might be helpful and simplify this code; is it helpful to think in terms of exactly equivalent footnotes (your updated logic) or nearly equivalent ( ~ old logic)?

geniza/corpus/models.py
blms: Exactly, it's the latter. There shouldn't be a digital footnote without annotations.

geniza/corpus/models.py
```diff
@@ -1337,38 +1337,50 @@ def merge_with(self, merge_docs, rationale, user=None):
...
+                            setattr(self_fn, attr, getattr(footnote, attr))
+                    self_fn.save()
+                    # reassign each annotation's footnote to the footnote on this doc
+                    for annotation in footnote.annotation_set.all():
+                        annotation.footnote = self_fn
+                        annotation.save()
+                except Footnote.DoesNotExist:
+                    # if there is no match, we are clear of any unique constaint violation and can
+                    # simply add the footnote to this document
+                    self.footnotes.add(footnote)
```
rlskoeser, line 1369: I think pulling this out separately (instead of using get or create) is clearer & better

rlskoeser (approved):
thanks for elaborating more in the comments; what you say about not changing the includes footnote check makes sense

Issue #1366, opened 2023-04-26
As a content editor, I want a separate field to record people's names and roles in each document, so that I can build a structured dataset of all people across the PGP.


kseniaryzhova:
### testing notes (round 3)

On the QA site admin:

- Edit an existing person and save. If there are no person records, create one, hit save, and then (this is important!) edit it and save it again.
- [x] Create a new person. Before hitting save, try to add a relationship with the previously edited person. If you begin typing that person's name, they should show up in the autocomplete list of people.

---

### testing notes (round 2)

Visit the [QA site admin](https://test-geniza.cdh.princeton.edu/admin) and add a new Person entry.

- [x] Start by adding a name for the person. The help text for the language should read: "Please indicate the language of most components of the person's name as you've indicated it here"
  - [x] The check box for the name should now read "display name" (instead of "primary")
- [x] Social roles should be listed alphabetically
- [x] The check box below that should read "Person page," and then for help-text, "Check box if this person should have a dedicated, public Person page on the PGP. If checked, please draft a public description below."
- [x] Under "Related People", the list of relationship types under the Relations field should be listed alphabetically

Next, edit or create a Document in the admin.

- [x] Related Fragments, Footnotes, and Related People sections should all be below transcription viewer

---

### testing notes

Visit the [QA site admin](https://test-geniza.cdh.princeton.edu/admin).

- [x] There should be a new section called "Entities" with subsections called People, Person social roles, Person-Document relationships, and Person-Person relationships
- [x] All of those except People should be pre-populated with the [lists from here](https://docs.google.com/spreadsheets/d/1RwdV_qMsXo1L-UOplU77ps2J1DSvKb6OKAFq4wOwCRo/edit?pli=1#gid=0)
- [x] You should be able to add a new person, with names in different languages, gender, social role, description, and footnotes
- [x] You should be able to add/edit social roles, Person-Document relationships and Person-Person relationships
  - [x] You should be able to add Hebrew and Arabic translations for each
  - [x] For social roles, there should also be a "display label" field in case you want to provide a slightly different label for public display
- [x] When editing or adding a Document, there should be a new section at the bottom of the "Change document" page called "Related People". You should be able to pick the person you created there.
  - [x] Once you save this relationship, the document should also show up on the Person's "Change person" page.
- [x] If you create a second person, on the "Add person" page, there should be a section called "Related People". you should be able to select the first person you created from there and specify the relationship between the two people.
[...]

PR #1412, opened 2023-07-05
Revisions to entities models (#1366)

blms:
## In this PR

Per https://github.com/Princeton-CDH/geniza/issues/1366#issuecomment-1622221304 via slack conversations:
- Some small revisions to entities models: help text, verbose names, and alphabetical sorting
- Tweak the mixed fieldsets/inlines logic to allow the ITT panel to be mixed (in the same order as it used to be)

## Questions

The hacky solution becomes significantly hackier here as I make an exception for the ITT panel, which is not technically a fieldset or inline, but essentially a template snippet being rendered in place. Is there a more Django way we can do this, like creating some kind of custom fieldset or inline for it?

The rationale behind this is to keep things from changing too drastically on the admin page without users necessarily understanding why. For example, Ksenia's slack message:

> why/when did we decide to move the footnotes up above the transcription editor in the document detail? I'm not sure I like the new version because it clumps the text together and at least the editor was a break. But also it communicates what's available on the page much more quickly (to have the editor visible first) and then you can kind of dig down into the footnotes

If the solution I've implemented here is too hacky and there's not a clear simple solution, I can ask about priority.


rlskoeser (commented):
I see what you mean about the fieldset and inline order being even more hacky; I think what I've usually seen is just customizing the model-specific admin template more, which might be more readable and explicit but ends up with a lot of logic in the template which seems dangerous for maintenance. 

I think I'd almost rather have the order configured by name (like you do for admin field edit order) and then implement that in the template - but I'm not sure how hard that is and not sure it's worth your time refactoring. As long as this is tested and commented really clearly it seems ok... but it would be good to look at how we might be impacted by django upgrades.

I wonder if there are any improvements if we can get to newer versions of django? I know they were working on making django admin easier to customize, but IDK if this level of customization was on their radar or if it's very common.

rlskoeser (approved):

Issue #1371, opened 2023-05-10
As a content editor, I want a way to filter documents by date in the admin for enhanced csv exports


kseniaryzhova:
### testing notes (QA -- round two)

On the [QA site admin](https://test-geniza.cdh.princeton.edu/admin), navigate to the list of documents.
- [x] The requested help text should now be visible for both date filters

### testing notes (QA)

On the [QA site admin](https://test-geniza.cdh.princeton.edu/admin), navigate to the list of documents.
- [x] There should now be two new filters, "Date from (CE)" and "Date to (CE)", on the right sidebar below the "By status" filter section
- [x] Inputting CE dates into these filters, and hitting the Enter key, should appropriately filter the list, only showing documents with document dates that fall within the range specified
- [x] After applying a filter, clicking "x Remove" underneath should remove the filter condition
- Note that inferred date has been split into a separate user story (#1417)

### dev notes
- [x] filter on document standard date
- ~~filter on datings with option to exclude inferred datings in filter~~ (moved to new user story #1452)

---

**Is your feature request related to a problem? Please describe.**
There is currently no way to filter by date in the admin search, even though we have dated documents and will have many more with inferred dates.

**Describe the solution you'd like**
We discussed having a date filter similar to that in the public where a user can enter their desired date ranges, as well as a way to differentiate bettwen inferred and dates on document.

kseniaryzhova:
@blms is it expected behavior that you can only enter one date (Start or end) and a time? So if you want to do a range (1000-1100) you have to put in 1000 and hit enter and then put in 1100 as the end date and hit enter.

blms:
@kseniaryzhova That is expected behavior, but I can definitely see how that's not ideal… it might be a limitation of the Django admin, though, since each filter is applied separately (similar to how you have to click a link on each of the other filters). If you'd like, we can mark this as "tested needs attention" and I can investigate and see if there's a better solution.

kseniaryzhova:
@blms Please put help text for the admin site date filter, along the lines of "hit enter after each date entry" or something like that to make it clear you have to hit enter each time after inputting start and end date.

kseniaryzhova:
@blms Help text looks great, had MR look at it too! Thanks, closing!

PR #1416, opened 2023-07-07
Allow filtering by date in admin with solr (#1371)

blms:
## In this PR

Per #1371:
- Allow filtering "Date from" and "Date to" in Document admin list view, using solr to take advantage of pre-processed dates


rlskoeser (approved):
Cool, hadn't seen this kind of admin filter before

Issue #1373, opened 2023-05-10
As a content editor, I want a separate field in the document detail page so that I can record place information mentioned in the document.


kseniaryzhova:
### testing notes (QA)

- [x] On the [QA site admin](https://test-geniza.cdh.princeton.edu/admin), there should now be a section called Places.
- [x] There should also be sections for "Document-Place relationships" and "Person-Place relationships", and these should be filled out with the agreed-upon place relationship types.
- [x] Visit the "Places" section and you should be able to add new Places: name(s), lat/long coordinates, related people, and footnotes for tracking the source of the information.
- [x] Add or edit a Document. You should see a new section, Related Places, and be able to search for and add related places with the dropdown(s).
- [x] Add or edit a Person. Again there should be a Related Places section that allows you to add Places to a Person.
- [x] When you add related places to documents and people, and go back to the Place page to edit, you should now see those documents and people show up under Related Documents and Related People on the Place page.
- Note that #1453 is a separate user story and not being tested here.
- Note that #1461 also applies to Places and can be added later to add documents directly to a Place page, but is not being tested here.

---

**Describe the solution you'd like**
I want a way to record the places mentioned in a document and to qualify this data, i.e. by specifying where a letter was written and to where it was delivered as well as the itineraries described in the letter. This will help us later track the movements of people and materials and record this data for use in visualization approaches.

kseniaryzhova:
@blms 
Finalized language for document + place relations: 
- Origin (written in (legal document), written from (letter))
- Destination (generally for letters)
- Places mentioned (excluding origin and destination) 
- Possibly the same as a place that appears in
- Was once believed to appear in but in fact doesn’t

blms:
@kseniaryzhova Could I suggest these simplifications? We could make these display labels if you need more information for data entry purposes in the admin.

- Origin
- Destination
- Mentioned
- Possibly mentioned
- Formerly believed to be mentioned

kseniaryzhova:
@blms works as it should! My one quip is the same as the people page, the fact that I can't add documents from the places page. This may be less relevant actually than it is for people since we'll mostly be browsing places.

PR #1419, opened 2023-07-12
Place entities (#1373)

blms:
## In this PR

Per #1373:
- Add Place entities to the backend, including inlines to attach them to related Documents and People
- Make some inlines generic to allow DRY code reuse


geniza/corpus/admin.py
```diff
@@ -373,9 +361,10 @@ def view_old_pgpids(self, obj):
         DocumentTextBlockInline,
         DocumentFootnoteInline,
         DocumentPersonInline,
+        DocumentPlaceInline,
     ]
     # mixed fieldsets and inlines: /admin/corpus/document/snippets/mixed_inlines_fieldsets.html
-    fieldsets_and_inlines_order = ("f", "f", "i", "f", "itt", "i", "i", "i")
+    fieldsets_and_inlines_order = ("f", "f", "i", "f", "itt", "i", "i", "i", "i")
```
rlskoeser, line 367: are you able to make sense of these and keep track of them when you're updating?

seems like comments at least would be helpful when you have this many of them

geniza/entities/models.py
```diff
@@ -276,3 +276,111 @@ class Meta:
...
+        decimal_places=4,
+        blank=True,
+        null=True,
+    )
+    longitude = models.DecimalField(
+        max_digits=7,
+        decimal_places=4,
+        blank=True,
+        null=True,
+    )
```
rlskoeser, line 296: I haven't used it before and am not sure how much it takes to integrate, but would it make sense to use a `PointField` from geodjango here? It seems like the default widgets might get you a lot of functionality that would be useful.

geniza/entities/models.py
```diff
@@ -276,3 +276,111 @@ class Meta:
...
+    # sources for the information gathered here
+    footnotes = models.ManyToManyField(Footnote, blank=True, related_name="places")
+
+    def __str__(self):
+        """
+        Display the place using its display name, if one is designated,
+        otherwise display the first name.
+        """
+        try:
+            return str(self.names.get(primary=True))
```
rlskoeser, line 306: Is at least one name required? should a primary name be required?

geniza/entities/templates/admin/entities/place/change_form.html
```diff
@@ -0,0 +1,10 @@
+{% extends "admin/change_form.html" %}
+
+{# Render mixed normal and inline fieldsets #}
+{% block field_sets %}
+    {% include "admin/corpus/document/snippets/mixed_inlines_fieldsets.html" %}
```
rlskoeser, line 5: should this be set at a higher level in a common template if it's used across multiple admin models?

rlskoeser (approved):
Seems reasonable.

blms:
Thanks for the review @rlskoeser! Your question about the primary name helped uncover some issues with the unique constraint for max one primary name per entity. First, you're right that there should also be a minimum of one. Second, I was able to cause some unique constraint violations with some stress testing of the admin (setting multiple primary names, hitting save, getting the validation error, and then unchecking all but one). 

So I decided to move that logic entirely into the form and out of the DB, which seems to generally be best practice for this kind of generic FK relations anyway. (Not sure if I should go ahead and remove the error catching in the str method now too, or just leave it in in case of a missed edge case…)

geniza/corpus/admin.py
```diff
@@ -469,9 +457,20 @@ def view_old_pgpids(self, obj):
...
+        "f",  # shelfmark, languages, description fieldset
+        "f",  # date on document fieldset
+        "i",  # DocumentDatingInline
+        "f",  # tags, status, order override fieldset
+        "itt",  # images/transcription/translation panel
+        "i",  # DocumentTextBlockInline
+        "i",  # DocumentFootnoteInline
+        "i",  # DocumentPersonInline
+        "i",  # DocumentPlaceInline
+    )
```
rlskoeser, line 473: 👍 thanks for documenting!

geniza/entities/models.py
rlskoeser: I agree that it's probably not worth the overhead of the additional dependencies. It's possible that PUL already has postgis installed since they are doing some geospatial work (or perhaps they're just using Solr spatial search, which I haven't had the chance to work with but seems cool)  - I'd be happy to ask what they're using if that would be helpful.

rlskoeser (approved):
Updates look good to me.

Issue #1070, opened 2022-09-11
As a content editor, I want to override the orientation of images displayed for a document so I can rotate images to display in logical orientation for readability/useability.


kseniaryzhova:
### testing notes (QA)

On the [QA admin site](https://test-geniza.cdh.princeton.edu/admin):
- [x] Navigate to a document with images. Rotate one or more of the images using the new rotation controls (may require a hard refresh to appear). The buttons should rotate the images in place
- [x] You should also be able to reorder the images freely
- [x] On save, the rotations should persist, and appear in the transcription/translation viewer in the admin
- [x] Rotations should also be reflected in the public site document detail page (may require a refresh)
- [x] Rotations should also be reflected in the search results thumbnails on the public site

---

**Is your feature request related to a problem? Please describe.**
There are some images (newest Bodleian batch) which display the document sideways. This is confusing and makes it hard to read documents and edit/add new transcriptions. See Additional context for a screenshot. 

**Describe the solution you'd like**
Just like the image order override, I would like for there to be an option in the admin site to rotate the displayed images to their logical positions. I would like for it to be intuitive and preferably linked to the image order override section (in my mind the two features go together as image display manipulation). 

**Additional context**
![image](https://user-images.githubusercontent.com/105385128/189538836-e3ef5892-a8be-4f7c-a157-f33c7c262188.png)

rlskoeser:
@kseniaryzhova is it reasonable to only allow increments of 90 degree rotations? And do you have any sense of which direction would be more logical or most common, rotate to the right or to the left first?

kseniaryzhova:
@rlskoeser 90 increment rotations at a time right? So I would turn twice to get an 180 degree turn right? (There's at least one instance where it need a 180 turn). As for which would be more common, right or left, I have no idea. The above example should be rotated 90 degrees to the left, but a quick search turned up equal instances of it needing to be rotated right or left. If it helps, this solution does not need to be user friendly - it's enough that the management team can do it, I don't envision this being of use to content admins/editors since it's an infrequent permanent fix to help with the display.

blms:
From @kseniaryzhova over slack:

> Oh yes only 90 degree turns is good. It's really just to turn an image permanently. Anything else people can use the little knob that does rotations already. This is just for like the Bodleian manuscripts that are published on the Bodleian site as sideways. Or sometimes another collection will have the verso with the main text upside down.

kseniaryzhova:
[...]

PR #1421, opened 2023-07-17
Add rotation override control to document model, admin (#1070)

blms:
## In this PR

Per #1070:
- Add rotation override field to Document
- Allow this to be set via JS/buttons in the django admin
- Ensure rotations are moved correctly when image order is changed
- Display rotated images on document detail page/editor
- Ensure OSD launches with the correct rotation

## Notes

- I thought about trying to couple rotation with the canvas URI, perhaps even keeping both the rotation override and the order override in a single data object, but this (array of ints) seemed like the simplest solution that would have the most minimal impact on code, and behave otherwise most similarly to the order override.
- There is a possibility of the rotation overrides getting out of sync in very rare edge cases, for example if rotated fragments are removed from a document and a different fragment is present or added. (Also, if fragments, order, or rotation are altered programmatically, the rotation may get out of sync.) Those edge cases seem rare enough that they aren't worth coding exceptions for and can be manually corrected in the admin.


geniza/corpus/models.py
```diff
@@ -501,6 +501,9 @@ class Document(ModelIndexable, DocumentDateMixin):
     image_order_override = ArrayField(
         models.URLField(), null=True, verbose_name="Image Order", blank=True
     )
+    image_rotation_override = ArrayField(
```
rlskoeser, line 504: Does it have to be an array field? That's the part that feels brittle to me. How does it update when the images are dragged to re-order?

What about using a dictionary in a jsonfield, so you only story the overrides that are specified and they are keyed on an image or canvas id?

geniza/corpus/models.py
```diff
@@ -792,28 +796,37 @@ def iiif_images(self, filter_side=False, with_placeholders=False):
...
+                iiif_images  # add any remaining images after ordered ones
+            )
+            # reassign to original variable so we can add rotation if necessary
+            iiif_images = ordered_images
+
+        # add any rotations from the rotation override to each dict entry, in order
+        for i, degrees in enumerate(self.image_rotation_override or []):
+            # ensure no errors if too many entries in the override array
+            if len(iiif_images.keys()) > i:
+                iiif_images[list(iiif_images.keys())[i]]["rotation"] = int(degrees)
```
rlskoeser, line 816: it's hard to read with this many nested `())[`

rlskoeser (commented):
Exciting to see this implemented; this generally looks reasonable to me but I do think using an array field for this is a brittle solution and a dict would be much better.

FWIW, I checked the spreadsheet where we spec'ed out the requirements for transcription support and how to implement with iiif, and it's possible to set a rotation override in a iiif manifest (I think this is the relevant portion of the docs: https://iiif.io/api/presentation/2.0/#rotation)  - although our note in the spreadsheet indicates there weren't any viewers at the time that supported this.

My understanding is that the rotations are (likely/potentially) document specific since the same canvas may hold more than one document written in different directions; but this is something to think about (and note somewhere?) if and when we get back to creating our own iiif manifests of PGP documents.

blms:
@rlskoeser I'm a bit stumped about what's going on with the migration unit tests here! It's failing on the migration `0038_document_dating_permissions` with the "Content Editor" group raising a `DoesNotExist` exception—I guess we haven't unit tested any migrations since then. Spent a couple of hours already trying to debug and I can't figure it out, was hoping maybe your expertise could help.

I did notice those groups were instantiated in a `common` app migration, so I tried updating the `common` dependencies to match a similar migration in another app (`0003_add_footnotes_permissions`), then tried using the most recent `common` migration in the dependencies, and neither worked…

It's odd because the footnotes migration tests are running just fine, but it seems like the migration test case is failing to run the `common` migrations when testing the `corpus` app's migrations. Have I overlooked something in my migrations?

rlskoeser:
I suspect this is the part you are missing:
https://github.com/Princeton-CDH/geniza/blob/main/geniza/common/migrations/0001_content_editor_group.py#L28-L34 

When you run migrations from start to finish the permissions aren't guaranteed to exist yet for any particular migration; if it works locally when you run new migrations but doesn't work in CI when it's running all of them then it's likely this problem.

blms:
@rlskoeser I thought that might be it as well, but I added that in both migrations and it's still not working! (It was already present in 0038.) It's odd—it's the `Group` that can't be found now, rather than the `Permission`.

https://github.com/Princeton-CDH/geniza/blob/55317505b0842129616673a6df00935fbae6b966/geniza/corpus/migrations/0038_document_dating_permissions.py
https://github.com/Princeton-CDH/geniza/blob/55317505b0842129616673a6df00935fbae6b966/geniza/corpus/migrations/0040_tag_merge_permissions.py

geniza/corpus/models.py
```diff
@@ -498,11 +498,8 @@ class Document(ModelIndexable, DocumentDateMixin):
     fragments = models.ManyToManyField(
         Fragment, through="TextBlock", related_name="documents"
     )
-    image_order_override = ArrayField(
-        models.URLField(), null=True, verbose_name="Image Order", blank=True
-    )
-    image_rotation_override = ArrayField(
-        models.IntegerField(), null=True, verbose_name="Image Rotation", blank=True
+    image_overrides = models.JSONField(
+        null=False, blank=True, default=dict, verbose_name="Image Order/Rotation"
```
rlskoeser, line 502: good call to combine these!

geniza/corpus/models.py
```diff
@@ -796,26 +793,31 @@ def iiif_images(self, filter_side=False, with_placeholders=False):
...
+                # include rotation: overridden value or 0 degrees
+                "rotation": int(override.get("rotation", 0)),
+            }
+            for canvas, override in sorted_overrides
+            if canvas in iiif_images
+        } or {}  # if condition is never met, instantiate empty dict (instead of set!)
+        ordered_images.update(
+            iiif_images  # add any remaining images after ordered ones
+        )
+        return ordered_images
```
rlskoeser, line 820: I'm having trouble following the logic here - I can't tell if it's the complexity or that the comments aren't quite enough

geniza/corpus/models.py
```diff
@@ -796,26 +793,31 @@ def iiif_images(self, filter_side=False, with_placeholders=False):
...
+        # if image_overrides not present, return list, in original order
+        if not self.image_overrides:
+            return iiif_images
 
-        return iiif_images
+        # order returned images according to override: first, sort overrides by "order"
+        sorted_overrides = sorted(
+            self.image_overrides.items(),
+            # get order if present; use ∞ as fallback to sort unordered to end of list
+            key=lambda item: item[1].get("order", float("inf")),
```
rlskoeser, line 804: this ∞ syntax is new to me, cool

rlskoeser (commented):
The refactor on the jsonfield to combine order and rotation is great.

Having trouble following the logic to apply the sorting overrides; wondering if it's more complicated than it needs to be.

geniza/corpus/models.py
blms: I see what you mean! I was having trouble following the comments myself. Couldn't think of a way to simplify the logic but hopefully the comments are clearer now.

blms:
@rlskoeser Hopefully I've improved the comments—now I'm just still completely stumped about the migration unit test failures. Everything _looks_ right, but it's still refusing to work! Wondering if it's some contingency I'm missing or if there's a bug in the migration unit test base class that I'm not seeing… 🤔

rlskoeser:
When I run these migration tests in isolation (e.g., `pytest -k TestDocumentImageOverrides`) they both pass. When I run all the tests in `test_corpus_migrations`, they fail. Wondering if there's a pytest ordering problem going on here? We're starting to get a lot of tests that want to run "last". I pushed a change to run these two second to last - that worked for me locally when running just the test corpus migrations; but it looks like it's still failing when running everything / when running on CI.  Maybe there's some order conflicts with other migration tests... Running locally I get a failure in a _different_ set of migrations!

My suggestions: look into the ordering a bit more, and also see if there's a pytest django fixture to reset the db so the order isn't as big of a problem. 

If you can't get ordering resolved, I don't think tests on migration logic should be a blocker. If it's easy, you could see if you could pull out and test the logic via a separate method/include, but otherwise I think you can mark these tests to be skipped with a comment explaining why and move forward.

rlskoeser:
@blms I tweaked the test ordering and the full test suite passes for me but still fails on github. Is that the case for you also? If so, maybe for now you could set it to skip when running on github actions with a comment about why it's disabled.

blms:
@rlskoeser Good call. I went ahead and did that since I'm experiencing the same behavior. However, now I'm seeing a completely different test fail in CI because it thinks the image override migration never happened… and only in one of the two GH Actions runs?! 😨

rlskoeser:
@blms how out of date is this from current develop branch? would it be worth updating to check if that improves/clarifies the situation with failing tests?

blms:
@rlskoeser I think that's the last of the unit tests problems solved! Should we say it's acceptable since the codecov misses are just the skipped migration tests (test code itself and associated migrations)? Or should I adjust codecov rules to skip it somehow?

rlskoeser (approved):

rlskoeser:
acceptable to me! congratulations & thanks for your persistence

Issue #1420, opened 2023-07-17
As a content editor, I want the location field for digital edition/translations to automatically populate from an existing edition/translation on the same source, so that I can save time manually re-entering it.


blms:
### testing notes (qa)

In the [QA admin site](https://test-geniza.cdh.princeton.edu/admin), navigate to a document.

- Add an Edition footnote with some value in the "location" field
- Add a new transcription with the transcription editor on the same source as the Edition footnote you created
- In the transcription editor, create some annotation blocks on the images
- [x] On the document, there should now be a new Digital Edition footnote with the same location as the Edition footnote
- [x] Try the same with a Translation footnote with a location, adding a translation on the same source in the translation editor. The newly created Digital Translation footnote should get the same location as the Translation footnote
- [x] Try this on a document with more than one Edition footnote on the same source, each having different locations. Using the transcription editor to create a new transcription on that source should _not_ populate the new Digital Edition footnote with a location, since it is ambiguous which Edition footnote should be used.

---

Requested by @mrustow over Slack. This should include:
- [x] A one-time migration to copy all `Footnote.location` values from edition to digital edition, and translation to digital translation, on the same Source
- [x] Logic to populate this automatically when users are creating digital content via annotations in the transcription/translation editor

kseniaryzhova:
@blms works as it should, thank you! Closing!

PR #1422, opened 2023-07-18
Populate new and existing digital footnotes' "location" field from corresponding non-digital footnote (#1420)

blms:
## In this PR

Per #1420:
- Migration to populate existing digital footnotes' `location` field from corresponding non-digital footnotes, if exactly one exists on the document + source pair
- Logic to do the same when a new digital footnote is created via annotation

## Notes

- As you will note in the code, there are cases where there are multiple corresponding non-digital footnotes on the same source+document, and in those cases we can't determine which location to use. My plan is to report on these, using a modified version of the migration as a shell script, after the migration on prod.


geniza/annotations/views.py
```diff
@@ -74,11 +76,24 @@ def parse_annotation_data(request):
         )
     except Footnote.DoesNotExist:
         source = Source.objects.get(pk=source_id)
+        # try to find a corresponding non-digital footnote
+        # (i.e. Translation for Digital Translation, Edition for Digital Edition)
```
rlskoeser, line 80: could you add one more line about the assumption here? am I understanding correctly that we're assuming if there's exactly one non-digital footnote for this source then the digital content is coming from the same location?

geniza/annotations/views.py
```diff
@@ -74,11 +76,24 @@ def parse_annotation_data(request):
...
+        corresponding_footnote = Footnote.objects.filter(
+            doc_relation__contains=corresponding_relation,
+            source__pk=source_id,
+            content_type=document_contenttype,
+            object_id=document_id,
```
rlskoeser, line 85: you might limit your query to footnotes that actually have a location set; from what I remember a lot of them do not, and handling at the query level might simplify the logic below

rlskoeser: oh, and I see you want one and only one result, otherwise it isn't useful...  I don't know if filtering on location makes that problematic, but another way to make sure you get one is to call `get` instead of filter and catch the too many results exception that Django will throw; might simplify the logic where you're setting the location below

rlskoeser (approved):

geniza/annotations/views.py
blms: Great ideas, thank you @rlskoeser!

Issue #1390, opened 2023-06-13
As a content editor, I want Person-Person relationship types visually sorted into their categories in the admin form, so that I can select them at a glance.


blms:
### testing notes (QA)

On the [QA site admin](https://test-geniza.cdh.princeton.edu/admin), create or edit a Person entry.

- [x] Under "Related People", the dropdown of types of relationships (the one labeled "Relation") should be sorted into categories such as business and property, extended family, etc.

### dev notes

optgroup with model choice field

kseniaryzhova:
@blms Noticed that I cannot add related documents from the person page, which seems like an issue if someone is working from their own personal archives and not going through the documents separately. Is this expected behavior or a bug?

![image](https://github.com/Princeton-CDH/geniza/assets/105385128/0a1498f0-1fe8-48a0-a30b-fc1be5e33cc0)

kseniaryzhova:
@blms Re: my comment on adding the additional people-to-people relationships (possibly the same as, not to be confused with). I can add the relationships, but I do not see a way to add the subheadings that would group these automatically. Otherwise, for the existing relationship, it does group automatically in the dropdown.

kseniaryzhova:
@blms ok, closing! Will update the other user stories once I've discussed with MR.

rlskoeser:
> The problem here is that we don't know what the reverse relationship should be, in other words if someone says "Parent" we don't have a system to have the other person say "Child". I wonder if it would make sense to have these reverse relationships created automatically—would love @rlskoeser's input here if this is sensible, since this is some advanced Django modeling! It might be a separate user story. (Something like "As a content admin I want the reverse relationships between people automatically populated, so I do not have to enter each relationship twice")

Auto-generating reverse relationships seems like a bad idea to me - more complicated and hard to make sure everything stays in sync.  We get the reverse lookup automatically if we want it, right? Can you leverage that somehow?

One thought is to add an optional reverse label for relationships that have reverse names (like parent/child); and/or a flag to mark relationships as symmetric, where label can be used both directions (e.g. spouse or friend).

Can you share some links to anything on the test site where I could see what this currently looks like, and help me think through by way of specific examples?

blms:
> Auto-generating reverse relationships seems like a bad idea to me - more complicated and hard to make sure everything stays in sync. We get the reverse lookup automatically if we want it, right? Can you leverage that somehow?

[...]

PR #1426, opened 2023-07-26
Categorize entries in person-person relations select (#1390)

blms:
## In this PR

Per #1390:
- Use `select` and `optgroup` to visually categorize (and alphabetize) person-person entity relations in Person form
- Per PGP-Performant meetings, these options are meant to be more or less set in stone, unless there is a significant need for a new/edited option, so the add/remove/change buttons are hidden from the inline


geniza/entities/admin.py
```diff
@@ -132,7 +172,13 @@ def get_form(self, request, obj=None, **kwargs):
...
         if obj:
             self.own_pk = obj.pk
-        return super().get_form(request, obj, **kwargs)
+
+        # remove add/change/delete buttons from "role" widget
+        form = super().get_form(request, obj, **kwargs)
+        form.base_fields["role"].widget.can_add_related = False
+        form.base_fields["role"].widget.can_change_related = False
+        form.base_fields["role"].widget.can_delete_related = False
+        return form
```
rlskoeser, line 181: would it be better to handle by user permissions, limiting which user groups have access to add/change/delete these?

rlskoeser (approved):
Seems reasonable, just wonder if it would be better to handle permissions at user/group level rather than overriding here.

geniza/entities/admin.py
blms: Good question, asked the team! I know originally their decision was to make this editable only from the admin list view, and not from the inlines, but I've asked if they'd prefer group-based permissions.

blms: Thanks for raising this, the team agrees group-based permissions are preferable!

blms:
@rlskoeser Going to go ahead and merge since all has been reviewed except the migration, and it's the same standard group permissions migration (including the `create_permissions` call!) that we've used in the past.

Issue #1381, opened 2023-05-24
As a content admin, I want a person with 10+ associated documents to automatically generate a front-end People Page, so that content editors do not have to manually make significant people pages live. 


richmanrachel:
### testing notes (QA)

Forgot to ever deliver this one for testing! It should continue to be true that a person record will only get a public person page if it has 10 related documents or the "person page" checkbox is checked on their edit page.

---

### dev notes
> can mitigate with help text and maybe also an admin filter to see who does/does not have a page so they could easily override those who wouldn't otherwise. + make the override field `list_editable`

---

**Is your feature request related to a problem? Please describe.**
We don't know how quickly/consistently data will be added related to people. So by building an automatic feature to make a person's page public if they cross a certain threshold, we will prevent artificial bottlenecks and get important people (who we may not even realize were important!) onto the front end.

**Describe the solution you'd like**
In addition to a manual boolean checkmark (or something equivalent) for turning on public pages for figures that may not reach the arbitrary 10 document threshold, we would like the association of people and documents to automatically trigger a public page publication.

**Additional context**
Perhaps we can write the code in such a way that if 10 ends up being too low of a number, we can change it? Or at least know how to find and access this feature?

kseniaryzhova:
@blms works as it should! Thank you, closing!

PR #1430, opened 2023-08-03
Automatically set Person.has_page = True if document count meets threshold (#1381)

blms:
## In this PR

Per #1381:
- On adding a Person-Document relation, if the Person does not have a Person page yet (i.e. their `has_page` property is `False`) and they have >= 10 documents associated, give them a page

## Notes

- I started this with an `m2m_changed` signal handler, but discovered more weirdness with m2m relations and Django: calling `.add` on either side of the relation programmatically fires `m2m_changed`, but does not call `PersonDocumentRelation.save`. On the other hand, saving using the form (which is likely the only way this will ever be done) does the exact opposite: it calls `PersonDocumentRelation.save`, but does not fire `m2m_changed`, the latter of which is especially surprising to me. 
  - Is it still worth including the signal handler? I removed it in d6975ac1fbce40a86f95c4cbc1afe6c351e95248 but maybe it's worth keeping around in case the programmatic calls ever happen.
- Since the Person feature still hasn't been released on prod and we aren't doing any kind of bulk ingest of them, I didn't bother with a migration

### DIFF
--- a/geniza/entities/apps.py
+++ b/geniza/entities/apps.py
@@ -1,5 +1,16 @@
 from django.apps import AppConfig
+from django.db.models.signals import m2m_changed
 
 
 class EntitiesConfig(AppConfig):
     name = "geniza.entities"
+
+    def ready(self):
+        # attach m2m_changed signal for person-document relations
+        from geniza.entities.models import PersonSignalHandlers
+
+        m2m_changed.connect(
+            PersonSignalHandlers.person_document_relation_changed,
+            sender="entities.PersonDocumentRelation",
+        )
+        return super().ready()
--- a/geniza/entities/apps.py
+++ b/geniza/entities/apps.py
@@ -1,16 +1,5 @@
 from django.apps import AppConfig
-from django.db.models.signals import m2m_changed
 
 
 class EntitiesConfig(AppConfig):
     name = "geniza.entities"
-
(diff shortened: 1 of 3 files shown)


rlskoeser:
@blms I meant to comment on this and forgot! (IDK if you actually tagged me or requested review yet?)

I would not actually store the `has_page` property unless it is manually checked. In the view (or, probably better, in the appropriate queryset) you filter on the minimum number of related documents or if the property is set. If a person is not found with that filter, then the page 404s and/or they aren't included in the list.

blms:
Great point, thank you @rlskoeser! This is really a frontend concern, and should be handled separately, as long as content editors know that the `has_page` property has no effect on people with more than 10 documents. Which makes sense because the original purpose of that property was to override it for important people with fewer than 10 documents. I'm going to go ahead and close this PR and revisit this when I start implementing the views and templates.

rlskoeser:
Agreed - I think you can mitigate with help text and maybe also an admin filter to see who does/does not have a page so they could easily override those who wouldn't otherwise. (will you / did you make the override field list editable?)

blms:
> maybe also an admin filter to see who does/does not have a page so they could easily override those who wouldn't otherwise. will you / did you make the override field list editable?

Good ideas both, I did not!

Issue #1438, opened 2023-08-10
Annotations on the document detail page do not respect reordering


blms:
### testing notes (qa)

On the [QA site](https://test-geniza.cdh.princeton.edu/):

- [x] Edit a document transcription or translation, and reorder some annotations by dragging and dropping text blocks in the right column. You should see a small confirmation message appear at the bottom of the screen indicating that the reorder was successful.
- [x] Visit that document's entry on the public site and hard refresh. The new order of annotations in that particular transcription/translation should persist.

---

The `content_html` property of a footnote is missing the `order_by` step.

kseniaryzhova:
@blms Works as it should, closing, thank you!

PR #1439, opened 2023-08-10
Respect annotation reordering in document detail page (#1438)

blms:
## In this PR

Per #1438:
- Add `order_by("content__schema:position", "created")` to the annotation set in `Footnote.content_html`, so that reordered annotations will appear in the correct order on the document detail page

## Questions

- Since this is a cached property, I'm wondering about cache busting when annotations are reordered. It seems to work for me locally when I hard refresh the doc detail page, but wondering if that's always the case / if there's a way to do it automatically from the backend?

### DIFF
--- a/geniza/footnotes/models.py
+++ b/geniza/footnotes/models.py
@@ -606,6 +606,7 @@ class Footnote(TrackChangesModel):
         # keyed on canvas uri
         # handle multiple annotations on the same canvas
         html_content = defaultdict(list)
+        # order by optional position property (set by manual reorder in editor), then date
         for a in self.annotation_set.all().order_by(
             "content__schema:position", "created"
         ):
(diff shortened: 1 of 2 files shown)


geniza/footnotes/models.py
```diff
@@ -606,7 +606,9 @@ def content_html(self):
         # keyed on canvas uri
         # handle multiple annotations on the same canvas
         html_content = defaultdict(list)
-        for a in self.annotation_set.all():
+        for a in self.annotation_set.all().order_by(
```
rlskoeser, line 609: probably worth a one-line comment here documenting what this is for

I checked, default ordering is by `created`; makes sense we wouldn't want to rely on an optional/custom parameter for sort

rlskoeser (approved):
Update looks good to me, makes sense.

The cached property only applies to the object instance, to avoid recomputing a dynamic property multiple times for generating a page (possibly not strictly needed in this case, unless we have logic that depends on whether there is content). When you reload the page you get a new instance of the object, so this caching isn't relevant. I suspect that's why it worked for you on a reload.

Issue #1429, opened 2023-08-02
As a content admin, I want both dates on document and inferred dates to merge when I merge duplicate PGPIDS so no data is lost when cleaning up duplicates. If there are two different dates on documents for the same PGPID, I want there to be an error message drawing my attention to the issue so I can choose the correct date or otherwise record the discrepancy.


kseniaryzhova:
**Describe the solution you'd like**
When duplicate PGP entries have dates entered in the inferred or date on document fields, I want this information preserved when I merge documents. Our model allows for multiple inferred dates but ostensibly there should only be one date on document. If there are two different dates on document for the same document entry, an error message should occur upon merging, prompting the content admin to resolve the issue and pick the correct date. This error message will likely only apply to a handful of cases, it's more for data cleanup.

kseniaryzhova:
@blms works as it should, closing!  Thank you!

PR #1444, opened 2023-08-16
Resolve document and inferred dates in document merge (#1429)

blms:
## In this PR

Per #1429:
- In document merge, handle dates
  - Raise ValidationError on conflicting standard/original document dates
  - If not in conflict, ensure no standard/original document dates are lost during merge
  - Add all inferred datings from merged documents

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -1296,13 +1296,30 @@ class Document(ModelIndexable, DocumentDateMixin):
             # handle document dates validation before making any changes;
             # mismatch should result in exception (caught by DocumentMerge.form_valid)
             if (
-                doc.doc_date_standard
-                and self.doc_date_standard
-                and self.doc_date_standard != doc.doc_date_standard
-            ) or (
-                doc.doc_date_original
-                and self.doc_date_original
-                and self.doc_date_original != doc.doc_date_original
+                (
+                    # both documents have standard dates, and they don't match
+                    doc.doc_date_standard
+                    and self.doc_date_standard
+                    and self.doc_date_standard != doc.doc_date_standard
+                )
+                or (
+                    # both documents have original dates, and they don't match
+                    doc.doc_date_original
+                    and self.doc_date_original
+                    and self.doc_date_original != doc.doc_date_original
+                )
+                or (
+                    # other document has original, this doc has standard, and they don't match
+                    doc.doc_date_original
+                    and self.doc_date_standard
+                    and doc.standardize_date() != self.doc_date_standard
+                )
+                or (
+                    # other document has standard, this doc has original, and they don't match
+                    doc.doc_date_standard
+                    and self.doc_date_original
+                    and self.standardize_date() != doc.doc_date_standard
+                )
             ):
                 raise ValidationError(
                     "Merged documents must not contain conflicting dates; resolve before merge"
@@ -1316,6 +1333,7 @@ class Document(ModelIndexable, DocumentDateMixin):
                 self.doc_date_standard = doc.doc_date_standard
             if doc.doc_date_original:
                 self.doc_date_original = doc.doc_date_original
+                self.doc_date_calendar = doc.doc_date_calendar
 
             # add inferred datings (conflicts or duplicates are post-merge
             # data cleanup tasks)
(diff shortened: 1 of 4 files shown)


geniza/corpus/models.py
```diff
@@ -1293,8 +1293,35 @@ def merge_with(self, merge_docs, rationale, user=None):
...
+                and self.doc_date_original != doc.doc_date_original
+            ):
+                raise ValidationError(
+                    "Merged documents must not contain conflicting dates; resolve before merge"
+                )
+
             # add any tags from merge document tags to primary doc
             self.tags.add(*doc.tags.names())
+
+            # if not in conflict, migrate dates to result document
```
rlskoeser, line 1314: this is either copying over dates where none exist or copying over exact duplicates? that makes sense, might be worth slightly expanding the comment

geniza/corpus/models.py
```diff
@@ -1293,8 +1293,35 @@ def merge_with(self, merge_docs, rationale, user=None):
...
+                doc.doc_date_standard
+                and self.doc_date_standard
+                and self.doc_date_standard != doc.doc_date_standard
+            ) or (
+                doc.doc_date_original
+                and self.doc_date_original
+                and self.doc_date_original != doc.doc_date_original
+            ):
+                raise ValidationError(
+                    "Merged documents must not contain conflicting dates; resolve before merge"
```
rlskoeser, line 1308: Is it possible for one document to have a standard date and the other to have an original date that do not match each other? (seems unlikely, but possible?) Should that be considered conflicting and merit review?

geniza/corpus/tests/test_corpus_models.py
```diff
@@ -1644,6 +1645,65 @@ def test_document_merge_with_log_entries(document, join):
...
+    # clone join twice for additional merges
+    join_clone = Document.objects.get(pk=join.pk)
+    join_clone.pk = None
+    join_clone.save()
+    join_clone_2 = Document.objects.get(pk=join.pk)
+    join_clone_2.pk = None
+    join_clone_2.save()
+
+    # create some datings; doesn't matter that they are identical, as cleaning
+    # up post-merge dupes is a data cleanup task. unit test will make sure that
```
rlskoeser, line 1660: Does this mean manual data cleanup?  Shouldn't be that hard to detect exact duplicates by implementing an `__eq__` method on the `Dating` object like we did on `Foonotes` (IIRC), but you all would have to determine how likely or valuable that is.

geniza/corpus/views.py
```diff
@@ -691,7 +692,17 @@
...
+
+        try:
+            primary_doc.merge_with(secondary_docs, rationale, user=user)
+        except ValidationError as err:
+            # in case the merge resulted in an error, display error to user
+            messages.error(self.request, err.message)
+            # redirect to this form page instead of one of the documents
+            return HttpResponseRedirect(
+                "%s?ids=%s"
+                % (reverse("admin:document-merge"), self.request.GET.get("ids", "")),
```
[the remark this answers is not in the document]
rlskoeser: this warning does seem like a reasonable thing to flag, can you address?

rlskoeser (approved):
Seems like a valuable addition to the merge functionality. (Hope the problem wasn't discovered by data loss!).

Could we add a note or checklist somewhere that would prompt checking the merge function when relevant models are updated that might have an impact on the merge logic?

geniza/corpus/views.py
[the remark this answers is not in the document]
blms: My thinking here was that the scheme, hostname, and path of the URL would always be handled by `reverse`, while the only part that can come from `self.request.GET` comes after the query string portion. But maybe that's still unsafe? Wasn't sure how else to refresh the page but with the error message.

geniza/corpus/tests/test_corpus_models.py
blms: Team agree this is fine as a manual data cleanup task, but can revisit if it starts to become a problem—thanks for the solution!

geniza/corpus/models.py
blms: Good call, I think it makes sense to flag these to users even if it technically won't result in any data loss and is rare enough.

rlskoeser (approved):

Issue #1431, opened 2023-08-07
Transcription and translation may become misaligned when resizing window


blms:
### testing notes (qa)

On the [QA public site](https://test-geniza.cdh.princeton.edu/):
- [x] Visit a document with both transcription and translation content. Open both of these side by side. Ordered lines should be aligned side-by-side, and resizing the window should retain the alignment, despite some lines wrapping differently.

---

Workaround is to collapse and re-expand one of the open panels.

https://github.com/Princeton-CDH/geniza/assets/7234006/d2449b0d-191d-4d2a-85a6-3a585f2fb057

kseniaryzhova:
@blms works as it should! Closing, thank you!

Issue #1409, opened 2023-06-27
Alignment between Arabic transcriptions and English translations is slightly off


blms:
### testing notes (qa)

On the [QA public site](https://test-geniza.cdh.princeton.edu/):
- [x] Visit a handful (5-10) of documents with both transcription and translation content, including both JA and Arabic documents. Open both the transcription and the translation. Numbered lines in the transcription and translation should be aligned side-by-side.

(Note that in some cases, such as PGPID 1018, one or the other side may contain content that is not matching, such as a paragraph on one side and numbered lines on the other. Alignment can only occur when both sides contain matching numbered lines.)

---

### testing notes (QA)

On the [QA public site](https://test-geniza.cdh.princeton.edu/):
- [x] Find a few documents with a transcription in Arabic and translation in English, such as PGPID 38281. Transcription and translation lines should be aligned properly.
- [x] Try the same on the QA site admin.

---

### dev notes

![Screenshot 2023-06-27 at 11 25 01 AM](https://github.com/Princeton-CDH/geniza/assets/7234006/008b66f3-d5e7-4b74-9005-1a30f688dadc)

It may not be possible to fix this without affecting the other languages, but worth investigating.

Example doc: https://geniza.princeton.edu/en/documents/38281/
Example doc on QA: https://test-geniza.cdh.princeton.edu/en/documents/38281/

kseniaryzhova:
@blms alignment was great for the admin site but not public for some reason, same document, 38281.
![image](https://github.com/Princeton-CDH/geniza/assets/105385128/e7451ae3-43e0-466e-83de-1ece6739a618)
![image](https://github.com/Princeton-CDH/geniza/assets/105385128/cff6633d-20d9-428f-b072-ea6ae96c93ee)

kseniaryzhova:
@blms It works for me, I think I should have @mrustow test on her Mac since there were differences before.

blms:
@mrustow before we reopen, can you give this one a try on Safari [on QA](https://test-geniza.cdh.princeton.edu/en/documents/38281/) after a hard refresh? (Cmd+Opt+R) Here's what I'm seeing there now:

![Screenshot 2023-08-16 at 11 34 37 AM](https://github.com/Princeton-CDH/geniza/assets/7234006/56c713bc-63f7-4caa-95a5-aca1429ce6be)

mrustow:
Yeah it's sooooo close, like a mm off — 

[...]

PR #1445, opened 2023-08-17
Various alignment bugfixes (#1409, #1431)

blms:
## In this PR

Per #1409:
- Fix issue where alignment was only applied to the first transcription/translation block, and only the first OL within those
  - Also apply this fix to alignment removal
- Fix issue where `min-height` was sometimes causing misalignment due to lack of `max-height`
- Fix issue where initial OL alignment for translation was exactly 1 pixel off
- Fix django admin CSS conflict issues

Per #1431:
- Apply alignment removal and re-alignment on window resize (if on desktop and both transcription and translation are open)

Other:
- I noticed the "li period hider" element was sometimes misaligned in various configurations, so I adjusted it in both the public and admin CSS

### DIFF
--- a/sitemedia/js/controllers/ittpanel_controller.js
+++ b/sitemedia/js/controllers/ittpanel_controller.js
@@ -59,29 +59,34 @@ export default class extends Controller {
                 // when transcription and translation are both opened, align their contents line-by-line
                 this.alignLines();
             } else {
-                // when one of those two toggles is closed, remove data-lines from each line (alignment no longer needed)
+                // when one of those two toggles is closed, remove data-lines from each line
+                // (alignment no longer needed)
                 if (this.hasTranscriptionTarget) {
-                    this.transcriptionTarget
-                        .querySelectorAll("li")
-                        .forEach((li) => {
+                    this.transcriptionTargets.forEach((target) => {
+                        target.querySelectorAll("li").forEach((li) => {
                             if (li) li.removeAttribute("data-lines");
                         });
+                    });
                 }
                 if (this.hasTranslationTarget) {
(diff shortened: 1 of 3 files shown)


sitemedia/js/controllers/ittpanel_controller.js
```diff
@@ -106,39 +87,86 @@ export default class extends Controller {
...
+                    }
+                }
+            }
+        }
+    }
+
+    removeAlignment() {
+        // remove alignment
+        // first remove data-lines from each line
+        if (this.hasTranscriptionTarget) {
```
rlskoeser, line 136: seems like some redundancy, wondering if there's a nice way to iterate over all like-elements; but that's a nice-to-have, not essential

sitemedia/js/controllers/ittpanel_controller.js
```diff
@@ -106,39 +87,86 @@ export default class extends Controller {
...
-                // set data-lines attribute on each li according to which is longer
-                edLines[i].setAttribute("data-lines", maxLineCount);
-                trLines[i].setAttribute("data-lines", maxLineCount);
-            }
+    handleResizeAlign() {
+        // on resize, remove alignment and realign using new heights
+        if (this.isDesktop() && this.transcriptionAndTranslationOpen()) {
+            this.removeAlignment();
+            this.alignLines();
         }
```
rlskoeser, line 170: nice

rlskoeser (approved):
Logic updates look good to me. I like the refactor, makes the resize handler really clean.

(Wow, 1 pixel off?! Someone has eagle eyes, glad you were able to resolve it.)

Issue #1446, opened 2023-08-21
Digital translation footnote in scholarship records behaving incorrectly, excluding other footnotes on source


blms:
### testing notes (qa)

- [x] On the QA public site, [PGPID 6955's scholarship records page](https://test-geniza.cdh.princeton.edu/en/documents/6955/scholarship/) should no longer display the below behavior, and should instead say "includes Digital Translation, Edition, Translation"
- [x] Check a handful of other documents with many scholarship records, including digital transcriptions/translations. The scholarship records page should never say "for X see" without further information below the word "see", and it should always list all document relations for a particular source, even if they are split across multiple footnotes.

---

On PGPID 6955, this record has no location or URL, yet says "for digital translation see" instead of "includes digital translation". Also, it is causing other footnotes on the source to be excluded entirely from the scholarship records.

![image](https://github.com/Princeton-CDH/geniza/assets/7234006/580adf58-4694-4b56-b49f-b734ec6cd269)

In admin:

![Screenshot 2023-08-21 at 10 10 45 AM](https://github.com/Princeton-CDH/geniza/assets/7234006/a7368c25-dd81-46f4-a639-4686ca18faf2)

kseniaryzhova:
@blms Works well, thank you! Closing!

PR #1447, opened 2023-08-21
Fix display of multiple footnotes on same source (#1446)

blms:
## In this PR

per #1446:
- Use template filter to check if any footnote on a source has a location or url
- Use template filter to collect doc relations for ALL footnotes on the source, not just displaying the first


rlskoeser (approved):
This all makes sense to me, and template tag seems like a reasonable solution in this context. (I presume that since this has been regrouped, we have a list and not a footnote queryset so we can't implement these as queryset methods.)

Issue #1442, opened 2023-08-16
As a content editor, I want a "no language" option when entering source languages (with help text) for unpublished transcriptions because the language will automatically be determined by the document languages already present on the doc detail pages.


kseniaryzhova:
### testing notes (QA)

On the [QA site admin](https://test-geniza.cdh.princeton.edu/admin), navigate to the Sources section. Add or edit a Source.
- [x] There should now be an "Unspecified (unpublished transcriptions)" option in the Source languages dropdown, as agreed upon in Slack.
- [x] Let me know how I should pass along the below links to sets of sources with possibly wrong languages, once the production release with this Unspecified option is completed. (New cleanup chore issue assigned to Ksenia? How do you all keep track of data cleanup tasks?)

---

**Describe the solution you'd like**
When selecting source languages for unpublished transcriptions, I would like a No Language option coupled with help text that specifies "The No Language option should be selected only in the case of unpublished transcriptions because the language of the transcription is already marked on the document detail page." (or something to that effect)

**Describe alternatives you've considered**
We've considered making language selection optional and not obligatory but we do need those source languages to help track translation languages and to help scholars navigate secondary source languages. We also considered selecting transcription languages just as any other source, but then the bibliographic display is confusing in that it displays the language of transcription in parentheses (MR does not like).

blms:
(Note: Unspecified language option isn't available yet, but posting this here in preparation for that.)

In order to help you all maintain consistency across the DB, I've come up with a handful of lists for data cleanup. We can make a new chore issue if that's helpful. I imagine most of these that are unpublished will need the new "Unspecified" language option, but don't want to assume that and programmatically reassign all of them if that's incorrect for any.

- [Sources with no language selected](https://geniza.princeton.edu/admin/footnotes/source/?all=&languages__isnull=True) (from before the constraint was added; note that all of these happen to be unpublished)
- [Sources marked as Judaeo-Arabic](https://geniza.princeton.edu/admin/footnotes/source/?all=&languages__id__exact=6) (I assume all of these are wrong?)
- [Unpublished sources marked as Hebrew](https://geniza.princeton.edu/admin/footnotes/source/?all=&languages__id__exact=3&source_type__id__exact=3)
- [Unpublished sources marked as Arabic](https://geniza.princeton.edu/admin/footnotes/source/?all=&languages__id__exact=8&source_type__id__exact=3)
[...]

Issue #1375, opened 2023-05-10
As a content editor, I want clear help text when adding a source to explain how to select the source language, so that it is done consistently for translations and transcriptions.


kseniaryzhova:
### testing notes (QA) round two

On the [QA site admin](https://test-geniza.cdh.princeton.edu/admin), navigate to the Sources section. Add or edit a Source.
- [x] Now, under "Languages", the help text should refer to "the Unspecified language option" for unpublished transcriptions, instead of instructing users to choose the language of the transcription text.

### testing notes (QA)

On the [QA site admin](https://test-geniza.cdh.princeton.edu/admin), navigate to the Sources section. Add or edit a Source.

- [x] Under "Source type", the help text should clearly communicate the need for different sources for unpublished transcriptions and translations.
- [x] Under "Languages", the help text should clearly communicate that only sources consisting entirely of transcription content should include transcription language.

---

### dev notes

Revised solution: Help text when adding a source to make it clear that unpublished translations and unpublished editions are two separate sources, and that the unpublished translation sources need to be set with the language of the translation.

---

**Is your feature request related to a problem? Please describe.**
Right now there is an undifferentiated unpublished source option, which means editions and translations are being tagged with this option. It's a requirement to select a source language, but for editions, we track the source language through the language tagging option on the document itself. For translations, however, we do want to know the language of scholarship.

**Describe the solution you'd like**
We want two separate source type options: unpublished translations and unpublished editions, to replaced the current "unpublished" option. For unpublished editions, there is no need to select a source language and it should not display in the footnote. However, for unpublished translations, the language of scholarship should be selected and should display in a footnote.

kseniaryzhova:
@blms so even if the translation and edition are on the same Google Doc, they would still be two different sources (with the same author, etc)?

blms:
@kseniaryzhova Yes, that’s the idea! Though, I think that would be the case either way; the difference is just that in my view, they should both have the “unpublished” source type, rather than one having “unpublished edition” and one having “unpublished translation”.

mrustow:
[...]

PR #1448, opened 2023-08-22
Add "unspecified" source language option (#1442)

blms:
## In this PR

Per #1442
- Add "unspecified" source language options for unpublished transcriptions
- Exclude from string representation/citation of sources

## Notes

This is an extremely minor PR, I'm only requesting review to make sure you sign off on the concept. The rationale is essentially that the team wants the source language requirement to remain in place to prevent editors from simply forgetting to choose a language, but also allow an option that accurately reflects the content of the source in the case of unpublished transcriptions (i.e. there is no language besides the document language).

More info here: https://github.com/Princeton-CDH/geniza/issues/1442


rlskoeser (commented):
Thanks for flagging this for my attention. 

I'm struggling a little with the concept; or maybe it's the language that bothers me. Unspecified language seems a little inaccurate to me - is it actually the document language? What happens if the document has multiple languages?  From what I remember, we model languages slightly differently for sources and documents; is there any incompatibility here that matters?  Is this a time saving measure to avoid redundant data entry?

This part of the issue caught my attention:
> We also considered selecting transcription languages just as any other source, but then the bibliographic display is confusing in that it displays the language of transcription in parentheses (MR does not like).

Is there a way to fix this on the citation logic instead? Omit languages for unpublished documents or similar?

I don't object to this; I just want to make sure we're not introducing a data discrepancy that may be problematic down the line (e.g., data exports/data analysis).  I think unpublished sources are really a different kind source anyway - they're not necessarily/typically a single unified source but more of a convenience container for related unpublished work.  

Would it be possible to make language optional for unpublished transcriptions only? Would that solve the problem in a simpler way?

blms:
@rlskoeser Thank you for helping think through this! It's a tricky one.

> Is there a way to fix this on the citation logic instead? Omit languages for unpublished documents or similar?

We can't omit languages from all unpublished document citations since some will be translations, and those will still appear in the translations dropdown and need to have a language visible there, so that the correct translation is possible to pick.

> Would it be possible to make language optional for unpublished transcriptions only? Would that solve the problem in a simpler way?

That was indeed what the team had originally come up with as a solution, but doing so would require changing the data model so that unpublished transcriptions are considered a separate "source type" from unpublished translations, because source language should still be required for unpublished translations so that they will appear in the translations dropdown. We discussed that in this issue: https://github.com/Princeton-CDH/geniza/issues/1375#issuecomment-1634682628 and Marina and I agreed that it's semantically too different from the other source types, as it collapses the distinction between source type and footnote doc relation for only that specific case. But I'm definitely willing to revisit if you disagree!

> Unspecified language seems a little inaccurate to me - is it actually the document language? 

Yes, so maybe it makes the most sense to rename it to something like "language of document", though that may be confusing as users might select it for other kinds of records without understanding its purpose.

The original solution was to just use help text to instruct users to select the language of the document (not the language of the scholarship) ONLY for sources that consist entirely of transcription content. But Ksenia brought up that this was confusing in https://github.com/Princeton-CDH/geniza/issues/1375#issuecomment-1676514267, and we came up with this alternate solution in the meeting.

> What happens if the document has multiple languages? From what I remember, we model languages slightly differently for sources and documents; is there any incompatibility here that matters?

These compatibility issues shouldn't be an issue, since these languages do not need to appear in citations or anywhere else on the public site. Citations only reflect the language of the scholarship in a source, not transcription content. In fact, source language is completely discarded in this case; it's just a question of how to best represent that to data entry folks.

> Is this a time saving measure to avoid redundant data entry?

Not exactly. It's more of a question of balancing the strictness of the requirement for a "language of scholarship" to be selected for the majority of sources, and the need to relax or reconfigure that restriction for this specific case where there is no scholarship in the source besides the document's contents reproduced as text. Redundant data entry is fine if it gets us a solution, as long as people are not also tagging the document language on published scholarship that happens to include transcriptions, or on translations. (This is what has been happening so far!)

rlskoeser:
Wow, complicated and confusing! You all have obviously thought about it more deeply than I have and spent more time with you. I agree with you and Marina that distinguishing between types of unpublished work doesn't make so much sense here, and totally agree with the importance of tracking language for translations.

If this is entirely an internal language terminology issue, then I trust you all to agree and decide on terminology that is meaningful and not likely to be confusing to the people doing data entry. 

The one place I can think of where source language _does_ appear is in the data exports, which do include language (I just checked - see [sources.csv on flat github](https://flatgithub.com/princetongenizalab/pgp-metadata?filename=data%2Fsources.csv&sha=595509a55fdc3c9cd0744307eaa9cb3010ce8ab6) ). I think including "not specified" there would be ambiguous and confusing. The options that occur to me are:
1. if language is "not specified" in the db, omit language in the data export 
2. if language is "not specified" in the db, pull languages from the associated documents (more accurate but potentially slow; more effort for you to implement)

First option seems reasonable to me for the short term at least (until we know if it's important to include); related languages should be available in the other exports.  (Although I think we do have a lingering problem that may not be tracked anywhere, which is that the source data export doesn't currently include an identifier or key that allows joining the related data exports.)

blms:
That sounds great to me, thank you @rlskoeser! I think you're right that the terminology is confusing. I'll pose the question on Slack since what we discussed in the meeting was "No language", which I also think is confusing (since it's actually inaccurate!). 

First solution for the exports make sense to me, at least for now as you say.

> Although I think we do have a lingering problem that may not be tracked anywhere, which is that the source data export doesn't currently include an identifier or key that allows joining the related data exports.

By related data exports, do you mean exports for related objects? Like a csv of exported footnotes can't easily be joined with a csv of exported sources? This is something I hadn't thought of at all (and I've definitely never tried joining CSVs before in this way!)

rlskoeser:
> By related data exports, do you mean exports for related objects? Like a csv of exported footnotes can't easily be joined with a csv of exported sources? This is something I hadn't thought of at all (and I've definitely never tried joining CSVs before in this way!)

Yes, that's exactly the problem - footnotes have a `source` field that includes the source citation, but some analysis of this data would be better done by joining the source.csv file, which has more granular-source level information. However, the source.csv doesn't include a corresponding citation field — and if even it did, I'm not sure it would be guaranteed unique.  Ideally we would have a source id or URI (semantic, not based on the db id) that could be added to both data exports for join purposes.

If this is something that could be in scope for you to work on, I'd love to get in your queue and think together about how we could define and generate a good, durable source key or uri.  (Might be easiest to add a slug field in the db, and prepopulate; I can share what I've done with this on other projects).

blms:
@rlskoeser I had forgotten to update this one! The team, specifically the two editors who are most frequently involved in editing sources, preferred "Unspecified (unpublished transcriptions)" to any other options given ("no language", "document language", "unpublished transcription content only" etc), and did not suggest any additional options, so I think we're good to go with that nomenclature.

I also went ahead and pushed the Source export language exclusion.

rlskoeser (approved):

Issue #887, opened 2022-05-25
Input date not always populating


mrustow:
### testing notes (QA)

On the [QA admin site](https://test-geniza.cdh.princeton.edu/admin), create a new document, and click "save".

- [x] On the [QA public site search](https://test-geniza.cdh.princeton.edu/en/documents/), search for that document. In the search result, its input date should read 2023 rather than "unknown".

---

Hello party people ;) On the public site, I just sorted by input date (great feature! thanks!) earliest to latest. The earliest ones were date "unknown," and very high PGPIDs, so high that they must have been input this month. After some digging it looks like the problem began May 4 (PGPID 35277), and its latest manifestation was today (PGPID 35363) — I got those input dates from the admin site, meaning they're in the admin site somewhere, just not displaying on the public site. Correction: not displaying in the search results/list view, but they are displaying in site view. 🤷🏻‍♀️

<img width="495" alt="image" src="https://user-images.githubusercontent.com/73319225/170370732-b1e06b66-5e4a-4db8-8bb7-711727827dba.png">

kseniaryzhova:
@blms this is still an issue, for example PGPID 36488 contains in the admin history that Alan added it Oct. 31 2022, but displays as unknown in the front end.

rlskoeser:
This is an indexing problem. I wonder if it's possible the initial indexing happens before the log entry documenting the input date is available?

I checked the index data for this one and input date was set correctly, so I indexed it and now the date appears on the search results.

I ran a solr queries on the production Solr for records with no input date set (`NOT input_year_i:[* TO *]`) and found 14 records; I went ahead and reindexed them.

This resolves the display problem, but won't solve the underlying bug (if there is one). It would be good to see if we can duplicate it on the test site — my first thought is try creating a record but not making any changes after the first save.  If that results in this behavior, then we can work on a fix — if there are no log entries then maybe we can fallback to the `created` attribute on the document (that should work for all new records; it will only be wrong for imported records, but those _should_ all have log entries.)

kseniaryzhova:
@blms works as it should! Thank you!

PR #1450, opened 2023-08-23
Use Document.created as fallback input date in index (#887)

blms:
## In this PR

Per #887:
- Use `Document.created` as a fallback when indexing input date, if the document's initial log entry is not yet created

## Notes

- Using the query from your comment in the issue, I found 364 documents on the public site that were missing input dates in the index, so it's definitely a common issue. (I also reindexed those documents on the public site so they are all fixed now.) But I think you are right that this will resolve it for all new records, which seems to be the only case where it happens anyway.

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -1207,7 +1207,8 @@ class Document(ModelIndexable, DocumentDateMixin):
         try:
             last_log_entry = list(self.log_entries.all())[-1]
         except IndexError:
-            # should only occur in unit tests, not real data
+            # occurs in unit tests, and sometimes when new documents are indexed before
+            # log entry is populated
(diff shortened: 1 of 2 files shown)


geniza/corpus/models.py
```diff
@@ -1207,7 +1207,8 @@ def index_data(self):
         try:
             last_log_entry = list(self.log_entries.all())[-1]
         except IndexError:
-            # should only occur in unit tests, not real data
+            # occurs in unit tests, and sometimes when new documents are indexed before
+            # log entry is populated
```
rlskoeser, line 1211: 👍

rlskoeser (approved):
great solution, thanks for finding and addressing

PR #1451, opened 2023-08-29
Index inferred dates, add admin filter to exclude inferred dates (#1371)

blms:
## In this PR

Per #1371:
- Index a document's full date range, including inferred datings, in solr, as a separate date range entry
- Add a filter "By date type" that allows one to use "document dates only" when filtering by dates using the other date range filters

## Notes

- Just as an FYI: `Document.dating_range` and `Document.solr_dating_range` are separated for when we will later need to sort by this range on the public frontend, as `dating_range` produces `PartialDate` objects from which we can generate `numeric_format` for sorting
- **The tricky issue**: I've managed to index this range in solr, but I'm now trying to figure out how to let admins filter on it in the django admin. Right now, the date filters are text input list filters, and they can only have one user input, a string value:
  https://github.com/Princeton-CDH/geniza/blob/15f1935e666caf3d19e29cd80ba6bea847865967/geniza/corpus/admin.py#L262-L268
  The team wants a way to just check a checkbox like "include inferred datings" to switch from the basic date filter to the one including inferred datings, but I'm not sure how to approach that within the Django admin toolkit. A new custom boolean filter that modifies the other two filters? A checkbox inserted into the date filter template manually? Something else?
  I know I'll ultimately need to modify `DateListFilter.queryset` to swap out `filter(document_date_dr=date_filter)` for `filter(document_dating_dr=date_filter)` on the SQS, just trying to think through the most sensible way to get there.


blms:
@rlskoeser I'm also at a complete loss about the last two failing tests. I can't replicate it in my local testing environment, and I don't see any way in the code that those two values could get swapped. Can you spot something I'm missing? I'm sure I'll feel foolish, but right now I'm coming up with nothing…

The only thing I can think of is that we're somehow hitting a datetime-related bug or difference between python 3.9.13 (version on github action) and 3.9.17 (my local), but that doesn't quite seem to explain the behavior and I can't find anything like that in the changelog.

rlskoeser:
@blms I'll try to find some time to pull this locally and investigate

blms:
@rlskoeser I think the root of the problem is that for some reason, the 3-digit years are getting leading zeros appended to the beginning of them in CI but not in my local environment. Therefore, the date comparisons were behaving differently because the strings themselves were different.

I was able to fix this by using `int(numeric_format)` for each date comparison, instead of comparing the strings. It does look like integer casting should work, based on how that numeric format is represented:
https://github.com/Princeton-CDH/geniza/blob/7ea098ac45078573d5a3af93bd4581a4b9007e66/geniza/corpus/dates.py#L56
I also allowed for either (with or without leading zeros) in the unit tests checking for the solr dates; I tested with solr and either works fine.

Hopefully this seems like a sensible solution!

---

I also went ahead and implemented the simplest version I could come up with for the "exclude inferred" filter, which is a filter that doesn't affect the queryset on its own but adds a request parameter that alters the behavior of the other date filters. I know it's a bit of a strange approach, so let me know if you'd prefer something else.

geniza/corpus/admin.py
```diff
@@ -283,11 +283,13 @@ def queryset(self, request, queryset):
 
             # use Solr to take advantage of processed date range fields
```
rlskoeser, line 284: I find it hard to read the or logic inside the queryset filter, can you set the field before you create and filter the queryset instead?

```suggestion
            # use Solr to take advantage of processed date range fields
            date_field = 'document_dated_dr' if request.GET.get("exclude_inferred", None) == "true" else 'document_dating_dr'
```
then you can set date filter options like this `date_filter_opts = {date_field: date_filter}` and pass them to the filter as kwargs `.filter(**date_filter_opts)`


Or maybe I'm misreading this because of the diff. Simpler might be to put the if/else at the outer level and just initialize the `sqs` two different ways

geniza/corpus/models.py
```diff
@@ -925,6 +925,65 @@ def title(self):
...
+    def dating_range(self):
+        """
+        Return the start and end of the document's possible date range, as PartialDate objects,
+        including standardized document dates and inferred Datings, if any exist.
+        """
+        # it is unlikely, but technically possible, that a document could have both on-document
+        # dates and inferred datings, so find the min and max out of all of them.
+        start_date = self.start_date if self.start_date else None
+        end_date = self.end_date if self.end_date else None
+        dating_range = [start_date, end_date]
```
rlskoeser, line 937: what are the values if they aren't set?

think this can be simpler:
```suggestion
        dating_range = [self.start_date or None, self.end_date or None]
```

geniza/corpus/models.py
```diff
@@ -925,6 +925,65 @@ def title(self):
...
+                dating_range[0] = start
+            # get end from standardized date range
+            end = PartialDate(split_date[1]) if len(split_date) > 1 else start
+            # use numeric format to compare to current max, replace if larger
+            end_numeric = int(end.numeric_format(mode="max"))
+            if dating_range[1] is None or end_numeric > int(
+                dating_range[1].numeric_format(mode="max")
+            ):
+                dating_range[1] = end
+
```
rlskoeser, line 962: This is a bit hard to follow (and seeing it is really making me wish that my experimental `undate` library was farther along and could be used for calculations like this!). Am I understanding correctly that you're calculating the min and max across all datings and date ranges? And you have to compare with the numeric values but return as partial dates?

I'm wondering if any of this logic could be shifted to the dating class or to partial date, but I also kind of think that isn't your problem to solve and if this is working and testing it's ok. (Maybe you can help me figure out how to translate to use cases for that `undate` should support.) My guess is that if we could implement gt and lt comparison methods on partial dates you could aggregate start and end dates and then just apply `min` and `max` - but there's probably some complexity I'm missing, and I don't think you need to implement this.

geniza/corpus/models.py
```diff
@@ -925,6 +925,65 @@ def title(self):
...
+            dating_range[1] is not None
+            and dating_range[1].isoformat(mode="max") != solr_dating_range[0]
+        ):
+            # max from dating_range, formatted YYYY-MM-DD or YYYY-MM or YYYY
+            solr_dating_range.append(dating_range[1].isoformat(mode="max"))
+        if not solr_dating_range:
+            return None
+        # if a single date instead of a range, start and end are the same
+        if len(solr_dating_range) == 1:
+            return solr_dating_range[0]
```
rlskoeser, line 983: is this equivalent?  (white space is probably not right)

```suggestion
        solr_dating_range = []
         for i, date in enumerate(self.dating_range()):
             if date: 
                # min/max from dating_range, formatted YYYY-MM-DD or YYYY-MM or YYYY
                solr_dating_range.append(date.isoformat(mode='max' if i == 1 else 'min'))
        if not solr_dating_range:
            return None
        # if a single date instead of a range, start and end are the same
        if solr_dating_range[0] == solr_dating_range[1]:
            return solr_dating_range[0]
```

geniza/corpus/models.py
```diff
@@ -925,6 +925,65 @@ def title(self):
...
+        # if there's more than one date, return as a range
+        return "[%s TO %s]" % tuple(solr_dating_range)
```
rlskoeser, line 985: It looks like you support `None` values for start/end in the dating range — I interpreted that as open-ended ranges, but maybe I'm misinterpreting because I don't have full context? Just wondering if instead of omitting the `None` values you should preserve them and convert them to `*` in the range value (as long as at least one value is present).

rlskoeser (approved):
Interesting stuff. Some questions and suggestions for you, but this generally looks workable to me.

geniza/corpus/models.py
blms: > Am I understanding correctly that you're calculating the min and max across all datings and date ranges? And you have to compare with the numeric values but return as partial dates?

Yes! Added some comments to clarify.

> (Maybe you can help me figure out how to translate to use cases for that `undate` should support.) My guess is that if we could implement gt and lt comparison methods on partial dates you could aggregate start and end dates and then just apply `min` and `max`

I think you're right, gt and lt between PartialDates would help—then we would just have to convert the string date range to PartialDates and do the comparison directly, instead of all the int casting and `numeric_format` calls.

geniza/corpus/models.py
blms: I don't think we support setting open-ended date ranges on documents, inferred or on-document; but we do support setting a single date as opposed to a range. And this is for indexing that backend data, not searching with an open-ended range. 

In case this is a source of confusion: I don't think we'll end up with `None` in min or max as a result of `Document.dating_range`, they're really just in that method as placeholder values for when `Document.start_date` or `Document.end_date` aren't returning something useful. (I actually think those methods just return `None` if not a date… but I wasn't 100% sure about their possible outputs.)

I think the only actual cases we'll encounter are `"start_date/end_date"` or just `"date"`, for both document date and inferred date, so `dating_range` could probably be simplified to reflect that. But essentially it will either return `(start, end)`, if there is some kind of range represented between those two fields, or `(date, date)` with the same date repeated, if there is exactly one date inferred/on-document, and therefore the range of possible dates for that document is exactly that narrow. Then, `solr_dating_range` might get one repeated date `(date, date)` as input, and if those turn out to be the same `date.isoformat(mode="max")` and `mode="min"` then we only want to index that document as having a single date, not an open-ended range. (Though if, for example, the input is just a year repeated `(YYYY, YYYY)`, then it should return a range like `[YYYY-01-01 TO YYYY-12-31]`.)

geniza/corpus/models.py
blms: Yes, great simplification!

rlskoeser (approved):
Thanks for simplifying and adding more comments, I think this is clearer now.

Issue #1311, opened 2023-02-08
Add polygon annotation boxes to the transcription editor 


kseniaryzhova:
### testing notes (QA -- round 2)

On the [QA site](https://test-geniza.cdh.princeton.edu/), go to add/edit a transcription or translation
- [x] The icons should now match the designs from Chelsea, and it should be clear which tool is selected
- [x] The help text should now mention clicking and dragging to create polygon annotations
  - I think I was able to reproduce the bug you mentioned, which is actually an underlying bug with Annotorious. I'm not sure there is much we can do there, but there is an [issue on Annotorious](https://github.com/annotorious/annotorious-openseadragon/issues/178) for it already.
  - It is caused when you hold shift and click once without dragging. It makes it impossible to use either of the tools. 
  - The fix is to simply refresh the page (no hard refresh needed). Then you will be able to click and drag again.
  - Hopefully, the updated help text will prevent people from encountering this, since it explicitly says to click _and drag_ now.

---

### testing notes (QA)

On the [QA site](https://test-geniza.cdh.princeton.edu/), go to add/edit a transcription or translation
- [x] There should now be two additional icons, one for a rectangle tool and one for a polygon tool, above each image, between the pin icon and the zoom control (may take a second to load)
- [x] The rectangle tool should be selected by default, and should work as usual
- [x] When the polygon tool is clicked, it should appear selected
- [x] The instructions on the right should be updated to explain how to use the polygon tool
- [x] Following the instructions (including a double-click at the end of drawing) should allow one to draw polygons that behave similarly to the rectangular annotations
- [x] You should be able to write a label and transcription/translation text for that polygonal region
- [x] Saving it should persist and show publicly, as with rectangular annotations
- [x] Leaving the page and returning to the editor should not cause the shape of the polygon to change or disappear

---

**Is your feature request related to a problem? Please describe.**
Text layout in Geniza documents is non-standard - there is diagonal text in the margins for example, and other non-rectangular shaped text. We want to draw polygonal annotation boxes to better map transcription text to image.

**Describe the solution you'd like**
I want to be able to reformat the annotation box on an image from a rectangle to a polygon to better fit non-standard text layouts. This should be a marked choice - the default should be the rectangular box. 

**Additional context**
Just one example: 
![image](https://user-images.githubusercontent.com/105385128/217596398-a456bbc7-4ecf-481f-922a-2844835606c7.png)

rlskoeser:
[...]

Issue #1221, opened 2022-11-08
Editing/deleting parts of annotation box titles results in unexpected behavior (no change or deleting entire annotation box)


kseniaryzhova:
### testing notes (QA)

- [x] On the [QA site](https://test-geniza.cdh.princeton.edu/) transcription/translation editor, the below bug should no longer occur.

---

**Describe the bug**
When I go to delete a portion of a title for an existing annotation box, it either does nothing or deletes the entire title and the annotation box. without warning. See video.

**To reproduce**
See video: https://clipchamp.com/watch/cyzDf9MOL0H
Testing in test site, PGPID 733.

**Expected behavior**
When I go to edit a title, I want my edits to be immediately reflected in the transcription editor (so I can choose to save or cancel, etc) and I do not want for the entire annotation box to be affected by an edit of the title.

blms:
@kseniaryzhova Tried following along with the video but I can't replicate this bug, it seems to work fine. Is it possible this was fixed by the most recent changes to the transcription editor?

kseniaryzhova:
@blms still an issue. Will send you a video via chat.

blms:
Small update on this... annotorious-OSD released my patch for this, the `disableDeleteKey`, in version 2.7.11. What I had forgotten entirely was that we were pinned to 2.7.6 because of a bug preventing our annotations from loading introduced in the following version.

I tracked down the bug today and opened a PR with a fix on their repo: https://github.com/annotorious/annotorious-openseadragon/pull/180. Both `disableDeleteKey` and my other bugfix work for me now locally, but it may be some time before we're able to get it into an Annotorious release and subsequently upgrade the dependency in our code.

kseniaryzhova:
@blms no issues, it works! Thank you, closing!

PR #1456, opened 2023-09-05
Annotorious upgrade, delete key fix (#1221), polygon tool (#1311)

blms:
## In this PR

- Fix a small issue with the enlarge button trying to initialize in the editor
- Upgrade annotorious, annotorious-toolbar, and annotorious-openseadragon to latest versions

Per #1221:
- Use the new `disableDeleteKey` option in Annotorious to fix a bug with editing annotation labels

Per #1311:
- Add a toolbar with the rectangle and polygon tools from Annotorious to allow creation of polygon image annotations within the editor


geniza/corpus/templates/corpus/snippets/document_transcription.html
```diff
@@ -115,14 +115,28 @@
...
                                 {% if edit_mode %}
+                                    {# popout and close button #}
                                     <button class="primary popout-close-button" type="button" aria-label="Close"><i class="ph-x"></i></button>
                                     <button class="popout-button" type="button">
                                         <svg>
                                             <use xlink:href="{% static 'img/ui/desktop/all/pop-out.svg' %}#pop-out-icon" />
                                         </svg>
                                         <span class="sr-only">Pop out</span>
                                     </button>
+                                    <fieldset class="toolbar">
```
rlskoeser, line 126: Why are these added to the geniza template instead of to tahqiq?

I don't object necessarily, just want to understand

geniza/corpus/templates/corpus/snippets/document_transcription.html
```diff
@@ -115,14 +115,28 @@
...
+                                        {# rectangle tool #}
+                                        <label class="rect-tool active-tool">
+                                            <input type="radio" checked />
+                                            <span class="sr-only">Rectangle tool</span>
```
rlskoeser, line 130: not sure how meaningful "rectangle tool" is; how about rephrasing to describe better what these do?

```suggestion
                                            <span class="sr-only">Select a rectangle</span>
```

geniza/corpus/templates/corpus/snippets/document_transcription.html
blms: I hadn't thought to do that! Was conceptualizing it as a Geniza feature so put it in Geniza, but I think it makes more sense to pass the toolbar element (in this case the fieldset) as a param to tahqiq, and instantiate the button elements there.

geniza/corpus/templates/corpus/snippets/document_transcription.html
blms: Will put this suggestion into the tahqiq update, thanks!

sitemedia/js/controllers/annotation_controller.js
```diff
@@ -89,7 +89,10 @@ export default class extends Controller {
 
         // enable annotorious-tahqiq
         const { config, canvasURL, manifestId, annotationContainer } = settings;
-        const anno = Annotorious(viewer);
+        const anno = Annotorious(viewer, { disableDeleteKey: true });
+
+        // get toolbar element
+        const toolbarContainer = this.element.querySelector(".tahqiq-toolbar");
```
rlskoeser, line 95: do you still need to bind this here? what does the annotation controller need to do with it?

blms: The annotation controller needs to pass it to tahqiq instantiation, but doesn't need to define a variable for it so I've simplified.

rlskoeser (approved):

Issue #1458, opened 2023-09-11
As a content admin, when viewing people-to-people relationships in person pages, I want reverse relationships to be visible, so that I don't inadvertently add a relationship twice.


blms:
### testing notes (QA)

on the [QA site admin](https://test-geniza.cdh.princeton.edu/admin/):

- [x] If you select one of the entries in the list of [Person-Person relationship types](https://test-geniza.cdh.princeton.edu/admin/entities/personpersonrelationtype/), you should be able to view and edit the converse relationship type name. All the ones from the spreadsheet should now be populated here.
- Add some Person entries and some Person-Person relationships between them
- [x] After adding a single relationship between people, both people should have the relationship represented on their page
  - [x] The first person should show the relationship you entered in the "Related People (input manually)" section, with the relationship type you selected
  - [x] When you go to the other person's Person page, you should see the "converse" relationship type in the "Related People (automatically populated)"

### dev notes 

- from discussion on https://github.com/Princeton-CDH/geniza/issues/1390
- need a new field on RelationshipType for reverse of relationship (probably come up with better nomenclature?)
  - reverse relationships listed [here](https://docs.google.com/spreadsheets/d/1RwdV_qMsXo1L-UOplU77ps2J1DSvKb6OKAFq4wOwCRo/edit?usp=sharing)
  - can do this in a data migration for these

blms:
@kseniaryzhova A handful of questions that have come up in the process of implementing this:

1. Should the reciprocal relations for “Maternal cousin” and “Paternal cousin” just be “Cousin”? 
   - Example: if Jane is my maternal cousin because she’s my mother’s sister’s daughter, then the relationship is identical on both sides—I’m also her maternal cousin. But if Jane is my mother’s brother’s daughter, then while she’s still my maternal cousin, I’m actually her _paternal_ cousin. In other words, if I add Jane in the DB as my maternal cousin, it’s ambiguous whether I’m Jane's maternal or paternal cousin.
2. For the nephew and niece relations, is there a reason why we only have aunt and uncle relations as actual relationship types, and nephew/niece only appear as reciprocal relations?
3. Is "Reverse name" and the following help text a good description for what we're talking about here? "Reciprocal" seems inaccurate (since it implies something done or given in return for something else, rather than a semantic complement)
   -  I'm using the term "[semantic converse](https://en.wikipedia.org/wiki/Converse_(semantics))" here after some research on wikipedia 😆 
[...]

Issue #1457, opened 2023-09-11
As a content admin, when adding people-to-people relationships in person pages, I want an added "ambiguity" category to the drop down so I can clarify when people are similar/not the same.


kseniaryzhova:
### testing notes (QA)

on the [QA site admin](https://test-geniza.cdh.princeton.edu/admin/):
- [x] The below options and category should now be in the people-to-people relationship dropdown menu

**Describe the solution you'd like**
I want an extra category added to the people-to-people relationship dropdown menu. It should be called "Ambiguity" and include the following options: 

- possibly the same as
- not to be confused with

kseniaryzhova:
@blms Perfect! Works as it should, thank you!

PR #1462, opened 2023-09-13
Person-Person: Add converse relationship types and display (#1458), add new "ambiguity" category (#1457)

blms:
## In this PR

Per #1458:
- Show read-only list on the Person admin for reverse Person-Person relationships
  - Rename existing one to "Related People (input manually)" and name the second "Related People (populated automatically)"
- Add field for "[converse](https://en.wikipedia.org/wiki/Converse_(semantics))" type names for relationship types
  - Populate `converse_name` from provided spreadsheet (for those relationships that are asymmetrical)
  - Display the `converse_name` on the People pages' read-only reverse Person-Person sections

Per #1457:
- Add new PersonPersonRelationType category called "Ambiguity" to track relationships of ambiguity, such as "possibly the same person" and "not to be confused with", and populate with those two types (both symmetrical so no converse name required)

## Notes

All language, naming, and labeling was discussed, iterated, and approved in [9/13 meeting](https://github.com/Princeton-CDH/geniza/issues/1458#issuecomment-1717989006)


rlskoeser (approved):
This all looks reasonable to me from a quick review of the code. I like the solution of adding an optional 'converse' relationship, I hope it works for what you all need!

Issue #1459, opened 2023-09-13
As a content admin, I want to be able to merge two (identical) people pages without losing any data


kseniaryzhova:
### testing notes (QA)

In the [QA site admin](https://test-geniza.cdh.princeton.edu/admin):
- [x] Visit the list of People. You should be able to select multiple people, then in the dropdown, choose "merge selected people"
- [x] You should be brought to a view showing a comparison between the people you selected, listing all the properties of both
- [x] If any relations are not present for a record (for example, document-person) this should also be indicated
- [x] You should be able to choose a primary person for the merge
- [x] When you merge them, you should be redirected to the merged record
- [x] On merge, descriptions should be combined; all names (except duplicates), relations (docs/people/places), and footnotes should appear on the merged record
  - [x] The display name of the primary person should be the display name of the merged record
- [x] Merging two people with different genders, or different social roles, should cause an error visible to the user
  - [x] The exception is when one of the genders is "Unknown" and the other(s) is/are identical, since unknown is treated as blank

---

**Describe the solution you'd like**
I want to be able to merge duplicate people pages in the same way we merge documents now (side-by-side comparison of entires, no data loss, choosing the main display name).

kseniaryzhova:
@blms testings works as it should. I did see a potential issue. When I merge two records, the display name becomes the associated name for every related document, even if some of them were under the other name pre-merge. Is there an easy way to maintain that difference? Just so you can see which names were associated with which documents? If it's too much of an investment, we can deal.

blms:
@kseniaryzhova Definitely possible, we could keep track of which name is the display name when the document-person association is created, then ensure that doesn't get modified when things are merged. I would estimate about 6 hours for it. Want to create a new user story for it and close this one if complete otherwise?

kseniaryzhova:
@blms We decided not to keep track of which name displays on the related documents page after the merge. Closing, thank you!

PR #1474, opened 2023-10-13
Person merge methods/form/view/templates (#1459)

blms:
## In this PR

Per #1459:
- Add methods, form, view, templates for merging 2 or more Person entities

Heavily adapted from document merge.


geniza/entities/models.py
```diff
@@ -146,6 +151,168 @@ def __str__(self):
...
+            for lang_code in language_codes
+        }
+
+        # collect names as strings (since that's the important part for comparison)
+        self_names = [str(name) for name in self.names.all()]
+
+        for person in merge_people:
+            # ensure any has_page overrides are respected
+            if person.has_page and not self.has_page:
+                self.has_page = True
```
rlskoeser, line 180: possibly simpler (I'm assuming has page should be true if it is true for any of the records being merged); true if either is true otherwise false

```suggestion
            self.has_page = self.has_page or person.has_page
```

geniza/entities/models.py
```diff
@@ -146,6 +151,168 @@ def __str__(self):
...
+
+            # ensure missing role and gender are migrated
+            if person.role and not self.role:
+                self.role = person.role
+            if self.gender == Person.UNKNOWN and person.gender != Person.UNKNOWN:
+                self.gender = person.gender
```
rlskoeser, line 186: am I understanding correctly, these are only set if not already present? (comment could be slightly clearer, maybe something like: migrate/copy role and gender if not already present)

does it matter if there's a conflict?

geniza/entities/models.py
```diff
@@ -146,6 +151,168 @@ def __str__(self):
...
+                if str(name) not in self_names:
+                    # if not duplicated, make non-primary and add
+                    name.primary = False
+                    name.save()
+                    self.names.add(name)
+
+            # add description if set and not duplicated
+            # for all supported languages
+            for lang_code in language_codes:
+                description_field = "description_%s" % lang_code
```
rlskoeser, line 202: is all of this to handle model translation fields?

geniza/entities/models.py
```diff
@@ -146,6 +151,168 @@ def __str__(self):
...
+                if person_description and person_description not in current_description:
+                    description_chunks[lang_code].append(
+                        "Description from merged entry:\n%s" % (person_description,)
+                    )
+
+            # combine person-person relationships
+            for to_relationship in person.to_person.all():
+                # start with "to" relations; merged person is "from_person"
+                # don't add relationships where the other person is self!
+                if to_relationship.to_person.pk != self.pk:
```
rlskoeser, line 214: could filter out self relations at the queryset level instead of in the loop

geniza/entities/models.py
```diff
@@ -146,6 +151,168 @@ def __str__(self):
...
+                    # first remove any doc relations; person footnotes should not have
+                    # doc_relation anyway, but just in case...
+                    if footnote.doc_relation:
+                        footnote.doc_relation = ""
+                        footnote.save()
+
+                    # then add to this person
+                    self.footnotes.add(footnote)
+
+        # combine aggregated content for text fields
```
rlskoeser, line 260: wondered why this was so far from where it's collected, but I see now there's a change in indentation - so this is after content has been aggregated from all records being merged?

geniza/entities/templates/entities/snippets/person_option_label.html
```diff
@@ -0,0 +1,80 @@
...
+    {% if person.gender %}
+        <div class="form-row">
+            <label>Gender</label><div>{{ person.gender }}</div>
+        </div>
+    {% endif %}
+    {% if person.description %}
+        <div class="form-row">
+            <label>Description</label><div>{{ person.description }}</div>
+        </div>
+    {% endif %}
```
rlskoeser, line 31: does this display all languages or only current language? should it display all, or all that have any content? wonder if there is a way to use model translate tabs to embed them, but I think then users would have to click to see if there's any content

geniza/entities/templates/entities/snippets/person_option_label.html
```diff
@@ -0,0 +1,80 @@
...
+    {% if person.documents.count %}
+        {% for rel in person.persondocumentrelation_set.all %}
+            <div class="form-row">
+                <label>Related document</label><div>
+                    <a target="_blank" href="{% url 'admin:corpus_document_change' rel.document.pk %}">
+                        {{ rel.document }}</a> ({{ rel.type }})</div>
+            </div>
+        {% endfor %}
+    {% endif %}
+    {% if person.to_person.count %}
```
rlskoeser, line 41: fwiw I think `.exists` should be sufficient and slightly cheaper to calculate; although it doesn't look like you're displaying anything outside the loop, so you probably don't need the conditional at all

geniza/entities/templates/entities/snippets/person_option_label.html
```diff
@@ -0,0 +1,80 @@
...
+        {% for rel in person.to_person.all %}
+            <div class="form-row">
+                <label>Related person</label>
+                <div>
+                    <a target="_blank" href="{% url 'admin:entities_person_change' rel.to_person.pk %}">
+                        {{ rel.to_person }}</a> ({{ rel.type }})
+                </div>
+            </div>
+        {% endfor %}
```
rlskoeser, line 50: If you want to be explicit about data not being present, you could use `{% empty %}` and output "no related persons" or something like that. (Hard to tell from the templates how much information is already on the form or  whether that's important to include.)

geniza/entities/models.py
```diff
@@ -146,6 +151,168 @@ def __str__(self):
...
+                self.role = person.role
+            if self.gender == Person.UNKNOWN and person.gender != Person.UNKNOWN:
+                self.gender = person.gender
+
+            # combine log entries (before name, since name used in log entries)
+            self._merge_logentries(person)
+
+            # combine names
+            for name in person.names.all():
+                if str(name) not in self_names:
```
rlskoeser, line 193: if any `name` records are duplicated and not moved, will they be left stranded in the db? or will the delete on the merged person record cascade and clean them up?

geniza/entities/models.py
```diff
@@ -146,6 +151,168 @@ def __str__(self):
...
+            for lang_code in language_codes:
+                description_field = "description_%s" % lang_code
+                person_description = getattr(person, description_field)
+                current_description = getattr(self, description_field)
+                if person_description and person_description not in current_description:
+                    description_chunks[lang_code].append(
+                        "Description from merged entry:\n%s" % (person_description,)
+                    )
+
+            # combine person-person relationships
```
rlskoeser, line 210: complicated - a little hard to follow!

rlskoeser (approved):
Complicated stuff! Glad the document merge approach is useful and can be adapted for this. Have a few comments and suggestions, but generally looks reasonable based on reviewing the code changes.

geniza/entities/templates/entities/snippets/person_option_label.html
blms: It will display only the current language, which is how the document merge currently works. I spent a bit of time trying to figure out how to loop through the different language options, but it's not trivial; we could hard code English, Arabic and Hebrew... But since it's already not working like this on the document merge, and it seems quite unlikely that a person would have a Hebrew or Arabic description but _not_ an English one, I think it's probably ok for now. (And we can add as a feature later if it ends up being needed.)

geniza/entities/models.py
blms: The latter!

geniza/entities/templates/entities/snippets/person_option_label.html
rlskoeser: I see. That sounds reasonable. The main thing is that we're not losing any content in any language when merging, and it sounds like we handle that in both cases.

This seems like a feature that modeltranslate ought to offer, but I just looked at the docs and don't see anything like it. I'm having trouble figuring out how to articulate what we need to even find if there's a relevant issue.

geniza/entities/models.py
blms: Yup, copied verbatim from document merge

geniza/entities/models.py
blms: Right, all the description chunks are collected for each merged person, and then combined at the end of the loop. This is also from the document merge verbatim.

geniza/entities/models.py
blms: I think it's probably worth raising a validation error if there's a conflict, just to make sure researchers don't make a mistake and combine people who shouldn't actually be merged.

geniza/entities/models.py
blms: Hopefully a little easier now, though I have some regrets about the naming scheme of "from" and "to"... it isn't the best, but maybe not worth changing at this point.

geniza/entities/views.py
```diff
@@ -0,0 +1,74 @@
...
+
+        try:
+            primary_person.merge_with(secondary_people, user=user)
+        except ValidationError as err:
+            # in case the merge resulted in an error, display error to user
+            messages.error(self.request, err.message)
+            # redirect to this form page instead of one of the people
+            return HttpResponseRedirect(
+                "%s?ids=%s"
+                % (reverse("admin:person-merge"), self.request.GET.get("ids", "")),
```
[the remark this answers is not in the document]
blms: Ha, it's showing this one again! But I think the [same answer](https://github.com/Princeton-CDH/geniza/pull/1444#discussion_r1307682129) applies from document merge.

geniza/entities/models.py
```diff
@@ -176,14 +177,21 @@ def merge_with(self, merge_people, user=None):
 
         for person in merge_people:
             # ensure any has_page overrides are respected
-            if person.has_page and not self.has_page:
-                self.has_page = True
+            self.has_page = self.has_page or person.has_page
 
-            # ensure missing role and gender are migrated
+            # migrate/copy role and gender if not already present, check for conflicts otherwies
```
rlskoeser, line 182: typo in your comment (Nick used to catch these in mine all the time)
```suggestion
            # migrate/copy role and gender if not already present, check for conflicts otherwise
```

geniza/entities/models.py
rlskoeser: curious what you would you use instead - have you thought of something more semantic?  agree it probably isn't worth changing

I've wondered if these relationships would be easier to express as RDF with subject, predicate, object (although that opens up a different can of worms...)

rlskoeser (approved):
Revisions look good to me. I think explicitly displaying the empty related documents and persons will be helpful when reviewing merges.

geniza/entities/models.py
blms: I didn't have anything in mind, and can't think of anything off the top of my head—which contributes to the inertia to not change it! 😆

PR #1485, opened 2023-10-23
Cleanup Unicode non-breaking space (\xa0) in text records (#1475)

blms:
## In this PR

Per #1475 
- When saving records, replace `\xa0` with ASCII space ` ` in Document descriptions, Names, and Annotation content, so that the wrong kind of space isn't persisted in the DB (as it breaks search)

## Notes/questions

- Troubles with the corpus migrations unit tests continue—again, only on GitHub actions, so I've skipped it (thus the coverage issues).
- I don't think we plan on making Names searchable on the public frontend, but I figured it would make sense to implement this for the benefit of admin search, does that make sense?
- I found exactly one Source record that has this. Definitely not worth a migration, but is it worth a cleanup in the `save` method? I think this would also only be for the benefit of admin search.
- Is this something that should be common to every model and apply to every `TextField`/`CharField`, or would that be overkill?


geniza/annotations/migrations/0005_annotation_cleanup_nbsp.py
```diff
@@ -0,0 +1,27 @@
...
+
+def cleanup_annotations(apps, schema_editor):
+    """Cleanup all annotations with unicode nbsp in their labels or values"""
+    Annotation = apps.get_model("annotations", "Annotation")
+    for anno in Annotation.objects.filter(content__icontains="\xa0"):
+        for body in anno.content["body"]:
+            if "label" in body:
+                body["label"] = body["label"].replace("\xa0", " ")
+            if "value" in body:
+                body["value"] = body["value"].replace("\xa0", " ")
```
rlskoeser, line 14: Do you need to normalize spaces here or anywhere else you're cleaning up these non-breaking spaces?  i.e., if there are non-breaking spaces and regular white spaces together, you'll end up with extra spaces. In most contexts I think it won't matter, since in both Solr and typical browser html display multiple spaces get displayed as a single space. The places I'm thinking of that matter are when we display transcriptions (we have a css style to preserve white space) and the corpus export.

Since it's a limited set you're cleaning up, and you've already looked at them, maybe you already know this isn't a concern.

rlskoeser (approved):
Implementation looks good to me

rlskoeser:
in response to your questions and notes

> * Troubles with the corpus migrations unit tests continue—again, only on GitHub actions, so I've skipped it (thus the coverage issues).

Ok, thanks for noting. I forget when/if it's acceptable to use an external utility method for a migration (I know it's not exactly in the spirit of migrations); wondering if that would be a better approach for testing the logic used in a migration without having to run the actual migration.

Another thought: perhaps we could decide on a point at which you could squash migrations. We have enough now that it might be a good idea; once you do, I think we could remove the previous migration tests and tell codecov to ignore them (if it doesn't automatically).  That would simplify the problems with ordering on any new migration tests, and also presumably speed up the test suite time a bit. Maybe after an upcoming release would be a good time for this?  (Honestly, you could probably do it anytime.)

> * I don't think we plan on making Names searchable on the public frontend, but I figured it would make sense to implement this for the benefit of admin search, does that make sense?

Makes sense to me.

> * I found exactly one Source record that has this. Definitely not worth a migration, but is it worth a cleanup in the `save` method? I think this would also only be for the benefit of admin search.

One source seems like something that is not worth writing code for - could you just clean it up manually, or ask someone else to?

> * Is this something that should be common to every model and apply to every `TextField`/`CharField`, or would that be overkill?

It seems like overkill to me - from what you've found so far, it seems like these are mostly from older records that were imported from other systems, so it doesn't seem like something that would be an ongoing problem.  The one reason that I would change my mind is if there's a simple way you could implement it in a reusable way that could be easily applied to all the fields that might need it, e.g. as a mixin or text/charfield subclass - but even that is probably not warranted here.

geniza/annotations/migrations/0005_annotation_cleanup_nbsp.py
blms: Thank you so much for catching this, it was something I completely overlooked, and in fact it did occur over a hundred times in annotations. Not sure about document descriptions but I'm sure it was present there as well. Used regex to improve the replacement in all cases!

blms:
> Another thought: perhaps we could decide on a point at which you could squash migrations. We have enough now that it might be a good idea; once you do, I think we could remove the previous migration tests and tell codecov to ignore them (if it doesn't automatically). That would simplify the problems with ordering on any new migration tests, and also presumably speed up the test suite time a bit. Maybe after an upcoming release would be a good time for this? (Honestly, you could probably do it anytime.)

@rlskoeser Makes sense to me, I'll look into it (never squashed more than a couple of migrations at a time in django). Do you mean just delete them from the test suite code, and how do you mean to tell codecov to ignore them?

PR #1490, opened 2023-10-25
Index rotations and display in search results (#1070)

blms:
## In this PR

Realized while writing testing notes for #1070 that rotated images did not appear in search results. So I made the following changes:
- Index rotation overrides as `*_is`, and zip into tuples in `get_result_document`, following the same pattern as labels
- Fallback to list of 0s (the same length of the list of images)
- Display rotated images in results using same method from ITT viewer

The nice thing about the fallback is that we don't have to reindex anything manually, as we know rotation is always 0 unless the override has been set, at which point the document will reindex on save.

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -1198,6 +1198,8 @@ class Document(ModelIndexable, DocumentDateMixin):
                 "language_code_s": self.primary_lang_code,
                 "language_script_s": self.primary_script,
                 # use image info link without trailing info.json to easily convert back to iiif image client
+                # NOTE: if/when piffle supports initializing from full image uris, we could simplify this
+                # code to index the full image url, with rotation overrides applied
                 "iiif_images_ss": [
                     img["image"].info()[:-10]  # i.e., remove /info.json
                     for img in images
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -1203,6 +1203,7 @@ class Document(ModelIndexable, DocumentDateMixin):
                     for img in images
                 ],
                 "iiif_labels_ss": [img["label"] for img in images],
+                "iiif_rotations_is": [img["rotation"] for img in images],
                 "has_image_b": len(images) > 0,
             }
         )
--- a/geniza/corpus/solr_queryset.py
+++ b/geniza/corpus/solr_queryset.py
@@ -52,6 +52,7 @@ class DocumentSolrQuerySet(AliasedSolrQuerySet):
         "translation_language_direction": "translation_language_direction_s",
         "iiif_images": "iiif_images_ss",
         "iiif_labels": "iiif_labels_ss",
+        "iiif_rotations": "iiif_rotations_is",
         "has_image": "has_image_b",
         "has_digital_edition": "has_digital_edition_b",
         "has_digital_translation": "has_digital_translation_b",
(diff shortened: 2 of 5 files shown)


rlskoeser (commented):
I've forgotten what we index for the iiif image; is it a full iiif image url? It seems like it would be simpler to apply at index time rather than combining and splitting and applying at display time.  

From what I remember of the viewer you had to have the rotation separate for a different reason, although I've forgotten exactly why.

blms:
> I've forgotten what we index for the iiif image; is it a full iiif image url?

@rlskoeser it's the image info JSON without `/info.json` at the end, and then that is used to instantiate a new `IIIFImageClient` object per image. Some examples:
```
['https://luna.manchester.ac.uk/luna/servlet/iiif/ManchesterDev~95~2~84385~116670', 'https://luna.manchester.ac.uk/luna/servlet/iiif/ManchesterDev~95~2~84390~116671']
['https://iiif-cloud.princeton.edu/iiif/2/e5%2F10%2F4c%2Fe5104c53230b49cea5f97b8e48e3a14f%2Fintermediate_file', 'https://iiif-cloud.princeton.edu/iiif/2/1c%2Fd4%2Fca%2F1cd4cababb4544fcb5af557276a09496%2Fintermediate_file']
['https://images.lib.cam.ac.uk/iiif/MS-TS-AS-00162-00068-000-00001.jp2', 'https://images.lib.cam.ac.uk/iiif/MS-TS-AS-00162-00068-000-00002.jp2']
```

Does it make sense to apply the rotations at index time and produce a different result? Would we need to perform network requests and use `IIIFImageClient` etc at index time instead of in `get_result_document`? I think the latter method originated in https://github.com/Princeton-CDH/geniza/pull/549 if it helps.

> From what I remember of the viewer you had to have the rotation separate for a different reason, although I've forgotten exactly why.

I'm thinking the reason was that for the `srcset`, images needed different sizes passed as params to the template filter, but always the same rotation variable from the overrides. So the cleanest template code ended up being to apply the template filter twice, once for size and once for rotation. But I don't remember exactly—there may have been a more pressing reason related to how the template filter works that I'm forgetting!

rlskoeser:
Thanks for the reminder - this is a design choice about piffle that I wish I could revise, it would be a lot easier to give it an arbitrary iiif image url and let it separate out the api endpoint from the image id and the rest of the url (it already has logic for parsing out the portions of the url). I think I did that because it's not always easy to separate them out.... although maybe that's not true, it ought to be based on the API. 🤔 

fwiw, if you apply rotation to a piffle image object you should be able to keep the same rotated image and then apply a different size.  In case it matters to the other code you mentioned.

Given the constraints this approach seems reasonable. I'll create an issue in piffle to say we should add a way to initialize from an arbitrary iiif image url - thinking of a static method, like `from_uri`.  (LMK if you have any suggestions)

If you would, please add a comment here that if/when piffle supports initializing from full image uris we could simplify this code to index the full image url with rotation applied.

rlskoeser (approved):
ok to merge once you document why the rotation has to be separate (limitation of current piffle)

rlskoeser:
@blms I should have checked the piffle docs instead of trusting my memory! Piffle already has the method I was proposing to use: `IIFImageClient.init_from_url`. If I get inspired and have time (ha) maybe I will take a stab at refactoring. You shouldn't refactor, but if something comes up where you're working on this part of the code again maybe we can take advantage of this.

Issue #1470, opened 2023-10-04
Automatic ingest of old/historic shelfmarks into the PGP for both backend and front end visibility


kseniaryzhova:
## testing notes 

In the [QA site admin](https://test-geniza.cdh.princeton.edu/admin/corpus/fragment/):
- [x] Spot-check some affected Fragments to make sure the historical shelfmarks are entered correctly
- [x] Check the affected CUL fragments to make sure the commas are fixed
- [x] Check the affected Bodleian fragments to make sure the catalog numbers are entered correctly

---

**Describe the solution you'd like**
Please automatically ingest the old/historic shelfmarks for the relevant shelfmarks so these old shelfmarks show up in search and display publically.

**Additional context**
[PGP Shelfmarks Guide - needs review by MR but nearly done](https://docs.google.com/spreadsheets/d/1YuRUX5L_cqPvhUJtyyp9kDOpQsXYa18Q9alqznUAl8c/edit?usp=sharing)

blms:
Some questions have come up as I’m working through this list:

1. Should anything go in for row 4, the note about ignoring certain parts of the historic shelfmark?
2. For row 30, does this example look right?
   - current shelfmark `L-G Ar. I.100`
   - historic shelfmark `West. Coll. (Westminster College) Ar. I.100; Frag. Cairens. (misc section) Ar. I.100`
3. Row 51 current shelfmark looks slightly wrong, at least I’m seeing `T-S A#.#` rather than `T-S A.#`. Would this be formatted like this?
   - current shelfmark: `T-S J2.4`
     - historic shelfmark: `TS Box J, 2, fol. 4`
   - current shelfmark: `T-S K6.150.2`
     - historic shelfmark: `TS Box K, 6, fol. 150.2`
4. Row 54 `T-S Misc. 35-36`, would this be formatted as such?
   - current shelfmark:`T-S Misc. 35.38`
   - historic shelfmark: `TS Loan 38`
5. Row 55, are we only doing specifically the one shelfmark `T-S 6H9–21`, or any that match the pattern “T-S number-letter-number” (e.g. `T-S 6J#`, `T-S 8J#`, `T-S 8K#`)? And are the numbers and letters separated with spaces? Here's what I have now:
   - current shelfmark: `T-S 6H9–21`
   - historic shelfmark: `TS 6 H 9–21; TS 6 H f. 9–21; TS Box 6 H f. 9–21`
[...]

PR #1519, opened 2024-01-09
Fragment historical shelfmark population and ingest (#1470)

blms:
## In this PR

Per #1470:
- Data migration that uses regex substitution to populate fragment historical shelfmarks based on [researchers' spreadsheet](https://docs.google.com/spreadsheets/d/1YuRUX5L_cqPvhUJtyyp9kDOpQsXYa18Q9alqznUAl8c/edit)
- One-time use management command to ingest specific catalog numbers into Bodleian fragment historical shelfmarks (after the migration) using a CSV provided by KR

## Notes

Since this is all code that will be run once—and in the case of the management command, will be removed after running—I figure it's fine not to unit test, and to do acceptance testing instead. I've also already run them (in a modified form to produce CSVs) and confirmed the output with MR and KR. But let me know if you think I should unit test any of this!


rlskoeser (approved):
Looks reasonable. Agree that it's not worth unit testing in this case.

Issue #1476, opened 2023-10-18
Issues with shelfmark scoped search


blms:
### testing notes

On the [QA site](https://test-geniza.cdh.princeton.edu/en/documents):
- [x] Searching for most shelfmarks using the syntax `shelfmark:"T-S 12.302"`, `shelfmark:"AIU II.A.73"`, etc should bring up the correct results
- [x] There should not be a bunch of irrelevant results containing the word "type" highlighted (this is what was happening before—and what this issue was intended to fix)
- [x] Both on the public site and admin, it should work roughly the same way (though the admin search is not ordered by relevance)

---

- Shelfmark scoped search brings up irrelevant results after the relevant ones. This was happening before the solr 9 upgrade, but since then, it seems to highlight the word "type" in each additional result. More here: https://github.com/Princeton-CDH/geniza/issues/1359#issuecomment-1719972954
- Searching for `shelfmark:"T-S 12.30"`, the relevance score seems like it may be wrong, since all the partial matches like "T-S 12.304" come up before "T-S 12.30 + JRL SERIES B 2367". Maybe it's a problem with joins. (confirmed, it is a problem with joins, unrelated.)
- Does not work at all in admin search

kseniaryzhova:
@blms works as it should, closing, thanks!

PR #1538, opened 2024-02-09
Allow deprecated edismax nested search for shelfmarks (#1476)

blms:
## In this PR

Per #1476:
- Add a new `uf` param to searches that contain nested `type=edismax` searches, as [since solr 7.2](https://solr.apache.org/guide/7_2/solr-upgrade-notes.html#solr-7-2) they will not work without it. (Shelfmark scoped search happened to be the one place where a `type=edismax` exists inside another `type=edismax` so far!)


rlskoeser (approved):
Wow, wild. Thanks for tracking this down and documenting so clearly with the comments in the code.

PR #1545, opened 2024-02-16
Omit duplicate footnote locations when listing scholarship records

rlskoeser:


blms (approved):
Yes, I think this resolves it! I wasn't able to find any examples where this didn't work.

Issue #1286, opened 2023-01-05
Bracket and other character search is functioning unpredictably 


kseniaryzhova:
### testing notes (round 2)

On the [QA public search](https://test-geniza.cdh.princeton.edu/en/documents/):
- [x] All of the below should still work
- [x] Now when searching for words _without_ the sigla in your search query, they should be highlighted in the search results, even when the result contains sigla
- [x] Searching _with_ sigla should still work as long as you use double quotes (it will not work unquoted) and should still highlight

### testing notes

On the [QA public search](https://test-geniza.cdh.princeton.edu/en/documents/):
- [x] Searching exactly for terms with sigla in them, like [`"وسـ[ـعرهـ]ـم"`](https://test-geniza.cdh.princeton.edu/en/documents/?q=%22%D9%88%D8%B3%D9%80%5B%D9%80%D8%B9%D8%B1%D9%87%D9%80%5D%D9%80%D9%85%22&docdate_0=&docdate_1=&sort=relevance), in double quotes should bring up the correct transcription(s) containing only results with those exact sigla in them
  - [x] The terms, with the sigla, should be highlighted in the result
- [x] Searching for the word [`وسعرهم`](https://test-geniza.cdh.princeton.edu/en/documents/?q=%D9%88%D8%B3%D8%B9%D8%B1%D9%87%D9%85&docdate_0=&docdate_1=&sort=relevance) without sigla or connectors, without double quotes, should bring up PGPID 9157 again, but ranked lower in relevance than those already containing that word without sigla
  - [x] The term with the sigla should NOT be highlighted in that result ~~(unfortunately not feasible with the current approach)~~
- [ ] The same general behavior should be true for other sigla, and work for Hebrew script as well

---

**Describe the bug**
Some of our transcription content is incomplete and there are often lacunae in words that are represented in the transcriptions via brackets and other chracters such as  < { ⟦ . ـ). Currently, adding brackets to search terms adds irrelevant results that are not the actual word but are just other texts with brackets/lacunae.

**To reproduce**
See video. But also (on the public search):
 1539 results for كتاب 
 4522 results for كت[ا]ب
 4454 results for كتـ[ـا]ب
1 result for just the connector ـ

**Expected behavior**
In a simple search (without quotations or boolean operators, etc), we want search results to include brackets and other characters such as  < { ⟦ . ـ if they're relevant. So كتاب should bring up searches for كتـ[ـا]ب as well and so on.

[...]

PR #1546, opened 2024-02-21
Allow searching for words with brackets and other sigla in transcriptions (#1286)

blms:
## In this PR

Per #1286:
- Allow searching for exact matches on words with brackets and other [transcription sigla](https://docs.google.com/document/d/1pyFajrGRYTzRsb2yFjfX0XmNqpz1kUO9mhlISGwfpxk/edit) inside them using double quotes (e.g. `"wo[r]d"` should match a transcription with `wo[r]d` in it verbatim)
  - Use whitespace tokenizer instead of standard tokenizer in transcription field to prevent sigla from breaking up words
  - Exclude double quoted exact searches from arabic-JA conversion
- Allow searches for words without sigla to match instances of the word with sigla inside, when not using double quotes (e.g. `word` should also match a transcription with `wo[r]d` in it)
  - Use a transcription cleaning method to index such words

### DIFF
--- a/geniza/corpus/solr_queryset.py
+++ b/geniza/corpus/solr_queryset.py
@@ -120,18 +120,24 @@ class DocumentSolrQuerySet(AliasedSolrQuerySet):
         # to avoid it being interpreted as a boolean
         search_term = self.re_shelfmark_nonbool.sub("BL or", search_term)
 
-        search_term = arabic_or_ja(search_term)
-
         # look for exact search, indicated by double quotes
         exact_queries = self.re_exact_match.findall(search_term)
         if exact_queries:
             # store unmodified query for highlighting
             self.highlight_query = search_term
             # limit any exact phrase searches to non-stemmed field
-            search_term = self.re_exact_match.sub(
-                lambda m: f"content_nostem:{m.group(0)}",
-                search_term,
+            exact_phrases = " ".join(
+                [
+                    f"content_nostem:{m}"
+                    for m in self.re_exact_match.findall(search_term)
+                ]
             )
+            # add in judaeo-arabic conversion for the rest (double-quoted phrase should NOT be
+            # converted to JA, as this breaks if any brackets or other sigla are in doublequotes)
+            remaining_query = arabic_or_ja(self.re_exact_match.sub("", search_term))
+            search_term = " ".join([exact_phrases, remaining_query]).strip()
+        else:
+            search_term = arabic_or_ja(search_term)
 
         # convert any field aliases used in search terms to actual solr fields
         # (i.e. "pgpid:950 shelfmark:ena" -> "pgpid_i:950 shelfmark_t:ena")
--- a/geniza/footnotes/models.py
+++ b/geniza/footnotes/models.py
@@ -1,3 +1,4 @@
+import re
 from collections import defaultdict
 from functools import cached_property
 from os import path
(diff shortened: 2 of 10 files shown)


solr_conf/conf/managed-schema
```diff
@@ -104,14 +104,14 @@
...
-      <tokenizer class="solr.StandardTokenizerFactory"/>
+      <tokenizer class="solr.WhitespaceTokenizerFactory"/>
       <filter class="solr.ICUFoldingFilterFactory"/>
       <filter class="solr.StopFilterFactory" words="stopwords.txt" ignoreCase="true"/>
       <filter class="solr.NGramFilterFactory" minGramSize="2" maxGramSize="20" />
       <filter class="solr.LowerCaseFilterFactory"/>
     </analyzer>
     <analyzer type="query">
-      <tokenizer class="solr.StandardTokenizerFactory"/>
+      <tokenizer class="solr.WhitespaceTokenizerFactory"/>
```
rlskoeser, line 114: did it work locally when you changed the configset for your existing Solr core and reindexed? I remember that changing field definitions sometimes used to cause problems

solr_conf/conf/managed-schema
```diff
@@ -104,14 +104,14 @@
     <fieldType name="transcription_text" class="solr.TextField" positionIncrementGap="100" multiValued="true">
     <analyzer type="index">
       <charFilter class="solr.HTMLStripCharFilterFactory"/>
-      <tokenizer class="solr.StandardTokenizerFactory"/>
+      <tokenizer class="solr.WhitespaceTokenizerFactory"/>
```
rlskoeser, line 107: Makes a lot of sense to me that we'd want a less aggressive tokenizer for transcription content, wish I had thought of that when I was working on tuning the search!

Will this cause any problems? e.g., is punctuation at the end of words still removed? (I forget where that typically happens in Solr, and I also don't know how much of a problem it would be for transcription content)

geniza/footnotes/models.py
```diff
@@ -652,6 +653,46 @@ def content_text(self):
         if self.content_html_str:
             return BeautifulSoup(self.content_html_str, features="lxml").get_text()
 
+    @staticmethod
+    def clean_text(text):
```
rlskoeser, line 657: I love how carefully you have documented and tested this. The one thing I wonder is if it would be better to implement this at the solr level. Will this need to be applied anywhere else? Does it need to be applied to query terms as well? 

It looks like all your regexes could be converted to a solr config using the pattern replace filter factory:
https://solr.apache.org/guide/solr/9_3/indexing-guide/charfilterfactories.html#solr-patternreplacecharfilterfactory

geniza/corpus/solr_queryset.py
```diff
@@ -120,18 +120,24 @@ def _search_term_cleanup(self, search_term):
...
-                search_term,
+            exact_phrases = " ".join(
+                [
+                    f"content_nostem:{m}"
+                    for m in self.re_exact_match.findall(search_term)
+                ]
             )
+            # add in judaeo-arabic conversion for the rest (double-quoted phrase should NOT be
+            # converted to JA, as this breaks if any brackets or other sigla are in doublequotes)
+            remaining_query = arabic_or_ja(self.re_exact_match.sub("", search_term))
```
rlskoeser, line 137: is this a bug / limitation in the `arabic_or_ja` implementation? or is it more of an acceptable/reasonable limitation? will there be any user-facing docs that need to be revised after these changes go live?

rlskoeser (approved):
Thanks for your work on this. Looks good to me. My one question is whether the cleaning should happen in solr instead of in python, but even if we think that is a good direction it doesn't need to be done now.

solr_conf/conf/managed-schema
blms: Yes!

solr_conf/conf/managed-schema
blms: Seems like it's actually not, but also doesn't seem to cause any problems. Since this field type is also used for translations, we can make searches like this and see how it behaves differently in that case (description with the standard tokenizer vs. translation with the whitespace—search term `illustrious judge` and `"illustrious judge"` produces this):

Screenshot 2024-02-22 at 4 54 27 PM

It might be worth creating a new field type for translations, actually, but as of now the tokenizer change doesn't seem to be causing problems.

geniza/footnotes/models.py
blms: I think the two primary use cases for now are: either you know exactly how it appears in the transcription, in which case you include sigla with double quotes; or you don't, in which case you just type the word without sigla or double quotes. The only case I can think of to apply it at query time would be if you wanted to include sigla in your search query, but wanted it to match transcriptions without sigla. But I'm not sure how common of a use case it is.

As for whether this is done in solr or python, it's a good question. Personally I don't have a strong preference, but leaning towards Python just because it's already done and maybe not worth the effort to reimplement if we don't need it anywhere else.

geniza/corpus/solr_queryset.py
blms: Team agreed it was an acceptable limitation but that it should be mentioned in the user-facing docs, specifically the "How to Search" page. Thanks for flagging and I'll make a chore issue to document it!

geniza/footnotes/models.py
rlskoeser: ohhh... I wasn't paying close enough attention to the solr changes on my first review; we're now indexing transcription text twice?

That's the one reason it might be worth shifting to solr, because it ought to be possible to support both of these search cases without completely duplicate fields.

That said, I don't think it's worth your time to revise the implementation now. If you can add a comment somewhere around here it might be helpful. If at a later point we're revisiting the indexing (as we may do for transcription once we have more content and if we want better language-specific support), then we might want to revisit this too.

blms: That makes sense, thanks! I'll add the comment.

PR #1555, opened 2024-03-01
Migrate regex from python to solr (#1286)

blms:
As you suggested @rlskoeser, moving the regex to solr was absolutely the way to go here! Not only did it simplify the indexing code and reduce the index size, but it also fixed a bug where a search term without sigla would not be highlighted in a transcription where that term only appears _with_ sigla.


rlskoeser (approved):
Wow! That's fantastic. I'm glad the regexes translated so easily to a solr character filter. Thanks for preserving all the comments explaining what's going on, super valuable.

Issue #1512, opened 2023-12-13
As a content admin, I want an event model that can be linked to multiple people, places, and documents, so that I can recreate a person's timeline and link related data in new ways.


kseniaryzhova:
### Testing notes (QA - round 2)

In the [QA site admin](https://test-geniza.cdh.princeton.edu/admin):
- [x] It should now be possible to add new events from a Person edit page.

### Testing notes (QA)

Visit the [QA site admin](https://test-geniza.cdh.princeton.edu/admin).
- [x] Edit a document. After a hard refresh you should now see a "related events" section near the bottom of the page, with a + button to add an event.
- [x] Clicking this button should bring up a popup in which you can create an event with a name and description.
- [x] Save the event, then "save and continue editing" the document. That event should now be stored under "related events".
- [x] Visit the Events section of the admin. The event you created should appear.
- [x] Edit the event you created. You should be able to attach people, places, and documents to the event.
- [x] Create a new Event from the Events section of the admin. You should not be able to save it without giving it a related document.
- [x] Create and save one with a related document. Visiting that document edit page should now show the event under "related events".
- [x] Edit a person and/or a place. There should now be a "related events" section on each. You should be able to attach related events from this section; however, you should not be able to add a _new_ event from these pages (i.e. no + button). 

---

**Describe the solution you'd like**
I want an event model that has a many to many relationship with people, places and documents. It will be most prevalent on the people pages as it tracks life events. Events are related to dates on documents but are ultimately a different type of information (and there can be many documents reporting an event, and even the date of the event based on documentary evidence is unclear).

blms:
Handful of questions as I'm working on this:

1. Looking back at the original data model diagram I made, I had planned for a field on Events called `certainty` that would allow either "Documentary" or "Inferred." Have we decided that all Events must be supported by documentary evidence, therefore we don't need that field? Or do we also want to allow inferred events?
   - If so, one possible approach to this (rather than a separate field) is that we just allow it to be attached to 0 documents. In that case it's self-evident that it's inferred rather than documentary.
2. Do we need a `notes` field, or is `description` sufficient since contextual data could go there? Or would it be helpful to have a notes field that only appears in the admin?
3. Do any Event relationships with other items need additional / contextual information? Like does an Event-Document relationship need any kind of `type` or `notes` field? (and Event-Place, Event-Person?) Or is it enough to just have the description and name of the event itself?
[...]

PR #1562, opened 2024-03-13
Event model (#1512)

blms:
## In this PR

Per #1512:
- Add an event model with M2M relations with documents, people, and places
- Add inline forms to Document, Person and Place admins for associating events
- When adding an event from the admin, require at least one document
  - Set `min_num` of Documents to 0 when creating an Event directly from a Document page; otherwise, `min_num=1`
  - Prevent new Events from being created via Person or Place admins, per team's request (also would complicate Event-Documents `min_num` code if this were allowed!)
- Auto-generate a date range for each event based on associated documents, if possible, but allow override
  - Find the min and max dates, mixing inferred and on-document, from all associated documents, and generate a standardized date range from that

Other:
- Convert standard date display function into static method for use outside of Document model (basically anywhere we have standardized dates/ranges)

## Notes

Looking especially for review on the Django admin forms trickery, the `standard_date_display` function changes, and the `documents_date_range` function


geniza/entities/models.py
```diff
@@ -68,6 +70,78 @@ def save(self, *args, **kwargs):
...
+                min = full_range[0]
+                if min is None or start_numeric < int(min.numeric_format(mode="min")):
+                    # store as PartialDate
+                    full_range[0] = start
+                # use numeric format to compare to current max, replace if larger
+                end_numeric = int(end.numeric_format(mode="max"))
+                max = full_range[1]
+                if max is None or end_numeric > int(max.numeric_format(mode="max")):
+                    # store as PartialDate
+                    full_range[1] = end
```
rlskoeser, line 136: I had to look back at the other one to remind myself; I remember some of our discussion on that before (and wish I had a better solution to offer you).  

the one thing I wonder is if this could be a utility method that you call in both places and call it on the same content - instead of calling here on the result of dating range, could you call in both places on the dating set instances? (either for one document or all of them).

My other question is a conceptual / semantic one: does any date always take precedence over `None` for either start/end bounds?

geniza/entities/admin.py
```diff
@@ -419,3 +478,66 @@ class PlacePlaceRelationTypeAdmin(TabbedTranslationAdmin, admin.ModelAdmin):
...
+    autocomplete_fields = ("document",)
+    fields = (
+        "document",
+        "document_description",
+        "notes",
+    )
+    extra = 0
+
+    def get_min_num(self, request, obj=None, **kwargs):
+        """Override min_num to change it conditionally based on how Event is being added"""
```
rlskoeser, line 496: I don't quite understand what this is doing or how it is used.  Does it only apply when creating a new event-document relationship?  (Your comments are almost enough but I can't quite get there)

geniza/corpus/dates.py
```diff
@@ -183,15 +183,15 @@ def original_date(self):
             [self.doc_date_original, self.get_doc_date_calendar_display()]
         ).strip()
 
-    @property
-    def standard_date(self):
+    @staticmethod
+    def standard_date_display(standard_date):
```
rlskoeser, line 187: Would this be better as a standalone utility method? it feels a bit odd attached to the document / document mixin now that you're also using it elsewhere.  (maybe slightly simpler as standalone instead of static, IDK)

geniza/entities/models.py
```diff
@@ -68,6 +70,78 @@ def save(self, *args, **kwargs):
...
+        validators=[RegexValidator(DocumentDateMixin.re_date_format)],
+    )
+
+    def __str__(self):
+        """Use the name field for string representations"""
+        return self.name
+
+    @property
+    def date_str(self):
+        """Return a formatted string for the event's date/range, for use in public site"""
```
rlskoeser, line 108: it would be helpful to add a little more content to the docstring to make explicit the logic / fallback order here

rlskoeser (approved):
Generally looks reasonable to me, just a few suggestions and questions.

geniza/entities/admin.py
blms: Updated comments to clarify. The idea here is that you can create an Event directly from a Document edit page, in an inline form, because all Events are supported by documentary evidence. For that reason, Events are also intended to always have at least one related Document. However, if we enforce that requirement at the DB level, or even in the admin using `min_num` out of the box, it would mean that you have to manually add the Event-Document relationship from the popup to save the Event without validation errors—even though you're creating it from within an inline on an associated Document's page, and therefore that relationship will be created as soon as you save the Document.

So it's more or less a convenience thing for content editors/admins who will be creating new Events directly from the popup on Document pages, so that they don't also have to choose that same Document in a dropdown within the popup to prevent validation errors.

The one caveat, which I've now noted, is that it's theoretically possible to create an Event from this inline popup and then _not_ save the Document, or delete the relationship with that new Event before saving the Document. In those cases, the Event would be orphaned, i.e. created without any relationship with a Document. But I figure that tradeoff is worth avoiding confusion ("why do I have to add a document to the Event when i'm creating the Event _from_ a document page?")

geniza/entities/models.py
blms: > the one thing I wonder is if this could be a utility method that you call in both places and call it on the same content - instead of calling here on the result of dating range, could you call in both places on the dating set instances? (either for one document or all of them).

Not exactly that, since we need to also consider each of the documents' `document_date` and not just their inferred `Dating` sets.  However, the comparison between two `PartialDate` ranges ends up being essentially the same code, so that can be factored out into a utility method! (I moved it to a static method on `PartialDate` so that it's clear the arguments should both be sets of `PartialDate` instances.)

> My other question is a conceptual / semantic one: does any date always take precedence over None for either start/end bounds?

Not sure I fully understand, but I think the answer is yes; we don't support unbounded date ranges, either on-document or inferred.

geniza/corpus/dates.py
```diff
@@ -126,6 +126,27 @@ def numeric_format(self, mode="min"):
         See :meth:`isoformat` for more details."""
         return self.isoformat(mode, "numeric")
 
+    @staticmethod
+    def get_date_range(old_range, new_range):
+        """Compute the min and max dates between two PartialDate ranges."""
```
rlskoeser, line 131: this is the widest possible range covered by the two ranges, correct? if you think of the date ranges as sets, this is the union rather than the intersection?

asking to make sure I understand - I think a slight wording change in the comment would make this clear.

rlskoeser (approved):
I'm glad my questions prompted a refactor on the date range calculation; it's good that at least some of that logic can be shared.

Thanks for clarifying the logic and complexity of the documents and events; it sounds reasonable enough. I'm glad you flagged the orphaned event possibility in the comments so there's some record.

Issue #965, opened 2022-07-20
As a transcription editor, I should see an error if I try to update an annotation with out of date content so that I don't overwrite someone else's changes.


rlskoeser:
## testing notes (QA)

On the QA site:
- [x] Edit a transcription or translation. Open the same one for editing in two different windows, in order to mimic two people working on the same transcription simultaneously.
  - [ ] If someone else is available to test actually editing simultaneously, that would be an even better test!
- [x] Scroll down to ensure the transcription is fully loaded in both windows.
- [x] Edit a block of text and save it in one of the windows. The second window should _not_ reflect this change.
- [x] Start editing the block of text in the second window, and then save it. You should receive an error that someone else edited the text and you must refresh to get the updated text.
- [x] Try deleting the block of text in the second window. You should get a similar error message.
- [x] Try refreshing the page in the second window. You should now see the updates that you made in the first window, and be able to save/delete freely.

---

Documenting this because I'm putting off implementing E-tags in the new annotation server code and we don't support it in tahqiq yet anyway. We can probably work around it with workflow / assignments, but it would be good to implement in the longer term.

## dev notes
- [x] add etag logic to annotation detail view; adapt solr last modified mixin to make a more generic version that will let us make use of django's existing conditional response behavior
- [x] update tahqiq to store and update etag when requesting annotations, and include etag in the request header when making changes (post, delete)
- [x] add warning/error display to transcription editor to report problems with saving or deleting records

kseniaryzhova:
@rlskoeser would you like to pass this off to Ben?

rlskoeser:
@kseniaryzhova we could discuss — I'm not sure how much of a priority this needs to be; you may be able to address with workflow. You might also want to get an estimate from Ben to help decide.

kseniaryzhova:
@blms Any thoughts? How long would this take? And how fatal would an override be? We don't often work on the same document, at least not on the transcription/translation, which is usually entered in one go, copied over from Google Docs.

blms:
@kseniaryzhova I would probably estimate somewhere in the ballpark of 15-20 hours for this. I would consider it low priority if it's not likely to be encountered. The potential error is if someone is working on a transcription/translation and it's edited by someone else in the meantime. If they click save, the other person's edits will be overwritten. Though, the overwritten edits should still be accessible via the automated backups to [the github repo](https://github.com/princetongenizalab/pgp-text).

kseniaryzhova:
[...]

PR #1565, opened 2024-03-19
Use Django conditional response to prevent mid-air collisions (#965)

blms:
## In this PR

Per #965:
- Add method to compute an Annotation's ETag (md5 hash of content)
- Include ETag in `compile` method for use in annotation lists (required in order to associate a single annotation with a single ETag when retrieved via a list)
- Add conditional response for `POST`/`DELETE` using Django `@condition` in a mixin, in order to return 412 when computed ETag does not match `If-Match` header
- Update Tahqiq to version in https://github.com/Princeton-CDH/annotorious-tahqiq/pull/32
- Improve styles for alerts, and make them last longer if they are errors

### DIFF
--- a/geniza/annotations/models.py
+++ b/geniza/annotations/models.py
@@ -1,3 +1,5 @@
+import hashlib
+import json
 import re
 import uuid
 from collections import defaultdict
@@ -193,7 +195,7 @@ class Annotation(TrackChangesModel):
         and the data will not be saved.
         """
         # remove any values tracked on the model; redundant in json field
-        for val in ["id", "created", "modified", "@context", "type"]:
+        for val in ["id", "created", "modified", "@context", "type", "etag"]:
             if val in data:
                 del data[val]
 
@@ -247,6 +249,14 @@ class Annotation(TrackChangesModel):
         else:
             return cleaned_html
 
+    @property
+    def etag(self):
+        """Compute and return an md5 hash of content to use as an ETag"""
+        # must be a string encoded as utf-8 to compute md5 hash
+        content_str = json.dumps(self.content, sort_keys=True).encode("utf-8")
+        # ETag should be wrapped in double quotes, per Django @condition docs
+        return f'"{hashlib.md5(content_str).hexdigest()}"'
+
     def compile(self, include_context=True):
         """Combine annotation data and return as a dictionary that
         can be serialized as JSON.  Includes context by default,
@@ -258,6 +268,11 @@ class Annotation(TrackChangesModel):
         anno = {}
         if include_context:
             anno = {"@context": "http://www.w3.org/ns/anno.jsonld"}
+        else:
+            # NOTE: ETag required here for inclusion in annotation list, which is how
+            # annotations are fetched during editing; need to associate each ETag with
+            # an individual annotation, for comparison with response ETag on POST/DELETE
+            anno = {"etag": self.etag}
 
         # define fields in desired order
         anno.update(
--- a/geniza/annotations/tests/test_annotations_views.py
+++ b/geniza/annotations/tests/test_annotations_views.py
@@ -8,9 +8,59 @@ from django.urls import reverse
 from pytest_django.asserts import assertContains, assertNotContains
 
 from geniza.annotations.models import Annotation
-from geniza.annotations.views import AnnotationResponse
+from geniza.annotations.views import AnnotationLastModifiedMixin, AnnotationResponse
 from geniza.corpus.models import Document
 from geniza.footnotes.models import Footnote, Source, SourceType
+
+
+class TestAnnotationLastModifiedMixin:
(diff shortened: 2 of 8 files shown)


geniza/annotations/models.py
```diff
@@ -247,6 +249,14 @@ def sanitize_html(cls, html):
         else:
             return cleaned_html
 
+    @property
+    def etag(self):
+        """Compute and return an md5 hash of content to use as an ETag"""
+        # must be a string encoded as utf-8 to compute md5 hash
+        content_str = json.dumps(self.content, sort_keys=True).encode("utf-8")
```
rlskoeser, line 256: is hashing `content` sufficient? is there anything stored outside of the content that should invalidate the etag?

geniza/annotations/models.py
```diff
@@ -247,6 +249,14 @@ def sanitize_html(cls, html):
...
+        # ETag should be wrapped in double quotes, per Django @condition docs
+        return f'"{hashlib.md5(content_str).hexdigest()}"'
```
rlskoeser, line 258: I didn't remember this and it isn't obvious to me from the docs that it needs to be wrapped this way. I would have thought the django middleware would generate the header as needed. I presume you have tested this and it does the right thing?

geniza/annotations/tests/test_annotations_views.py
```diff
@@ -8,11 +8,61 @@
...
+        mixin = AnnotationLastModifiedMixin()
+        # nonexistent annotation should return None
+        assert mixin.get_last_modified(request=None, pk=1234) == None
+        # otherwise should return last modified
+        assert (
+            mixin.get_last_modified(request=None, pk=annotation.pk)
+            == annotation.modified
+        )
+
+    def test_disaptch(self, admin_client, annotation):
```
rlskoeser, line 34: typo in your test method name :-) 
```suggestion
    def test_dispatch(self, admin_client, annotation):
```

geniza/annotations/views.py
```diff
@@ -30,6 +31,38 @@ class ApiAccessMixin(AccessMixin):
     raise_exception = True  # return an error instead of redirecting to login
 
 
+class AnnotationLastModifiedMixin(View):
```
rlskoeser, line 34: was a little surprised you need a whole mixin for this, but I guess it's nice to have the conditional logic encapsulated in one place and separate from the main annotation detail view

geniza/annotations/views.py
```diff
@@ -30,6 +31,38 @@ class ApiAccessMixin(AccessMixin):
...
+        except Annotation.DoesNotExist:
+            return None
+
+    def get_last_modified(self, request, *args, **kwargs):
+        """Return last modified :class:`datetime.datetime`"""
+        try:
+            anno = Annotation.objects.get(pk=kwargs.get("pk"))
+            return anno.modified
+        except Annotation.DoesNotExist:
+            return None
```
rlskoeser, line 51: is there a way to get the annotation object once instead of twice?

rlskoeser (approved):
Looks good, glad to see this implemented.

geniza/annotations/models.py
blms: I'm going off the decorator's docstring (which appears on hover in my editor)—and you're right! Looks like Django would wrap it anyway, but I suppose we'll save the tiny amount of computation required to do that.

https://github.com/django/django/blob/ae10146793dca0e0594c7acdee20ca3810983f39/django/views/decorators/http.py#L94-L97

rlskoeser: It's more the readability of the code than the computation. Thanks for checking!

geniza/annotations/models.py
blms: Nothing else can be changed by the editor, except for `modified`, which shouldn't invalidate the etag if the content doesn't also change. (I think `canonical`, `via`, and `footnote` are theoretically mutable, but can't be changed by the editor and would require database access or the django console to change.)

rlskoeser: That makes sense! Thanks for confirming. Worth noting this in a comment in the code somewhere? (worth noting somewhere to keep an eye on this in case it ever changes / try to keep it from changing)

geniza/annotations/views.py
blms: It may not be necessary; was going off the initial idea [here](https://github.com/Princeton-CDH/geniza/issues/965#issue-1311261872) but I guess it sort of lost the generic-ness and became about annotations… I suppose the only way to make the mixin generic would be the dispatch function alone, which is pretty small.

geniza/annotations/views.py
blms: Refactored a bit based on this question and the one about the mixin—let me know if you think the changes seem reasonable.

Issue #1566, opened 2024-03-19
As a public site user, I want a page to view details about an individual person, so that I can learn more about their life and presence in documents.


blms:
### testing notes (QA)

On the QA site, navigate to the [list of people](https://test-geniza.cdh.princeton.edu/en/people).
- [x] Click on a person with a link. It should navigate to a Person Detail page.
- [x] This page should match the design below. (No life events, bibliography, citation, or permalink yet.)
- [x] The tabs for related documents, people, and places should be disabled, but they should show accurate counts.

---

Design here: https://www.figma.com/file/mPu59EJcveUnltjlll3twI/Princeton-Geniza-Project?type=design&node-id=768-10996&mode=design&t=d1QOs8BP588SiFnE-4

kseniaryzhova:
@blms looks great, thank you, closing!

PR #1567, opened 2024-03-20
Add basic Person Details page (#1381, #1566)

blms:
## In this PR

Per #1566:
- Add a basic person details page (using designs by Chelsea) with metadata (names, description, social role, etc)
- Include tabs for relationships pages but make them inactive for now

Per #1381:
- Use `get_queryset` to disable pages for people with less than 10 associated documents, unless `has_page=True`

## Notes

Just looking for a general sanity check on the templates here, since it's been some time since I've done design implementation.


geniza/entities/templates/entities/person_detail.html
```diff
@@ -0,0 +1,70 @@
+{% extends 'base.html' %}
+{% load static i18n render_bundle_csp fullurl %}
+
+{% block meta_title %}{{ page_title }}{% endblock meta_title %}
+{% block meta_description %}{{ page_description }}{% endblock meta_description %}
+{% block extrameta %}
+    {% spaceless %}
+        {% for lang_code in PUBLIC_SITE_LANGUAGES %}
+            {% language lang_code %}
+                <link rel="alternate" hreflang="{{ lang_code }}" href="{% fullurl "entities:person" person.pk %}" />
```
rlskoeser, line 10: not strictly related but question: I saw an alert in google search console that we have alternate links without a canonical; do you know if we're declaring a canonical link anywhere in the header or how we would determine that one? do we have to declare one language as canonical?

geniza/entities/tests/test_entities_models.py
```diff
@@ -294,6 +294,11 @@ def test_merge_with_log_entries(self):
             in moved_log.change_message
         )
 
+    def test_get_absolute_url(self):
+        # should get person page url in user's language by pk
+        person = Person.objects.create()
+        assert person.get_absolute_url() == "/en/people/%s/" % person.pk
```
rlskoeser, line 300: Can you add a slug field for use in urls? I have a strong preference for not using database id numbers for public site urls.  We can discuss more if needed/helpful - I haven't done on a multilingual site before, so you might want multiple slugs. We have logic on other projects to track past slugs when it's possible they could change, so you can resolve if something changes (both for datasets and for webapp urls).

geniza/entities/templates/entities/person_detail.html
```diff
@@ -0,0 +1,70 @@
...
+        </section>
+        {% if person.description %}
+            <section class="description">
+                <h2>
+                    {# Translators: label for person description / bio #}
+                    {% translate 'Description / Bio' %}
+                </h2>
+                <p>{{ person.description }}</p>
+            </section>
+        {% endif %}
```
rlskoeser, line 68: probably depends on your designs and decisions made elsewhere, but should a person have a permalink similar to what we have on document pages?

geniza/entities/templates/entities/person_detail.html
```diff
@@ -0,0 +1,70 @@
...
+                    </dt>
+                    <dd>{{ person.role }}</dd>
+                {% endif %}
+            </dl>
+            {# secondary metadata #}
+            <dl class="metadata-list secondary">
+                {% if person.names.count > 1 %}
+                    <dt>
+                        {# Translators: label for alternative names for a person #}
+                        {% blocktranslate count counter=person.names.count|add:-1 trimmed %}
```
rlskoeser, line 46: would it make sense to implement a custom queryset for non-primary names so you could easily filter without all the extra logic here in the template?

geniza/entities/tests/test_entities_models.py
rlskoeser: I recognize that we're using database ids for PGPIDs but those are an exception to me since we're preserving / continuing an existing identifier system that existed previous to PGPv4 and also because there's not a reliable way we could generate short, brief, stable slug-type identifiers for documents.

rlskoeser (approved):
Template code looks reasonable to me. Been a while since I had to look at all the translation stuff! Seems like you've got clear comments everywhere.  LMK if there's more feedback you're looking for.

geniza/entities/templates/entities/person_detail.html
blms: Good question! Looks like on document pages, we have `document.permalink` as the canonical page, and that property is generated by specifically stripping out the language from the URL. Not sure if that's how it's meant to be, since that will cause an immediate redirect to one with a language (presumably to whatever the crawler's language is), and therefore isn't necessarily stable in content. It looks like Google in particular does not like 3xx redirects on canonical pages, and prefers the result of the redirect as the canonical URL. So it may make sense to choose a language (presumably English) as canonical instead, on both documents and people.

rlskoeser: oh wow, is that what Google has been complaining about with the non-canonical?! good catch. I wonder if we can just add the `rel=nofollow` attribute to tell bots not to crawl it?  I do think that for SEO purposes our English pages should probably be the canonical version...

My intention was a language-agnostic permalink for citation purposes. I don't like the idea of privileging one language as canonical when citing the resource; if you're citing a PGP document, are you citing the English version of the description of that document? (maybe, sometimes) Are you citing the document itself? If you and someone else both reference a PGP document but looked at different versions of the localized site, should they be different urls or not? I can do some looking around and asking to see if there are any best practices for multilingual digital scholarship. Would be interesting to hear what Geniza Lab team members think about this.  (I'm thinking of the permalink more as a URI, if that make sense.)

I'm not sure if the same considerations apply to people as they do to documents.

geniza/entities/templates/entities/person_detail.html
blms: Good question, I'll put it on the agenda for our next meeting.

rlskoeser (approved):
Looked through the changes since my last review, everything seems reasonable.

Glad you resolved the tests you had to disable in github actions previously.

geniza/annotations/tests/test_annotations_migrations.py
```diff
@@ -86,7 +86,7 @@ def test_footnote_created(self):
         ).exists()
 
 
-@pytest.mark.second_to_last
+@pytest.mark.order("second_to_last")
```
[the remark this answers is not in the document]
blms: Yes, although to be precise it's actually a complete library change! This switches the library from the no longer maintained [`pytest-ordering`](https://github.com/ftobia/pytest-ordering) to its successor, [`pytest-order`](https://github.com/pytest-dev/pytest-order/).

blms:
Thanks @rlskoeser! Leaving open for now so that we can get the last of the search issues tested and release to production before merging any of the new public-facing design implementation to the dev branch.

geniza/annotations/tests/test_annotations_migrations.py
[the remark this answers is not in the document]
rlskoeser: I saw the change in the requirements file but thought I'd wait for your response. Did not know that the first was not being maintained anymore! Good to know.

Issue #1570, opened 2024-03-25
Incorrect words are highlighted in complete word quotation search (Hebrew script)


kseniaryzhova:
### testing notes

In the [QA site](https://test-geniza.cdh.princeton.edu/en/documents/):
- [x] The below search `"אלממ"` should highlight as expected in search results, when using double quotes.

---

**Describe the bug**
Quotation search of words I know exist in the corpus brings up inexact matches as well.

**To reproduce**
2. Search for `"אלממ"` in double quotes
3. Look at results 8, 10, 16 (1 and 3 are weird too)

**Expected behavior**
Quotation search in any script should bring up exact matches only.

blms:
This is a challenging solr highlighting issue. We have an exact match (unstemmed) field, and exact searches are matched against that. 

But in the highlighting portion of that query, the highlighter uses edismax to search all query fields:

https://github.com/Princeton-CDH/geniza/blob/782fdd94906c61d6e466c93ae23f94e52212bc65/geniza/corpus/solr_queryset.py#L161-L167

Also, the `content_nostem` field isn't present in the highlighted fields:

https://github.com/Princeton-CDH/geniza/blob/782fdd94906c61d6e466c93ae23f94e52212bc65/geniza/corpus/views.py#L176-L195

Simply adding `content_nostem` with `.highlight()` doesn't do the trick, no matter how I mess with the highlight query formation.

@rlskoeser Any ideas off the top of your head as to how I might approach this? Is there a way we can get the exact match highlighted when partial matches are also present in the result?

rlskoeser:
@blms I suspect that you are not getting highlighting back on `content_nostem` because it's indexed but not stored; I'm looking at the fields in the managed-schema file because I had to refresh my memory on this.  You do have to have an active keyword search on the field as well.  Even if you change that field so solr will store them I'm not sure if that fully solves your problem, because right now you have a single `content_nostem` field - I think you would need to revise copy field behavior and define separate fields (`description_nostem` and `transcription_nostem`); then you have to query across all of them, return highlighting for all of them, and then display in priority order (nostem matches first if there are any, then the other copy of the field).

It does make me feel like we're approaching the problem wrong. It seems to go against how Solr is meant to be used, and it it seems increasingly complicated; although I do remember that it took us a while to arrive at the nostem field as a solution for this problem.  

Because of this complexity there may be some trade-offs with the exact searching and highlighting that will be difficult to solve.

blms:
[...]

PR #1573, opened 2024-04-03
Highlight nostem matches on exact searches (#1570)

blms:
## In this PR

Per #1570:
- Index and store nostem fields
- Split `content_nostem` into `description_nostem` and `transcription_nostem`
- Convert exact searches from scoped `content_nostem:` to `OR` between scoped description/transcription nostem fields
- Highlight new description/transcription nostem fields
- Replace `description` and `transcription` highlights in highlights dict with `description_nostem` and `transcription_nostem` highlights when searching for exact matches

### DIFF
--- a/geniza/corpus/solr_queryset.py
+++ b/geniza/corpus/solr_queryset.py
@@ -76,6 +76,8 @@ class DocumentSolrQuerySet(AliasedSolrQuerySet):
         "has_digital_translation": "has_digital_translation_b",
         "has_discussion": "has_discussion_b",
         "old_shelfmark": "old_shelfmark_bigram",
+        "transcription_nostem": "transcription_nostem",
+        "description_nostem": "description_nostem",
     }
 
     # regex to convert field aliases used in search to actual solr fields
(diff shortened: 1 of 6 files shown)


geniza/corpus/solr_queryset.py
```diff
@@ -256,8 +259,17 @@ def get_highlighting(self):
...
         for doc in highlights.keys():
-            if "transcription" in highlights[doc]:
+            # _nostem fields should take precedence over stemmed fields in the case of an
+            # exact search; in that case, replace highlights for stemmed fields with nostem
+            if is_exact_search and "description_nostem" in highlights[doc]:
+                highlights[doc]["description"] = highlights[doc]["description_nostem"]
+            if is_exact_search and "transcription_nostem" in highlights[doc]:
+                highlights[doc]["transcription"] = [
+                    clean_html(s) for s in highlights[doc]["transcription_nostem"]
+                ]
```
rlskoeser, line 271: I had been thinking that you would include all the highlights in the context and then the template could display them in priority order.  I think it probably doesn't matter, maybe it's a little cleaner to do here.

rlskoeser (approved):
I trust you if you say this fixes the problem you've been seeing.

As we've discussed, I still think it's a brittle and overly complicated approach. I think the separate fields and highlighting means that you can't use exact searching in combination with other kinds of searches that would normally result in highlighting - although it sounds like that's the trade off we want at this point.

You've documented this clearly and tested the expected functionality, which is really helpful and will make it easier if and when we revisit the implementation later.

Issue #1571, opened 2024-03-27
As a public site user, I want to be able to filter people in the browse page by gender, social role, and relation to documents, so that I can narrow down my browse results


blms:
### testing notes (QA)

On the QA site, navigate to the [list of people](https://test-geniza.cdh.princeton.edu/en/people).
- [x] Click "Filters". The filter button should change colors and open a panel of filters.
- [x] There should be filters for gender, social role, and document relations. Each option should have counts.
- [x] Clicking filters should narrow down the results.
- [x] Clicking multiple options within the same filter should broaden the results to "people matching any of the selected criteria" for that filter.

---

Design here: https://www.figma.com/file/mPu59EJcveUnltjlll3twI/Princeton-Geniza-Project?type=design&node-id=776-80765&mode=design&t=G94VYkwBxY12RovA-4

kseniaryzhova:
@blms works as it should! Closing!

PR #1574, opened 2024-04-08
Add options to filter the people browse page (#1571)

blms:
## In this PR

Per #1571: 
- Add filters to person list view

Other:
- Fix some lighthouse errors
- Fix codecov issue by adding token to workflow

## Review notes

- Mostly looking for a review of the list view class, and secondarily the form class
- On the view class, for a while I was getting an error where selecting one facet of a multi-valued field (in this case, document relation type) would 0 out the counts for all the other facets, even though some of the objects had more than just one type in the DB. I resolved this with just a brand new queryset [using `pk__in`](https://github.com/Princeton-CDH/geniza/pull/1574/files#diff-ca444522251a1984d3a28f9a948dc24af332f0127c782deea36c650fa7a31a6aR241-R244) but maybe there's a better way.


geniza/entities/views.py
```diff
@@ -176,27 +179,105 @@ class PersonListView(ListView):
...
 
+        form = self.get_form()
+        # bail out if form is invalid
+        if not form.is_valid():
+            return people.none()
+
+        # filter by each supported field
+        search_opts = form.cleaned_data
+        self.applied_filter_count = 0
+        if "gender" in search_opts and search_opts["gender"]:
```
rlskoeser, line 206: a bit more readable and efficient to do this way:
```suggestion
        if search_opts.get("gender"):
```

geniza/entities/views.py
```diff
@@ -176,27 +179,105 @@ class PersonListView(ListView):
     # Translators: description of people list/browse page
     page_description = _("Browse people present in Geniza documents.")
     paginate_by = 50
+    form_class = PersonListForm
+    applied_filter_count = 0
+
+    # fields to facet
+    facet_fields = ["gender", "role__name", "persondocumentrelation__type__name"]
```
rlskoeser, line 186: Just to make sure I'm understanding, you're doing db-only search and faceting? If so, seems like it's worth a brief comment to make that explicit.

Did you decide that would be easier than implementing with solr? Or is it because you're only filtering and not doing keyword search?

rlskoeser (commented):
I think I understand your question about the facet and I could see how that would happen with a db query - django joins on the row that matches and skips the other ones. It's not immediately obvious to me how you could do that without querying as you're doing.

I do wonder if instead of a facet field it would work to use a model choice field backed by a model, and then all the form has to do is update the counts; you could assume zero if not set/passed in.  I think this would save you the extra query where you're getting all the zero-count values.

I have a multi-choice field populated by a model in the prosody archive code base, although it doesn't include counts: https://github.com/Princeton-CDH/ppa-django/blob/main/ppa/archive/forms.py#L246-L255

geniza/entities/views.py
blms: Exactly the latter—only filtering, so a system with analysis and indexing didn't seem necessary. I didn't actually consider using solr! Maybe it could have saved time.

As for the other piece of the model choice field, I worked a bit on that and I seem to essentially end up at the same number of queries. I might be missing something, but it didn't feel worth it to spend any more time on. But I'll definitely revisit if it has performance problems, so thank you for the idea and the link to ppa-django!

Issue #1614, opened 2024-07-02
As a public site user, I want the documents search page to match the designs of the People and Places browse pages, so that I am not confused by inconsistency.


blms:
## testing notes (QA)

On the QA public site, navigate to the [documents search page](https://test-geniza.cdh.princeton.edu/en/documents/)
- [x] It should match the designs below
- [x] Filters and sort should work as before
- [x] Should include a "no image" placeholder when there are no images
- See also: #1347

---

Designs here: https://www.figma.com/design/mPu59EJcveUnltjlll3twI/Princeton-Geniza-Project?node-id=1480-126818&t=H3S1pUpx52ZsBXpn-4

richmanrachel:
Looks great! Closing :)

PR #1618, opened 2024-07-18
Implement the document search page redesign (#1614)

blms:


blms:
Though some designs will be revised, merging this PR to prevent blocking regex search, as none of the revisions involve functionality.

Issue #1716, opened 2025-01-09
Manage python requirements in a pyproject.toml file


rlskoeser:
This update should replace top-level `requirements.txt` and all the text files under `requirements/` folder.

It may be helpful to refer to this packaging guideline https://learn.scientific-python.org/development/guides/packaging-simple/ and the official pyproject documentation https://packaging.python.org/en/latest/guides/writing-pyproject-toml/

Here's an example pyproject.toml file from one of our other CDH django projects: https://github.com/Princeton-CDH/mep-django/blob/main/pyproject.toml

You will have to update the github actions python installation for this change.

richmanrachel:
Mohamad finished this.

PR #1792, opened 2025-05-06
Packaging using toml

mabdellatif88:
**Associated Issue(s):** #

### Changes in this PR
To setup the app, this PR proposes to [replace](https://github.com/Princeton-CDH/geniza/issues/1716) requirements files with toml file for packaging and setup.

- Replacing the requirements directory by pyproject.toml file
- Updating the workflow scripts and documentation text to use it instead

### Evaluation
_After replacing the requirements files by the toml file, the health of the app is evaluated using two measures._

- [ ] Listing all the installed packages in the python environment using the toml file (after the change) and comparing it to the one generated using the requirements files (before the change). The later is a subset of the former
- [ ] All the tests run healthy on the branch


scripts/README.md
```diff
@@ -3,7 +3,7 @@
 This directory contains stand-alone scripts associated with
 Princeton Geniza Project that are not part of the web application proper.
 
-Requirements for these scripts can be found in `requirements/scripts.txt`.
+Requirements for these scripts can be found in `scripts/scripts_requirements.txt`.
```
rlskoeser, line 6: Ah, I'd forgotten this one even existed. Glad you handled it.

pyproject.toml
```diff
@@ -1,5 +1,107 @@
+[build-system]
+requires = ["setuptools", "setuptools-scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "geniza"
+description = "Cairo Geniza - Python/Django web application"
+requires-python = ">=3.8"
```
rlskoeser, line 8: Is 3.8 still the minimum? (Not that it probably matters very much)

rlskoeser (approved):
All the changes look good to me. What a nice confirmation to have all the GitHub Actions passing.

pyproject.toml
mabdellatif88: May be, not sure. I think it is ok to announce 3.9 as the minimum and work towards a more recent version of python (e.g. 3.12) going forward, it is ok to leave 3.8 behind. Thanks for noticing we still have 3.8 written here! we should update it to 3.9 for now

Issue #1822, opened 2025-06-30
Upgrade djiffy to 0.9.2 and run django migrations on QA and production instances


blms:
@rlskoeser due to the stepped approach required for the djiffy upgrade, we'll need to do this before @mabdellatif88 completes his upgrades. I figured I could use the opportunity to walk him through the deploy and release process.

My thinking is:
- [x] I'll open a PR with the dependency upgrade
- [x] We'll merge that, and I'll get on a call with Mohamed and walk him through the process of a QA deploy

We could then just do a production release, or I could write testing notes and we could go through the whole testing process as well. Not sure if it makes sense to test at that point, besides checking that IIIF images still work, since we'll need to test it again once we upgrade to 1.0.0 with the rest of the upgrades.

I assume this would ultimately be a patch release.

Let me know what you think!

blms:
@rlskoeser Sorry for the confusion! I mean this (from [the djiffy readme](https://github.com/Princeton-CDH/djiffy?tab=readme-ov-file#upgrading-to-v100)):

> If you have used any version of djiffy prior to v1.0.0, then it is required to upgrade to v0.9.2 first, and run all migrations on your Django project. The Django project must be running Django version 3.2 at this point.
> 
> After you have run those migrations, you can upgrade to any Django version 3.2+, and install djiffy v1.0.0.

If I recall correctly, we need to run these migrations on all Geniza instances—meaning production as well—before we complete the upgrade to djiffy 1.0.0, and thus, django 4. So I think we'll need to do a patch production release in order to run the djiffy 0.9.2 migrations.

rlskoeser:
@blms oh, right, because of the json field migration! Yes, then a small release with the upgrade makes sense to me.

blms:
Closing this as it's live in QA right now. We will need to do a production release after the other two issues in QA pass (#1793 and #1812).

Issue #1821, opened 2025-06-30
resolve attrdict / addict dependency issues


rlskoeser:
Some of the CDH python libraries used in the geniza codebase depend on attrdict, but that is no longer being maintained and is not supported for newer versions of Python.  In at least one case I migrated a library to depend on addict instead, but then later learned there was an attrdict3 (I believe) that might be a drop-in replacement for attrdict (as long as it's being maintained).  Probably worth a little bit of looking at the options to see which is more likely to be maintained on the longer term.

- [x] decide whether to use addict or attrdict3 for geniza code
- [x] add to geniza python dependencies
- [x] update code to use the new version

blms:
- @mabdellatif88 and I found that neither is particularly recently updated (2020 for addict, 2022 for attrdict3).
- However, we're only using attrdict for two things:
  1. Checking the type of piffle data to see if it's an `AttrMap`, and converting it to a dict for json serialization if so
  2. Generating piffle-like data for unit tests
  - Since piffle uses addict, I think it makes the most sense to use that here too.

Issue #853, opened 2022-05-11
Upgrade the python, Django, wagtail stack for Geniza 


rlskoeser:
## Testing notes (QA)

On the [test site](https://test-geniza.cdh.princeton.edu/), test the following categories to ensure that they still function as before.

In cases where things don't work as expected, check if they work the same way on the [production site](https://geniza.princeton.edu/) (except for destructive actions that would impact production data!). If they are also broken in production, file a new bug issue. If they are only broken in the test site, leave a comment here.

Note that production data was synced around 1:00pm ET on July 15, so any data changes made since then won't be reflected.

#### General
- [x] Use the light/dark switch
- [x] Use the language switch
- [x] Check all content pages (About, How To). The visual styles and content should match production
- [x] In the [Wagtail CMS](http://test-geniza.cdh.princeton.edu/cms), you should be able to add, edit, and delete content pages
- [x] Header and footer should appear as in production; links should work as expected

#### Search pages
- [x] general and regex searches on Documents
- [x] search results should show any: description, metadata, images, transcription/translation content
- [x] keyword search, filters, and sort on People
- [x] grid view/list view switcher on People
- [x] map and sort for Places
- [x] clicking on Place markers should scroll you to the place

#### Document detail pages
- [x] Navigate to a document page. Its look should match production, data should not be missing (including shelfmark, clickable tags list, "what's in the PGP")
- [x] image viewer: zoom, pan, rotate
- [x] navigate between multiple transcriptions and translations
- [x] open transcriptions and translations side by side and they should align
- [x] use the transcription editor; it should work as before (barring instances of the bug #1824)
- [x] Add a new transcription or translation and the Source autocomplete that appears should work
- [x] Select Bibliography tab
- [x] Related Documents tab

#### Person detail pages
- [x] Description, life events, citation, link, clickable social role, alternate names should all appear if present
- [x] Related People network graph; pan, zoom, click (when links appear), hover lines
- [x] Related Documents list/sorting
- [x] Related People list/sorting
- [x] Related places map
- [x] Related places list/sorting

#### Place detail pages
- [x] Map, coordinates, other name
- [x] Related documents list/sorting
- [x] Related people list/sorting

#### General
- [x] Filtering, searching, and sorting sections for Documents, Fragments, Sources, Footnotes
- [x] Merge tags in the Tags list (select w/ checkbox, then dropdown at the top)
- [x] Merge documents: data visible should match production site; merge should succeed
[...]

PR #1823, opened 2025-07-01
Upgrade djiffy to 0.9.2 for interim jsonfield migration

blms:
**Associated Issue(s):** #1822

Requesting review from @rlskoeser on this one to make sure the additional required changes make sense; and from @mabdellatif88 to keep him informed and give him the opportunity to ask questions, and get a sense of how I typically structure these upgrade PRs.

### Changes in this PR

- Upgrade djiffy to 0.9.2 so that we can complete the jsonfield migration before the 1.0 upgrade
   - Bump piffle to required 0.6, resolve `piffle.iiif`
   - Remove rdflib dependency, which was conflicting (we don't use it directly)
   - Pin `django-autocomplete-light` until it is unpinned in djiffy
   - Regenerate fixtures (`ui_ux_test_documents.json`): the old fixtures were using the old `jsonfield` djiffy model which stored json as strings, so they have now been converted to proper django JSONFields where the json is stored as actual json
- Add `attrdict` to dependencies as it's no longer a sub-dependency of one of the upgraded packages. We will need to replace it in #1821
- Fix a broken unit test `test_iiif_images_iiifexception`

### Notes

- @mabdellatif88 I think the unit test `test_iiif_images_iiifexception` that we'd deleted on the call yesterday actually was written incorrectly, and I'm not sure how it passed before. We `except` the `IIIFException` in the tested function, so it shouldn't be raised, and `assertRaises` should be false. I think it is useful to keep after all, since what we're actually testing is the "log at level WARN" when an exception _is_ raised.
- Fixture generation isn't currently working due to some foreign key conflicts between generated fixture data and data introduced in migrations. If fixtures ever need to be regenerated, this code will likely need to be updated. (in this PR I replaced the old entries manually)

### DIFF
--- a/geniza/corpus/fixtures/ui_ux_test_documents.json
+++ b/geniza/corpus/fixtures/ui_ux_test_documents.json
@@ -1,50 +1,44 @@
 [
   {
     "fields": {
-      "created": "2022-01-20",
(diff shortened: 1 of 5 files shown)


mabdellatif88:
Glad the [maintenance work](https://github.com/Princeton-CDH/geniza/issues/853) cleared off the dust on all these
Everything you did here makes sense to me, I have no handover-related questions. However, I have a comment on a modification you made to the toml file. I noticed you pinned django-autocomplete-light to 3.9.7. Is there a reason why not >= instead of pinning (==)?

PR #1842, opened 2025-07-30
Upgrade bugfix

mabdellatif88:
**Associated Issue(s):** #853 

### Changes in this PR
_Fix the behavior of checkmarks when updating a footnote_

- Changed ul to div in 3 spots

### Notes

- Only frontend (JS) modifications

### Reviewer Checklist
The expected behavior is tested

- [x] (manually) on Chrome
- [x] (manually) on Safari
- [x] The change didn't result in breaking any of the existing tests

### DIFF
--- a/sitemedia/js/admin.js
+++ b/sitemedia/js/admin.js
@@ -41,11 +41,13 @@ window.addEventListener("DOMContentLoaded", () => {
     // Digital Edition + Edition checkboxes
     let existingDocRelations = document.querySelectorAll(
         // ^='footnote' captures both Document and Source footnote inlines
-        "div[id^='footnote'] td.field-doc_relation ul"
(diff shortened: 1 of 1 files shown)


blms (changes requested):
Testing locally, this seems to work well on the Footnote edit page, but runs into errors in the Source and Document edit pages.

To replicate the error: create a new Source or Document record through the admin locally, and scroll down to the Footnotes table. Try checking a checkbox there and you'll get a JS error.

Maybe the structure of these sections have changed in a different way?

Not sure if this is right, but maybe `td.field-doc_relation div` should be `td.field-doc_relation > div` in those two places. (`> div` will match only the the immediate child, whereas `div` could match children of children)

Issue #808, opened 2022-04-25
clean up unnecessary attribution formatting code and methods


rlskoeser:
we have a `format_attribution` templatetag but we don't really need it anymore; currently only referenced in the view that generates a local manifest

additional context:
> that's all it's used for, so should probably not be a template tag. I believe it was formerly used for displaying attributions in the document detail template, but we stopped displaying that per `Document` and instead show attributions for each `Fragment`. So I think the local manifest is the only place it's used, and the local manifest is also the only reason we still have a `Document.attribution()` function.

_Originally posted by @blms in https://github.com/Princeton-CDH/geniza/pull/790#issuecomment-1108864636_

kseniaryzhova:
@rlskoeser is this still relevant?

rlskoeser:
@kseniaryzhova yes, but low priority / cdh maintenance (don't assign to Ben)

richmanrachel:
@mabdellatif88 - where does this stand?

mabdellatif88:
Out of scope of the upgrade, needs dedicated and separate effort. I can do it

richmanrachel:
@mabdellatif88 - sounds good. no need to rush this unless it has implications I'm unaware of.

blms:
Code organization change only, testing not required. Closed by merge of #1850 into develop. 👍

PR #1850, opened 2025-08-07
Moving (and refactoring) format_attribution

mabdellatif88:
**Associated Issue(s):** [#808](https://github.com/Princeton-CDH/geniza/issues/808) 

### Changes in this PR

- Rm format_attribution from templatetags
- Adding format_attribution (refactored) to views' DocumentManifestView

### Reviewer Checklist

- [x] Tests pass
- [x] Compatible with the most recent develop

### DIFF
--- a/geniza/corpus/templatetags/corpus_extras.py
+++ b/geniza/corpus/templatetags/corpus_extras.py
@@ -121,18 +121,6 @@ def iiif_info_json(images):
 
 
 @register.filter
-def format_attribution(attribution):
-    """format attribution for local manifests (deprecated)"""
-    (attribution, additional_restrictions, extra_attrs_set) = attribution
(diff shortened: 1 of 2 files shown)


blms (commented):
~~Looks good! But now checking it out locally, I'm wondering if I was wrong in saying that we want to preserve the HTML 😬 Looking at how it's actually used, it's just used in the `attribution` field of IIIF manifests like [this one](https://geniza.princeton.edu/documents/19559/iiif/manifest/), which is supposed to be human readable.~~

~~How much effort to simplify this into just joining all of these strings by a space in between, instead of any HTML or newlines, and updating the unit tests to expect that format?~~

blms (approved):
Scratch my previous review, just took another look at [the Presentation 2.0 spec](https://iiif.io/api/presentation/2.0/#property-values-in-html) and it explicitly permits HTML in the `attribution` line. So I think we're good to go! ✅

Issue #1868, opened 2025-08-19
As a content editor/admin, I want the "Region" model and "Containing region" field renamed to "Geographic area", so it is less confusing with the existing Region checkbox.


blms:

blms:
Verified re-labeling in test site, closing

PR #1869, opened 2025-08-19
Rename containing region "geographic area" (#1868)

blms:
**Associated Issue(s):** #1868

### Changes in this PR

- Use "verbose name" and "verbose name plural" to relabel the admin module for Regions to "Geographic area"
- Use "verbose name" to relabel the foreign key `containing_region` field on Place
- Rename the metadata export column `containing_region` to `geographic_area`

### Notes

- Used verbose name rather than table/column rename operations to prevent unnecessary database schema changes, since it's only a nomenclature change
- The changes to `outd["url_admin"] ` in the metadata export are unrelated, automated formatting by pre-commit hooks


mabdellatif88 (approved):
Looks good

Issue #1800, opened 2025-05-12
Footnotes without doc relations can display incorrectly


blms:
### Testing notes (QA)

On the [test public site](https://test-geniza.cdh.princeton.edu/):
- [ ] Navigate to a document with a footnote with no doc relations, such as PGPID 469, and navigate to its "Selected bibliography" page. It should not have an empty gray square below the citation (see below screenshot)

---

### dev notes 

- A user tried to add digital edition and translation content to a document.
- It appears that document may have already had a footnote on that source record without any doc relation checked. That footnote has no log entries, but its database id (`23100`) suggests it is significantly older than the ones created recently (`24673`, `24675`, `24677`). Unclear how or when it was created.
- The user attempted to add the digital translation twice (`24675`, `24677`); the first time, it appears the footnote did not get the "Digital Translation" relation attached (`24675`)
- There is also a bug where footnotes with no relation type add an empty "tag" bubble to the select bibliography page
   ![Image](https://github.com/user-attachments/assets/734fdadc-b893-47fb-a24a-05d323f45339)

Until the bug is fixed, this could possibly occur again in the following footnotes with no doc relations:

<details><summaryFootnotes with no doc relations</summary>

- Footnote of BL OR 5542.34 ([PGPID 469](https://geniza.princeton.edu/admin/corpus/document/469/change/))
- Footnote of BL OR 5566B.7 ([PGPID 489](https://geniza.princeton.edu/admin/corpus/document/489/change/))
- Footnote of Bodl. MS heb. c 50/14 ([PGPID 555](https://geniza.princeton.edu/admin/corpus/document/555/change/))
- Footnote of Bodl. MS heb. c 56/10 ([PGPID 557](https://geniza.princeton.edu/admin/corpus/document/557/change/))
- Footnote of Bodl. MS heb. d 66/88 ([PGPID 585](https://geniza.princeton.edu/admin/corpus/document/585/change/))
- Footnote of Bodl. MS heb. d 68/101 ([PGPID 589](https://geniza.princeton.edu/admin/corpus/document/589/change/))
- Footnote of Bodl. MS heb. f 56/48 ([PGPID 611](https://geniza.princeton.edu/admin/corpus/document/611/change/))
- Footnote of Bodl. MS heb. f 56/48 ([PGPID 611](https://geniza.princeton.edu/admin/corpus/document/611/change/))
[...]

PR #1877, opened 2025-08-29
Prevent empty doc relation from displaying (#1800)

blms:
**Associated Issue(s):** #1800

### Changes in this PR

Per #1800:
- Prevent empty `doc_relation` on Footnotes from being displayed in and below citations

### DIFF
--- a/geniza/corpus/templatetags/corpus_extras.py
+++ b/geniza/corpus/templatetags/corpus_extras.py
@@ -176,7 +176,7 @@ def all_doc_relations(footnotes):
     relations = set()
     for fn in footnotes:
         relations.update(
-            set([n.strip() for n in fn.get_doc_relation_display().split(",")])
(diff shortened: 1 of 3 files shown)


geniza/corpus/templatetags/corpus_extras.py
```diff
@@ -176,7 +176,7 @@ def all_doc_relations(footnotes):
     relations = set()
     for fn in footnotes:
         relations.update(
-            set([n.strip() for n in fn.get_doc_relation_display().split(",")])
+            set([n.strip() for n in fn.get_doc_relation_display().split(",") if n])
```
mabdellatif88, line 179: Good you are checking n is not empty, but it would be better if you change the name of the variable into something that is more readable

mabdellatif88 (changes requested):
Something minor for readability: change the variable name. Otherwise, you'd be all set

Issue #1171, opened 2022-10-07
Changing languages after navigating to a document via search redirects to search results page


blms:
Steps to replicate:
1. Login to admin so that you can see the language switcher
2. Navigate to the public site search and make a search
3. Click on a document detail link from search results
4. Change languages using the language switcher
5. You will be redirected back to search results

This is likely related to Turbo, as the language switcher is outside of the turbo frame being updated.

kseniaryzhova:
@blms Is this easy to fix? Because it does seem to be an issue with people trying to use the site in other languages.

blms:
@kseniaryzhova I think it might be a tricky one, it might require some restructuring of how we use Turbo. I would estimate around 6 hours. Sounds like it might be high priority though?

PR #1889, opened 2025-09-22
Allow the language switcher to reload asynchronously (#1171)

blms:
**Associated Issue(s):** #1171

### Changes in this PR

- Create a simple template-rendering view for the language switcher, accessible at `/language-switcher`
- Pull root `ul` outside of the template and attach it to a new Stimulus controller
- On turbo page loads in the `main` turbo-frame, use the Stimulus controller to reload the language switcher asynchronously, such that it can get the current page URL in other languages despite being outside the `main` turbo-frame


mabdellatif88:
Looking good

mabdellatif88 (approved):

Issue #1871, opened 2025-08-20
As a front end user, I want a more specific error message to indicate I have hit a solr regex complexity limit, so I know how to adjust my query.


blms:
### testing notes (QA)

On the [QA public site](https://test-geniza.cdh.princeton.edu/):
- [x] Perform a Regex search with `.{0,19}` between two characters, such as `A.{0,19}B`. You should see an error in red text, with a message indicating that your query may be too complex and to try using alternatives to large ranges.

---

A user has triggered the Solr error `org.apache.lucene.util.automaton.TooComplexToDeterminizeException: Determinizing [regex] would require more than 10000 effort.`

This appears to be a hard limit in Solr, which is configurable in Lucene (`DEFAULT_DETERMINIZE_WORK_LIMIT`) but not solr.

An example regex that hits this error is `a.{0,19}t`, which, due to our full-field matching requirements, is transformed into `.*a.{0,19}t.*`.

Currently the user gets the generic solr regex error message: "Error retrieving search results. Check regular expression for errors such as unescaped or unclosed brackets or parentheses."

Ideally, we should tell the user when query complexity is what caused an error. However, due to https://github.com/Princeton-CDH/parasolr/issues/88, we currently can't see the specific error message that Solr encountered via Parasolr.

If the required Parasolr modification is too big of a lift, a possible solution would be just including some text like this in the generic regex error message: "The regex query may also be too complex to compute; try using `.*` instead of large numeric ranges like `.{0,20}`."

Issue #1845, opened 2025-08-04
Invalid regex `.*{0,10}` raises server error instead of form validation error


blms:
### testing notes (QA)

On the [QA public site](https://test-geniza.cdh.princeton.edu/):
- [x] Navigate to the document search and perform a regex search. In the query include `.*{0,10}`. You should see an error in red text on the search page explaining why this does not work.

---

`.*{0,10}` is invalid regex as `.*` is not quantifiable, and it raises a server error on PGP regex search.

We should catch this on the application side as part of the [regex search form validation routine](https://github.com/Princeton-CDH/geniza/blob/cd060ddc72daa15084194090833e42bf5fc9ca12/geniza/corpus/forms.py#L412).

PR #1893, opened 2025-09-22
Validate nested regex quantifiers (#1845) and update error message (#1871)

blms:
**Associated Issue(s):** #1845, #1871 

### Changes in this PR

Per #1845:
- Add validation to prevent nested regex quantifiers, such as `*{0,12}` or `+*`, using python regex's built in error handling

Per #1871:
- Update generic regex error message to indicate that complexity errors (such as large numeric ranges) may be the culprit

### DIFF
--- a/geniza/corpus/forms.py
+++ b/geniza/corpus/forms.py
@@ -362,6 +362,14 @@ class DocumentSearchForm(RangeForm):
         "has_discussion": "has_discussion",
     }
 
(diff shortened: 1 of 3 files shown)


mabdellatif88 (approved):
It is always a good idea to add try except, have a hierarchy of specific and generic errors, and write negative as well as positive testing. The fix looks good

Issue #1916, opened 2025-10-28
All tags deleted on new tag + backspace in the admin tags field


mrustow:
### Testing notes (QA)

On the [QA site admin](https://test-geniza.cdh.princeton.edu/admin):
- [x] Navigate to a document with existing tags. Start typing a tag name in the field, and select one from the dropdown. Hit backspace and it should only remove the tag you just added (and let you continue typing/edit what you typed originally)
- [x] You should also be able to remove a tag by clicking the little X button on it
- [x] Saving the document should not cause any of the original tags to be unexpectedly removed

---

In the admin interface, when I type a tag and then try to delete it (for instance if I select the wrong one from the drop-down), hitting the delete key results in all the tags in the tags field getting deleted. 

Here's a screen recording (already shared with Ben on Slack):

https://github.com/user-attachments/assets/17bf1459-df7b-4bad-a6d5-1b7113d27910

amelbensalim:
Hitting backspace still removes all the tags, they only reappear when I refresh the page (without saving). Will send a video of this on slack.

amelbensalim:
Working very smoothly! Thanks so much!

PR #1919, opened 2025-10-29
Fix "clear all" tag bug in admin by monkey-patching select2 (#1916)

blms:
**Associated Issue(s):** #1916

### Changes in this PR

- Apply a monkey-patch to select2 at runtime to fix [this bug](https://github.com/select2/select2/issues/3335), present in upstream select2 (via django-autocomplete-light), where backspacing a single new entry on a multi-select combobox will clear all entries from it

### DIFF
--- a/sitemedia/js/admin.js
+++ b/sitemedia/js/admin.js
@@ -147,6 +147,30 @@ window.addEventListener("DOMContentLoaded", () => {
             }
         });
     }
+
+    // hook into django-autocomplete-light initialization event
+    // and monkey-patch select2 allowClear bug. more info:
+    // https://github.com/select2/select2/issues/3335#issuecomment-1218072422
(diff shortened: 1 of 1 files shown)


mabdellatif88 (approved):
I am approving for the time-being. However, Instead of having to monkey-patch the JS library select2, I recommend creating a chore task to replace [select2](https://github.com/select2/select2) with another one that is better and more regularly maintained (e.g. [tom-select](https://github.com/orchidjs/tom-select) or [bootstrap-multiselect](https://github.com/davidstutz/bootstrap-multiselect)).

blms:
@mabdellatif88 Definitely agree that would be ideal. Unfortunately as long as we use `django-autocomplete-light`, we are tied to using `select2`. (it's not a direct dependency of our project but a sub-dependency of DAL) Maybe the maintainers of that project will switch it at some point though.

Issue #1974, opened 2026-03-23
As a content editor, I would like LtR scripts to be compatible with the transcription editor, so that I can enter Latin-script transcriptions


amelbensalim:
### testing notes (QA)

On the [staging site](https://test-geniza.cdh.princeton.edu/en/):
- [x] Navigate to add or edit a transcription or translation
- [x] When you open the text editor to edit the contents of one, you should see new buttons for left-to-right and right-to-left directionality
- [x] The buttons should correctly reflow the text RTL or LTR
- [x] Saving should persist changes to the directionality, including when you navigate back to the document detail page
- [x] You should now be able to save a transcription in a LTR language
- [x] You should now be able to save a translation in a direction that does not match the Source language (i.e. if the source is in english you should be able to save an RTL translation, and if the source is in Hebrew you should be able to save an LTR one)
- [x] Note: the styles are optimized for Latin script in LTR, and Arabic or Hebrew script in RTL. It may look a bit strange if you try to align Latin as RTL or Arabic/Hebrew as LTR. If there are other legitimate cases where rendering is wrong and we need to make an exception, please note them here!
- [x] Search for the document(s) you added/modified transcriptions and translations for (by PGPID might be easiest, like `pgpid:100`), in the [document search](http://test-geniza.cdh.princeton.edu/en/documents). In search result snippets, the transcription and translation should reflect the alignment/directionality you chose.

---

**Is your feature request related to a problem? Please describe.**
In Ottoman studies, transcriptions are conventionally written using the Latin-script based modern Turkish. Currently, our transcription editor is hard coded for right-to-left scripts like Arabic and Hebrew, and it cannot display left-to-right scripts. Moreover, this would benefit other later Geniza documents that are written with Latin-scripted languages. 
Slack thread on this issue: 

<img width="1704" height="857" alt="Image" src="https://github.com/user-attachments/assets/6088f6ff-d330-480c-8fbf-341b0c3caf0d" />

**Describe the solution you'd like**
We would like the transcription editor to be able to accepted both LtR and RtL scripts, similar to how the translation editor currently does.

blms:
We have a handful of different ways we could approach this. I’m leaning towards **option 4** but want to make sure I lay out the options and reasoning.

1. One option would be the same way we approached it for translation: the Source record would indicate the text direction. However, many documents have transcriptions from Sources that don’t match the direction of the text currently. About 3,480 transcribed documents have transcriptions from Sources that are LTR (about 185 Source records).
[...]

PR #1975, opened 2026-04-01
Add LTR and RTL buttons to transcription/translation text editor (#1974)

blms:
**Associated Issue(s):** #1974

### Changes in this PR

- Update `annotorious-tahqiq` to 1.6.0 and include `directionality` TinyMCE plugin to get the new LTR and RTL buttons in the text editor
- Add `dir` to allowed attributes for HTML sanitization
- Ensure attributes like `dir` are preserved during list flattening (and add a small amount of fallback behavior)
- Update `HTMLLineNumberParser`, which is used only in solr indexing, to propagate the `dir` attribute down to all `li` elements in the indexed HTML
   - Search results often return highlighted transcription/translation snippets with _just_ the `li`s and no surrounding list—this ensures they get the correct directionality
- Update all CSS to support transcription/translation block-level `dir` (and also make it a little more DRY and generalized using `inline` properties)

### Reviewer Checklist

- [x] Run `npm install` to get the latest version of `annotorious-tahqiq` that includes the new buttons

### DIFF
--- a/geniza/annotations/models.py
+++ b/geniza/annotations/models.py
@@ -109,7 +109,7 @@ class Annotation(TrackChangesModel):
 
     # allowed tags and attributes for annotation body content HTML
     ALLOWED_TAGS = ["del", "li", "ol", "p", "span", "sup", "i"]
-    ALLOWED_ATTRIBUTES = ["lang"]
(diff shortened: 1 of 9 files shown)


mabdellatif88 (approved):
Looks good!

Issue #1977, opened 2026-04-09
LTR transcriptions have weird spacing when window is smaller than fullscreen


kseniaryzhova:
### testing notes (QA)

On the [staging site](https://test-geniza.cdh.princeton.edu/en/):
- [x] Add a transcription and a translation—both LTR—to a document. (you could even try copying and pasting the transcription and translation [from here](https://geniza.princeton.edu/en/documents/35086)!)
- [x] Save them and navigate back to the document details. They should align correctly.

---

**Describe the bug**
Line spacing for LTR transcriptions is off when the browser window is smaller than full screen. Namely the lines are too close together especially if it's a long transcription that is 4 lines or longer.

**To reproduce**
1. Go to ['this PGPID'](https://geniza.princeton.edu/en/documents/35086/)
2. Make sure transcription and translation panels are open.
3. Make the browser window smaller than full screen.

**Expected behavior**
We want the line spacing of all transcriptions and translations to adjust automatically when the size of the browser window changes.

**Screenshots**
<img width="1433" height="818" alt="Image" src="https://github.com/user-attachments/assets/d6a11621-5764-446a-a90d-c5ea976c634d" />

**Device information**
 - Device: [e.g. iPhone 6, Samsung Galaxy S10, MacBook Pro]
 - OS and version: [e.g. macOS 10.14, Windows 10, iOS 8.1, Android 9]
 - Browser and version: [e.g. Firefox 72, Chrome 79, Safari 13]
 - Site version [e.g. 0.24.1 - check in the site footer]

**Additional context**

kseniaryzhova:
@blms
Translation does not align with transcription, even after a hard refresh. This is the full screen size now with all three panels open at once. https://test-geniza.cdh.princeton.edu/en/documents/5585/
<img width="1683" height="614" alt="Image" src="https://github.com/user-attachments/assets/88a15dd7-460f-450a-a9e6-db6885770f0a" />

blms:
@kseniaryzhova I believe that's because the translation source is listed as being in Hebrew, so the Hebrew font size and line spacing are being applied to the translation. Is that a legitimate use case for Latin script translations? If so, we can make an exception, but we should make the decision.

kseniaryzhova:
@blms thanks for catching, I re-tested with an LTR-language source translation and it seemed to work. I don't foresee this being an issue because secondary sources tagged with a language will have the translations in that language. Closing, thank you!

PR #1978, opened 2026-04-09
Fix alignment line height for nested p elements inside li (#1977)

blms:
**Associated Issue(s):** #1977

### Changes in this PR

- Ensures `p` elements nested inside `li` get the correct standardized transcription/translation line height, when there are both a transcription and translation present, so that total height of text per `li` is correctly calculated. This fixes a regression introduced [here](https://github.com/Princeton-CDH/geniza/blob/c98709b1faea7c0ada7621177bb6faf330eb3a2b/sitemedia/scss/components/_transcription.scss#L1569-L1574) in 94b4786180ed58a70a810ca7bc1b09d3076f5b3a.
  - (generally these should still get the `body` line-height, but it needs to be overridden in the specific case that we're aligning transcription and translation, i.e. both are visible)
- Adds code to sanitization process to unwrap `<p>` inside `<li>`, ensuring any `dir` attribute on the `<p>` is preserved and propagated back up

### DIFF
--- a/geniza/annotations/models.py
+++ b/geniza/annotations/models.py
@@ -265,7 +265,8 @@ class Annotation(TrackChangesModel):
 
         # if resulting text has any span elements with no attributes, remove them;
         # if resulting text has any p elements inside li, unwrap them
-        if "<span>" in cleaned_html or "<p>" in cleaned_html:
+        # (leave out ">" from opening tag in case of attributes)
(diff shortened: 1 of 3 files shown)


mabdellatif88 (changes requested):
Minor additional guard: please strip away html tags from the input text

mabdellatif88 (approved):
Looks good!

PR #1982, opened 2026-04-21
Validating input dates to error out on future dates

### LINKED ISSUE #1973: As a content admin, I should not be able to put in a date for a document or a person page that is AFTER the current date, so that I can avoid adding incorrect dates and typos to the database.
### testing notes

- [ ] Create a new document with future CE date => PGP should stop you
- [ ] Edit an existing document; change its CE date to a future one => PGP should stop you
- [ ] Create a new document with future Date on document => PGP should stop you
- [ ] Edit an existing document change its Date on document to a future one => PGP should stop you
- [ ] Create a new person with future CE date => PGP should stop you
- [ ] Edit an existing person; change its CE date to a future one => PGP should stop you

### dev notes

- Update `DocumentDateMixin.clean` to raise `ValidationError` for dates in the future:
   - For the date to check, use `self.doc_date_standard` if it's been set; if not, but `self.doc_date_original` has been set, compute the standard date using `self.standardize_date(update=False)`
   - If the user entered a date range, make sure to check only the latter date of the range; ranges are entered with a `/` between two dates, so you should be able to split on that character
   - Let's say the date you're checking is `date_to_check`; get it into ISO format using `PartialDate(date_to_check).isoformat(mode="max", fmt="full")`
   - To make the comparison, get a python date object from it using `date.fromisoformat`, and compare it against `date.today()`, raising a `ValidationError` if it is later

---

**Is your feature request related to a problem? Please describe.**
Right now you can add any date to documents and people, which means typos aren't caught. There's been a few dates added in the future and we want to avoid that.

**Describe the solution you'd like**
If someone tries to enter a date after today's date (today being the day you are entering the information) the PGP should not allow you to save and will show an error message indicating the issue (dates in the future are not valid). This should apply across all calendars supported by the PGP: CE, Seleucid, Hijri, AM, Kharaji.

mabdellatif88:
**Associated Issue(s):** resolves #1973

### Changes in this PR

- Validating documents' dates
- Testing attempting to input future dates
- Validating person's dates
- Testing attempting to add a person with a future date

### Reviewer Checklist

- [x] Create a new document with future CE date
- [x] Edit an existing document; change its CE date to a future one
- [x] Create a new document with future Date on document
- [x] Edit an existing document change its Date on document to a future one
- [x] Create a new person with future CE date
- [x] Edit an existing person; change his/her CE date to a future one
- [x] Tests pass 100%

### DIFF
--- a/geniza/corpus/dates.py
+++ b/geniza/corpus/dates.py
@@ -244,6 +244,16 @@ class DocumentDateMixin(TrackChangesModel):
             raise ValidationError("Original date is required when calendar is set")
         if self.doc_date_original and not self.doc_date_calendar:
             raise ValidationError("Calendar is required when original date is set")
+        date_to_check = self.doc_date_standard
+        if not self.doc_date_standard and self.doc_date_original:
+            date_to_check = self.standardize_date(update=False)
+        if (
+            date.fromisoformat(
+                PartialDate(date_to_check).isoformat(mode="max", fmt="full")
+            )
+            >= date.today()
+        ):
+            raise ValidationError("Can't input a date in the future!")
 
     def standardize_date(self, update=False):
         """
@@ -283,7 +293,6 @@ class DocumentDateMixin(TrackChangesModel):
                     end = start
                 else:
                     end = PartialDate(date_parts[1])
-
                 self._parsed_date = {"start": start, "end": end}
             except ValueError:
                 # ignore if it can't be parsed (records before validation added)
--- a/geniza/corpus/dates.py
+++ b/geniza/corpus/dates.py
@@ -236,6 +236,7 @@ class DocumentDateMixin(TrackChangesModel):
         return self.get_document_date(self.doc_date_standard, self.original_date)
 
     def get_doc_date(self):
+        """Utility method to validate / vet document's date"""
         date_to_check = (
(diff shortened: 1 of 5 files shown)


geniza/entities/models.py
```diff
@@ -471,6 +471,26 @@ def __str__(self):
         except Name.DoesNotExist:
             return str(self.names.first() or super().__str__())
 
+    def clean(self):
+        """
+        Require date_str is not a future date.
+        If date_str is a range, require its end not tobe a future date
+        """
+        # date_to_check = datetime.strptime(self.date_str, '%d %B %Y CE').date()
```
blms: Commented out code

geniza/corpus/dates.py
```diff
@@ -235,6 +235,28 @@ def document_date(self):
         """Property: formatted display of combined original and standardized dates"""
         return self.get_document_date(self.doc_date_standard, self.original_date)
 
+    def get_doc_date(self):
```
blms, line 238: Missing docstring (that's why sphinx complained in the PR checks)

blms (commented):
Code looks good, and works as intended! Just a couple of tiny things and then I think this'll be ready to go 👍

blms (approved):
Changes look good, thanks!

Issue #1986, opened 2026-05-04
Admin error, cannot select/reorder thumbnails


blms:
### Testing notes (QA)
On the test site admin:
- [x] Navigate to https://test-geniza.cdh.princeton.edu/admin/corpus/document/25837/change/
- [x] You should see both the placeholders and the real IIIF images in the "image order/rotation overrides" section
- [x] You should be able to reorder them with click-and-drag
- [x] You should also be able to use the checkmarks on all of them, and rotation buttons on the IIIF images
- [x] Saving should persist the changes
- [x] Reordering images and placeholders should reorder transcription content on them
- [x] In the transcription editor section, you should still be able to move transcription content from placeholders to images the normal way

---

The admin edit page for this document presents an error: https://geniza.princeton.edu/admin/corpus/document/25837/change/

```
Uncaught TypeError: Cannot read properties of null (reading 'classList')
    at admin.js:353:82
```

mabdellatif88:
What are the steps to re-produce?

blms:
@mabdellatif88 The JS error in the issue description appears on page load on [the linked admin page](https://geniza.princeton.edu/admin/corpus/document/25837/change/). It prevents the checkboxes and rotation controls in the "Image order/rotation overrides" section from rendering after the first placeholder ("T-S NS 327.45 recto"). It also prevents the images from being reordered using click and drag.

mabdellatif88:
I don't see it (neither locally nor on prod) even with Chrome Dev tool's JS console (all I can see there is a couple of irrelevant warnings)

blms:
@mabdellatif88 That's odd. Locally you may need a newer prod db clone (there should be a dump from the other day in `/home/conan/migrate` on `cdh-geniza1`). Could you send a screenshot of what this section looks like on prod for you, on this record? (PGPID 25837)

<img width="1268" height="599" alt="Image" src="https://github.com/user-attachments/assets/960b9314-0584-4655-b59c-16659da13518" />

mabdellatif88:
Same as the one you shared. I don't see the error in both. Is it only visible in the JS console?

blms:
Yes, the `Uncaught TypeError: Cannot read properties of null (reading 'classList') at admin.js:353:82` is only visible in the JS console.

The screenshot shows the accompanying behavior bug, which is the missing checkboxes under the last four images, and the missing rotation controls under the last three.

You're seeing that behavior bug but nothing in the JS console?

[...]

Issue #1942, opened 2025-11-26
As a content editor, I want to reorder placeholders on fragments in the rotation/order override section, so that I can interleave their transcriptions with those for existing images.


blms:
### Testing notes (QA)

On the [QA site admin](https://test-geniza.cdh.princeton.edu/admin):
- [x] Navigate to a document with content on placeholders, such as PGPID 451, PGPID 4288, and PGPID 7030.
- [x] The placeholder images that have transcription or translation content should show up in the reordering section.
- [x] You should be able to reorder them if there is more than one image with content on it, and when you save changes, the order should be reflected in the transcription/translation section.
- [x] Spot check migrated transcription and translation content from the content that was added before placeholders were attached to individual fragments, via this CSV: [migration-report.csv](https://github.com/user-attachments/files/24295746/migration-report.csv)
   - Check them against the [published PGP site](https://geniza.princeton.edu/) to ensure the content is in the correct order and no content is missing.
   - Confirm they can still be viewed/edited/reordered in the transcription/translation editor.
   - Note that they may not be assigned to the correct shelfmark or image—especially on joins—since they were previously unassociated with any fragment.

---

For example:
- I have two fragments, A and B
- A has images and B has placeholders
- I need the transcription to be ordered B recto, A recto, B verso, A verso

This is not possible in the current implementation.

amelbensalim:
Looking great overall!

 I have a few q's related to the last testing task: 

- On the published site, in cases where it seems images were later added, do those require manual fixing to associate the transcription/lation content to the new images? (apologies if im forgetting a discussion we already had on this, or if this issue will allow us to easily resolve this) Example: https://geniza.princeton.edu/en/documents/620/

- Still on the published site, is it expected for the editor to display a separate third placeholder related to an existing transcription when I open the editor to add a second?: 
Example: 
https://geniza.princeton.edu/en/documents/446/transcribe/317/
<img width="2251" height="1318" alt="Image" src="https://github.com/user-attachments/assets/208c607e-d217-4b67-97d3-cc2817f5adc2" />

<img width="2242" height="1315" alt="Image" src="https://github.com/user-attachments/assets/b02f1596-239e-4ddc-ba05-0d92d1547ed7" />

- From this^ editor, I could not reorder the placeholders 
- Other than these things, I did not notice anything major like missing content when spot checking

blms:
[...]

PR #1988, opened 2026-05-04
Ensure admin pages handle mixed placeholders/IIIF (#1986)

blms:
**Associated Issue(s):** #1986

### Changes in this PR

Per #1986:
- The JS error in this issue resulted from the work in #1942 to allow reordering placeholders. That work was done assuming that all annotation content per fragment would either be on placeholders _OR_ iiif images, but didn't account for both.
- That resulted in the `thumbnails` section no longer rendering placeholder thumbnails once the fragment had been populated with IIIF images. Thus, the JS script failed on trying to match "annotated canvases for reordering" with "rendered thumbnails on fragments".
- This PR ensures that `Fragment.thumbnail` always renders any annotated placeholders as well as any IIIF images, so that the reordering section and the thumbnails are never out of sync.

### DIFF
--- a/geniza/corpus/models.py
+++ b/geniza/corpus/models.py
@@ -2038,33 +2038,68 @@ class TextBlock(models.Model):
 
     def thumbnail(self):
         """iiif thumbnails for this TextBlock, with selected images highlighted"""
-        canvases = []
-        if not self.fragment.iiif_images():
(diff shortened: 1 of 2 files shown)


mabdellatif88 (approved):
Looking good!

PR #1990, opened 2026-05-07
Disallowing invalid date ranges 

### LINKED ISSUE #1981: As a content admin, I should not be able to put in a date range for a document or a person page of which the end date is older than the start one, so that I can avoid adding incorrect dates and typos to the database.
**Is your feature request related to a problem? Please describe.**
Right now you can add any date to documents and people, which means typos aren't caught. There's been a few date ranges with end older than start added and we want to avoid that.

**Describe the solution you'd like**
If someone tries to enter a date range of which end is older than start, the PGP should not allow you to save and will show an error message indicating the issue (dates in the future are not valid). This should apply across all calendars supported by the PGP: CE, Seleucid, Hijri, AM, Kharaji.

mabdellatif88:
**Associated Issue(s):** resolves #1981

### Changes in this PR

- Disallowing invalid date ranges for documents module
- Disallowing invalid date ranges for people module

### Reviewer Checklist

- [x] Create a new document with CE date range whose end is earlier than the start
- [x] Edit an existing document; change its CE date to a date range whose end is earlier than the start
- [x] Create a new document with (non CE) date range whose end is earlier than the start
- [x] Edit an existing document change its (non CE) date range whose end is earlier than the start
- [x] Create a new person with date range whose end is earlier than the start
- [x] Edit an existing person; change his/her (non CE) date range whose end is earlier than the start
- [x] Tests pass 100%

### DIFF
--- a/geniza/corpus/dates.py
+++ b/geniza/corpus/dates.py
@@ -252,10 +252,25 @@ class DocumentDateMixin(TrackChangesModel):
         date_to_check = (
             self.end_date if not date_to_check and self.parsed_date else date_to_check
         )
+        start_to_return, end_to_return = None, None
+        range_splitted = (
+            self.doc_date_original.split("/") if self.doc_date_original else None
+        )
+        if range_splitted and len(range_splitted) > 1:
+            start = PartialDate(range_splitted[0])
+            end = PartialDate(range_splitted[1])
+            start_to_return = date.fromisoformat(
+                start.isoformat(mode="max", fmt="full")
+            )
+            end_to_return = date.fromisoformat(end.isoformat(mode="max", fmt="full"))
         return (
-            date.fromisoformat(date_to_check.isoformat(mode="max", fmt="full"))
-            if date_to_check
-            else None
+            (
+                date.fromisoformat(date_to_check.isoformat(mode="max", fmt="full"))
+                if date_to_check
+                else None
+            ),
+            start_to_return,
+            end_to_return,
         )
 
     def clean(self):
@@ -267,9 +282,11 @@ class DocumentDateMixin(TrackChangesModel):
             raise ValidationError("Original date is required when calendar is set")
         if self.doc_date_original and not self.doc_date_calendar:
             raise ValidationError("Calendar is required when original date is set")
-        date_to_check = self.get_doc_date()
+        date_to_check, start_date, end_date = self.get_doc_date()
         if date_to_check and date_to_check > date.today():
             raise ValidationError("Can't input a date in the future!")
+        if start_date and end_date and end_date < start_date:
+            raise ValidationError("End date can't be earlier than start date!")
 
     def standardize_date(self, update=False):
         """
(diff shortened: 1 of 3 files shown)


geniza/corpus/dates.py
```diff
@@ -252,10 +252,25 @@ def get_doc_date(self):
         date_to_check = (
             self.end_date if not date_to_check and self.parsed_date else date_to_check
         )
+        start_to_return, end_to_return = None, None
+        range_splitted = (
+            self.doc_date_original.split("/") if self.doc_date_original else None
```
blms: Should this be `date_to_check.split("/")` instead of `self.doc_date_original.split("/")`?

blms (changes requested):
I was able to save a new document with the CE date range `2026/2024`.

Image

geniza/corpus/dates.py
mabdellatif88: - Thanks for catching the range error, I'll investigate it and add more test if turns out necessary
- No, date_to_check above may be coming from either a range or a single date. The line I added at 256 is intended only for ranges

blms: I see. In that case, should this be either `self.doc_date_standard` (CE date field) or an already-converted `self.doc_date_original` (historic calendar date)? I'm thinking it's not running right now since I only entered a CE date and not a historic calendar date, so `range_splitted` is being set `None`.

blms (commented):
I wonder if we should be checking the bare `doc_date_original` at all or if we should only ever validate it post-standardization?

I ask because I went to edit an existing document with an original date, PGPID 27263, and "Date on document" is `1 Nisan 1485`, with the Seleucid calendar. I figured to test the range I should try `1 Nisan 1485/1 Nisan 1484`, but that resulted in:
```
ValueError at /admin/corpus/document/27263/change/
invalid literal for int() with base 10: '1 Nisan 1485'
```
Which I guess makes sense, because that's a full date with a month's name, not a formatted MM-DD-YYYY. 

Maybe for such cases with historic dates, we only check against the standardized date post-`convertdate`, and if it isn't able to successfully convert it, we don't bother with validation (since we can't really validate against something non-standardized)?

Let me know what you think!

blms:
I tried every other test case and they all worked well, so hopefully that issue with historic calendars should be the last hurdle. Otherwise this is looking good 👍

mabdellatif88:
That is what I tried initially, but standardization loses the order info of a range, it passes the oldest date as start and the youngest/most recent date as the end, so couldn't really validate after standardization unless we agree to modify the standardization itself. Another alternative is to add a different validation on the format, in other words, error out if the entered date doesn't comply with yyyy-mm-dd (only in the case of CE dates)

blms:
> That is what I tried initially, but standardization loses the order info of a range, it passes the oldest date as start and the youngest/most recent date as the end, so couldn't really validate after standardization unless we agree to modify the standardization itself.

Ah, I see. Maybe if `doc_date_original` contains a `/` then we convert each date on either side of the `/` separately, and then compare them to each other after conversion? (Or maybe we only run the validation on `doc_date_original` if it matches yyyy, yyyy-mm, or yyyy-mm-dd, and otherwise ignore it and assume it's valid)

> Another alternative is to add a different validation on the format, in other words, error out if the entered date doesn't comply with yyyy-mm-dd (only in the case of CE dates)

We should definitely do that for `doc_date_standard`, to match the help text, good call!

blms (approved):
Changes look good, all works well now! Thanks for the thorough unit tests as well.

Issue #2014, opened 2026-08-17
As an admin site user, I would like to be able to filter for transcriptions on placeholder images, so that they can be easily updated


amelbensalim:
### Testing notes (QA)

On the [test site admin](https://test-geniza.cdh.princeton.edu/admin/):
- [x] Navigate to the Documents list. In the right sidebar there should be two new filters: one for _all_ documents with transcriptions on placeholder images, and one for documents where there are _also_ existing IIIF images (i.e. transcriptions that can already be moved to images)
- [x] Clicking on the first filter should pull up documents with any transcription or translation content on placeholders
- [x] Clicking on the second filter should instead pull up only documents for which transcription or translation content lives on placeholders **AND** real IIIF images are available

---

There are a host of transcriptions that were uploaded on placeholder images when images were not available. As more have been digitized/made available, particularly from CUDL, we would like to go through them and ensure all the images and transcriptions are properly matched. Having an admin site filter would make the task much easier and more sustainable over time

Princeton-Geniza-Lab:
@blms this seems to be on the testing site already, did you want us to go ahead and test? And do you have testing notes?

blms:
@Princeton-Geniza-Lab I didn't realize a staging deploy had been completed successfully! Yes, please go ahead and test. I have added testing notes to the issue descirption.

kseniaryzhova:
@blms works well, thank you! Closing! (Had to switch accounts because in my PGL account I couldn't click the checkboxes, may not have the right permissions)

PR #2015, opened 2026-08-18
Add placeholder image filters to admin (#2014)

blms:
**Associated Issue(s):** #2014

### Changes in this PR

- Add a filter to the Document admin list view that shows documents with any transcription or translation content on placeholder images
- Add a second filter in the same place that only shows documents that have content on placeholder images AND that have real IIIF images—in other words, placeholder content that can be moved to real images


mabdellatif88 (approved):
Looks good and works well!

